PR #6, opened 2019-08-03
Validator Set Hash

ebuchman:
Adds a `validator::Set` struct containing a sorted list of validators and a `hash()` method to compute the merkle root.

Includes a test vector generated from the Go codebase.


ebuchman:
Merging for now. Will treat CI next

Issue #10, opened 2019-08-12
Errors parsing the response or Client RPC


amanusk:
I am trying to use the client RPC to interact with the basic `abci-cli kvstore` application.
I have run `tendermint node` and `abci-cli kvstore` from two terminals.
RPC calls with curl as in the examples [here](https://tendermint.com/rpc/?shell#introduction) all work as expected.

To try and broadcast a transaction from rust I use the following code:
```
use tendermint::rpc::Client;

fn main() {
    let client = Client::new(&"tcp://127.0.0.1:26657".parse().unwrap()).unwrap();

    let j = "abcd";

    let tx = tendermint::abci::transaction::Transaction::new(j.as_bytes().to_owned());
    let response = client.broadcast_tx_sync(tx).unwrap();
    println!("{:?}", response);
}
```

I am getting the error 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("invalid type: integer `0`, expected a string at line 5 column 13") }', src/libcore/result.rs:1084:5
```

When calling `curl 'localhost:26657/abci_query?path=""&data="abcd"&prove=false'` afterwards, I do get a response where I can see that `abcd` was indeed stored successfully, so the problem is with parsing the response on the client side.

From some poking around it seems that this has to do with the `Code` field in the response, perhaps it should be `"0"` and not `0`.

Any help solving this is appreciated. Is this a problem of the RPC documentation?

tarcieri:
@amanusk it's possible the Tendermint RPC API changed to return an integer here rather than a string. If so the code parser needs to be modified... ideally to accept both.

PR #13, opened 2019-08-14
Fix abci::Code parser to accept a string or an int

amanusk:
Addresses issue #10
Added tests for both cases


amanusk:
@tarcieri,
Thanks, didn't consider the option of parsing both cases.
Added implementation inspired by [this](https://users.rust-lang.org/t/deserialize-a-number-that-may-be-inside-a-string-serde-json/27318/3).

Not sure error handling is correct.

Also added a test for both cases

PR #14, opened 2019-08-15
RPC: Update hash to 32 byte value.

amanusk:
Update tests to new standard


amanusk:
Running the simple kvstore example, I am getting an error on parsing the result
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("expected 40-character hex string, got \"88D4266FD4E6338D13B845FCF289579D209C897823B9217DA3E161936F031589\" at line 9 column 3") }', src/libcore/result.rs:1084:5
```

It seems hashes have been update from 20 to 32 bytes. https://github.com/tendermint/tendermint/blob/master/CHANGELOG.md#v0260

This PR updates the hash and the tests

tarcieri:
@ebuchman what are your thoughts on compatibility here? should we just target the latest Tendermint release only?

ebuchman:
> should we just target the latest Tendermint release only?

Yes, probably for now. We'll figure out a backwards compatibility story a bit later on ...

src/abci/code.rs
```diff
@@ -87,8 +87,8 @@ impl<'de> Deserialize<'de> for Code {
             where
                 E: serde::de::Error,
             {
-                match val.parse::<u32>() {
-                    Ok(val) => self.visit_u32(val),
+                match val.parse::<u64>() {
```
ebuchman: Why did this change?

src/abci/transaction/hash.rs
```diff
@@ -11,7 +11,7 @@ use subtle::{self, ConstantTimeEq};
 use subtle_encoding::hex;
 
 /// Size of a transaction hash in bytes
-pub const LENGTH: usize = 20;
+pub const LENGTH: usize = 32;
```
ebuchman: I introduced HASH_SIZE in https://github.com/interchainio/tendermint-rs/blob/master/src/merkle.rs#L6 ... we'll need to de-duplicate eventually but for now it's fine ....

tests/rpc.rs
```diff
@@ -155,7 +155,7 @@ mod endpoints {
 
         assert_eq!(
             &response.hash.to_string(),
-            "75CA0F856A4DA078FC4911580360E70CEFB2EBEE"
+            "88D4266FD4E6338D13B845FCF289579D209C897823B9217DA3E161936F031589"
```
ebuchman: Do the values of these not matter at all?

src/abci/code.rs
amanusk: For better readability.
This is related to a previous PR #13
In either cases `visit_u64` method is called, so thought it would be better if it is explicit in the code.

tests/rpc.rs
amanusk: As I understand, these tests only check that parsing work correctly, and do not test the actual correctness of the hash.
Also, the hash is of the sent transaction and not of the response right?

amanusk:
Any other fixes required?

tests/rpc.rs
ebuchman: > Also, the hash is of the sent transaction and not of the response right?

That's right

PR #17, opened 2019-08-19
Allow lower case for abci::data

amanusk:
While for `fmt` both lower case and upper case encoding are possible:
https://github.com/interchainio/tendermint-rs/blob/b88febc173ece044da043b98469f3faa110ef41e/src/abci/data.rs#L44

When decoding upper case from rpc response, there is an error:
```
---- endpoints::broadcast_tx_sync stdout ----
thread 'endpoints::broadcast_tx_sync' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("bad encoding at line 6 column 24") }', src/libcore/result.rs:1084:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```


ebuchman:
Do we need to be using `subtle_encoding` for the hex here? The `hex` crate handles upper case and lower case: https://docs.rs/hex/0.3.2/hex/fn.decode.html

ebuchman:
Oh I see, subtle_encoding is Tony's crate - @tony-iqlusion would it be better to just use it when we're dealing with keys ?

tarcieri:
@ebuchman you can add another crate if you want, but note that `subtle-encoding` also handles upper and lower case hex, and the impl to do either is already available on this type's `FromStr` impl:

https://github.com/interchainio/tendermint-rs/blob/566dfb6/src/abci/data.rs#L42

To use it, change the `Deserialize` impl to something like:

```rust
impl<'de> Deserialize<'de> for Data {
    fn deserialize<D: de::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        use de::Error;
        let string = String::deserialize(deserializer)?;
        string.parse().map_err(D::Error::custom)
    }
}
```

amanusk:
Thanks @tarcieri, updated the method as you suggested.
Although unlikely, there can still be an error if the data is encoded with both lower and upper case chars.

PR #26, opened 2019-09-07
mvp commit validation

ebuchman:
First take at verifying a commit using the existing vote and validator structures.

Note the commit structure in Tendermint is scheduled to change: https://github.com/tendermint/tendermint/issues/1648

It might be better to have a version of this function that is generic in the vote type.

### DIFF
--- a/src/amino_types/vote.rs
+++ b/src/amino_types/vote.rs
@@ -102,7 +102,7 @@ impl block::ParseHeight for CanonicalVote {
 }
 
 impl CanonicalVote {
-    fn new(vote: Vote, chain_id: &str) -> CanonicalVote {
+    pub fn new(vote: Vote, chain_id: &str) -> CanonicalVote {
         CanonicalVote {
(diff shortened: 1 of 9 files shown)


ebuchman:
This is mostly replaced by https://github.com/interchainio/tendermint-rs/pull/31 but we'll probably want some of those to_amino methods for the actual implementations so we can get the sign bytes right. cc @Liamsi

Liamsi:
Yes, I can probably re-use some of the `to_amino` methods. It might be slightly more idiomatic to implement the `From` trait instead to convert to the amino types though:
https://doc.rust-lang.org/std/convert/trait.From.html

A slightly more annoying usage of amino occurs when creating [the header hash](https://github.com/tendermint/tendermint/blob/134fe2896275bb926b49743c1e25493f6b24cc31/types/block.go#L393) as we need to encode field by field.

ebuchman:
Replaced by #31 and #36

PR #31, opened 2019-09-08
MVP Generic Sequential & Skipping Lite Client Verifiers

ebuchman:
This is an MVP of generic lite client verifiers for the sequential and skipping (ie. bisection) cases. It defines traits for the key data types (`Header`, `Validators`, `Validator`, `Commit`, `Vote`) and exposes the minimal amount of info to express the core validation.


ebuchman:
Notes about the bisecting version:

- we should just call it "SkippingVerifier" since it skips blocks. It does not have a notion of provider or request so the bisection actually takes place outside this module
- we don't have the validator set for the commit we're verifying
- we need the votes to return validator ids and we need to be able to fetch validators by id so we can verify the votes

The latter points are good justification for keeping the address in the actual vote structure included in the commit, as per https://github.com/tendermint/tendermint/issues/3596#issuecomment-486919668. Otherwise, lite clients would have to also download the current validator set to know if the votes are from any validators they already trust.

lite/src/skipping.rs
```diff
@@ -0,0 +1,94 @@
...
+
+            // check vote is valid from validator
+            if !val.verify(vote.sign_bytes(), vote.signature()) {
+                return Err(Error::InvalidSignature);
+            }
+            signed_power += val.power();
+        }
+
+        // check the signers account for +1/3 of the voting power
+        if signed_power * 3 <= total_power * 1 {
```
ancazamfir: IIRC, the ADR had some security factor to allow to require more than "at least correct validator from old set has signed the new set".  Something like:
```suggestion
        if signed_power * 3 <= total_power * slack {
             // with 1 <= slack <= 2 
```

ebuchman: Oh good point. That should probably be configured as a parameter in the SkippingVerifier

lite/src/sequential.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn verify_commit<C>(&self, commit: C) -> Result<(), Error>
+    where
+        C: Commit,
+    {
+        let total_power = self.state.next_validators.total_power();
+        let mut signed_power: u64 = 0;
+
+        // NOTE that the vals and commit have a 1-to-1 correspondance.
+        // This means we don't need the validator IDs or to do any lookup,
+        // we can just zip the iterators.
```
ancazamfir: IMO we should still verify it, i.e. check same number/ length of vals and commits.

lite/src/skipping.rs
```diff
@@ -0,0 +1,94 @@
...
+{
+    /// Returns the time after which the current state expires
+    /// and the verifier must be reset subjectively.
+    fn expires(&self) -> Time {
+        self.state.header.bft_time() + self.trusting_period
+    }
+
+    /// Verify takes a header, a commit for the header,
+    /// and the next validator set referenced by the header.
+    /// Note we do not know the correct validator set for this commit -
```
ancazamfir: We still have a list of validators for this commit and even if we don't trust them yet IMO we should still perform the basic validation (similar to the one done for sequential). Or it can be done once we know at least one correct validator has signed.

lite/src/sequential.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Verify takes a header, a commit for the header,
+    /// and the next validator set referenced by the header.
+    /// Without knowing this next validator set, we can't
+    /// really verify the next header, so we make verifying
+    /// the current header conditional on receiving that validator set.
+    /// We could break this up into two steps, one where we verify the header,
+    /// and one where we receive and validate the next validators,
+    /// but for now we combine them. Returns an error if verification fails.
+    fn verify<C>(&self, now: Time, header: H, commit: C, next_validators: V) -> Result<(), Error>
```
milosevic: Why we need next_validators here when we have it as `TrustedState.next_validators`?

lite/src/sequential.rs
```diff
@@ -0,0 +1,102 @@
...
+            // skip absent and nil votes
+            // NOTE: do we want to check the validity of votes
+            // for nil ?
+            let vote = match vote_opt {
+                Some(v) => v,
+                None => continue,
+            };
+
+            // check vote is valid from validator
+            if !val.verify(vote.sign_bytes(), vote.signature()) {
```
milosevic: Where do we verify that signed bytes corresponds to the right BlockID?

lite/src/skipping.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Returns an error if verification fails.
+    fn verify<C>(self, now: Time, header: H, commit: C, next_validators: V) -> Result<(), Error>
+    where
+        C: Commit,
+    {
+        // ensure the state is not expired.
+        if self.expires() < now {
+            return Err(Error::Expired);
+        }
+
```
milosevic: Should we have check that `header.Height > self.state.header.height() + 1`, i.e., that we are not in sequential case?

lite/src/skipping.rs
```diff
@@ -0,0 +1,94 @@
...
+        // ensure the next validators in the header matches what was supplied.
+        if header.next_validators_hash() != next_validators.hash() {
```
milosevic: Shouldn't it be `header.validators_hash != next_validators.hash()`?

lite/src/sequential.rs
```diff
@@ -0,0 +1,102 @@
...
+            return Err(Error::NonSequentialHeight);
+        }
+
+        // ensure the validators in the header matches what we expect from our state.
+        if header.validators_hash() != self.state.next_validators.hash() {
+            return Err(Error::InvalidValidators);
+        }
+
+        // ensure the next validators in the header matches what was supplied.
+        if header.next_validators_hash() != next_validators.hash() {
```
milosevic: Shouldn't it be header.validators_hash != next_validators.hash()?

lite/src/skipping.rs
milosevic: Agree with Anca. We still need to verify that commit is valid with respect to its validator set. If +2/3 of validators from this set signed correctly and it is confirmed by +1/3 from the trusted val set, then it's fine.

lite/src/types.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/// Vote contains the data to verify a validator voted correctly in the commit.
+/// In an ideal world, votes contain only signatures, and all votes are for the same
+/// message. For now, Tendermint votes also sign over the validator's local timestamp,
+/// so each vote is for a slightly different message.
+/// Note the Vote must also know which validator it is from.
+pub trait Vote {
+    fn validator_id(&self) -> ValID;
+    fn sign_bytes(&self) -> Bytes;
```
milosevic: How do we know what is the value vote is for? We only have sign bytes.

lite/src/types.rs
```diff
@@ -0,0 +1,103 @@
...
+    type Validator: Validator;
+
+    /// Hash of the validator set.
+    fn hash(&self) -> Hash;
+
+    /// Total voting power of the set
+    fn total_power(&self) -> u64;
+
+    /// For iterating over the underlying validators.
+    fn into_vec(&self) -> Vec<Self::Validator>;
```
milosevic: Minor detail: it seems we are always using it as iterator so that might be more appropriate type, but maybe vector makes more sense for future extensions.

milosevic:
Very nice and clean.

lite/src/skipping.rs
ebuchman: Not sure I understand. According to the spec, we don't verify the commit according to +2/3 of its actual validator set, but only according to +1/3 of our last trusted validator set. You're saying we should do both ? Presumably we would do the +1/3 of our trusted set first, and if that checks out, then we would do the +2/3 of the actual set. But if the later check failed, it would mean that one of our trusted validators is misbehaving ...

Seems like we may want to go back to the spec on this, unless I'm understanding something wrong

lite/src/sequential.rs
ebuchman: `next` is confusing :(

The `trusted_state.next_validators` is with respect to `trusted_state.header`.

So if the `trusted_state.header` is H, the `trusted_state.next_validators` is for H+1. but the `next_validators` we receive in this verify method are actually for `H+2` ...

lite/src/sequential.rs
ebuchman: We check that the `commit.header_hash()` matches the hash of the header. This header is a component of the block id. It's the responsibility of the implementor of vote to ensure the sign bytes include the correct block id. I thought I could get away with not even surfacing the block id here, since the light client really only cares about the header hash, the block id is kind of an implementation detail, but i may have been a bit overzealous ...

lite/src/types.rs
ebuchman: The assumption is that when the implementor mapped their actual vote type to this trait, they only included votes for the block id, and votes for nil or other block ids were excluded.

lite/src/skipping.rs
ebuchman: Don't think so - this is about the validators for the block after this header we're verifying.

lite/src/types.rs
tarcieri: You could probably use `impl Iterator<Item=&Self::Validator>` as the return type here, then call `iter()` in an implementation

lite/src/skipping.rs
ancazamfir: the +1/3rd check is just to decide if we should trust the actual vals in skipping case. In the sequential case we don't need to do this check. But in both cases we should check +2/3rds of actual vals signed + all the other checks (hashes, proper content for header, votes, etc). 
The meaning of the <2/3rds signing should be the same for both cases.

ancazamfir: > Seems like we may want to go back to the spec on this, ...

agreed.

milosevic: It seems that's not clear enough. In my head it should be as Anca said. Hopefully TLA+ will minimise this kind of comments :)

lite/src/sequential.rs
milosevic: I am worried with the case that vote looks like correct one but it is not for the right blockID. Don't have precise scenario yet.

ebuchman: Like it's for the right header but the wrong block parts ? I am worried about it too, but in this design it's supposed to be taken care of by the code that implements the traits for real data structures - ie. it's that code's responsibility to only include votes that have the right block id, so by the time this code runs, we can assume that all votes are for the right block id.

ebuchman: Like it's for the right header but the wrong block parts ? I am worried about it too, but in this design it's supposed to be taken care of by the code that implements the traits for real data structures - ie. it's that code's responsibility to only include votes that have the right block id, so by the time this code runs, we can assume that all votes are for the right block id.

lite/src/types.rs
ebuchman: Can you provide a bit more detail here, not sure I fully understand.

If I do `fn into_iter(&self) -> Iterator<Item = &Self::Validator>;` I get 

```
the size for values of type `dyn std::iter::Iterator<Item = &<V as types::Validators>::Validator>` cannot be known at compilation time
```

Also `impl` doesn't seem to be allowed directly in the return type? Ie.

```
error[E0562]: `impl Trait` not allowed outside of function and inherent method return types
  --> lite/src/types.rs:48:28
   |
48 |     fn into_iter(&self) -> impl Iterator<Item = Self::Validator>;
```

tarcieri: Aah, my bad, guess you can't use `impl Trait` in a trait definition, in which case you'd need a concrete iterator type.

In that case, you could potentially use an associated type e.g. `type ValidatorIter`.

One note about `into_iter` as a name: that's the name used by the [`IntoIterator` trait](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html), which is for iterators that consume `self`. The convention for a borrowing iterator is `fn iter(&self) -> MyIter`

melekes:
The two main differences between `SequentialVerifier` and `SkippingVerifier` seem to be:

1. sequential height check
2. amount of voting power needed to trust a new header

Can we maybe offload (1) to the caller and make (2) a parameter, so that we can merge `SequentialVerifier` and `SkippingVerifier` into `Verifier`

lite/src/skipping.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        // ensure the commit matches the header.
+        if header.hash() != commit.header_hash() {
+            return Err(Error::InvalidCommitValue);
+        }
+
+        // ensure that +1/3 of the trusted validators signed correctly
+        self.verify_commit(commit)
+    }
```
melekes: Shouldn't we update the trusted state here? Otherwise a caller will have to recreate a Verifier with each new header

lite/src/sequential.rs
Liamsi: Can we find a better name for `next_validators` s.t. it's less confusing then? Similarly with `header`. 
Maybe `next_header` and `next_validators_for_header` (if that's right). That would indicate that: we are expecting the next header in the sequence.

ebuchman:
Ok, some significant refactoring:

- No more state. Just pure functions that return errors.
- Focus here is now purely on "verifying commits". So anything to do with "last header" and "next validators" has been moved out. This means that:
  - Callers are responsible for passing sequential headers to the `verify` function and increasing but not necessarily sequential headers to the `verify_trusting`.
  - Callers are responsible for getting and checking `next_validators` before verifying the next header.
- Sequential and skipping are mostly consolidated.

lite/src/skipping.rs
ebuchman: Actually how should we handle the `trustlevel` here? Do we define an enum and give users a few levels (eg. 1/3, 1/2, 2/3), or should we let them pass in any integer, and then scale the 3 accordingly?

lite/src/sequential.rs
ebuchman: Done.

lite/src/skipping.rs
ebuchman: This is obviously necessary lol. Done.

lite/src/sequential.rs
ebuchman: should be clearer now

lite/src/verifier.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // The vals and commit have a 1-to-1 correspondance.
+    // This means we don't need the validator IDs or to do any lookup,
+    // we can just zip the iterators.
+    let vals_iter = vals_vec.into_iter();
+    let commit_iter = commit_vec.into_iter();
+    for (val, vote_opt) in vals_iter.zip(commit_iter) {
+        // skip absent and nil votes
+        // NOTE: do we want to check the validity of votes
```
melekes: if we're going to punish for them, then yes?

lite/src/verifier.rs
```diff
@@ -0,0 +1,181 @@
...
+    H: Header,
+    V: ValidatorSetLookup,
+    C: Commit,
+{
+    if let Err(e) = validate_vals_and_commit(header, &commit, &validators) {
+        return Err(e);
+    }
+
+    // ensure that +1/3 of last trusted validators signed correctly
+    if let Err(e) = verify_commit_trusting(&validators, &commit) {
```
melekes:
```suggestion
    if let Err(e) = verify_commit_trusting(&last_validators, &commit) {
```

lite/src/verifier.rs
```diff
@@ -0,0 +1,181 @@
...
+    commit: C,
+    last_validators: V,
+    validators: V,
+) -> Result<(), Error>
+where
+    H: Header,
+    V: ValidatorSetLookup,
+    C: Commit,
+{
+    if let Err(e) = validate_vals_and_commit(header, &commit, &validators) {
```
ancazamfir:
```suggestion
    if let Err(e) = validate(header, &commit, &validators) {
```
and remove call to `verify_commit_full()` below?

lite/src/types.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// Vote contains the data to verify a validator voted correctly in the commit.
+/// In an ideal world, votes contain only signatures, and all votes are for the same
+/// message. For now, Tendermint votes also sign over the validator's local timestamp,
+/// so each vote is for a slightly different message.
+/// Note the Vote must also know which validator it is from.
+pub trait Vote {
+    fn validator_id(&self) -> ValID;
+    fn sign_bytes(&self) -> Bytes;
```
Liamsi: Shouldn't this sign_bytes method take a `chain_id` too? Or is the implementer supposed to have this in the trait implementation? If so, should the trait also include a `fn chain_id(&self) -> Id` method to make this more explicit?

Liamsi: Had a quick convo with @ebuchman: we probably want to keep the abstractions and here as general as possible. Ideally, they should expose as little implementation details as possible. E.g. the traits should also work for a lite client that only follows a single chain. Hence, the trait should not need to know about this idea.

lite/src/types.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub last_header: H, // height H-1
+    pub validators: V,  // height H
+}
+
+/// Need to do something better here :)
+pub type Height = u64;
+pub type Hash = u64; // TODO
+pub type Time = u64; // TODO
+pub type Bytes = u64; // TODO
+pub type ValID = u64; // TODO
```
Liamsi: If I understand correctly, we still want to figure out what these types should be. As the rest of this package is purely traits, should these types should probably be as general as possible, too. (meaning should they also be traits only describing the expected behaviour and no concrete implementation)? 

Here are some suggestions:

- `ValID` could also be a trait with one method `fn id(&self) -> &[u8];` (`as_bytes` is probably more idiomatic) and the size / length of the id  `fn len() -> usize` (currently it's always 20)
- `Height` could also be a similar trait: so far it should only come with one method: `fn value(self) -> u64`
- `Bytes` 🤔 should either be dropped and `sign_bytes` could directly return `&[u8]`, or the name should be different/more specific and called sth like `Signable` or simply `AsBytes` if we do not want to express that this will be signed; It should also have a single method `as_bytes` or `value`
- similar with `Hash`
- `Time` solely seems to be used to determine if a header is expired for that it seem there are two different concepts: that of a duration and that of a point/instant in time; a trait could come with the necessary subset of Instant and Duration of the [std::time](https://doc.rust-lang.org/std/time/index.html) module (?)

Liamsi: OK, these can be concrete instead.

lite/src/verifier.rs
ebuchman: Why these changes ?

ancazamfir: `validate()` already calls `validate_vals_and_commit()` and `verify_commit_full()`. 
Also to reflect that all checks done by sequential client are also done by the skipping one.

ebuchman: Ah, you mean `verify` not `validate` ?

I agree this might be nicer, but also there's a performance trade off - for the skipping ("trusting") case, I don't think we want to bother verifying the full commit until we verify that +1/3 of the old validator set signed. If we had it just call `verify` first, then we'd verify the full commit before we even discover if we can trust it according to the trusting method.

lite/src/verifier.rs
ebuchman: We can't necessarily slash for them, because if they're not valid we don't know who signed them. We could "punish" peers if they send us commits with bad nil votes by just disconnecting from them, but it's not clear this actually poses any real risk to us and is worth doing since it seems we can just ignore them.

lite/src/types.rs
ebuchman: Add this to the comments

lite/src/sequential.rs
ebuchman: Added a note about this in the Vote comments

lite/src/types.rs
ebuchman: Wow that worked! Thanks! https://github.com/interchainio/tendermint-rs/pull/31/commits/f521f048d6a25b4ffd0efcf2c788ba17d7dcf95d

cc @spacejam

ebuchman: Eek this is probably going to over complicate implementations - I haven't figured out yet how a concrete type (eg. the `validator::Set`) can actually implement this without a trait object. It seems we have to implement ExactSizedIterator for it, but haven't made that work easily yet.

lite/src/verifier.rs
ancazamfir: > Ah, you mean `verify` not `validate` ?

Yes, sorry I meant `verify()`.

> I agree this might be nicer, but also there's a performance trade off - for the skipping ("trusting") case, I don't think we want to bother verifying the full commit until we verify that +1/3 of the old validator set signed. If we had it just call `verify` first, then we'd verify the full commit before we even discover if we can trust it according to the trusting method.

I think the complexity is same for both. But to address your concern, can't we call `verify_commit_trusting` follow by `verify`?

lite/src/skipping.rs
josef-widder: If we give a fixed selection of trust levels we could test / model check all the levels and give more complete guarantees. Perhaps we should also allow trust level of 1 (all validators have signed). I will have to adapt the English spec to allow that. (Right now there is a '>' and I should change that to ">=" with a +1 term)

ebuchman:
After starting to work on tests I found myself essentially implementing a lot of types and functionality that already exists in the tendermint-rs and in #36, so starting to think that means we will actually test this using the real types and it's not worth it to duplicate all that in mocks. Not sure really how to interpret this yet, whether it means the the abstractions here aren't as useful as hoped or if they're maybe not abstract enough. The mocks would be a bit lighter than the real types, but it's not clear the effort of implementing them is really worth it ... would love more opinion.

So I would vote to merge this as is without tests (unless there are still unaddressed review comments but I think I got everything) and to work on the tests along with the trait implementations in #36.

Liamsi:
@tarcieri: Do you usually merge or squash PRs here? The commit history on master looks like a mix of both.

tarcieri:
@Liamsi I personally prefer merge commits, because they give you a cryptographically authenticated history of how code landed on `master`

Liamsi:
Thanks @tarcieri. That makes sense. The team seems in favour of squashing though (like in the golang tendermint repo). And I think there is also merit in having a cleaner history on master. The original commit history is still around in the Pull request. Well ... as long as github is around I guess :-P 

In any case I'm in favour of consistency. Will stick to squash merging then.

ebuchman:
> cryptographically authenticated

With SHA1. And trusting github. The git model is so broken it hurts my brain: https://ebuchman.github.io/posts/2019-01-26-software-collab-part1/ . So I air on the side of clean commit history - it's especially nice to have every commit correspond to just one PR so you can see how everything happened, and otherwise restrict merging to master so anyone can see that each merge got the approvals if they want to. Would love to see more tooling developed around authenticating that merges happened properly and that squash merges correspond to the same patches as the set of commits, but alas, we're not there yet.

tarcieri:
> With SHA1.

[Yup](https://www.reddit.com/r/crypto/comments/e1krjh/gnupg_uses_by_default_sha1_20_bytes_output_and/f8r5wtr/) 😜

PR #36, opened 2019-09-18
Light client: trait impls

Liamsi:
This PR will have implementations of the traits created in #31 and a few tests.


tendermint/src/block/header.rs
```diff
@@ -70,6 +71,106 @@ pub struct Header {
...
+        byteslices.push(data_hash_enc.as_slice());
+        byteslices.push(validator_hash_enc.as_slice());
+        byteslices.push(next_validator_hash_enc.as_slice());
+        byteslices.push(consensus_hash_enc.as_slice());
+        byteslices.push(app_hash_enc.as_slice());
+        byteslices.push(last_result_hash_enc.as_slice());
+        byteslices.push(evidence_hash_enc.as_slice());
+        byteslices.push(proposer_address_enc.as_slice());
+
+        Hash::Sha256(simple_hash_from_byte_slices(byteslices.as_slice()))
```
Liamsi: This looks a bit too exhaustive but aims to do exactly the same as: https://github.com/tendermint/tendermint/blob/134fe2896275bb926b49743c1e25493f6b24cc31/types/block.go#L393

This could probably be simplified substantially. Maybe, `simple_hash_from_byte_slices` had a similar logic as a Hasher in golang (`Write` the slices and then `Sum` to create the tree once and return the hash)?

tendermint/src/validator.rs
```diff
@@ -30,14 +35,37 @@ impl Set {
...
+        ))
+    }
+
+    fn total_power(&self) -> u64 {
+        self.validators.iter().fold(0u64, |total, val_info| {
+            total + val_info.voting_power.value()
+        })
+    }
+
+    fn into_vec(&self) -> Vec<Self::Validator> {
```
ebuchman: Tried changing this to `iter` in the original traits. It compiles there, but not sure how to implement it yet. See https://github.com/interchainio/tendermint-rs/pull/31#discussion_r330849769.

Maybe should just revert the change since it might not be worth it for now, unless you have an easy way to deal with this.

tendermint/src/block/header.rs
ebuchman: Definitely seems excessive.

Good idea re a write based API. Though I don't think you usually see that for Merkle trees? At least the Merkle tree version needs to track the boundaries of each write, where as a normal hasher just concatenates everything together.

tendermint/src/lite/types.rs
```diff
@@ -76,6 +75,8 @@ pub trait Commit {
     /// we ignore absent votes and votes for nil here.
     /// NOTE: we may want to check signatures for nil votes,
     /// and thus use an ternary enum here instead of the binary Option.
+    // TODO figure out if we want/can do an iter() method here that returns a
+    // VoteIterator instead of returning a vec
```
ebuchman: See https://github.com/interchainio/tendermint-rs/pull/31#discussion_r330845421

tendermint/src/validator.rs
ebuchman: reverted

tendermint/src/lite/types.rs
ebuchman: reverted

tendermint/src/vote.rs
```diff
@@ -67,6 +71,31 @@ impl Vote {
...
+impl lite::Vote for Vote {
+    fn validator_id(&self) -> account::Id {
+        self.validator_address
+    }
+
+    fn sign_bytes<B>(&self, mut sign_bytes: &mut B)
+    where
+        B: BufMut,
+    {
+        // TODO: 1) everytime we encode sth. an error can occur. Change the trait to return a result
```
ebuchman: Why should we ever have an error here? Assuming we control what's being encoded. And how would we handle it gracefully? Again, seems there might not be a choice but to panic.

tendermint/src/vote.rs
```diff
@@ -67,6 +71,31 @@ impl Vote {
...
+        // instead to enable proper error handling.
+        // 2) Figure out where the chain_id should come from (if sign_bytes remains on Vote, the
```
ebuchman: Good point - may have been thinking we'd use something else here, eg. the CanonicalVote, which contains the ChainID

tendermint/src/vote.rs
```diff
@@ -92,6 +121,11 @@ impl Type {
     pub fn to_u8(self) -> u8 {
         self as u8
     }
+
+    /// Serialize this type as a 32-bit unsigned integer
+    pub fn to_u32(self) -> u32 {
```
ebuchman: Added this because clippy didn't like `to_u8() as u32`

tendermint/src/amino_types/vote.rs
```diff
@@ -49,6 +51,30 @@ impl Vote {
     }
 }
 
+impl From<&vote::Vote> for Vote {
+    fn from(vote: &vote::Vote) -> Self {
+        Vote {
+            vote_type: vote.vote_type.to_u32(),
+            height: vote.height.value() as i64, // TODO potential overflow :-/
+            block_id: Some(BlockId {
```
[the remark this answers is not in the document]
Liamsi: Hmm, I don't remember tbh. These amino types are quite old. Let me check if the encoding matches if remove the `Option`.

tendermint/src/amino_types/vote.rs
```diff
@@ -49,6 +51,30 @@ impl Vote {
...
+                hash: match vote.block_id.hash {
+                    Hash::Sha256(h) => h.to_vec(),
+                    _ => vec![],
+                },
+                parts_header: match &vote.block_id.parts {
+                    Some(parts) => Some(PartsSetHeader::from(parts)),
+                    None => None,
+                },
+            }),
+            round: vote.round as i64,
```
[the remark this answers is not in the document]
Liamsi: Probably not.

tendermint/src/amino_types/vote.rs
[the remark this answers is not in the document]
Liamsi: So, it looks like one can simply add the `#[prost(message)]` for a nested struct if it is an `Option`. Otherwise, it seems a little more effort is necessary. e.g. I locally changed the field's type to `BlockId` instead of `Option<BlockId>`. Then the compiler (due to prost) will complain:

```
error[E0599]: no method named `as_ref` found for type `amino_types::block_id::BlockId` in the current scope
  --> tendermint/src/amino_types/vote.rs:22:28
   |
22 | #[derive(Clone, PartialEq, Message)]
   |                            ^^^^^^^ method not found in `amino_types::block_id::BlockId`
   | 
  ::: tendermint/src/amino_types/block_id.rs:11:1
   |
11 | pub struct BlockId {
   | ------------------ method `as_ref` not found for this
   |
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `as_ref`, perhaps you need to implement it:
           candidate #1: `std::convert::AsRef`

error[E0308]: mismatched types
  --> tendermint/src/amino_types/vote.rs:22:28
   |
22 | #[derive(Clone, PartialEq, Message)]
   |                            ^^^^^^^
   |                            |
   |                            expected struct `amino_types::block_id::BlockId`, found enum `std::option::Option`
   |                            this match expression has type `amino_types::block_id::BlockId`
   |
   = note: expected type `amino_types::block_id::BlockId`
              found type `std::option::Option<_>`

error[E0599]: no method named `get_or_insert_with` found for type `amino_types::block_id::BlockId` in the current scope
  --> tendermint/src/amino_types/vote.rs:22:28
   |
22 | #[derive(Clone, PartialEq, Message)]
   |                            ^^^^^^^ method not found in `amino_types::block_id::BlockId`
   | 
  ::: tendermint/src/amino_types/block_id.rs:11:1
   |
11 | pub struct BlockId {
   | ------------------ method `get_or_insert_with` not found for this

error[E0308]: mismatched types
  --> tendermint/src/amino_types/vote.rs:22:28
   |
22 | #[derive(Clone, PartialEq, Message)]
   |                            ^^^^^^^ expected struct `amino_types::block_id::BlockId`, found enum `std::option::Option`
   |
   = note: expected type `amino_types::block_id::BlockId`
              found type `std::option::Option<_>`

error: aborting due to 4 previous errors

Some errors have detailed explanations: E0308, E0599.
For more information about an error, try `rustc --explain E0308`.
error: Could not compile `tendermint`.
´´´

File: tendermint/src/block/header.rs [hunk 9]
```diff
@@ -83,11 +85,89 @@ impl lite::Header for Header {
...
     }
 
     fn hash(&self) -> Hash {
-        unimplemented!()
+        let mut version_enc = vec![];
+        // TODO: if there is an encoding problem this will
+        // panic (as the golang code would):
+        // https://github.com/tendermint/tendermint/blob/134fe2896275bb926b49743c1e25493f6b24cc31/types/block.go#L393
+        // https://github.com/tendermint/tendermint/blob/134fe2896275bb926b49743c1e25493f6b24cc31/types/encoding_helper.go#L9:6
+        // Instead, handle errors gracefully here.
```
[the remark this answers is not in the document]
Liamsi: > Why would there ever be an encoding error and what would it mean to handle it gracefully? 

I can't think of a scenario where there could be and encoding error here (while encoding for  hashing). If there is any case, it still feels wrong to me to panic instead of returning an error and let the caller decide if it is correct to panic (or handle the error in another way). 
If we can not hash the header, this probably means a bug in the encoding library (or another bigger problem) and the caller should probably decide to panic anyways. I just feel it isn't the job of the `hash` method to decide.

tendermint/src/amino_types/vote.rs
```diff
@@ -49,6 +51,30 @@ impl Vote {
...
+            timestamp: Some(TimeMsg::from(vote.timestamp)),
```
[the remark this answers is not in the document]
Liamsi: See above.

tendermint/src/amino_types/vote.rs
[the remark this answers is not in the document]
Liamsi: OK, it looks like this can be fixed by adding in a required attribute: https://github.com/danburkert/prost/blob/9551f2852e414df72339634087c46ce5a08e85b2/prost-derive/src/field/message.rs#L14-L56

I'm looking into that.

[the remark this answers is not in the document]
Liamsi: Adding the `required` attribute, e.g.:
```rust
#[prost(message, required, tag = "4")]
pub block_id: BlockId,
```
makes the compiler happy 😄but breaks some encoding test-vectors 🤔 

```
thread 'amino_types::vote::tests::test_sign_bytes_compatibility' panicked at 'assertion failed: `(left == right)`
  left: `[15, 34, 0, 42, 11, 8, 128, 146, 184, 195, 152, 254, 255, 255, 255, 1]`,
 right: `[13, 42, 11, 8, 128, 146, 184, 195, 152, 254, 255, 255, 255, 1]`', tendermint/src/amino_types/vote.rs:317:9
```
Hmm, this feels somehow familiar. Will investigate further.

tendermint/src/vote.rs
Liamsi: Thanks!

tendermint/src/amino_types/vote.rs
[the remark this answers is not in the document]
Liamsi: OK, here is the catch:

In golang one can simply not initalize a field. Golang won't complain. And amino will treat this as sth. that can be skipped (and hence is optional in the protobuf sense). 

Now on the rust side you have to init a field (unless you make it an option). If you make it an Option, the encoding behaves like amino (None will be not encoded at all). If you have a regular field (e.g. BlockId), and add a `required` attribute, prost(-amino) will treat as `required` (in the protobuf sense) and signal, we got fiel blah of type blubb but nothing in there (for blockId this is fieldnumber 4 and hence `4 << 3 | 2 = 32` followed by a `0x00` for indicating that this was empty. 
Amino actually does the same but rolls back if it found out there is nothing (via that zero 0x00) unless told otherwise via `writeEmpty` (at least it seems to work [now](https://github.com/tendermint/go-amino/pull/287)):
https://github.com/tendermint/go-amino/blob/fbf776258498dbb3aa6637ca82c7fe5c7255fd2f/binary-encode.go#L496-L524

Alt Text

tendermint/src/lite/types.rs
```diff
@@ -96,7 +99,9 @@ pub trait Commit {
 /// is only necessary to avoid slashing in the multi chain context.
 pub trait Vote {
     fn validator_id(&self) -> Id;
-    fn sign_bytes(&self) -> &[u8];
+    fn sign_bytes<B>(&self, sign_bytes: &mut B)
+    where
+        B: BufMut;
```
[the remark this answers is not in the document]
Liamsi: Maybe I did run into issues regarding the lifetime. I'll try to see if I can simply return ` &[u8]` as previously.

tendermint/src/block/header.rs
```diff
@@ -111,3 +191,10 @@ pub struct Version {
     )]
     pub app: u64,
 }
+
+fn encode_bytes(bytes: &[u8]) -> Vec<u8> {
+    let mut chain_id_enc = vec![];
```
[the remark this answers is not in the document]
Liamsi: All other encoding methods are named `encode_*`. But it is a private helper function. Happy to rename it.

tendermint/src/block/header.rs
```diff
@@ -83,11 +85,89 @@ impl lite::Header for Header {
...
+        let mut num_tx_enc = vec![];
+        prost_amino::encoding::encode_varint(self.num_txs, &mut num_tx_enc);
+        let mut total_tx_enc = vec![];
+        prost_amino::encoding::encode_varint(self.total_txs, &mut total_tx_enc);
+        let mut last_block_id_enc = vec![];
+        amino_types::BlockId::from(&self.last_block_id)
+            .encode(&mut last_block_id_enc)
+            .unwrap();
+        let mut last_commit_hash_enc = vec![];
+        if let Some(last_commit_hash_bytes) = self.last_commit_hash.as_bytes() {
```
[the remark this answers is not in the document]
Liamsi: Done (there is a helper `encode_hash` now).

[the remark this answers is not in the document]
Liamsi: Done (there is a helper encode_hash now).

tendermint/src/vote.rs
Liamsi: Same reason as above. Yes, probably we should panic if can't encode this but I would prefer to leave it to the caller to decide (especially as this aims to be a generic trait). But maybe I'm overthinking this.

tendermint/src/lite/types.rs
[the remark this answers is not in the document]
Liamsi: Actually, the problem is that the slice we want return here is created from a temp value inside that  method (instead of sth borrowed from parameters of that method, in this case only `self`). e.g. this does not work:

```rust
    fn sign_bytes(&self) -> &[u8] {
        let mut sign_bytes = vec![];
        CanonicalVote::new(AminoVote::from(self), "TODO")
            .encode(&mut sign_bytes)
            .unwrap();
        sign_bytes.as_slice()
    }
```

On simple way to returning something would be to return a `Vec<u8>` instead of `&[u8]`. As it isn't reference this should work without any problems.
A much more complex alternative might be to add a field that caches the the sign bytes. Not sure the latter works or would be worth the effort (?).

tendermint/src/vote.rs
ebuchman: Can we wrap the CanonicalVote in something that also contains the ID and the signature?

Liamsi: I think that makes sense. Will create a separate struct tomorrow. Feel free to merge this (if you want), I won't forget this change.

tendermint/src/block/header.rs
Liamsi: Note to myself: add back the test that checks the hash here:
https://github.com/interchainio/tendermint-rs/blob/262e252a04cb7d7b660a829f3240e7196ad955b9/tendermint/tests/rpc.rs#L167-L168

Liamsi: done in https://github.com/interchainio/tendermint-rs/pull/36/commits/e686c457e2913c11737e65305a146a216dc66729

tendermint/src/vote.rs
Liamsi: Done in c941c95. Name of the new wrapping type `SignedVote`.

Liamsi:
Note that @yihuang added and impl for lite::Commit which I've merged (and then slightly modified) in #63.

tendermint/src/validator.rs
```diff
@@ -36,6 +41,31 @@ impl Set {
     }
 }
 
+impl lite::ValidatorSet for Set {
+    type Validator = Info;
+
+    fn hash(&self) -> Hash {
+        // TODO almost the same as above's pub fn hash(self) -> merkle::Hash
```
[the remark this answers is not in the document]
Liamsi: Yes, we can actually just delete the `pub fn hash(self) -> merkle::Hash` from `impl Set`. As far as I can see, it isn't used anywhere.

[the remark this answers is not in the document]
Liamsi: Added in: 23042bb8d31439983d09d19cf28a62cf8062bf0a

Liamsi:
@ebuchman Can we merge this soon?

ebuchman:
@Liamsi I'm unfortunately out of commission this week so will leave it to you. Thanks @yihuang for your contributions, much appreciated!!

Liamsi:
OK, thanks. I'll merge in a bunch of (reviewed) changes from #63 into `lite_impl`. Will add more tests and if everything looks OK, merge all this into your branch `bucky/lite`. Think we should still do another review on `bucky/lite` -> `master`. Then we have a proper (skipping) light client to use.

Liamsi:
Actually, no. Will merge your intial traits PR #31 to master and base the follow-up PRs (this one #36 and #63) to master too.

tendermint/src/lite/verifier.rs
```diff
@@ -127,7 +127,8 @@ where
         };
 
         // check vote is valid from validator
-        if !val.verify_signature(vote.sign_bytes(), vote.signature()) {
+        let sign_bytes = vote.sign_bytes();
+        if !val.verify_signature(&sign_bytes, vote.signature()) {
```
ancazamfir: Lines 121-133 are duplicated in verify_commit_trusting() and here. Maybe a function that takes a validator and a vote and returns error be useful to both? I know in verify_commit_trusting() we do an extra check but that could be done before or after.

tendermint/src/vote.rs
```diff
@@ -67,6 +69,47 @@ impl Vote {
     }
 }
 
+/// SignedVote is the union of a canoncialized vote, the signature on
```
ancazamfir:
```suggestion
/// SignedVote is the union of a canonicalized vote, the signature on
```

tendermint/src/lite/verifier.rs
Liamsi: Good catch. There is a subtle difference in those lines though: in one case the validator is looked up in the set in the other it part of the looped elements (also some of the lines are part of the control flow via `None => continue,`).  This could be deduplicated (completely) if we do not `zip` the iterators and do a lookup in both cases. I'll leave this as is for now.

ancazamfir: > There is a subtle difference in those lines though: in one case the validator is looked up in the set in the other it part of the looped elements (also some of the lines are part of the control flow via None => continue,). 

Yes, I saw that but was thinking that once there is a validator and a vote the checks should be the same (most the loop body). But I looked closer and it's not much left in common to justify a separate function.

Another question, for `verify_commit_full()` we do assume: `The vals and commit have a 1-to-1 correspondence`; is this checked elsewhere? I was looking for a check that verifies that `val.address == vote.validator_address`. Or `val.verify_signature()` will take care of it?

ancazamfir: Also, for `vals_iter.zip(commit_iter)`, I believe this works as we want as long as there is no nil validator in the set (I think this is the case but I have a hard time navigating the Rust code :(  ) ...otherwise according to the doc: ` If the first iterator returns None, zip will short-circuit and next will not be called on the second iterator.`

Liamsi:
Thanks for the review @brapse and @ancazamfir! I would also like to merge this as is. A few tests are in a follow-up PR #63 which I didn't merge in here as this PR would grow quite large.

tendermint/src/lite/verifier.rs
Liamsi: The only check is that length matches. Also AFAIR the commits / validators are sorted. That means if the the len matches there is the desired correspondence. But I'll double check and follow up in #63. It seems this is more confusing than helpful maybe we should deduplicate and do the additional lookups instead.

Issue #41, opened 2019-09-23
RPC endpoints: compatibility tests are not running


Liamsi:
I was looking into the RPC endpoints and noticed that the integration tests are not running anymore. 

I think that is because the [`tests`](https://github.com/interchainio/tendermint-rs/tree/master/tests) directory got moved out of the module and is now a separate directory. Moving them back into a module (e.g. the `tendermint` directory fixes this). Alternatively, `tests` could be a separate crate referred to by the root Cargo.toml.

(related #38 #4)

tarcieri:
@Liamsi the RPC integration tests that require a live `gaiad` instance are all marked as `#[ignore]` as a live `gaiad` instance may or may not be available. See notes here:

https://github.com/interchainio/tendermint-rs/blob/master/tests/integration.rs

tl;dr:

```
cargo test -- --ignored
```

Liamsi:
Yes, noticed that too. But also the non-ignored fixture-based tests do not run; e.g. https://github.com/interchainio/tendermint-rs/blob/06eab13d173e51c20464104e8c6cbe6b4cffe2f9/tests/rpc.rs#L161-L166

Or is this also on purpose?

tarcieri:
Aah, yes you're right, `tests` should get moved to `tendermint/tests`

Liamsi:
OK, thanks! Will include that in the PR and will push in a sec. 

We could instead also make `tests` a separate crate. Similarly to what prost does here:
https://github.com/danburkert/prost/tree/master/tests
and
https://github.com/danburkert/prost/blob/9551f2852e414df72339634087c46ce5a08e85b2/Cargo.toml#L26
But we can discuss this separately or do so when `tests` grows larger.

Liamsi:
Resolved via #42

PR #42, opened 2019-09-23
SignedHeader should contain commit

### LINKED ISSUE #38: RPC incompatibility: SignedHeader does not contain Commit
The struct for the /commit rpc response is currently different from the go-lang implementation. This could be on purpose: commits were or are about to be refactored ([to consist of only signatures](https://github.com/tendermint/tendermint/issues/1648)). 
The light client will talk to the rpc endpoint exposed by tendermint. Hence, the types should now match the go implementation (or introduce a new compatible type exclusively for the light client if desired).

Current `SignedHeader` in Rust:
https://github.com/interchainio/tendermint-rs/blob/5844dc5c0faa537b4225c5221cd863285b97d8ce/tendermint/src/rpc/endpoint/commit.rs#L44-L47

`SignedHeader` in golang:
https://github.com/tendermint/tendermint/blob/51dc810d041eaac78320adc6d53ad8b160b06601/types/block.go#L708-L713

encapsulated in the `ResultCommit ` response:
https://github.com/tendermint/tendermint/blob/51dc810d041eaac78320adc6d53ad8b160b06601/rpc/core/types/responses.go#L34-L37

by this endpoint:
https://github.com/tendermint/tendermint/blob/51dc810d041eaac78320adc6d53ad8b160b06601/rpc/core/blocks.go#L317

ref: #31 #36

Liamsi:
resolves #38 and #41


tendermint/src/block/commit.rs
```diff
@@ -4,11 +4,12 @@ use crate::{block, Vote};
 use serde::{Deserialize, Serialize};
 use std::{ops::Deref, slice};
 
-/// Last commit to a particular blockchain: +2/3 precommit signatures.
+/// Commit contains the evidence that a block was committed by a set of validators.
```
ebuchman:
```suggestion
/// Commit contains the justification (ie. a set of signatures) that a block was committed by a set of validators.
```

Best not to use `evidence` since it has another meaning within blocks.

PR #54, opened 2019-10-30
adr-002 lite client

ebuchman:
Lite client ADR. A bit backwards because we've already got WIPs for the implementation, but trying to get this written up before we make much more progress there.

Link for nicer reading: https://github.com/interchainio/tendermint-rs/blob/bucky/adr-lite/docs/architecture/adr-002-lite-client.md


docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,439 @@
...
+
+We take up the components of the diagram in term.
+
+### State
+
+The lite node state contains the following:
+
+- current height (H)
+- last header (H-1)
+- current validators (H)
```
ValarDragon: Current validator pubkeys, or all validator pubkeys and weights?

(I think it suffices for them to have >2/3 + 1 of validator pubkeys and weights, though the lite client then has to track their stake updates & and total stake.)

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,439 @@
...
+
+It also includes some configuration, which contains:
+
+- trusting period
+- initial list of full nodes
+- method (sequential or skipping)
+- trust level (if method==skipping)
+
+The node is initialied with a trusted header for some height H-1
+(call this header[H-1]), and a validator set for height H (call this vals[H]).
```
ValarDragon: It only actually needs to be initialized with the validator set hash, since it can then get the validator data from the network. (Unless this is more difficult in implementation)

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,439 @@
...
+If (3) returns a conflicting header, verify the header by requesting the
+corresponding commit and running the verification of (2). If the verification
+passes, there is a fork, and evidence should be published so the validators get
+slashed. We leave the mechanics of evidence to a future document. For now, the
+lite client will just log an error and exit. If the verification fails, it
+means the backup that provided the conflict is bad and should be removed.
+
+#### Skipping Sync
+
+Skipping sync is essentially the same as sequential, except for a few points:
```
ValarDragon: Is there a spec / security analysis I can read for this? It seems unsafe to me without full node updates or validity proofs. (SNARKs)

ebuchman: spec is here, should have linked to it properly: https://github.com/tendermint/tendermint/blob/master/docs/spec/consensus/light-client.md

docs/architecture/adr-002-lite-client.md
ebuchman: true enough - we could take it even further and say its just initialized with a height and header hash, and everything is fetched from a full node. that would be the most light weight, but still before anything can be started, we'd need to fetch the header and the validators, and we could call that fully populated state "initialization". so I guess it's a difference between what the user has to pass in for initialization and what state has to be initialized before the protocol can really start

docs/architecture/adr-002-lite-client.md
ebuchman: We need all pubkeys and weights because we dont know in advance which ones will be signing the commit

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,439 @@
...
+
+Most of the lite client logic resides in the verifier, where commits for a
+header are actually verified. In order to abstract over all the data in the
+Tendermint data structures, we use a set of traits that include only the
+information absolutely necessary for the lite client. From this perspective, we
+have the following traits.
+
+#### Header
+
+A Header has only a height, time, and validator sets, and can be identified by its hash:
```
ValarDragon: This looks way cleaner than before!

Doesn't it also need one hash for all state machine data?

ebuchman: Yes but that's not relevant to the Tendermint side of the lite client protcol - this is really just about verifying updates to the validator set and the new headers they sign. once those updates are trusted, then you can look inside at app hashes to verify state queries

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,439 @@
...
+However, this turns out to be much more difficult than expected due to the lack
+of size information in the `Iter` trait. It's possible to use another associated
+type, eg. `type ValidatorIter: ExactSizeIterator<Item = Self::Validator>;`, but it just
+complicates things significantly. If this becomes a performance issue (ie. we
+end up having to copy the vector of validators), we can address it then.
+
+#### Commit
+
+A commit is a collection of votes that we will want to iterate over so we can
+verify each vote.
```
ValarDragon: Can we make the API s.t. we can upgrade to aggregate signatures w/o an API change in the lite client?

Essentially make one function that goes from validator index to boolean representing "has signed", and one aggregate signature struct. Verify checks that 2/3rds of stake has signed, and that aggregate signature.verify(<signed pubkeys>) verifies.

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,439 @@
...
+- Simple peering strategy
+- Clear separation between verification logic and the actual data types
+
+### Negative
+
+- Abstract traits requires more coding, more opportunity for bugs in trait implementation
+
+### Neutral
+
+- Certain validity checks are ommitted since they have little bearing
```
ValarDragon: Out of curuosity, what is being omitted?

Is it just checking that the proposer & round were valid? (If so, totally agreed with that being unnecessary)

docs/architecture/adr-002-lite-client.md
ebuchman: I thot about this a bit. I feel it might be a bit too early for that. This thing is already pretty abstract and I'm not sure it will pay yet to make it even more so.

Also, come to think of it, will aggregate sigs work with the skipping method? If I have an aggregate signature, can I check if +1/3 of some other validator set (not necessarily the validator set that produced the sig) signed it?

docs/architecture/adr-002-lite-client.md
ebuchman: I outlined a few points in the `Validation` section above. Didn't mention round and proposer, but those are additional things that aren't checked. Mentioned things like LastCommitHash, BlockID.PartsHeader, and Time

docs/architecture/adr-002-lite-client.md
ValarDragon: Oh I see where I was confused, this is defining a trait not a struct. Perhaps we can change this sentence to the following:

```
A Header must contain a height, time, and validator sets. It can be uniquely identified by its hash:
```

(The core contribution being removing the word only)

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+
+    fn header_hash(&self) -> Hash;
+    fn into_vec(&self) -> Vec<Option<Self::Vote>>;
+}
+```
+
+Note the `Option` here. When it is `None`, it indicates that either:
+
+- there was no vote from this validator
+- the validator voted nil
```
Liamsi: A link to what "voted nil" means would be helpful if someone wanted to use this library, is mainly a Rust dev (who doesn't often deal with `nil`s), and not familiar with this edge case in the tendermint side? Looks like the best description is actually also in adr25?

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+depiction of what is required for lite client verification, and surfaces certain
+design issues in the underlying Tendermint blockchain (eg. the `BlockID` issue
+referenced above).
+
+This abstraction may also facilitate test generation, as we will not need to
+generate complete Tendermint data structures to test the lite client logic, only
+the elements it cares. While this provides a lot of flexibility in mocking out
+the types, we must be careful to ensure they match the semantics of the actual
+Tendermint types, and that we still test the verification logic sufficiently for
+the actual types.
```
Liamsi: Clearly describing the rationale behind the decision to keep things abstract (and rather work with traits) 👍

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+bytes that are being signed. When those message bytes and the signature are passed to 
+`validator.verify_signature` for the corresponding validator, it should return true:
+
+```rust
+pub trait Vote {
+    fn validator_id(&self) -> ValID;
+    fn sign_bytes(&self) -> &[u8];
+    fn signature(&self) -> &[u8];
+}
+```
```
Liamsi: Do we need to add a comment here how sign_bytes can work without having a chainId parameter?

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+
+Some things are left explicitly unvalidated as they have minimal bearing on the correctness of the lite client.
+These include:
+
+- LastCommitHash
+	- In the skipping case, it's not possible to verify the header refers to the correct previous block without reverting to the sequential case. So in the sequential case, we don't validate this either. If it's incorrect, in indicates the validators are misbehaving, though we can only detect it as a lite client if there's a fork.
+- BlockID
+	- As mentioned, this includes a merkle root of the entire block and is not verifiable without downloading the whole block, which would defeat the purpose of the lite client!
+- Time
+	- Verifying the time would require us to download the commit for the previous block, and to take the median of the timestamps from that commit. This would add significant overhead to the lite client (an extra commit to validate for every block!). If the time is incorrect, in indicates that the validators are explicitly violating the protocol rules in a detectable way which full nodes should detect in the first place and shouldn't forward to lite clients, so there would probably be bigger issues at foot.
```
Liamsi:
```suggestion
	- Verifying the time would require us to download the commit for the previous block, and to take the median of the timestamps from that commit. This would add significant overhead to the lite client (an extra commit to validate for every block!). If the time is incorrect, it indicates that the validators are explicitly violating the protocol rules in a detectable way which full nodes should detect in the first place and shouldn't forward to lite clients, so there would probably be bigger issues at foot.
```

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+{
+    pub last_header: H, // height H-1
+    pub validators: V,  // height H
+}
+```
+
+And the main method for verifying and updating the state should look like:
+
+```rust
+pub fn verify_sequential<H, V, C>(&mut self,
```
Liamsi: The method's name and signature does not reflect that it also updates the (trusted) state. Which slightly confusing with the sentence in front of this snippet.

docs/architecture/adr-002-lite-client.md
ValarDragon: You have a better impression of this than I do, but it feels to me like aggregate signatures shouldn't be too far out? Related projects are already using them (LibraBFT, Nervos (https://github.com/cryptape/overlord/) a TM fork)

To check an aggregate signature, you must have all the signing public keys. Is this 1/3rd check needed for `checkSupport`? If so, I believe that it is fine, since you are provided the signing public keys in order to verify.

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+(call this header[H-1]), and a validator set for height H (call this vals[H]).
+
+The node may be initialized by the user with only a height and header hash, and
+proceed to request the full header and validator set from a full node. This
+reduces the initialization burden on the user, and simplifies passing this
+information into the process, but for the state to be properly initialized it
+will need to get the correct header and validator set before starting the lite
+client syncing protocol.
+
+The configuration contains an initial list of peers.
```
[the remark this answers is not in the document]
ValarDragon: These can be other lite clients though, right?

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+
+Here we describe the happy path:
+
+1) Request header[H], commit[H], and vals[H+1] from the primary, and check that they are well formed and from the correct height
+2) Pass header[H], commit[H], vals[H], and vals[H+1] to the verification library, which
+
+  will:
+    - check that vals[H] and vals[H+1] are correctly reflected in the header
+    - check that the commit is for the header
+    - check that +2/3 of the validators correctly signed the header hash
```
ValarDragon: A minor point I hadn't thought of before, not an actual suggestion for an immediate change:

The lite client could instead check that `1/3 + c` signed the header hash, and you get a sliding scale of guarantees in exchange for better efficiency. 

* When `c > 0`, the lite client is guaranteed that upon seeing conflicting satisfying commits, the `1/3` assumption has been broken. Upon learning this, the client can feel free to panic / abandon the chain until manual override, etc.
* For `epsilon < c < 1/3`, the lite client increases its odds of catching the double-signing node.
* When `c > 1/3`, the lite client is guaranteed that upon seeing conflicting satisfying commits, it can create a fraud proof for at least one entity.

Since seeing a successful fork is such an extreme event, it seems likely to me that many lite clients may find it sufficient to be aware of the fork and deal w/ it via social consensus, and to not have to be able to submit the fraud proof.

(This whole point is moot once aggregate signatures are in. Then theres no advantage to only verifying a partial amount)

docs/architecture/adr-002-lite-client.md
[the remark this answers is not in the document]
Liamsi: Not sure tbh. I think they might be. But as far as I understand the light clients are initialised with a set of fullnodes? Unless we define an API for light clients to query each other etc?

docs/architecture/adr-002-lite-client.md
ebuchman: Ok so a number of things have come to my attention that suggest this abstraction isn't good enough:

1) Doesn't support aggregate sigs
2) Doesn't support batch verification (whoops!)
3) Testing is actually not much easier than using the actual Tendermint data types.

So all this leads me to believe that the commit trait should be abstracted further to not expose individual signatures but only information about whether say X% signed. In the interest of making some progress in the short term though I think we'll move forward as is and open an issue to deal with this at a later date.

As for the +1/3 check, we have the following situation:
- I trust validator set V
- I get a commit C from validator set V', where V != V'
- I want to check if +1/3 of V is in V' and signed C 
- the +1/3 from V may comprise an arbitrary percentage of V' (ie. they could be 1% or 99%, it doesn't matter). 

Can this be done for an aggregate sig?

docs/architecture/adr-002-lite-client.md
ebuchman: Cool - want to open an issue on the spec repo for this idea? We can also try to formally model the distinction in these guarantees: https://github.com/tendermint/spec

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,454 @@
...
+
+The state is considered "expired" if the difference between the current time and
+the time from the trusted header is greater than a configurable "trusting
+period". If at any point the state is expired, the node should log an error and
+exit - it's needs to be manually reset.
+
+
+### Manager
+
+For lack of a better name. The Manager co-ordiantes the syncing and is the
```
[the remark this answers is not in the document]
ebuchman: Noted here https://github.com/interchainio/tendermint-rs/issues/56

docs/architecture/adr-002-lite-client.md
ValarDragon: Yes this can be done. 
The aggregate signature will contain a list of who signed it, so you have the same situation as the non aggregate case. (The verifier has a set of signers, and has to check if 1/3 of V is in this set)

docs/architecture/adr-002-lite-client.md
ValarDragon: Done! https://github.com/tendermint/spec/issues/59

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,474 @@
...
+- [ADR-025](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-025-commit.md)
+- [Validator Signing Spec](https://github.com/tendermint/tendermint/blob/master/docs/spec/consensus/signing.md)
+- [Tendermint consensus specification](https://arxiv.org/abs/1807.04938)
+
+We may want to be more strict here in distinguishing between these cases. 
+On the one hand, 
+[upcoming changes](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-025-commit.md)
+to the commit structure will prevent votes from being included if they are from the
+wrong block.
+On the other, we may want the validator to verify the votes for nil - even
```
Liamsi:
```suggestion
On the other hand, we may want the validator to verify the votes for nil - even
```

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,474 @@
...
+    - header[H-1]
+    - vals[H]
+
+Here we describe the happy path:
+
+1) Request header[H], commit[H], and vals[H+1] from the primary, and check that they are well formed and from the correct height
+2) Pass header[H], commit[H], vals[H], and vals[H+1] to the verification library, which
+
+  will:
+    - check that vals[H] and vals[H+1] are correctly reflected in the header
```
Liamsi: For proper formatting:
```suggestion

    - check that vals[H] and vals[H+1] are correctly reflected in the header
```

docs/architecture/adr-002-lite-client.md
```diff
@@ -0,0 +1,474 @@
...
+    - check that the commit is for the header
+    - check that +2/3 of the validators correctly signed the header hash
```
Liamsi: Everywhere else you specify which header we are talking about (by explicitly listing a height `H` via `header[H]`). I think it would be good be consistent in this listing, too. (same for validators)

ebuchman:
Ok I addressed all the feedback. Also updated it to match the latest version of the code and included some notes about that. Added a proper image too.

ebuchman:
I don't understand why this is failing. The latest master has been merged in and it passes locally ...

ebuchman:
Ok we should update this per https://github.com/interchainio/tendermint-rs/pull/105 ... removed the Validator trait!

ebuchman:
Ok thanks to @jibrown we actually have a really nice diagram of the data flow for the whole thing now! Might need a few updates to the text to align with the new image though.

ebuchman:
Pushed an ADR refactor. ADR002 is now an index into light client ADRs, and the core verification is ADR 003. Still going to do some cleanup/alignment on the new ADR 003 and just leave the others as WIPs to be done in future PRs.

Liamsi:
The clippy messages look somewhat familiar. I think they should be fixed in the latest master.

ebuchman:
Traits are good. Just want to go through the "implementing traits" part and then this is good to merge and I will work on the other two (CLI and fork detection)

ebuchman:
Ok, I think this is finally done. It's been realigned with the latest code and describes only what's already been implemented in the `lite` and `lite_impl`. Let's merge.

PR #55, opened 2019-10-30
Address clippy warnings

ebuchman:
I couldn't figure out how to remove the `extern crate` with the rename here. I read https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/path-clarity.html#renaming-crates, but couldn't get it working. Possibly it's because I needed to edit *alot* more files, but didn't feel I should have to do that. So I also tried https://doc.rust-lang.org/1.28.0/cargo/reference/unstable.html#rename-dependency, but couldn't get it working either. So I just turned off the lint in one spot ...

Also the `config_toml_parser` could be broken out a bit but didn't seem like a big deal to just disable the lint for now.

### DIFF
--- a/tendermint/src/lib.rs
+++ b/tendermint/src/lib.rs
@@ -18,7 +18,12 @@
     html_root_url = "https://docs.rs/tendermint/0.10.0"
 )]
 
+// NOTE(EB): can't figure out how to easily remove the extern crate per Rust2018 upgrade ...
+#[allow(unused_extern_crates)]
 extern crate prost_amino as prost;
+
+// NOTE(EB): can't figure out how to easily remove the extern crate per Rust2018 upgrade ...
+#[allow(unused_extern_crates)]
 #[macro_use]
 extern crate prost_amino_derive as prost_derive;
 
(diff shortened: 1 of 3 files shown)


tendermint/src/lib.rs
```diff
@@ -18,7 +18,12 @@
     html_root_url = "https://docs.rs/tendermint/0.10.0"
 )]
 
+// NOTE(EB): can't figure out how to easily remove the extern crate per Rust2018 upgrade ...
+#[allow(unused_extern_crates)]
 extern crate prost_amino as prost;
```
tarcieri: Probably the easiest way to actually fix this is through the Cargo.toml:

```toml
prost = { package = "prost-amino", version = "0.4.0" }
prost-derive = { package = "prost-amino-derive", version = "0.4.0" }
```

tarcieri:
You may also want to add:

```rust
#![warn(rust_2018_idioms)]
```

...to the top of `lib.rs`

tendermint/src/lib.rs
ebuchman: I wasn't able to get this working. When I try it and then comment out these extern crate lines, I get tons of errors.

ebuchman:
> You may also want to add:

I'm using `cargo clippy --all --all-targets -- -Dwarnings -Drust-2018-idioms` which is what's being run in CircleCI, which presumably achieves the same result?

tendermint/src/lib.rs
tarcieri: @ebuchman oh, I bet it hasn't been updated to the 2018 edition, so `prost-amino-derive` is still dependent on the `extern crate` directives.

Probably the best solution would be to switch to upstream `prost` when Amino has been fully replaced with protos.

ebuchman: lol cannot wait!

PR #58, opened 2019-11-06
Upgrade to `zeroize` v1.0

tony-iqlusion:


tony-iqlusion:
Was curious to see if edits through the GitHub web UI would meet the GPG requirement. They show up as `[ Verified ]`, but don't seem to be satisfying it.

PR #59, opened 2019-11-08
simple_hash_from_byte_slices takes a Vec instead of slice

### LINKED ISSUE #57: simple_hash_from_byte_slices should take a Vec instead of slice
Currently we have `fn simple_hash_from_byte_slices(byte_slices: &[&[u8]]) -> Hash`

But this leads to some complexities in the few places it's so far used: 
- https://github.com/interchainio/tendermint-rs/blob/10a3026cfbd4ab333752d344645a4f07ecdc29f7/tendermint/src/validator.rs#L23-L35
- https://github.com/interchainio/tendermint-rs/pull/36/files#diff-a1644bb72a1c83c54d82738ef71ce095R126 (ie. the need for a separate variable followed by `as_slice` would go away)

Liamsi:
resolves #57

### DIFF
--- a/tendermint/src/merkle.rs
+++ b/tendermint/src/merkle.rs
@@ -8,9 +8,11 @@ pub const HASH_SIZE: usize = 32;
 /// Hash is the output of the cryptographic digest function
 pub type Hash = [u8; HASH_SIZE];
 
-/// Compute a simple Merkle root from the arbitrary byte arrays
-pub fn simple_hash_from_byte_slices(byte_slices: Vec<Vec<u8>>) -> Hash {
-    simple_hash_from_byte_slices_inner(byte_slices.as_slice())
+/// Compute a simple Merkle root from vectors of arbitrary byte vectors.
+/// The leaves of the tree are the bytes of the given byte vectors in
+/// the given order.
+pub fn simple_hash_from_byte_vectors(byte_vecs: Vec<Vec<u8>>) -> Hash {
+    simple_hash_from_byte_slices_inner(byte_vecs.as_slice())
 }
 
 // recurse into subtrees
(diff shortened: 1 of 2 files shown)


Liamsi:
I think this is good enough here. 

See 
https://github.com/interchainio/tendermint-rs/compare/lite_impl...lite_impl_simple_merkle_merged?expand=1
how this would affect the usage of `simple_hash_from_byte_slices`.

tendermint/src/merkle.rs
```diff
@@ -8,16 +8,21 @@ pub const HASH_SIZE: usize = 32;
 /// Hash is the output of the cryptographic digest function
 pub type Hash = [u8; HASH_SIZE];
 
-/// Compute a simple Merkle root from the arbitrary sized byte slices
-pub fn simple_hash_from_byte_slices(byte_slices: &[&[u8]]) -> Hash {
+/// Compute a simple Merkle root from the arbitrary byte arrays
+pub fn simple_hash_from_byte_slices(byte_slices: Vec<Vec<u8>>) -> Hash {
```
tarcieri: Those aren't slices 😉

Liamsi: Changed to `simple_hash_from_byte_vectors`  in 937ba67.

I think a better name for the the method  would rather be `compute_merkle_root_from_leaves` or `compute_merkle_root_from_leaf_values` instead of `simple_hash_from_[some_type]`. 
This would actually say what the method does instead of what type it expects as input (which the method signature clearly states). What do you think? CC @ebuchman

Liamsi:
Thanks for the review @thanethomson! 

Yes, I think it would be worth turning simple merkle into a trait soon. It also would be cool if the hashing algo could be easily swappable. I also looked a bit into this implementation: https://github.com/filecoin-project/merkle_light/tree/master/src but the amount of generality seems a bit like overkill for simple merkle. I'll open an issue for this and a separate PR. Although it seems merging this is also blocked by the GPG check :-/ (although all commits are signed, not sure what is going on here ...)

tendermint/src/merkle.rs
ebuchman: You're right about better naming happy to see this changed.

PR #62, opened 2019-11-11
Implement ``Default`` trait for tendermint data types

yihuang:
Implements ``Default`` trait for tendermint data types, useful for writing tests.
It also implements several utility traits.

### DIFF
--- a/tendermint/src/account.rs
+++ b/tendermint/src/account.rs
@@ -15,7 +15,7 @@ use subtle_encoding::hex;
 const LENGTH: usize = 20;
 
 /// Account IDs
-#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord, Default)]
 pub struct Id([u8; LENGTH]);
 
 impl Id {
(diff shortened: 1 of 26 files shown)


tendermint/src/account.rs
```diff
@@ -15,7 +15,7 @@ use subtle_encoding::hex;
 const LENGTH: usize = 20;
 
 /// Account IDs
-#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord, Default)]
```
tarcieri: For anything cryptographic / security-critical, I don't think it makes sense to add a `Default`  impl which initializes it to what's effectively an invalid value solely for the purposes of testing.

If your goal with this is to do something like `Id::default()` for tests, note that you could also do `Id::new(Default::default())` to get the same effect.

yihuang: Yeah, kinda agree with you, this use of  ``Default`` trait is purely due to laziness ;D
I think I should just create some custom data templates for tests.

yihuang:
I modified the patch to only contain meaningful ``Default`` instances now.

PR #64, opened 2019-11-12
Implement utility traits for tendermint data types

yihuang:
I removed the ``Default`` instances that don't make sense.

### DIFF
--- a/tendermint/src/block/id.rs
+++ b/tendermint/src/block/id.rs
@@ -16,7 +16,7 @@ pub const PREFIX_LENGTH: usize = 10;
 /// as well as the number of parts in the block.
 ///
 /// <https://github.com/tendermint/tendermint/blob/master/docs/spec/blockchain/blockchain.md#blockid>
-#[derive(Serialize, Deserialize, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+#[derive(Serialize, Deserialize, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord, Default)]
 pub struct Id {
     /// The block's main hash is the Merkle root of all the fields in the
     /// block header.
--- a/tendermint/src/hash.rs
+++ b/tendermint/src/hash.rs
@@ -28,6 +28,12 @@ pub enum Hash {
     Null,
 }
 
+impl Default for Hash {
+    fn default() -> Hash {
+        Hash::Null
+    }
+}
+
 impl Hash {
     #[allow(clippy::new_ret_no_self)]
     /// Create a new `Hash` with the given algorithm type
--- a/tendermint/src/node/info.rs
+++ b/tendermint/src/node/info.rs
@@ -62,6 +62,11 @@ pub struct ProtocolVersionInfo {
 pub struct ListenAddress(String);
 
 impl ListenAddress {
+    /// Construct `ListenAddress`
+    pub fn new(s: String) -> ListenAddress {
+        ListenAddress(s)
+    }
+
     /// Convert `ListenAddress` to a `net::Address`
     pub fn to_net_address(&self) -> Option<net::Address> {
         // TODO(tarcieri): validate these and handle them better at parse time
@@ -99,6 +104,12 @@ pub enum TxIndexStatus {
     /// Index is off
     #[serde(rename = "off")]
     Off,
+}
+
+impl Default for TxIndexStatus {
+    fn default() -> TxIndexStatus {
+        TxIndexStatus::On
+    }
 }
 
 impl From<TxIndexStatus> for bool {
--- a/tendermint/src/time.rs
+++ b/tendermint/src/time.rs
@@ -1,8 +1,10 @@
 //! Timestamps used by Tendermint blockchains
 
 use crate::error::{Error, ErrorKind};
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, SecondsFormat, Utc};
 use serde::{Deserialize, Serialize};
+use std::fmt;
+use std::str::FromStr;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tai64::TAI64N;
 
@@ -36,10 +38,29 @@ impl Time {
         Ok(Time(DateTime::parse_from_rfc3339(s)?.with_timezone(&Utc)))
     }
 
+    /// Return an RFC 3339 and ISO 8601 date and time string with 6 subseconds digits and Z.
+    pub fn to_rfc3339(&self) -> String {
+        self.0.to_rfc3339_opts(SecondsFormat::Nanos, true)
+    }
+
     /// Convert this timestamp to a `SystemTime`
     pub fn to_system_time(&self) -> Result<SystemTime, Error> {
         let duration_since_epoch = self.duration_since(Self::unix_epoch())?;
         Ok(UNIX_EPOCH + duration_since_epoch)
+    }
+}
+
+impl fmt::Display for Time {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f, "{}", self.to_rfc3339())
+    }
+}
+
+impl FromStr for Time {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Time::parse_from_rfc3339(s)
     }
 }
 
--- a/tendermint/src/abci/code.rs
+++ b/tendermint/src/abci/code.rs
@@ -18,6 +18,12 @@ pub enum Code {
     Err(u32),
 }
 
+impl Default for Code {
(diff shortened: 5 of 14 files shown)


tendermint/src/account.rs
```diff
@@ -15,7 +15,7 @@ use subtle_encoding::hex;
 const LENGTH: usize = 20;
 
 /// Account IDs
-#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord, Default)]
 pub struct Id([u8; LENGTH]);
```
tarcieri: This was the main one I said didn’t make sense 😜 (but also anything with a hash)

tendermint/src/public_key.rs
```diff
@@ -91,6 +91,13 @@ impl PublicKey {
     }
 }
 
+impl Default for PublicKey {
+    fn default() -> PublicKey {
+        static DUMMY_BYTES: [u8; 32] = [0; 32];
+        PublicKey::Ed25519(ed25519::PublicKey::new(DUMMY_BYTES))
```
tarcieri: This also doesn’t make any sense, and is a particularly pathological case as it’s a special point on the elliptic curve.

tendermint/src/time.rs
```diff
@@ -36,13 +38,38 @@ impl Time {
         Ok(Time(DateTime::parse_from_rfc3339(s)?.with_timezone(&Utc)))
     }
 
+    /// Returns an RFC 3339 string, such as 1996-12-19T16:39:57-08:00.
+    pub fn to_rfc3339(&self) -> String {
+        self.0.to_rfc3339()
```
tarcieri: If you want to support this, it should be Z-normalized, which I think it is but that shiuld be reflected in the comment.

yihuang:
* Removed ``Default`` instance for ``Account::Id``, ``PublicKey``, ``validator::Info``.
* Fixed ``Time`` format to use ``.to_rfc3339_opts(SecondsFormat::Micros, true)``

tendermint/src/abci/transaction/hash.rs
```diff
@@ -28,6 +29,15 @@ impl Hash {
     }
 }
 
+impl Default for Hash {
+    fn default() -> Hash {
+        // hash of empty data
+        let mut bytes = [0u8; LENGTH];
```
tarcieri: A `Hash` of all zero bytes is generally not semantically meaningful and IMO should not be the `Default`.

(I have seen security vulnerabilities arising from zero-value hashes in Go)

tendermint/src/time.rs
```diff
@@ -36,13 +38,38 @@ impl Time {
...
     /// Convert this timestamp to a `SystemTime`
     pub fn to_system_time(&self) -> Result<SystemTime, Error> {
         let duration_since_epoch = self.duration_since(Self::unix_epoch())?;
         Ok(UNIX_EPOCH + duration_since_epoch)
     }
 }
 
+impl Default for Time {
+    fn default() -> Time {
+        Time::unix_epoch()
```
tarcieri: I'm not sure a "default time" makes sense

tendermint/src/abci/transaction/hash.rs
yihuang: Makes sense, just removed.

tendermint/src/time.rs
yihuang: Makes sense, just removed.

yihuang:
I've changed the format of `Time` to `SecondsFormat::Nanos`, according to [tendermint implementation](https://github.com/tendermint/tendermint/blob/master/types/canonical.go#L13).

tendermint/src/node/info.rs
```diff
@@ -58,10 +58,15 @@ pub struct ProtocolVersionInfo {
 }
 
 /// Listen address information
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Default)]
```
tarcieri: It seems like this could have a more useful default (like, say, `0.0.0.0`, or whatever Tendermint does)

yihuang: Agree, I think the default is "tcp://127.0.0.1:26656";

yihuang: But, again, a concret network address might not suit for `Default` trait. Maybe I remove this instance.

tendermint/src/hash.rs
```diff
@@ -28,6 +28,12 @@ pub enum Hash {
     Null,
 }
 
+impl Default for Hash {
+    fn default() -> Hash {
+        Hash::Null
```
tarcieri: Does this actually make sense? Honestly I think this whole concept should be replaced with `Option<Hash>` instead...

tendermint/src/node/info.rs
```diff
@@ -33,7 +33,7 @@ pub struct Info {
 }
 
 /// Protocol version information
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Default)]
```
tarcieri: Are zeroes for all of these version values semantically meaningful?

tendermint/src/version.rs
```diff
@@ -2,7 +2,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt::{self, Display};
 
 /// Tendermint version
-#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
```
tarcieri: Should the default version really be empty string?

tendermint/src/block/header.rs
```diff
@@ -74,7 +74,7 @@ pub struct Header {
 /// application.
 ///
 /// <https://github.com/tendermint/tendermint/blob/master/docs/spec/blockchain/blockchain.md#version>
-#[derive(Serialize, Deserialize, Clone, Debug)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
```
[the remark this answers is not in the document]
yihuang: I think `Default` supposed to some universal default value for the data type, like unit in a monoid, should not change any time in the future (it would be a breaking change).

tendermint/src/block/id.rs
```diff
@@ -16,7 +16,7 @@ pub const PREFIX_LENGTH: usize = 10;
 /// as well as the number of parts in the block.
 ///
 /// <https://github.com/tendermint/tendermint/blob/master/docs/spec/blockchain/blockchain.md#blockid>
-#[derive(Serialize, Deserialize, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+#[derive(Serialize, Deserialize, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord, Default)]
```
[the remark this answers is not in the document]
yihuang: `block::Id` support empty value, that's a good candidate for a `Default` value.

tendermint/src/hash.rs
yihuang: Agree, but that probably should belong to another PR?

tendermint/src/version.rs
yihuang: Removed.

tendermint/src/node/info.rs
yihuang: Removed.

tendermint/src/block/header.rs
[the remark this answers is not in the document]
yihuang: Removed.

yihuang:
> Perhaps I'm a bit jaded by (in some cases security) vulnerabilities relating to Go zero values, which it seems is how `Default` is being leveraged in the cases I pointed out in this review.
> 
> If you're looking for a way to easily construct test fixtures, I think `Default` is the wrong tool for the job. Alternatively, in the test suite itself, you can define a trait _specifically for test fixtures_ in the _test suite itself_, and then impl it however you want!
> 
> ```rust
> trait TestFixture {
>     fn test_fixture() -> Self;
> }
> 
> impl TestFixture for Hash {
>     fn test_fixture() -> Hash {
>         Hash::Null
>     }
> }
> ```
> 
> In general, I think it's a bad idea for anything that's a testing-related convenience to leak outside of the test context, especially when it could potentially lead to security issues in production.
> 
> Is that what most of these are for?

Yeah, actually we've changed the test code to use more properly constructed mocking data.
But what's left in this patch now IMO are quite reasonable ;D

tarcieri:
@yihuang looks like this needs a GPG signature

yihuang:
> @yihuang looks like this needs a GPG signature

Done.

PR #65, opened 2019-11-13
Parse validators field in genesis api

yihuang:
`genesis` endpoint response has a `validators` field, which is similar to the one from `validators` endpoint, we can re-use the data type for this, only difference is the field name of `voting_power` becomes `power`.

### DIFF
--- a/tendermint/tests/support/rpc/genesis.json
+++ b/tendermint/tests/support/rpc/genesis.json
@@ -20,6 +20,17 @@
           ]
         }
       },
+      "validators": [
+        {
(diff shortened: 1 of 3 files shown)


tendermint/tests/support/rpc/genesis.json
```diff
@@ -20,6 +20,7 @@
           ]
         }
       },
+      "validators": [],
```
Liamsi: Do you mind adding in a validator here? You copy&paste the from genesis example in https://docs.tendermint.com/master/rpc/#/Info/genesis

yihuang: Done

PR #67, opened 2019-11-13
impl lite::Commit for commit::SignedHeader

yihuang:


tendermint/src/rpc/endpoint/commit.rs
```diff
@@ -1,6 +1,6 @@
 //! `/commit` endpoint JSONRPC wrapper
 
-use crate::{block, rpc};
+use crate::{amino_types::vote::CanonicalVote, block, lite, rpc, vote::SignedVote, Hash};
```
Liamsi: The rpc module currently does not depend on `amino_types`. I think, if possible, we should try to keep it that way. Even if that means introducing another type  under tendermint/block (or similar).

Liamsi:
Did the suggested changes here: https://github.com/interchainio/tendermint-rs/pull/63/commits/e1a7560e4dfbeaaa3f9013e2141b7296e33f2b5d

yihuang:
Great, thanks.

tendermint/src/rpc/endpoint/commit.rs
```diff
@@ -46,3 +46,26 @@ pub struct SignedHeader {
...
+impl lite::Commit for SignedHeader {
+    type Vote = SignedVote;
+    fn header_hash(&self) -> Hash {
+        self.commit.block_id.hash
+    }
+    fn into_vec(&self) -> Vec<Option<Self::Vote>> {
+        let chain_id = self.header.chain_id.to_string();
+        let mut votes = self.commit.precommits.clone().into_vec();
+        votes
+            .drain(..)
```
ebuchman: @yihuang Why use `drain` here rather than `iter` ?

yihuang: Yeah, my fault, not familiar enough with idiomatic rust then. But I believe this code already modified by @Liamsi in the final merged version ;D

PR #68, opened 2019-11-13
impl lite::ValidatorSetLookup

yihuang:


tendermint/src/validator.rs
```diff
@@ -62,6 +62,15 @@ impl lite::ValidatorSet for Set {
...
 
+impl lite::ValidatorSetLookup for Set {
+    fn validator(&self, val_id: account::Id) -> Option<Self::Validator> {
+        self.validators
+            .iter()
+            .cloned()
+            .find(|val| val.address == val_id)
+    }
+}
+
```
Liamsi: Would be cool to add a little test  for this. Happy to do so after merging.

Liamsi: done in: 23042bb8d31439983d09d19cf28a62cf8062bf0a

PR #69, opened 2019-11-18
Signing bytes need to encoded with encode_length_delimited

yihuang:
So we have identical signing bytes as golang version to verify signatures.

### DIFF
--- a/tendermint/src/amino_types/message.rs
+++ b/tendermint/src/amino_types/message.rs
@@ -1,3 +1,6 @@
+use prost_amino::encoding::encoded_len_varint;
+use std::convert::TryInto;
+
 /// Extend the original prost::Message trait with a few helper functions in order to
 /// reduce boiler-plate code (and without modifying the prost-amino dependency).
 pub trait AminoMessage: prost_amino::Message {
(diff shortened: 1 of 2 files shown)


tendermint/src/vote.rs
```diff
@@ -101,7 +101,9 @@ impl lite::Vote for SignedVote {
     }
 
     fn sign_bytes(&self) -> Vec<u8> {
-        AminoMessage::bytes_vec(&self.vote.to_owned())
+        let mut buf = Vec::with_capacity(self.vote.encoded_len() + 4);
```
Liamsi: 4 bytes should be correct but it would still be cleaner to take the length and its encoded length. Prost provides a method for that. As far as I remember it was called `encoded_len_varint`. Let me double check.

Liamsi: This should work (given the right imports):
```suggestion
        let vote_len = self.vote.encoded_len();
        let len_delimiter_len =
            prost_amino::encoding::encoded_len_varint(vote_len.try_into().unwrap());
        let mut buf = Vec::with_capacity(self.vote.encoded_len() + len_delimiter_len);
```
I think it would be better to add a method to the `AminoMessage` trait. Sth. like `length_delimited_bytes_vec`. But OK to merge for now.

tendermint/src/amino_types/message.rs
```diff
@@ -15,6 +17,20 @@ pub trait AminoMessage: prost_amino::Message {
...
+    /// Encode prost-amino message as length delimited.
+    ///
+    /// Warning: Only use this method, if you are in control what will be encoded.
+    /// If there is an encoding error, this method will panic.
+    fn bytes_vec_length_delimited(&self) -> Vec<u8>
+    where
+        Self: Sized,
+    {
+        let len = self.encoded_len();
+        let mut res = Vec::with_capacity(len + encoded_len_varint(len as u64));
```
Liamsi: Awesome, thanks!

Why did you decide against `len.try_into().unwrap()` (as suggested in https://github.com/interchainio/tendermint-rs/pull/69#discussion_r347431877) instead of `len as u64`? I think, the latter can silently overflow while the first would panic.

yihuang: Nice trick, re-commited.

Liamsi: Awesome, thanks!

PR #70, opened 2019-11-19
Implement several utility trait

yihuang:
* `Debug` of `Error` is for printing error.
* `Serialize/Deserialize` of `validator::Set` is because it's a part of the trusted state, which needs to be stored somewhere.

### DIFF
--- a/tendermint/src/validator.rs
+++ b/tendermint/src/validator.rs
@@ -11,7 +11,7 @@ use signatory_dalek::Ed25519Verifier;
 use subtle_encoding::base64;
 
 /// Validator set contains a vector of validators
-#[derive(Debug)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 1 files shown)


tendermint/src/lite/types.rs
```diff
@@ -102,6 +102,7 @@ pub trait Vote {
     fn signature(&self) -> &[u8];
 }
 
+#[derive(Debug)]
```
Liamsi: Didn't this change already make into `lite_impl_simple_merkle_merged` via https://github.com/interchainio/tendermint-rs/commit/9e00e99c2143f66a5dd61d80cb5c292a77eefbc6#diff-b1b6c7ff77dbf13838f76b8403715185?

yihuang: My mistake, just rebased this PR, only one line of change left.

PR #71, opened 2019-11-19
Fix AbciQuery to support error response.

yihuang:
For example:

`{"response": {"code": 1, "log": "account lookup failed: account not found", "info": "", "index": "0", "key": null, "value": null, "proof": null, "height  ": "0", "codespace": ""}}`

### DIFF
--- a/tendermint/src/rpc/endpoint/abci_query.rs
+++ b/tendermint/src/rpc/endpoint/abci_query.rs
@@ -58,9 +58,11 @@ impl rpc::Response for Response {}
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct AbciQuery {
     /// Response code
+    #[serde(default)]
     pub code: Code,
 
     /// Log value
+    #[serde(default)]
     pub log: Log,
 
     /// Info value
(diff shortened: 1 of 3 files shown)


tendermint/src/rpc/endpoint/abci_query.rs
```diff
@@ -69,24 +71,40 @@ pub struct AbciQuery {
...
-    pub value: String,
+    pub value: Option<String>,
 
     /// Proof (if requested)
     pub proof: Option<Proof>,
 
     /// Block height
-    pub height: block::Height,
+    #[serde(default, deserialize_with = "serializers::parse_height_option")]
+    pub height: Option<block::Height>,
```
Liamsi: So these are only optional now as they are None in case of an error response.

tendermint/src/rpc/endpoint/abci_query.rs
```diff
@@ -69,24 +71,40 @@ pub struct AbciQuery {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json;
+
+    #[test]
+    fn parse_query_responses() {
+        let error_rsp = r#"{"response": {"code": 1, "log": "account lookup failed: account not found", "info": "", "index": "0", "key": null, "value": null, "proof": null, "height": "0", "codespace": ""}}"#;
+        let success_rsp = r#"{"response": {"value": "some value"}}"#;
```
Liamsi: Doesn't a successful response also include `code` etc? Let me double check.

yihuang: It seems it depends solely on the abci app's implementation, and since it's protobuf, every field is optional.

Liamsi: It might be good to update our fixtures in the rpc to contain both JSON files, the happy path and the error case:
https://docs.tendermint.com/master/rpc/#/ABCI/abci_query

yihuang:
It's fixed in current master, closing now.

PR #72, opened 2019-11-19
add mock testing for lite client

yihuang:
Add some mocking data to test lite client.

### DIFF
--- a/tendermint/tests/support/lite/basic.json
+++ b/tendermint/tests/support/lite/basic.json
@@ -1,0 +1,78 @@
+{
+    "signed_header": {
+        "header": {
+            "version": {
+                "block": "10",
+                "app": "0"
+            },
+            "chain_id": "test-chain-y3m1e6-AB",
+            "height": "1",
+            "time": "2019-11-19T07:56:45.46738Z",
+            "num_txs": "0",
+            "total_txs": "0",
+            "last_block_id": {
+                "hash": "",
+                "parts": {
+                    "total": "0",
+                    "hash": ""
+                }
+            },
+            "last_commit_hash": "",
+            "data_hash": "",
+            "validators_hash": "1F0F3E080DCE3253C88394F2CE479C074C5A25FE94A6C8560C76B6F2281BD685",
+            "next_validators_hash": "1F0F3E080DCE3253C88394F2CE479C074C5A25FE94A6C8560C76B6F2281BD685",
+            "consensus_hash": "048091BC7DDC283F77BFBF91D73C44DA58C3DF8A9CBC867405D8B7F3DAADA22F",
+            "app_hash": "2E419E5A3CDD1F718EA956E9AF91E57FD4FD5D8C8DDEA43FB3D8728E8E41852D",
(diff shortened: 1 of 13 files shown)


tendermint/src/lite.rs
```diff
@@ -3,3 +3,122 @@ pub mod verifier;
...
+        .unwrap();
+        let validators = last_validators.clone();
+        verify_trusting(
+            header,
+            signed_header,
+            validator::Set::new(last_validators),
+            validator::Set::new(validators),
+        )
+        .expect("verify_trusting failed");
+    }
```
Liamsi: Very cool! Thanks. I'm wondering if we either should have simpler rust-only tests first. Or, simplify this (extract all json strings to vars/constants and have only the verification logic in the test). Also, there should be a comment explaining how the JSON was generated. 

Furthermore, @Shivani912 is working on generating similar tests: https://github.com/Shivani912/tendermint/tree/shivani/language-agnostic-tests/types/lite-client/tests

yihuang: I simply copy-pasted my local test node's json rpc response, only to test the basics work.
It's great to have more sophisticated test cases.

Liamsi: I think, these kind of tests currently tend to live in https://github.com/interchainio/tendermint-rs/tree/master/tendermint/tests as JSON files: https://github.com/interchainio/tendermint-rs/tree/master/tendermint/tests/support/rpc

I think it would increase the readability here if we'd extract the JSON into vars in the test (e.g. ` static VALID_HEADER: &str =  r#"{ ...}"#`

The test seems to fail on my machine btw.

tendermint/src/lite.rs
```diff
@@ -3,3 +3,122 @@ pub mod verifier;
...
+        "height": "1",
+        "time": "2019-11-19T07:56:45.46738Z",
+        "num_txs": "0",
+        "total_txs": "0",
+        "last_block_id": {
+          "hash": "",
+          "parts": {
+            "total": "0",
+            "hash": ""
+          }
```
Liamsi: This will panic: 
https://github.com/interchainio/tendermint-rs/blob/ef7410c184c736f032f24296c24cf921175582be/tendermint/src/amino_types/block_id.rs#L36-L37
called e.g. by:
https://github.com/interchainio/tendermint-rs/blob/ef7410c184c736f032f24296c24cf921175582be/tendermint/src/block/header.rs#L106

yihuang: This is a bug fixed in another PR, https://github.com/interchainio/tendermint-rs/pull/66.

yihuang:
- Move the test data into separate json files.
- Rebased on https://github.com/interchainio/tendermint-rs/pull/66.

tendermint/tests/support/lite/basic.json
```diff
@@ -0,0 +1,78 @@
+{
+    "signed_header": {
```
Liamsi: Thanks for moving the test JSON here 👌

Issue #75, opened 2019-11-26
RPC client uses incorrect encoding in abci_query request


mappum:
The RPC client encodes the `data` field for the `abci_query` endpoint into a JSON array of byte values, but the Tendermint side expects a hex string. This means every request fails, even with an empty data field.

mappum:
Closed by #77

PR #77, opened 2019-11-27
Fix abci_query RPC endpoint

mappum:
This is a fix for #75, although it's not passing the integration test yet since Tendermint is responding with "0" for the `height` field but `block::Height` enforces that height is >= 1. Other unrelated integration tests are also failing, other RPC request/response types are possibly out of date or just overly strict.

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -24,6 +24,9 @@ authors = [
     "Thane Thomson <thane@interchain.io>"
 ]
 
+[package.metadata.docs.rs]
+all-features = true
+
 [badges]
 circle-ci = { repository = "interchainio/tendermint-rs" }
 
@@ -49,6 +52,3 @@ zeroize = { version = "1.0", features = ["zeroize_derive"] }
 
 [dev-dependencies]
 serde_json = "1"
-
-[package.metadata.docs.rs]
-all-features = true
--- a/tendermint/src/block/height.rs
+++ b/tendermint/src/block/height.rs
@@ -11,7 +11,7 @@ use std::{
 /// 
 /// A height of 0 represents a chain which has not yet produced a block.
 #[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
-pub struct Height(u64);
+pub struct Height(pub u64);
 
 impl Height {
     /// Get inner integer value. Alternative to `.0` or `.into()`
--- a/tendermint/src/block/height.rs
+++ b/tendermint/src/block/height.rs
@@ -7,25 +7,27 @@ use std::{
 
 /// Block height for a particular chain (i.e. number of blocks created since
 /// the chain began)
+/// 
+/// A height of 0 represents a chain which has not yet produced a block.
 #[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub struct Height(u64);
 
 impl Height {
     /// Convert `u64` to block height.
-    ///
-    /// Note that 0 is not a valid block height.
+    /// 
+    /// Note that this method will never error and is just for backwards
+    /// compatibility.
     pub fn try_from_u64(n: u64) -> Result<Self, Error> {
-        // Minimum height is 1
-        if n > 0 {
-            Ok(Height(n))
-        } else {
-            Err(ErrorKind::OutOfRange.into())
-        }
+        Ok(Height(n))
     }
 
     /// Convert `i64` (used in e.g. Amino messages) to block height.
     pub fn try_from_i64(n: i64) -> Result<Self, Error> {
-        Self::try_from_u64(n as u64)
+        if n >= 0 {
+            Ok(Height(n as u64))
+        } else {
+            Err(ErrorKind::OutOfRange.into())
+        }
     }
 
     /// Get inner integer value. Alternative to `.0` or `.into()`
@@ -65,7 +67,7 @@ impl From<i64> for Height {
 
 impl From<u64> for Height {
     fn from(n: u64) -> Height {
-        Self::try_from_u64(n).unwrap()
+        Height(n)
     }
 }
 
--- a/tendermint/src/block/height.rs
+++ b/tendermint/src/block/height.rs
@@ -1,6 +1,7 @@
 use crate::error::{Error, ErrorKind};
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
 use std::{
+    convert::TryFrom,
     fmt::{self, Debug, Display},
     str::FromStr,
 };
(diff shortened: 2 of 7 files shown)


tarcieri:
@mappum I'm beginning to think it might be good to allow a `Height` of 0. It is representative of something meaningful: a chain in a state before it has produced an initial block, which in particular would help resolve this issue without making invasive changes to Tendermint KMS:

https://github.com/tendermint/kms/issues/369

mappum:
@tarcieri I added the 0-height change.

However, it's still hard to use the integration tests because there are various differences between responses on different chains. It would be nice if it could be tested against an easy-to-spin-up local testnet (e.g. `tendermint node --proxy_app=kvstore`), since testing against `gaiad` has the external requirement of figuring out what network it should be on. Currently, `gaiad` for Cosmos mainnet is on an older version of Tendermint, so I'm not sure if there's some gaia testnet that the integration tests target.

tarcieri:
> I added the 0-height change.

Great!

> However, it's still hard to use the integration tests because there are various differences between responses on different chains.

Yeah it's been a big pain point in the past. If you can think of a better strategy (particularly one that works in CI) I'd be all for it.

mappum:
I would vote for either making the integration test run something like this:
```bash
tendermint init --home=/some/tmp/dir
tendermint node --proxy_app=kvstore # (or counter or something)
```

(Then remove the Tendermint data after the tests run).

This could either run a `tendermint` binary included in this repo, or download it from Github releases (it could use the system's installed `tendermint` but this would mean people may run tests against the wrong version and have to worry about managing it themselves).

Until then, we could change the integration tests to target this kind of network but still expect the user to run it manually.

tarcieri:
@mappum there's a bit of prior art for this with the test harness used by the KMS: 

- https://github.com/tendermint/kms/blob/master/Dockerfile
- https://github.com/tendermint/kms/commit/5427d0fe076fa2b388594a7d69fc578504017e02

You could potentially try to use the same Docker image on CircleCI?

mappum:
@tony-iqlusion I can make that change in a later PR, for now I changed the integration tests to target a local node for a throwaway network. Are you OK with merging this change? (I'd very much like to use this fix so I can use the RPC client).

tendermint/Cargo.toml
```diff
@@ -46,9 +49,7 @@ tai64 = { version = "3", features = ["chrono"] }
 toml = { version = "0.5" }
 uuid = { version = "0.7", default-features = false }
 zeroize = { version = "0.9" }
+hex = "0.4.0"
```
tarcieri: There's a hex-compatible crate already in the dependencies:

```
use subtle_encoding::hex;
```

tendermint/src/block/height.rs
```diff
@@ -7,25 +7,27 @@ use std::{
...
 #[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub struct Height(u64);
 
 impl Height {
     /// Convert `u64` to block height.
-    ///
-    /// Note that 0 is not a valid block height.
+    /// 
+    /// Note that this method will never error and is just for backwards
+    /// compatibility.
```
tarcieri: I can try to address it in a followup PR, but it might be good to just break compatibility here, make the above struct `pub struct Height(pub u64);` for convenience, and add `From` impls between `Height` and `u64`

tendermint/Cargo.toml
mappum: :heavy_check_mark:

mappum:
I realized the encoding for the response should be changed too, since Tendermint always converts the `key` and `value` fields to base64 strings (but the request is in hex, lol).

Let me know if I should make any other changes.

tarcieri:
@mappum looks like it needs a GPG signature

mappum:
Signed last commit.

tarcieri:
@mappum looks like it requires all commits to be signed. Perhaps squash and sign the resulting commit.

tendermint/src/block/height.rs
```diff
@@ -7,25 +7,27 @@ use std::{
...
-        if n > 0 {
-            Ok(Height(n))
-        } else {
-            Err(ErrorKind::OutOfRange.into())
-        }
+        Ok(Height(n))
     }
 
     /// Convert `i64` (used in e.g. Amino messages) to block height.
     pub fn try_from_i64(n: i64) -> Result<Self, Error> {
```
tarcieri: Perhaps this should be a proper `TryFrom` now

tendermint/src/block/height.rs
```diff
@@ -7,25 +7,27 @@ use std::{
...
 pub struct Height(u64);
 
 impl Height {
     /// Convert `u64` to block height.
-    ///
-    /// Note that 0 is not a valid block height.
+    /// 
+    /// Note that this method will never error and is just for backwards
+    /// compatibility.
     pub fn try_from_u64(n: u64) -> Result<Self, Error> {
```
tarcieri: I don't think this method needs to exist anymore since the `From` impl always works. Or if you want to keep it around, it should probably just be `new` which returns `Self`

tendermint/tests/integration.rs
```diff
@@ -28,7 +29,12 @@ mod rpc {
     #[test]
     #[ignore]
     fn abci_query() {
-        // TODO(tarcieri): write integration test for this endpoint
+        let key = Path::from_str("unpopulated_key").unwrap();
```
tarcieri: This could alternatively be:

```suggestion
        let key = "unpopulated_key".parse().unwrap();
```

tendermint/src/block/height.rs
mappum: Should I also get rid of the existing panic-able `From<i64>`?

tarcieri: That'd probably be for the best (at the time this was written `TryFrom`/`TryInto` weren't stable)

tendermint/src/block/height.rs
mappum: ✅

tendermint/tests/integration.rs
mappum: ✅

tendermint/src/block/height.rs
mappum: ✅

mappum:
Made the changes, will wait to squash so it's easier to review.

tendermint/src/block/height.rs
```diff
@@ -1,33 +1,19 @@
...
     fmt::{self, Debug, Display},
     str::FromStr,
 };
 
 /// Block height for a particular chain (i.e. number of blocks created since
 /// the chain began)
+/// 
+/// A height of 0 represents a chain which has not yet produced a block.
 #[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub struct Height(u64);
```
tarcieri: You can probably make this `pub` now that it's no longer enforcing an invariant:

```suggestion
pub struct Height(pub u64);
```

mappum: ✅

mappum:
Now all comments are addressed and all commits signed.

tarcieri:
:shipit:

PR #79, opened 2019-11-28
Represent `last_commit` as `Option<Commit>`

yihuang:


yihuang:
> At least a brief description that motivates the changes would be cool next time :-) (even if this a PR isn't to master). Something along the lines of:
> 
> > Without these changes parsing an rpc response with an empty `last_commit` (e.g. on the first block), would err.
> 
> I've tested this locally... CI still doesn't run for PRs from forks :-/
> Thanks again @yihuang!

Thanks, next time, I'll write some description ;D

PR #80, opened 2019-11-28
update signatory to v0.16.0

linfeng-crypto:
update signature to version 0.16 and use signature crate 1.0.0-pre.1


.gitignore
```diff
@@ -9,3 +9,5 @@ Cargo.lock
 # These are backup files generated by rustfmt
 **/*.rs.bk
 
+# IDE
+.idea
```
tarcieri: As a fellow IntelliJ user, I'd recommend  defining a `[core.excludesfile]` in your `~/.gitconfig` so the `.gitignore` doesn't have to incorporate errata for every editor/IDE's metadata/temp files.

linfeng-crypto:
#82

PR #82, opened 2019-12-02
Problem: fails to compile with signatory 0.16

linfeng-crypto:
solution: update signature to version 0.16 and use signature crate 1.0.0-pre.1

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -37,8 +37,10 @@ prost-amino-derive = { version = "0.4.0" }
 rand_os = { version = "0.1" }
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1" }
-signatory = { version = "0.12", features = ["ed25519", "ecdsa"] }
-signatory-dalek = { version = "0.12" }
+signatory = { version = "0.16", features = ["ed25519", "ecdsa"] }
+signatory-dalek = { version = "0.16" }
+signatory-secp256k1 = { version = "0.16" }
+signature = { version = "1.0.0-pre.1", default-features = false }
 sha2 = { version = "0.8", default-features = false }
 subtle = "2"
 subtle-encoding = { version = "0.3", features = ["bech32-preview"] }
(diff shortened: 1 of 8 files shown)


Liamsi:
Could you re-commit with a GPG signed commit and force push? Otherwise we can't merge currently.

Liamsi:
Since recently this started failing with: 

```
  Compiling ed25519-dalek v1.0.0-pre.3
error[E0432]: unresolved import `rand`
  --> /Users/ismail/.cargo/registry/src/github.com-1ecc6299db9ec823/ed25519-dalek-1.0.0-pre.3/src/ed25519.rs:14:5
   |
14 | use rand::{CryptoRng, RngCore};
   |     ^^^^ use of undeclared type or module `rand`

error[E0432]: unresolved import `rand`
  --> /Users/ismail/.cargo/registry/src/github.com-1ecc6299db9ec823/ed25519-dalek-1.0.0-pre.3/src/secret.rs:22:5
   |
22 | use rand::{CryptoRng, RngCore};
   |     ^^^^ use of undeclared type or module `rand`

error: aborting due to 2 previous errors

```
This is unrelated to the changes made here. I ran into the same issue in https://github.com/interchainio/tendermint-rs/pull/63#issuecomment-563572455
I would still hold off merging this until we resolved the issue that causes this.

tarcieri:
Looks like it's due to an upstream release of `ed25519-dalek`. I would guess that including it in `Cargo.toml` with the `rand` feature enabled would fix it (or locking to the previous prerelease).

Liamsi:
Thanks @tarcieri! That works! I'll fix master in a bit and then merge in this PR.

Liamsi:
Just for my understanding. This is due to the `rand` feature which used to be a default feature and is not anymore?

tarcieri:
I think it's an upstream regression

tarcieri:
@linfeng-crypto can you rebase?

Liamsi:
Superseded by #89

PR #83, opened 2019-12-10
Fix build on master

Liamsi:
list d25519-dalek dep, adapt to renaming (simple_merkle hickup), and fmt:

 - #77 introduced an fmt issue
 - #36 was not up-to date regarding changes which were merged to master (renaming
   of simple_merkle)
 - new upstream release of made build of master fail (solution: explicitly add it
 ed25519-dalek (1.0.0-pre.3) with rand feature fixes that

### DIFF
--- a/tendermint/src/block/header.rs
+++ b/tendermint/src/block/header.rs
@@ -1,5 +1,5 @@
 //! Block headers
-use crate::merkle::simple_hash_from_byte_slices;
+use crate::merkle::simple_hash_from_byte_vectors;
 use crate::{account, amino_types, block, chain, lite, Hash, Time};
 use prost::Message;
 use {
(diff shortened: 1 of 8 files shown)


tendermint/src/block/header.rs
```diff
@@ -122,25 +122,25 @@ impl lite::Header for Header {
...
-        byteslices.push(validator_hash_enc.as_slice());
-        byteslices.push(next_validator_hash_enc.as_slice());
-        byteslices.push(consensus_hash_enc.as_slice());
-        byteslices.push(app_hash_enc.as_slice());
-        byteslices.push(last_result_hash_enc.as_slice());
-        byteslices.push(evidence_hash_enc.as_slice());
-        byteslices.push(proposer_address_enc.as_slice());
-
-        Hash::Sha256(simple_hash_from_byte_slices(byteslices.as_slice()))
+        let mut byteslices: Vec<Vec<u8>> = vec![];
```
ebuchman: We should fix this variable name at some point :)

PR #63, opened 2019-11-11
Light client: further trait impls, improvements, and tests

Liamsi:
#59 should be merged first ~~(hence a draft)~~
ref #36 

**Update:** Since this PR was opened (mostly to show the impact of #59), it grew to contain several further PRs added in by @yihuang (big thanks). These added 2 missing impls, resolved a few encoding edge cases and simple (json-based) tests. 

Particularly:
- #67 
- #68 
- #66 
- #69 
- #70
- #72 
- #79

Additionally this PR introduces a `TrustLevel` trait via bd26e74 and adds a `verify` method that is closer to the (current) spec / go-code in 017c1ba (with a few tests from JSON).

PR #84, opened 2019-12-10
Light client: further trait impls, improvements, and tests

Liamsi:
Supersedes  #63: clean-up the commit history and rebased the changes of #63 on master.

From #63's description:
> ref #36 
> 
> **Update:** Since this PR was opened (mostly to show the impact of #59), it grew to contain several further PRs added in by @yihuang (big thanks). These added 2 missing impls, resolved a few encoding edge cases and simple (json-based) tests. 
> 
 Particularly:
 - #67 
 - #68  (can be ignored for reviewing this PR as the `ValidatorSetLookup` trait was deleted later)
 - #66 (this as well as the following two mainly deal with decoding JSON )
 - #69 
 - #70
 - #72  (can be ignored for reviewing this PR because most of its changes got reverted in subsequent commits)
 - #79 (deals with decoding JSON)
> 
> Additionally this PR introduces a `TrustLevel` trait via bd26e74 and adds a `verify` method that is closer to the (current) spec / go-code in 017c1ba (with a few tests from JSON).


tendermint/src/block/header.rs
```diff
@@ -89,61 +94,75 @@ impl lite::Header for Header {
...
+        );
+        byteslices.push(self.last_commit_hash.as_ref().map_or(vec![], encode_hash));
+        byteslices.push(self.data_hash.as_ref().map_or(vec![], encode_hash));
+        byteslices.push(encode_hash(&self.validators_hash));
+        byteslices.push(encode_hash(&self.next_validators_hash));
+        byteslices.push(encode_hash(&self.consensus_hash));
+        byteslices.push(self.app_hash.as_ref().map_or(vec![], encode_hash));
+        byteslices.push(self.last_results_hash.as_ref().map_or(vec![], encode_hash));
+        byteslices.push(self.evidence_hash.as_ref().map_or(vec![], encode_hash));
+        byteslices.push(bytes_enc(self.proposer_address.as_bytes()));
```
Liamsi: rename `byteslices `

Liamsi:
TODOs:
 - [ ] rebase on master
 - [ ] deduplicate by adding a verification method to the `Commit` trait (which takes a validator-set) instead of doing: https://github.com/interchainio/tendermint-rs/blob/7b264815946b540d2b49f0c47ed2a47c45c5387e/tendermint/src/lite/verifier.rs#L120-L133 and https://github.com/interchainio/tendermint-rs/blob/7b264815946b540d2b49f0c47ed2a47c45c5387e/tendermint/src/lite/verifier.rs#L163-L184
- [ ] rename https://github.com/interchainio/tendermint-rs/pull/84#discussion_r356915744 (and other occurrences)?

tendermint/src/lite/verifier.rs
```diff
@@ -34,6 +37,37 @@ where
...
+        if h2.validators_hash() != h1_next_vals.hash() {
+            return Err(Error::InvalidNextValidatorSet);
+        }
+    } else {
+        // ensure that +1/3 of last trusted validators signed correctly
+        if let Err(e) = verify_commit_trusting(&h1_next_vals, commit, trust_level) {
+            return Err(e);
+        }
+    }
+    verify_commit_full(&h2_vals, commit)
```
ancazamfir: maybe?
```suggestion
    verify_header_and_commit(h2, sh2.commit, h2_vals)
```

tendermint/src/lite/verifier.rs
```diff
@@ -66,38 +100,40 @@ where
...
     C: Commit,
+    L: TrustLevel,
 {
     // NOTE it might be more prudent to do the cheap validations first
     // before we even call verify_commit_trusting, but not doing that
     // makes the code cleaner and allows us to just call verify directly.
 
     // ensure that +1/3 of last trusted validators signed correctly
-    if let Err(e) = verify_commit_trusting(&last_validators, &commit) {
+    if let Err(e) = verify_commit_trusting(&last_validators, &commit, trust_level) {
```
ancazamfir: Is this function supposed to work for adjacent headers as well? For those we need a different check, as it is the case in `verify()`. It is not clear to me why we need both `verify()` and `verify_trusting()`.
Can't we use only `verify()` which in the case of sequential is always called with sequential (non skipping) headers while in the case of bisection with both skipping and sequential headers?

Liamsi: Thanks @ancazamfir. I think we can delete `verify_trusting`. I shamelessly adapted some naming from the current go-code without completing it yet (and `verify_trusting` is remainder of that incomplete refactoring). Sorry for the confusion.

I think in this PR it make sense to adapt the naming and logic of the spec instead. In particularly what @milosevic  has been working on here: https://github.com/tendermint/spec/blob/4f7c55507cb99c35c7774c583f182859283f3017/spec/consensus/light-client.md

tendermint/src/lite/verifier.rs
Liamsi: Done in 3d49482. Thanks!

tendermint/src/lite/verifier.rs
Liamsi: Deleted `verify_trusting` for now in 3d49482

tendermint/tests/support/lite/header_tests_failing.json
```diff
@@ -0,0 +1,169 @@
+{
+  "test_cases": [
```
Liamsi: The files with the suffix `_failing` currently aren't ran anywhere! These are test-cases that currently fail ... I included them for the sake of transparency and with the goal to make them pass (or understand why they do not apply in Rust-land while they do in Go, if the case).

tendermint/tests/lite.rs
```diff
@@ -0,0 +1,123 @@
...
+    run_test_cases(cases);
+}
+
+#[test]
+fn header_tests_verify() {
+    let cases: TestCases = serde_json::from_str(&read_json_fixture("header_tests")).unwrap();
+    run_test_cases(cases);
+}
+
+fn run_test_cases(cases: TestCases) {
```
Liamsi: Compare: https://github.com/Shivani912/tendermint/blob/1b0fcf645b1ca67a93a708674db806ce7c8f630e/types/lite-client/tests/lite_client_test.go#L13-L58

tendermint/src/lite/verifier.rs
```diff
@@ -55,84 +87,33 @@ where
...
-        };
-
-        // check vote is valid from validator
-        let sign_bytes = vote.sign_bytes();
-        if !val.verify_signature(&sign_bytes, vote.signature()) {
-            return Err(Error::InvalidSignature);
-        }
-        signed_power += val.power();
-    }
+    let signed_power = commit.voting_power_in(vals)?;
```
Liamsi: @ancazamfir @ebuchman Does this work for you? I think it's closer to what is in the spec and should also be less confusing. It's used here and below in `verify_commit_trusting` (we do lookups in both cases now).

referencing related comments: https://github.com/interchainio/tendermint-rs/pull/36#discussion_r352581861

tendermint/src/lite/types.rs
```diff
@@ -71,19 +81,22 @@ pub trait Validator {
...
     fn header_hash(&self) -> Hash;
 
-    /// Return the underlying votes for iteration.
-    /// All votes here are for the correct block id -
-    /// we ignore absent votes and votes for nil here.
-    /// NOTE: we may want to check signatures for nil votes,
-    /// and thus use an ternary enum here instead of the binary Option.
-    // TODO figure out if we want/can do an iter() method here that returns a
-    // VoteIterator instead of returning a vec
-    fn into_vec(&self) -> Vec<Option<Self::Vote>>;
```
Liamsi: The `into_vec` or any possibility to iterate over the underlying votes is not used in the abstract verifier code. The parts of the code that needed this is now moved into `voting_power_in` (see below). I think this makes the code more readable and closer to the spec.

Liamsi:
I'd appreciate another pass of reviews here. I started working on putting this all together but it would be nice to merge this soon and continue with the remaining parts of the light client (on-top of these changes):
 - initial config for light client 
 - requesting from multiple peers (networking)
 - storing state (memory, DB)
 - detecting misbehaviour, gossiping misbehaviour
 - bisection
 - creating command line tool (light client daemon)

tendermint/tests/support/lite/commit_tests_failing.json
```diff
@@ -0,0 +1,551 @@
...
+              "data_hash": "",
+              "validators_hash": "75B9F27F25F3515EF7CA61DD42C272CCC280C537E6A9ED493710EB361CB90B0D",
+              "next_validators_hash": "75B9F27F25F3515EF7CA61DD42C272CCC280C537E6A9ED493710EB361CB90B0D",
+              "consensus_hash": "048091BC7DDC283F77BFBF91D73C44DA58C3DF8A9CBC867405D8B7F3DAADA22F",
+              "app_hash": "41CAFAE31CC70F5801FA1016A2DD54A9BCB8201B5B389919FE9976762532C516",
+              "last_results_hash": "",
+              "evidence_hash": "",
+              "proposer_address": "01F527D77D3FFCC4FCFF2DDC2952EEA5414F2A33"
+            },
+            "commit": null
```
Liamsi: Currently, this (`null` commit) won't even parse correctly.

tendermint/tests/support/lite/header_tests_failing.json
```diff
@@ -0,0 +1,169 @@
...
+            "proposer_priority": "0"
+          }
+        },
+        "trusting_period": "10800000000000",
+        "now": "2019-11-02T15:05:00Z"
+      },
+      "input": [
+        {
+          "signed_header": {
+            "header": null,
```
Liamsi: Currently, this (null `header`) won't even parse correctly. (In the SignedHeader struct a `header` isn't optional).

tendermint/src/lite/verifier.rs
```diff
@@ -34,8 +32,64 @@ where
...
+    }
+
+    if h2.header().height() == h1.header().height().increment()
+        && h2.header().validators_hash() != h1_next_vals.hash()
+    {
+        return Err(Error::InvalidNextValidatorSet);
+    }
+    verify_commit_trusting(h1_next_vals, h2.commit(), trust_threshold)
+}
+
```
ancazamfir: Looks good. Just a bit more cleanup maybe. It seems that `verify()` and `check_support()` have some dup code. Here is the summary of checks we do for a signed header:

1. is header properly formatted and signed -> this is `verify_header_and_commit()`. In order to check this we verify/ call:
   -  hashes: `validate_vals_and_commit()`
   -  more than 2/3rds have signed: `verify_commit_full()`
2. should the header be trusted -> this is `check_support()` and the checks are:
   - is the header expired?
   - should the validator set that signed the header be trusted?:  
       - for sequential we just check h1 next validator set is same as  h2 validator set. 
       - for skipping we check > 1/3rd have signed -> this is `verify_commit_trusting()`

`verify()` seems to put all these checks together so I think it should just call:
- `verify_header_and_commit()`
- `check_support()`

However, in the spec we do the two checks at different times, mainly because of the nature of bisection that comes back to already verified (as in 1. above) headers. In the spec:
- `verify()` performs 1.,  same as `verify_header_and_commit()` here
- `checkSupport()` performs 2, same as `check_support()` here

Some suggestions, maybe we could have some comments in the beginning of the file that describe this call flow. And maybe order the functions so the code is easier to read from top to bottom.

tendermint/src/lite/verifier.rs
ancazamfir: Yes, it looks better imo.

tendermint/src/lite/verifier.rs
Liamsi: Thanks @ancazamfir! Very useful feedback! 

Yeah, I was thinking of deleting the current existing `verify` function (which lends the naming and logic from the current go implementation) and follow the naming in the spec. I just need to update the tests to call `verify_header_and_commit()` and `check_support()` instead of the current `verify`. Then current `verify` can be deleted and name of `verify_header_and_commit` can be changed to `verify` which then should take a `SignedHeader` instead of (now) a separate `Header` and an associated `Commit`.  Passing in a SignedHeader instead makes the relation of the params clearer, too.

I spend quite some time on trying to get rid of the dependencies from the tendermint crate: Time, Id, Hash. Time was simply replaced by SystemTime. But with Id (which basically needs to implement the `Eq` trait) the mixture of pure generic functions on the one hand, and, traits with associated types (e.g. ValidatorId, ValidatorSet) and methods that return these, on the other hand (e.g. `fn validator(&self, val_id: Id) -> Option<Self::Validator>`) is a really interesting exercise but not an easy one (at least not for me). Almost felt like a very difficult puzzle / riddle :-D Will come back to this later. And do the changes you've mentioned now.

Liamsi: Deleted `verify`, renamed `verify_header_and_commit` to now `verify`. Added a comment saying that you are expected to have called `verify` before calling calling `check_support`. More documentation coming later in the module documentation with an example (render using `cargo doc --open --no-deps`). Thanks again :-)

tendermint/src/lite/verifier.rs
```diff
@@ -34,8 +44,38 @@ where
...
+    trust_threshold: L,
+    trusting_period: Duration,
+    now: SystemTime,
+) -> Result<(), Error>
+where
+    S: SignedHeader,
+    V: ValidatorSet,
+    L: TrustThreshold,
+{
+    if let Err(err) = expired(h1.header(), trusting_period, now) {
```
ebuchman: Is there a way to get rid of the `if let` here and just oneline it - ie. return if its an error, otherwise do nothing? `?` ?

tendermint/src/lite/verifier.rs
```diff
@@ -34,8 +44,38 @@ where
...
+        return Err(err);
+    }
+
+    if h2.header().height() == h1.header().height().increment()
+        && h2.header().validators_hash() != h1_next_vals.hash()
```
ebuchman: What about when the vals do match, shouldn't we just return Ok then?

tendermint/src/lite/types.rs
```diff
@@ -71,19 +91,22 @@ pub trait Validator {
...
+    /// Compute the voting power of the validators that correctly signed the commit,
+    /// have according to their voting power in the passed in validator set.
+    /// Will return an error in case an invalid signature was included.
+    ///
+    /// This method corresponds to the (pure) auxiliary function int the spec:
+    /// `votingpower_in(signers(h.Commit),h.Header.V)`.
+    fn voting_power_in<V>(&self, vals: &V) -> Result<u64, Error>
```
ebuchman: Now that we do this, can't we remove the `Validator` and `Vote` traits all together? They don't seem to be used in this package anymore (just in tests ...) !

tendermint/src/lite/verifier.rs
ebuchman: Great. But won't `votingpower_in` check the signatures again redundantly? We can always address this later, but should track it.

tendermint/src/lite/verifier.rs
```diff
@@ -1,24 +1,34 @@
...
+//! // TODO: add a proper example maybe showing how a `can_trust_bisection`
+//! // looks using the types and methods in this crate/module.
+//!```
+
+use crate::lite::{Commit, Error, Header, SignedHeader, TrustThreshold, ValidatorSet};
+use std::time::{Duration, SystemTime};
 
 /// Returns an error if the header has expired according to the given
 /// trusting_period and current time. If so, the verifier must be reset subjectively.
 /// NOTE: this doesn't belong here. It should be called by something that handles whether to trust
```
ebuchman: Think this comment might be stale?

tendermint/src/lite/verifier.rs
```diff
@@ -55,84 +95,34 @@ where
...
     C: Commit,
 {
     let total_power = vals.total_power();
-    let mut signed_power: u64 = 0;
-
-    let vals_vec = vals.into_vec();
-    let commit_vec = commit.into_vec();
-
-    if vals_vec.len() != commit_vec.len() {
+    if vals.len() != commit.votes_len() {
```
ebuchman: Should we make this part of `validate_vals_and_commit` instead ?

Also, should we make that function a method on `Commit` and then we can drop the need for a `votes_len` and hide the validity checks from this package, as per the spec comment `Additional checks should be done in the implementation to ensure header is well formed`

tendermint/src/validator.rs
```diff
@@ -51,13 +45,33 @@ impl lite::ValidatorSet for Set {
         })
     }
 
-    fn into_vec(&self) -> Vec<Self::Validator> {
-        self.validators.to_vec()
+    fn validator(&self, val_id: account::Id) -> Option<Self::Validator> {
+        self.validators
+            .iter()
+            .cloned()
```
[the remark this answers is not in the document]
yihuang: Oh, this is my code, yeah, switch `cloned` and `find` should more efficient.

tendermint/src/lite/verifier.rs
Liamsi: Yes, indeed. I've opened #96 to track feedback and Todos!

tendermint/src/rpc/endpoint/commit.rs
```diff
@@ -46,3 +50,84 @@ pub struct SignedHeader {
     /// Commit containing signatures for the header
     pub commit: block::Commit,
 }
+
+impl lite::SignedHeader for SignedHeader {
```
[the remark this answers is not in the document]
Liamsi: Moved to `block::signed_header` in https://github.com/interchainio/tendermint-rs/pull/84/commits/5891bae0e41890d58be6fecb484b65f277ef9c59
Might change, hence tracked in #96.

tendermint/src/validator.rs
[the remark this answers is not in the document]
Liamsi: Adressed in: https://github.com/interchainio/tendermint-rs/pull/84/commits/71d28085d388d5dce438459455d09f9694f688c1

tendermint/tests/lite.rs
```diff
@@ -0,0 +1,137 @@
...
+    signed_header: SignedHeader,
+    validator_set: Set,
+    next_validator_set: Set,
+}
+
+pub struct DefaultTrustLevel {}
+impl lite::TrustThreshold for DefaultTrustLevel {}
+
+fn read_json_fixture(name: &str) -> String {
+    fs::read_to_string(PathBuf::from("./tests/support/lite/").join(name.to_owned() + ".json"))
```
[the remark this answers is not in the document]
Liamsi: Added in https://github.com/interchainio/tendermint-rs/pull/84/commits/2f906797591122558852d3f422b357812104487f

Not sure that's really that useful though.

tendermint/src/lite/verifier.rs
Liamsi: Yes, but we'd need to provide a way to from `SystemTimeError` to lite::Error. Tracked in #96.

tendermint/src/lite/verifier.rs
Liamsi: removed in https://github.com/interchainio/tendermint-rs/pull/84/commits/a855701a649dcc097b7e1c66cf52bbe01374978e

tendermint/src/lite/types.rs
Liamsi: Vote can be removed now (done in https://github.com/interchainio/tendermint-rs/pull/84/commits/d907da622afdf5689f35e8d11d9e702419964f5a) 
Validator is currently used via `ValidatorSet`'s associated type, namely because of `fn validator(&self, val_id: Id) -> Option<Self::Validator>;`

tendermint/src/lite/verifier.rs
Liamsi: Good point! tracked in #96

PR #85, opened 2019-12-11
Upgrade to hyper v0.13; use async/await

tony-iqlusion:
The crate itself builds with these changes, but the tests don't.

I'm not sure how to write tests for this (it's my first time using `async`/`await` in Rust 😅)

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -32,9 +32,11 @@ circle-ci = { repository = "interchainio/tendermint-rs" }
 
 [dependencies]
 bytes = "0.4"
+bytes_0_5 = { version = "0.5", package = "bytes" }
 chrono = { version = "0.4", features = ["serde"] }
 failure = "0.1"
-hyper = { version = "0.10" }
+http = "0.2"
+hyper = "0.13"
 prost-amino = { version = "0.4.0" }
 prost-amino-derive = { version = "0.4.0" }
 rand_os = { version = "0.1" }
@@ -52,4 +54,5 @@ zeroize = { version = "1.0", features = ["zeroize_derive"] }
 ed25519-dalek = {version = "1.0.0-pre.3", features = ["rand"]}
 
 [dev-dependencies]
+futures = "0.3"
 serde_json = "1"
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -54,5 +54,5 @@ zeroize = { version = "1.0", features = ["zeroize_derive"] }
 ed25519-dalek = {version = "1.0.0-pre.3", features = ["rand"]}
 
 [dev-dependencies]
-futures = "0.3"
 serde_json = "1"
+tokio = "0.2"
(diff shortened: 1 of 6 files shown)


tony-iqlusion:
I tried the obvious thing (and what I'd do in TypeScript): add `async` to my tests. Unfortunately...

```
error: async functions cannot be used for tests
```

tony-iqlusion:
Well, I got it to compile locally, but the tests don't run:

```
---- rpc::net_info stdout ----
thread 'rpc::net_info' panicked at 'no current reactor', /Users/bascule/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.4/src/io/driver/mod.rs:243:21
```

Liamsi:
I briefly looked into `#[tokio::test]` but it seams that the whole body of the test would needs to be async (which isn't the case for the current integration tests). 

Instead I got sth like this working locally:

```rust
mod rpc_asyc {
    use tendermint::rpc::Client;
    use tokio::runtime::Builder;
    use core::future::Future;

    fn block_on<F: Future>(future: F) -> F::Output {
        Builder::new()
            .basic_scheduler()
            .enable_all()
            .build()
            .unwrap()
            .block_on(future)
    }
    
    #[test]
    fn my_test() {
        let cli = block_on(Client::new(&"tcp://127.0.0.1:26657".parse().unwrap())).unwrap();
        let net_info = block_on(cli.net_info()).unwrap();
        assert!(net_info.listening);
    }
}
```

Liamsi:
FWIW I created a PR against your branch with minimal changes here: #87 (feel free to ignore/close it if you settle with another aproach)

tarcieri:
Cool, I merged #87 into this branch. It appears to be working 🎉 

I think it might be worth fixing the HTTP error handling too, which I'll take a quick stab at

tony-iqlusion:
Added some basic non-panicking HTTP error handling in a50b503.

Unfortunately I don't have a local Tendermint to test against at the moment so I'm not sure it actually works. @Liamsi were the tests passing locally for you?

After `cosmoshub-3` launches I'll give it a try against that.

Liamsi:
>  @Liamsi were the tests passing locally for you?

Most of the integration tests pass but passed two didn't  (`abci_info`, `genesis`). 
- genesis fails with: called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("invalid digit found in string at line 11 column 26") }
 - abci_info with: called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("missing field `last_block_height` at line 9 column 5") }

tony-iqlusion:
@Liamsi aah cool, well I'm guessing those weren't regressions then, so at least we're no worse off.

Can you approve this PR if you think it looks good?

Liamsi:
Yes, I think it's OK to fix those regressions in another PR.

PR #87, opened 2019-12-11
Fix async tests by using tokio runtime

Liamsi:
Uses tokio runtime builder to create a `block_on` method (depends on #85).

Tested locally by re-enabling some of the ignored tests: most pass, some do not. Looks like the JSON format changed in the meantime.

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -32,9 +32,11 @@ circle-ci = { repository = "interchainio/tendermint-rs" }
 
 [dependencies]
 bytes = "0.4"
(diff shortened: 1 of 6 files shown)


Liamsi:
Thanks for reviewing @xla!

> The only thing that we need to be aware of is, if there is another runtime in the system which doesn't share the same pool, it can lead to halts and locks.

Is this sth. we should be concerned about in the context of this PR? Is that easy to get wrong in tests (create multiple Runtimes with distinct thread-pools)? I do not understand enough about the internals of the runtime to understand when we would need to worry about this.

Liamsi:
This got manualy merged into #85. Closing.

PR #89, opened 2019-12-11
Upgrade to `signatory` v0.17

tarcieri:


tarcieri:
note: this supersedes #82

PR #90, opened 2019-12-11
Replace `rand_os` with `getrandom` (closes #23)

tarcieri:
`rand_os` is just a wrapper for `getrandom`. I believe the long-term plan is to deprecrate it entirely.


tarcieri:
rustfmt is fixed

PR #92, opened 2019-12-11
Release/v0.11.0

ebuchman:
Added everything from v0.10. Moving forward we need to enforce PRs include these themselves. Perhaps we should adopt the conventions from Tendermint Core as we're accustomed to it. 

Bumped the version in the Cargo.toml and the lib.rs - anything else needed? Not sure exactly what's required in a release.

Also should we wait to include #84?


Liamsi:
> Also should we wait to include #84?

I don't think it is necessary wait for this. We can do a next release as soon as we have a fully working light client instead?

Liamsi:
> Bumped the version in the Cargo.toml and the lib.rs - anything else needed? Not sure exactly what's required in a release.

As far as I remember that's were the only necessary changes in the repo (maybe @tarcieri has sth to add). I followed the procedure described [here](https://doc.rust-lang.org/cargo/reference/publishing.html) to publish the crate (but it's been a while).

Liamsi:
> Moving forward we need to enforce PRs include these themselves. Perhaps we should adopt the conventions from Tendermint Core as we're accustomed to it.

Good point! Yeah, let's follow all the established conventions from tendermint core where it makes sense 🎉

ebuchman:
Great I opened it for review/approval!

CHANGES.md
```diff
@@ -1,3 +1,63 @@
...
+- Remove `secret_connection` and `ring` as dependencies (moved to KMS repo)
+  ([#60])
+- `tai64` from `2` to `3` ([#22])
+- `zeroize` from `0.9` to `1.1` ([#74, #89])
+- `hyper` from `0.10` to `0.13` ([#85])
+- `signatory` from `0.12` to `0.17` ([#89])
+- `subtle-encoding` from `0.3` to `0.5` ([#47])
+- `uuid` from `0.7` to `0.8` ([#91])
+- replace `rand_os` with `getrandome` ([#90])
+- pin to `ed25519-dalek` `1.0.0-pre.3` ([#83])
```
Liamsi: This change was reverted by updating to signatory 0.17 (which now [does the pinning](https://github.com/tendermint/signatory/pull/185) for us). Do you still want to include it here?

CHANGES.md
```diff
@@ -1,3 +1,63 @@
...
+
+- Remove `secret_connection` and `ring` as dependencies (moved to KMS repo)
+  ([#60])
+- `tai64` from `2` to `3` ([#22])
+- `zeroize` from `0.9` to `1.1` ([#74, #89])
+- `hyper` from `0.10` to `0.13` ([#85])
+- `signatory` from `0.12` to `0.17` ([#89])
+- `subtle-encoding` from `0.3` to `0.5` ([#47])
+- `uuid` from `0.7` to `0.8` ([#91])
+- replace `rand_os` with `getrandome` ([#90])
```
Liamsi:
```suggestion
- replace `rand_os` with `getrandom` ([#90])
```

CHANGES.md
tarcieri: Yeah, I effectively reverted that and opened an upstream issue here (which has since been fixed): https://github.com/dalek-cryptography/ed25519-dalek/issues/108

tendermint/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name       = "tendermint"
-version    = "0.10.0" # Also update `html_root_url` in lib.rs when bumping this
+version    = "0.11.0" # Also update `html_root_url` in lib.rs when bumping this
 license    = "Apache-2.0"
 homepage   = "https://www.tendermint.com/"
 repository = "https://github.com/tendermint/kms/tree/master/crates/tendermint"
```
Liamsi: Huh, didn't I update the repository here? Replace with 
```suggestion
repository = "https://github.com/interchainio/tendermint-rs/tree/master/tendermint"
```

CHANGES.md
ebuchman: Duh thanks will remove.

tarcieri:
:shipit:

PR #98, opened 2019-12-17
Fix serializer and deserializer not symmetric

yihuang:
These deserializers intents to parse some empty data as `None`, because golang version of tendermint would encode these empty values as some default values.
But since the serializers would encode `None` as `null`, which makes them not symmetric.
So the solution is make these deserializers also support decoding `null` as `None`.

### DIFF
--- a/tendermint/src/block.rs
+++ b/tendermint/src/block.rs
@@ -59,15 +59,18 @@ where
         pub signatures: Option<CommitSigs>,
     }
 
-    let commit = TmpCommit::deserialize(deserializer)?;
-    if commit.block_id.is_none() || commit.signatures.is_none() {
-        Ok(None)
+    if let Some(commit) = <Option<TmpCommit>>::deserialize(deserializer)? {
+        if let Some(block_id) = commit.block_id {
+            Ok(Some(Commit {
+                height: commit.height,
+                round: commit.round,
+                block_id,
+                signatures: commit.signatures.unwrap_or_else(|| CommitSigs::new(vec![])),
+            }))
+        } else {
+            Ok(None)
+        }
     } else {
-        Ok(Some(Commit {
-            height: commit.height,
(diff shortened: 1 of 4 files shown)


tendermint/src/block.rs
```diff
@@ -51,13 +51,16 @@ where
         precommits: Option<Precommits>,
     }
 
-    let commit = TmpCommit::deserialize(deserializer)?;
-    if commit.block_id.is_none() || commit.precommits.is_none() {
-        Ok(None)
+    if let Some(commit) = <Option<TmpCommit>>::deserialize(deserializer)? {
```
Liamsi: Cool, thanks! I want to double check if the tendermint rpc will return an empty commit (likely), or, if this can be treated as invalid input instead.

yihuang: An empty commit happens in the `last_commit` field in the first block.

ebuchman:
Thanks for more fixes. Do you think you could add some tests to help illustrate the issue better? Would probably be good to get a bunch of json data for blocks so we can tests all these encoding issues for first block, blocks with/without other hashes. Maybe in `tendermint/tests/support/rpc/block.json` ?

yihuang:
> Thanks for more fixes. Do you think you could add some tests to help illustrate the issue better? Would probably be good to get a bunch of json data for blocks so we can tests all these encoding issues for first block, blocks with/without other hashes. Maybe in `tendermint/tests/support/rpc/block.json` ?

I've added the unit tests which reveals the issue, and rebased against current master.
Currently, I only concern with `Block` and `TrustedState`, but I believe there are other several cases have a similar issue.
This issue happens when you want to serialization some data types into storage and load/deserialize them later.

greg-szabo:
This is implemented with #283 . Thanks @yihuang for your patience with this. Closing.

PR #102, opened 2019-12-23
Add verify_new_header

yihuang:
Add `verify_new_header` to combine the verifications of lite client logic when syncing new header.
Changed `TrustedState` to trait.

### DIFF
--- a/tendermint/src/block/signed_header.rs
+++ b/tendermint/src/block/signed_header.rs
@@ -4,7 +4,7 @@
 use crate::{
     block, hash, lite,
     lite::types::Validator,
-    lite::{Error, ValidatorSet},
+    lite::{ValidatorSet, VerifyError},
     vote::SignedVote,
 };
 use serde::{Deserialize, Serialize};
@@ -58,7 +58,7 @@ impl lite::Commit for SignedHeader {
         self.commit.block_id.hash
     }
 
-    fn voting_power_in<V>(&self, validators: &V) -> Result<u64, Error>
+    fn voting_power_in<V>(&self, validators: &V) -> Result<u64, VerifyError>
     where
         V: ValidatorSet,
     {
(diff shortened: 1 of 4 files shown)


ebuchman:
Thanks for this, but take a look at the new approach in https://github.com/interchainio/tendermint-rs/pull/114.

Note sure the `Error -> VerifyError` is helpful since `Verify` prefix doesn't really tell us much. We should make some improvements to the error systems though - we've started tracking some ideas in #110 . The `impl Display` from this PR would be good to preserve too if you want to submit that in another PR. Will close for now, but let us know if you think something else from this should be included.

yihuang:
The `Error` is a conflict with `std::error::Error`, the `Display` trait is a requirement of `Error` trait, that's it.
I'll look at the new approach, thanks.

tarcieri:
To avoid `Error` conflicting with `std::error::Error`, you can refer to the latter using its fully qualified name.

If you use a crate like [`thiserror`](https://crates.io/crates/thiserror) to impl `Display` and `std::error::Error` using a proc macro, you don't even have to worry about referring to `std::error::Error` directly, as it will use the proc macro to write the impl for you.

yihuang:
> To avoid `Error` conflicting with `std::error::Error`, you can refer to the latter using its fully qualified name.
> 
> If you use a crate like [`thiserror`](https://crates.io/crates/thiserror) to impl `Display` and `std::error::Error` using a proc macro, you don't even have to worry about referring to `std::error::Error` directly, as it will use the proc macro to write the impl for you.

If you wrap it inside your own `Error` type, you can implement whatever trait your like for it, but if you use the `tendermint-rs` one directly and hoping it has some trait which is not implemented, there's nothing you can do. I just think there is no reason the `Error` type is not an instance of `std::error::Error`.

tarcieri:
It definitely should `impl std::error::Error`. I'm just saying you don't need to change its name to do that.

Issue #96, opened 2019-12-16
Light client impl followup 


Liamsi:
Tracking TODOs for the Light client impl

Feedback / Review of #84 
 - [x] Move additional checks  to `validate_vals_and_commit` and make it function of `Commit`(?): https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358019467 - done in #109
 - [ ] redundant check for sigs: https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358018987, https://github.com/tendermint/spec/pull/61#discussion_r358015568
 - [ ] https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358008529 (and general organizational issues: e.g. crate vs module etc)
   - remove dependencies on tendermint types (account::Id, Hash)
- [ ] add example to doc: https://github.com/interchainio/tendermint-rs/pull/84#discussion_r357989612
- [x] `From<SystemTimeEror> for lite::Error` and  https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358017746. Done in #109
- [x] https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358018000 (return OK?) - Done in #109
- [x] remove `Validator` trait ? https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358018253 - done in #105 
- [x] how to handle if/when votes don't contain validator id? probably worth discussing in separate issue: https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358008934 - done, added a comment in #109 
- [ ] next steps: https://github.com/interchainio/tendermint-rs/pull/84#issuecomment-565229546
- [ ] rename types.rs to interface.rs - https://github.com/interchainio/tendermint-rs/pull/100#issuecomment-567903975 
- [ ] consider Client type and have can_trust be a method on the Client. Also put trust_threshold and trusting_period in the trusted state? https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360484471
- [ ] enforce off-by-one issues in the type system if possible, eg. TrustedState::new taking header h-1 and validators for h. https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360338190. Also verify taking vals for the right height https://github.com/interchainio/tendermint-rs/pull/100/#discussion_r360493104.
- [ ] Error::RequestFailed should wrap underlying requester errors. https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360341870
[...]

PR #104, opened 2019-12-27
MockHeader and test_is_within_trust_period

ebuchman:
Initial mock and test for the Header and `is_within_trust_period`.

I hesitated on mocking the Commit because it ultimately requires Validator, which has a `verify_signature` method, which I think we'd like to avoid needing to test this core logic (ie. https://github.com/interchainio/tendermint-rs/issues/96#issuecomment-569360555).

We should consider some more changes to the traits to eliminate the need for this. It will make mocking everything much easier! 

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/block/signed_header.rs
+++ b/tendermint/src/block/signed_header.rs
@@ -1,12 +1,7 @@
 //! SignedHeader contains commit and and block header.
 //! It is what the rpc endpoint /commit returns and hence can be used by a
 //! light client.
-use crate::{
-    block, hash, lite,
-    lite::types::Validator,
-    lite::{Error, ValidatorSet},
-    vote::SignedVote,
-};
+use crate::{block, hash, lite, lite::Error, validator::Set, vote::SignedVote};
 use serde::{Deserialize, Serialize};
 
 /// Signed block headers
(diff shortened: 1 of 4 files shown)


tarcieri:
> We should consider some more changes to the traits to eliminate the need for this. 

I would definitely recommend using traits for these sorts of interfaces and making your mocks a struct which impls the trait you'd like to mock.

Libra pervasively uses this pattern, and the other added benefit is it makes the code more easily (re)composable.

ebuchman:
@tarcieri we're already using traits for everything. The issue here is that mocking the trait would be too involved, because we need to mock a `verify_signature` method, and it's a method that's not even used in the module! 

I think I solved it here: https://github.com/interchainio/tendermint-rs/pull/105. The solution was to make one trait (ValidatorSet) an associated type of the other (Commit). That eliminates the need for the `verify_signature` method to be exposed in the traits, because the Commit can access the underlying functionality of its associated ValidatorSet trait without being restricted to the trait interface. Wahoo!

Liamsi:
Merging this into #96's branch. Will fix the build later and go through the changes of #107.

PR #105, opened 2019-12-28
remove Validator trait; make ValidatorSet associated to Commit

ebuchman:
As per #96, this removes the Validator trait by making ValidatorSet an associated type of Commit. As a bonus, we eliminate dependency on the `Id` type from the light module.

This allows implementations of Commit to access underlying functionality of a validator set (like its validators and signature verification with them) without requiring that functionality to be part of the light client module. This should also make it much easier to write tests for the module now, since we don't need to mock a `verify_signature`.

I'm not sure if this is the best way to do things, but it seems to be along the right lines ...

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/lite/types.rs
+++ b/tendermint/src/lite/types.rs
@@ -7,7 +7,6 @@
 // as_bytes(&self) -> &[u8] method. It's unlikely that a hash won't be
 // representable as bytes, or an Id (that is basically also a hash)
 // but this feels a a bit like cheating
-use crate::account::Id;
 
 use crate::block::Height;
 use crate::Hash;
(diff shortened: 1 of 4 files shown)


tendermint/src/lite/types.rs
```diff
@@ -1,14 +1,6 @@
...
 //! verification logic in `super::verifier` for a light client.
 
-// TODO can we abstract this away and use a generic identifier instead ?
-// Ie. something that just implements Eq ?
-// (Ismail): a really easy solution would be have a trait that expects an
-// as_bytes(&self) -> &[u8] method. It's unlikely that a hash won't be
-// representable as bytes, or an Id (that is basically also a hash)
-// but this feels a a bit like cheating
-use crate::account::Id;
-
```
Liamsi: NIce!

Liamsi:
OK, merging this into #104 and #104 into #96.

Issue #113, opened 2019-12-29
Light Client Testing


ebuchman:
The light client is coming together well, and has some integration tests starting from JSON data generated from Tendermint in Go to test the `verify` and `check_support` functions (see `tests/lite.rs`).

Now that the abstractions are cleaner, we should be adding unit tests to the `lite` module that use simple mocks for all the traits. These mocks should look basically like what we find in the TLA+ spec, ie. where validator ids are simply integers and a commits just contain the ids of the validators that signed. 

As per #110 and #111, we should simplify the public API and have a single primary public function for trying to update the trusted state without making any intermediate requests. The bisection function should use this function. Then we shouldn't need to test `verify` and `check_support` independently, as they'd be covered by this new function. Then we'd have two kinds of tests:

1) tests of the new function which cover a "single step". this would be effectively what IBC would use. We'd test for errors like expiration, invalid data, not enough voting power, etc, and we'd test various non-error cases
2) tests for bisection (ie. the current `verify_header` function).  We'd build some mock of a blockchain and instantiate a Requester with it so we can test bisection against different possible existing chains. We'd want to ensure the set of requested headers is correct, as well as all verification results.

In both cases, the tests should map clearly to possible traces from TLC, so that we can generate new tests by running TLC. 

We also need a third class of tests:

3) tests for the specific implementations of the traits, primarily of `voting_power_in`. This would check various actual instantiations of the Tendermint Commit structure and a Validator Set. Would need to ensure nil-votes don't count towards the voting power, and that we error on invalid signatures and on signatures for the wrong block id.

Liamsi:
The context for:

> these mocks should look basically like what we find in the TLA+ spec, ie. where validator ids are simply integers and a commits just contain the ids of the validators that signed.

and 

> In both cases, the tests should map clearly to possible traces from TLC, so that we can generate new tests by running TLC.

seems to be this: https://github.com/interchainio/verification/tree/develop

ebuchman:
Some of the description here is obsolete by now, but the general fact that we need the testing is not. Also this is prophetic:

> tests for the specific implementations of the traits, primarily of voting_power_in. This would check various actual instantiations of the Tendermint Commit structure and a Validator Set. Would need to ensure nil-votes don't count towards the voting power, and that we error on invalid signatures and on signatures for the wrong block id.

Given #282 is exactly this bug, which of course we're not testing ...

Issue #111, opened 2019-12-29
Light Client Spec related follow up


ebuchman:
Replacement for elements of #96 more specific to specification. These need to be updated synchronously with the spec:

- [ ] capture the validation functions more explicitly in the spec and align with their use in the code
- [ ] redundant check for sigs (https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358018987, https://github.com/tendermint/spec/pull/61#discussion_r358015568)
- [x] Improve naming of variables like `h1_xxx` and `h2` (https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360342976, https://github.com/interchainio/tendermint-rs/pull/100#discussion_r36034392)
- [x] how to get latest height from a node? use height=0 ? Could get rid of get_smaller_or_equal (https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360470376)
- [ ] improve logic of check_support to better clarify the 3 conditions: ok (update state), real error (faulty node or system), not-enough-voting-power (not real error, just need to bisect)
- [x] possible move validity checks from check_support to valdiate_vals_and_commit ...
- [x] add a single function that can be called by IBC that doesn't use a requester but otherwise does all checks necessary to update the store with a new trusted state. Ideally this function can be used by a function that does use the Requester

ebuchman:
Alot of this was addressed on the Rust side in #114

ebuchman:
These may be obsoleted by now but we should double check

brapse:
All these  requirements look implementation specific except the last one which will be better captured by a standalone issue @milosevic is working on. Closing for now but feel free to re-open with clarifications.

Issue #110, opened 2019-12-28
Light Client Rust related follow up


ebuchman:
Replacement for elements of #96 more specific to Rust implementation style and architecture.

We should update the ADR with the relevant changes from here as well.

Types:
- [x] Make `validate_vals_and_commit` a method on Commit (https://github.com/interchainio/tendermint-rs/pull/84#discussion_r358019467)
- [x] Use `validate_next_vals` !
- [ ] Introduce a `Client` type to hold `Store` and `Requester` (?)  (https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360484471)
- [ ] Put `trust_threshold` and `trusting_period` in the Store (?) (https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360484471)
- [ ] Enforce off-by-one issues in the type system, eg.:
    - [ ] `TrustedState::new` taking header h-1 and validators for h (https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360338190)
    - [ ] `verify` taking vals for the right height (https://github.com/interchainio/tendermint-rs/pull/100)
- [x] make `trust_level` satisfy Copy or Clone and don't pass by reference (https://github.com/interchainio/tendermint-rs/pull/100)
- [ ] memoize results of methods that don't take args, eg. `.hash()` (https://github.com/interchainio/tendermint-rs/issues/96#issuecomment-567003346)
- [ ] use an enum for height to get the latest state instead of special-casing 0 (#118)
- [ ] introduce a simple event type and let bisection function (or a Client structure it's a method on) take a closure that will fire events. this can be used for logging so we don't need a print statement or logger inside the functions

Module Structure:
- [ ] Remove dependency on `tendermint` types, eg. `Hash`
- [ ] Make it a crate ? 
- [ ] Rename types.rs to interface.rs (https://github.com/interchainio/tendermint-rs/pull/100#issuecomment-567903975)
- [ ] Add example to doc 
- [x] Simplify the public API and make it harder/impossible to use incorrectly
[...]

PR #114, opened 2019-12-29
New public API for the light client

ebuchman:
As per some points in #110, #111, and #113 , I think I've hit upon a much simplified logic for the light client and the bisection. This diverges from the spec and thus shouldn't be merged yet, but I think we can update the spec to match this and have something easier to reason about all around.

The essential idea is to have two primary public functions: 

- `verify_and_update_single`: This attempts to verify a single header and update the store. It does not use a requester (all data is provided as args). It performs all checks and verification. It is intended for use by IBC handlers.
- `verify_and_update_bisection`: This attempts to verify and update to some height, using a requester to request intermediate heights as necessary, according to bisection. It is intended for use by a light node. 

Both functions use a common pure function called `verify_single` which is not exposed, but which performs the core verification logic to determine if we can update from one header to another. It is not exposed because it does not check for expiry. Expiry checking is the responsibility of the public functions.

As per #113, we can now focus most of our testing on `verify_single`, as this is the work horse of both public functions. For `verify_and_update_single`, we should additionally test for expiry and for properly updating the store. For `verify_and_update_bisection`, we should additionally test for expiry and properly updating the store with the right set of intermediate headers.

Note that by making all the validation and verification explicit in `verify_single`, we eliminate the need for the `verify` and `check_support` functions. Instead we call `verify_commit_full` and `verify_commit_trusting` directly in `verify_single`. We can then delete all of `verify`, `check_support`, `can_trust_bisection` and `verify_header`. None of the functions in `verifier.rs` should be exposed to users of the crate.

Note for simplicity I'm also assuming here we can get the latest trusted state by requesting height 0 (see #111).

* [ ] Wrote tests
* [ ] Updated CHANGES.md


tendermint/src/lite.rs
```diff
@@ -1,7 +1,12 @@
 //! Core logic and traits of a light client.
 
+// TODO: don't expose the verifier, just the types and the public API.
+// the verifier has low level functions that can be used incorrectly.
+
+pub mod public;
```
Liamsi: I think for now it is cool to have that distinction and naming (public vs the rest). We should come up with another name though (maybe `verification`?) and not export the other modules which contain methods that can be used incorrectly.

ebuchman: Yep, just didn't get around to unexposing the other pieces. Are you suggesting we have a `verifier` and a `verification` module and only expose the latter ? I also don't think it'd be that bad if we consolidated the verifier.rs and public.rs into a single `verifier` module and just expose only the specific functions in `lite.rs`

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+) -> Result<(), Error>
+where
+    TS: TrustedState<LastHeader = SH, ValidatorSet = C::ValidatorSet>,
+    SH: SignedHeader<Commit = C>,
+    C: Commit,
+    L: TrustThreshold,
+    S: Store<TrustedState = TS>,
+{
+    // fetch the latest state and ensure it hasn't expired
+    let trusted_state = store.get(Height::from(0))?;
```
Liamsi: OK, we need to update the docu comment for Store::get if we settle for the 0 means latest semantic.

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+    // verify the untrusted commit
+    verify_commit_full(untrusted_vals, untrusted_sh.commit())
+}
+
+/// Attempt to update the store to the given untrusted header.
+/// Ensures our last trusted header hasn't expired yet, and that
+/// the untrusted header can be verified using only our latest trusted
+/// state from the store.
+/// This function is primarily for use by IBC handlers.
+pub fn verify_and_update_single<TS, SH, C, L, S>(
```
Liamsi: I think the naming is much clearer.

tendermint/src/lite/public.rs
ebuchman: Yep, not 100% sure how to manage process around changing semantics from a PR. Probably should happen in an ADR/spec first, be agreed there, and then percolate. Lots of things are happening backwards but thats life and we're learning.  Anyways, it's tracked in #111, and we should probably get input from folks on the spec side before we accept/merge this.

tendermint/src/lite.rs
Liamsi: > Are you suggesting we have a verifier and a verification module and only expose the latter ? 

Nope, I just think `public` is confusing and there should only be one module dealing with verification (`verifier` or `verification`).

>  I also don't think it'd be that bad if we consolidated the verifier.rs and public.rs into a single verifier module and just expose only the specific functions in lite.rs

Yes, that sounds right! 👍 (and was more what I meant, too)

ebuchman:
Awesome thanks - would love a review from @ancazamfir / @josef-widder / @milosevic before we merge since we'll then need to update the spec to match. If it seems good, then here we'd still need:

- [ ] Remove the old functions
- [ ] Update docs for the `.get(0)` semantics and remove `get_smaller...`
- [ ] Unexpose all the other functions in `lite.rs`

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+{
+    // Fetch the latest state and ensure it hasn't expired.
+    // Note when calling itself recursively, this check is redundant,
+    // but leaving it keeps things simple ...
+    let trusted_state = store.get(Height::from(0))?;
+    let trusted_sh = trusted_state.last_header();
+    is_within_trust_period(trusted_sh.header(), trusting_period, now)?;
+
+    // fetch the header and vals for the new height
+    let untrusted_sh = &req.signed_header(untrusted_height)?;
```
milosevic: Where do we check if downloaded header is within trusted period?

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+    SH: SignedHeader<Commit = C>,
+    C: Commit,
+    L: TrustThreshold,
+    R: Requester<SignedHeader = SH, ValidatorSet = C::ValidatorSet>,
+    S: Store<TrustedState = TS>,
+{
+    // Fetch the latest state and ensure it hasn't expired.
+    // Note when calling itself recursively, this check is redundant,
+    // but leaving it keeps things simple ...
+    let trusted_state = store.get(Height::from(0))?;
```
milosevic: Do we want to check for a case where we already have a header for untrusted_height in the store?

tendermint/src/lite/public.rs
ebuchman: We don't. The thinking is that our local `now` is close to the latest blockchain time, and if `now` is within the trusted period, so is the blockchain time. Also once the new height is stored, it will be checked for expiry the next time we try to update.

That said, it's probably not a bad idea to check that its within trusting period from the last trusted state, or, what should be equivalent, that its less than `now + X`, where X is our allowed variance between our local time and the blockchains latest bft time.

tendermint/src/lite/public.rs
ebuchman: We do a check like this in `verify_single`, where the height we're trying to update to must be greater than our latest height. If not we error. I think this can be enforced by the consumer of this package, so that we that we only try to sync to higher heights. But that should be made clearer in the API and I could go either way on it ...

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+        req,
+        store,
+    )?;
+
+    // Recursive call to update to the original untrusted_height.
+    verify_and_update_bisection(
+        untrusted_height,
+        trust_threshold,
+        trusting_period,
+        now,
```
milosevic: Are we safe passing at this point in time now? It has passed quite some time since original call.

ebuchman: I think the use of `now` in the recursion is redundant since we're not getting a new source of `now`, unless we want this API to include a clock or somethign that can provide multiple instances of now. Hence the comment above:

```
    // Fetch the latest state and ensure it hasn't expired.
    // Note when calling itself recursively, this check is redundant,
    // but leaving it keeps things simple ...
```

We could refactor a bit so that there's an inner bisection function that is called recursively that doesn't take the now. Happy to do that if it makes things clearer

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+    let trusted_height = trusted_state.last_header().header().height();
+    if untrusted_height <= trusted_height {
+        // TODO: return err
+    };
+
+    // validate the untrusted header against its commit, vals, and next_vals
+    let untrusted_header = untrusted_sh.header();
+    let untrusted_commit = untrusted_sh.commit();
+    validate_vals_and_commit(untrusted_header, untrusted_commit, untrusted_vals)?;
+    validate_next_vals(untrusted_header, untrusted_next_vals)?;
```
milosevic: Why nextVals are not checked as part of validate_vals_and_commit function? This function is just supposed to check if hashes of these data structures match data in the header, so it might be simpler doing all those kind of checks in the same function.

tendermint/src/lite/public.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    // if the new height is not sequential, check if we can skip
+    if untrusted_height > trusted_height.increment() {
+        let trusted_vals = trusted_state.validators();
+        verify_commit_trusting(trusted_vals, untrusted_commit, trust_threshold)?;
```
milosevic: In case of sequential headers we need to check also if 'if trusted_header.NextValidatorsHash != untrusted_header.ValidatorsHash'.

tendermint/src/lite/public.rs
ebuchman: yep for sure these should be consolidated. we actually weren't even calling validate_next_vals before

tendermint/src/lite/public.rs
ebuchman: Good catch!

tendermint/src/lite/public.rs
ebuchman: I added a comment about this in 60c8b16059c21ecbd72319ded47dcb68b2428cb0

tendermint/src/lite/public.rs
ebuchman: Addressed in 60c8b16059c21ecbd72319ded47dcb68b2428cb0 with better docs.

tendermint/src/lite/public.rs
ebuchman: Addressed in 60c8b16059c21ecbd72319ded47dcb68b2428cb0 by breaking out an inner recursive function that doesn't take `now`. This should make it clear we only check `now` once.

tendermint/src/lite/public.rs
ebuchman: done in 93209f432246031e23e0e59a2c19f178feebe789

tendermint/src/lite/public.rs
ebuchman: fixed in 39be6313e4861ecc836ebe47e8e11a2512f88115

ebuchman:
Ok, I merged https://github.com/interchainio/tendermint-rs/pull/115 into this so we have some tests, and addressed all the feedback above. There's a bunch of TODO's in the PR that should be broken out into one or more issues, but I think we can otherwise move forward with merging.

tendermint/src/lite/verifier.rs
```diff
@@ -36,80 +37,30 @@ where
...
-
-/// Validate the validators and commit against the header.
+/// Validate the validators, next validators, and commit against the header.
 // TODO(EB): consider making this a method on Commit so the details are hidden,
 // and so we can remove the votes_len() method (that check would be part of the
 // methods implementation). These checks aren't reflected
 // explicitly in the spec yet, only in the sentence "Additional checks should
 // be done in the implementation to ensure header is well formed".
-fn validate_vals_and_commit<H, V, C>(header: &H, commit: &C, vals: &V) -> Result<(), Error>
+pub fn validate_signed_header_and_vals<H, V, C>(
```
xla: Seeing a name including an `and` is typically a sign that the name just reiterates the steps inside the function body instead of conveying the higher level utility/purpose provided.

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
-/// based on the given trusted state, otherwise returns an Error.
-fn can_trust_bisection<TS, SH, C, L, S, R>(
-    trusted_state: &TS,    // h1 in spec
-    untrusted_header: &SH, // h2 in spec
+// Verify a single untrusted header against a trusted state.
+// Includes all validation and signature verification.
+// Not publicly exposed since it does not check for expiry
+// and hence it's possible to use it incorrectly.
+// If trusted_state is not expired and this returns Ok, the
+// untrusted_sh and untrusted_next_vals can be considered trusted.
```
xla: These comments should be proper doc strings prefixed with `///`.

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
     SH: SignedHeader<Commit = C>,
     C: Commit,
     L: TrustThreshold,
-    S: Store<TrustedState = TS>,
-    R: Requester<SignedHeader = SH, ValidatorSet = C::ValidatorSet>,
 {
-    // can we trust the still untrusted header based on the given trusted state?
-    match check_support(
-        trusted_state,
+    // ensure the new height is higher
```
xla: For what code is this comment?

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
+    verify_commit_full(untrusted_vals, untrusted_sh.commit())
+}
 
-    verify(&pivot_header, &pivot_vals)?;
+/// Attempt to update the store to the given untrusted header.
+/// Ensures our last trusted header hasn't expired yet, and that
+/// the untrusted header can be verified using only our latest trusted
+/// state from the store.
+/// This function is primarily for use by IBC handlers.
+pub fn verify_and_update_single<TS, SH, C, L, S>(
```
xla: Why is the side-effect of adding to the store part of this function? Wouldn't it be a better separation to decouple these two concerns.

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
+///     - any requests fail
+///     - requested data is inconsistent (eg. vals don't match hashes in header)
+///     - validators did not correctly commit their blocks
+/// This function is recursive: it uses a bisection algorithm
+/// to request data for intermediate heights as necessary.
+/// Ensures our last trusted header hasn't expired yet, and that
+/// data from the untrusted height can be verified, possibly using
+/// data from intermediate heights.
+/// This function is primarily for use by a light node.
+pub fn verify_and_update_bisection<TS, SH, C, L, R, S>(
```
xla: A similiar `and` case. I'd recommend this to be a just `update_bisection` - the error returned will communicate if verification failed.

tendermint/src/lite/verifier.rs
```diff
@@ -284,42 +246,100 @@ pub fn verify_header<TS, SH, C, L, S, R>(
...
+    // significantly from `now`.
+
+    // inner recursive function which assumes
+    // trusting_period check is already done.
+    _verify_and_update_bisection(untrusted_height, trust_threshold, req, store)
+}
+
+// inner recursive function for verify_and_update_bisection.
+// see that function's docs.
+fn _verify_and_update_bisection<TS, SH, C, L, R, S>(
```
xla: This `_` is not a good sign. If this is internal then the two intents can be split and made into distinct functions. Unless I miss a point why this needs to be in one.

tendermint/src/lite.rs
```diff
@@ -4,4 +4,9 @@ pub mod types;
 pub mod verifier;
 
 pub use self::types::*;
-pub use self::verifier::*;
+
+// TODO: don't expose this once the json tests
+// switch to using one of the other functions
```
xla: Do we have a follow-up for this in our meta issues?

tendermint/src/lite/verifier.rs
```diff
@@ -126,25 +77,11 @@ where
...
-    validate_vals_and_commit(header, commit, validators)?;
-
-    // ensure that +2/3 validators signed correctly
-    verify_commit_full(validators, commit)
-}
-
 /// Verify that +2/3 of the correct validator set signed this commit.
-/// NOTE: these validators are expected to be the correct validators for the commit.
-fn verify_commit_full<C>(vals: &C::ValidatorSet, commit: &C) -> Result<(), Error>
+/// NOTE: these validators are expected to be the correct validators for the commit,
```
xla:
```suggestion
/// NOTE: These validators are expected to be the correct validators for the commit,
```

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
-    // Run bisection: try again with a pivot header whose height is in the
-    // middle of the trusted height and the desired height (h2.height).
-
-    let trusted_header = trusted_state.last_header();
-    let trusted_height: u64 = trusted_header.header().height().value();
-    let untrusted_height: u64 = untrusted_header.header().height().value();
-    let pivot: u64 = (trusted_height + untrusted_height) / 2;
-    let pivot_header = req.signed_header(pivot)?;
-    let pivot_vals = req.validator_set(pivot)?;
+    // verify the untrusted commit
```
xla:
```suggestion
```
Comment doesn't add more context.

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
+    store: &mut S,
+) -> Result<(), Error>
+where
+    TS: TrustedState<LastHeader = SH, ValidatorSet = C::ValidatorSet>,
+    SH: SignedHeader<Commit = C>,
+    C: Commit,
+    L: TrustThreshold,
+    S: Store<TrustedState = TS>,
+{
+    // fetch the latest state and ensure it hasn't expired
```
xla:
```suggestion
    // Fetch the latest state and ensure it hasn't expired.
```

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +124,120 @@ where
...
-        req,
-        store,
-    )?;
-    // Add header (and corresponding next validators) to trusted state store:
-    let untrusted_next_vals = req.validator_set(untrusted_header.header().height().increment())?;
-    let untrusted_state = TS::new(untrusted_header, &untrusted_next_vals);
-    store.add(&untrusted_state)?;
 
-    Ok(())
+    // the untrusted header is now trusted. update the store
```
xla:
```suggestion
    // The untrusted header is now trusted. update the store.
```

tendermint/src/lite.rs
ebuchman: Yes its in #110

tendermint/src/lite/types.rs
```diff
@@ -136,25 +142,29 @@ pub trait Store {
...
+    /// https://github.com/interchainio/tendermint-rs/issues/118
     fn get(&self, h: Height) -> Result<&Self::TrustedState, Error>;
-
-    /// Retrieve the trusted signed header with the largest height h' with h' <= h, if it exists.
-    /// If it does not exist return an error.
-    fn get_smaller_or_equal(&self, h: Height) -> Result<Self::TrustedState, Error>;
 }
 
-#[derive(Debug, PartialEq)]
+// NOTE: Copy/Clone for convenience in testing ...
```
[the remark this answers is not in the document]
ebuchman: Possibly, but also an explanation of why we need to derive those traits and a reminder to maybe find a way to not need to do that ?

tendermint/src/lite/verifier.rs
ebuchman: what would you propose as a better option? the name is currently naming all the arguments (where header and commit together comprise a signed_header). How about `validate_state` or `validate_untrusted` ?

tendermint/src/lite/verifier.rs
ebuchman: This function isn't supposed to be exposed which is why they're not.

It's marked `pub` because it's needed in the tests, but it shouldn't be exposed in the `lite.rs` defining the public API for the crate. Of course currently it is because it's used in an integration test, but fixing that is tracked separately.

tendermint/src/lite/verifier.rs
ebuchman:
```suggestion
```

it's stale

tendermint/src/lite/verifier.rs
ebuchman: What about:

```suggestion
    // All validation passed successfully. Verify the validators correctly committed the block.
```

? Or you think not needed at all?

tendermint/src/lite/verifier.rs
ebuchman: Good question I've gone back and forth on it. It's also currently the thing blocking the fix in the integration tests that would allow us to not expose `verify_single` because we need to mock the store there! But the idea is for this lib to provide all the functionality for maintaining a light client correctly, which includes making updates to the store. We don't want users to forget to update the store or to update it incorrectly. 

That said, forgetting is probably not really an issue, and updating incorrectly we currently don't prevent since they own the store, but if we say "we'll handle all the updates" it might help.

Also, it's partially for symmetry with the bisection function, which updates the store as it goes. We could alternatively return the set of trusted headers and remove all `store` mgmt from this lib and leave it to the user.

In any case, I think we should take this up in a separate issue ...

tendermint/src/lite/verifier.rs
ebuchman:
```suggestion
    // The untrusted header is now trusted; Update the store
```

tendermint/src/lite/verifier.rs
ebuchman: So just `internal_update_bisection`?

tendermint/src/lite/verifier.rs
xla: Your suggestion does indeed give more context and should be kept.

tendermint/src/lite/verifier.rs
xla: If the rename above goes through this can just be `verify_and_update_bisection`.

tendermint/src/lite/verifier.rs
xla: Follow-up issue sounds perfect to continue this conversation and would capture my main review comment point.

tendermint/src/lite/verifier.rs
xla:
```suggestion
pub fn validate_untrusted<H, V, C>(
```

tendermint/src/lite/types.rs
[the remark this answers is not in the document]
xla: Then we should track it as a TODO/follow-up if we think it should be addressed.

tendermint/src/lite/verifier.rs
```diff
@@ -153,6 +85,9 @@ where
 
     // check the signers account for +2/3 of the voting power
     if signed_power * 3 <= total_power * 2 {
+        // TODO(EB): Use a different error from
+        // verify_commit_trusting else bisection
+        // will happen when the commit is actually just invalid!
```
ancazamfir: Can we fix this?

tendermint/src/lite/verifier.rs
```diff
@@ -184,98 +119,118 @@ where
...
-    // middle of the trusted height and the desired height (h2.height).
-
-    let trusted_header = trusted_state.last_header();
-    let trusted_height: u64 = trusted_header.header().height().value();
-    let untrusted_height: u64 = untrusted_header.header().height().value();
-    let pivot: u64 = (trusted_height + untrusted_height) / 2;
-    let pivot_header = req.signed_header(pivot)?;
-    let pivot_vals = req.validator_set(pivot)?;
+    // All validation passed successfully. Verify the validators correctly committed the block.
+    verify_commit_full(untrusted_vals, untrusted_sh.commit())
```
ancazamfir: I would move this after `validate_untrusted()` and even combine them as they perform signed header validation. 
Also isn't `verify_commit_full()` same as `verify_commit_trusting()` with different `trust_threshold`.

tendermint/src/lite/verifier.rs
```diff
@@ -284,42 +239,100 @@ pub fn verify_header<TS, SH, C, L, S, R>(
...
-
-    // Get the highest trusted header with height lower than sh2's.
-    let sh1_trusted = store.get_smaller_or_equal(height)?;
-    can_trust_bisection(
-        &sh1_trusted,
-        &sh2,
+    // this get is redundant the first time.
+    // TODO: possibly refactor so this func takes and returns
+    // trusted_state.
+    let trusted_state = store.get(Height::from(0))?;
```
ancazamfir: Yeh, this is a bit inefficient as we keep getting the trusted state from the store for all pivots that we just added in L#311. It would be maybe better if the trusted state is passed in as parameter. And easier to read.

The other question is about store implementation. I expect there is some background routine that runs periodically and purges expired headers from the store. This would interfere with bisection runs and needs to be analyzed as getting a trusted header from the store might fail.

tendermint/src/lite/verifier.rs
ancazamfir: Yeah, actually something is not right. First, I think it is important that `verify_single()` does both `validate_untrusted()` and `verify_commit_full()` first. Any errors from these functions are fatal and bisection should not proceed. As it is right now, `verify_single()` may return early with error from `verify_commit_trusting()` which causes the bisection caller to start bisecting even if this header/commit is not fully verified. For the second issue, see next comment...

tendermint/src/lite/verifier.rs
```diff
@@ -284,42 +239,100 @@ pub fn verify_header<TS, SH, C, L, S, R>(
...
+    let trusted_sh = trusted_state.last_header();
+
+    // fetch the header and vals for the new height
+    let untrusted_sh = &req.signed_header(untrusted_height)?;
+    let untrusted_vals = &req.validator_set(untrusted_height)?;
+    let untrusted_next_vals = &req.validator_set(untrusted_height.increment())?;
+
+    // check if we can skip to this height and if it verifies.
+    match verify_single(
```
ancazamfir: ...second issue: looks like we call verify_single() multiple times for the untrusted signed header. We expect this to be called with different (higher) trusted singed header but still we do the untrusted header verification (independent of the trusted one) multiple times and this is expensive.

tendermint/src/lite/verifier.rs
ebuchman: Yes it's tracked in #110

tendermint/src/lite/verifier.rs
ebuchman: So I was actually thinking the opposite, which was that we shouldn't bother trying to fully verify a future header (which is expensive) until we know we can skip to it.

I suppose it's a question of compute vs bandwidth. But even if we did the verification first, a full node who wanted to trick us into using a lot of bandwidth could always give us headers that are well signed by themselves so we'd waste time verifying them and not find out until much later that they're not actually valid from previous ones. So it doesn't really make sense to me to be verifying the commit of a header until we know we can skip to it and that was one of the key points I wanted to effect in these changes. 

Sorry, I should have written that up more explicitly somewhere.

ebuchman: And yes, verify_commit_full and _trusting are the same now except the later uses the threshold trait and the former is just fixed to 2/3

tendermint/src/lite/verifier.rs
ebuchman: Yes I thought it would be easiest to read and write first like this. And also figured stores would have caching so it wouldn't be inefficient. But see discussion with xla about not passing in store into these functions https://github.com/interchainio/tendermint-rs/issues/110#issuecomment-569973976

Been thinking about a Client state machine and functions that return TrustedState so that it's the users responsibility to make the store updates but they can't do it wrong.

Pruning should be guaranteed to not interfere - ie. it can only apply outside the trusting period.

tendermint/src/lite/verifier.rs
ancazamfir: > So I was actually thinking the opposite, which was that we shouldn't bother trying to fully verify a future header (which is expensive) until we know we can skip to it.

yes, but the problem is that if we know we can skip it, we start recursive bisection  which involves downloading and fully verifying all the pivots...so both bandwidth and cpu.

tendermint/src/lite/verifier.rs
ebuchman: We're only doing the validation, which is just hash checks. the hashes can be memoized so this would be cheap equality testing.

We do end up running verify_commit_trusting multiple times which is expensive and it would be interesting to see how we can optimize that but it will start to mess with our traits again

tendermint/src/lite/verifier.rs
ebuchman: as discussed, this is ok, because the in between headers are being verified fully and hence trusted, even if the node turns out to be bad at the end of the run, so that bandwidth and cpu isn't really wasted.

PR #115, opened 2019-12-29
Lite: complete test mocks and start some tests

ebuchman:
First take at some tests for the new `verify_single` function, ala #113 

This completes an implementation of mocks for the core types for the lite module and shows how to use them for testing. Still unmocked are the `Store` and `Requester`.

There's probably much to be improved here, including the test names and how we organize them, but this seems like a decent start.

This also surfaced some issues with error types:
- when used to verify a commit (rather than just checking if we can skip), `voting_power_in` can't detect signatures from validators not in the current validator set
- we should have two distinct errors for InsufficientVotingPower - one when it's insufficient to skip, and one when it's insufficient to verify! Otherwise we might start bisecting when really the commit is just invalid ...

For now we're just duplicating some of the mocks that are already in the `verifier` tests module. Probably `verifier.rs` and `public.rs` will be consolidated down to `verifier` in #114 and this will be deduplicated then.

* [x] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/lite/public.rs
+++ b/tendermint/src/lite/public.rs
@@ -220,10 +220,6 @@ mod tests {
         Hash::new(Algorithm::Sha256, &hashed).unwrap()
     }
 
-    fn fixed_hash() -> Hash {
-        Hash::new(Algorithm::Sha256, &Sha256::digest(&[5])).unwrap()
-    }
-
     // vals are just ints, each has power 1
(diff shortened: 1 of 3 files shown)


ebuchman:
I'm going to merge this into bucky/spec-reorg locally and just push there. avoiding squash merge while other PRs are depending on this branch oh man...

PR #116, opened 2019-12-29
lite: impl Requester, Store, and a basic syncing daemon

ebuchman:
Started in a new crate as per #110 https://github.com/interchainio/tendermint-rs/issues/110#issuecomment-569536778

Includes a basic test for the Requester against a running tendermint node.

Not really sure how to use async stuff yet so I just copied the `block_on` thing from the integration tests for now ...

Also includes a basic daemon that uses bisection to sync against a local node from hard coded values - this is working against a my local node.

* [ ] Wrote tests
* [ ] Updated CHANGES.md


xla:
@ebuchman As the Requesters async parts will be primarily around I/O [async-std](https://async.rs/) might be a better fit here.

tarcieri:
At least for now, `async-std` is fairly new and immature/rough-around-the-edges, lacking an ecosystem of higher level libraries (e.g. `tower`, `hyper`), and at least anecdotally I’ve heard mixed reports about how well using `async-std` with a `tokio` reactor actually works in practice.

FWIW, I was interested in `async-std` for awhile (see e.g. https://github.com/interchainio/tendermint-rs/issues/2#issuecomment-522251641 ) but much of what I liked about it ergonomically also landed in tokio 0.2.

ebuchman:
Merged in #114 and #117

ebuchman:
Merged in master and fixed conflicts. Also did some minor cleanup.

We can probably merge this more or less as is for now but we'll want to follow up with:

- move `subjective_init` into the `lite` module, make it generic over the traits, and add the extra validation as noted
- fix package structure (some impls for `lite` are in the new `tendermint-lite` crate while others are just in `tendermint`. Maybe they should all be in a module together in `tendermint` but `lite` itself should be factored out into its own crate and then we just need a crate for a `tendermint-lite-node` binary ? 
- refactor main.rs to be a CLI and take config & args

tendermint-lite/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "tendermint-lite"
+version = "0.1.0"
+authors = ["Ethan Buchman <ethan@coinculture.info>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tendermint = { path = "../tendermint" }
+tokio = "0.2"
```
Liamsi: Cool that there is (again) a light client crate now.  Is the plan to move everything from `tendermint/src/lite` into this crate too?

tendermint-lite/src/requester.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[test]
+    fn test_val_set() {
+        let client = block_on(rpc::Client::new(&"localhost:26657".parse().unwrap())).unwrap();
+        let req = RPCRequester::new(client);
+        let r1 = req.validator_set(5).unwrap();
+        let r2 = req.signed_header(5).unwrap();
+        assert_eq!(r1.hash(), r2.header().validators_hash());
+    }
+}
+*/
```
Liamsi: What's up with this? Why is it commented out?

tendermint-lite/Cargo.toml
ebuchman: I'm starting to think the `tendermint/src/lite` should be its own crate and the non main stuff here should be in the `tendermint` crate

tendermint-lite/src/requester.rs
ebuchman: It's an integration test requiring a full node and i wasn't sure how to indicate that ...

tendermint-lite/Cargo.toml
Liamsi: OK, I see. For binaries it is good practice to add the `Cargo.lock` to the repo, too.

tendermint-lite/src/requester.rs
Liamsi: I think adding a `#[ignore]` to the test and adding one sentence about why this is disabled and what it needs to be able to run would be cool.

An example of this is:

https://github.com/interchainio/tendermint-rs/blob/1cc2071cead14533e71a69e8db4051eb0ea09c60/tendermint/tests/integration.rs#L3-L32

tendermint-lite/src/requester.rs
```diff
@@ -0,0 +1,84 @@
...
+mod tests {
+    use super::*;
+    use tendermint::lite::types::Header as LiteHeader;
+    use tendermint::lite::types::Requester as LiteRequester;
+    use tendermint::lite::types::SignedHeader as LiteSignedHeader;
+    use tendermint::lite::types::ValidatorSet as LiteValSet;
+    use tendermint::rpc;
+
+    // TODO: integration test
+    #[test]
```
Liamsi:
```suggestion
    #[test]
    #[ignore]
```

tendermint-lite/src/requester.rs
```diff
@@ -0,0 +1,84 @@
...
+pub fn block_on<F: Future>(future: F) -> F::Output {
+    Builder::new()
+        .basic_scheduler()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(future)
+}
+
+/*
```
Liamsi:
```suggestion
```

tendermint-lite/src/requester.rs
Liamsi:
```suggestion
```

tendermint-lite/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "tendermint-lite"
+version = "0.1.0"
+authors = ["Ethan Buchman <ethan@coinculture.info>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tendermint = { path = "../tendermint" }
```
Liamsi: Shouldn't this be a proper dependency instead of a local path?

tarcieri: You can do both...
```suggestion
tendermint = { version = "0.11", path = "../tendermint" }
```

PR #117, opened 2019-12-30
first take at hooking up a node

ebuchman:
First take at hooking it all up into a running light node!

It currently fails I think because we're computing header hashes incorrectly ...

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint-lite/src/main.rs
+++ b/tendermint-lite/src/main.rs
@@ -1,4 +1,4 @@
-use tendermint::hash::Algorithm;
+use tendermint::hash;
 use tendermint::lite;
 use tendermint::lite::Error;
 use tendermint::lite::{
@@ -43,7 +43,8 @@ fn main() {
     let req = Requester::new(client);
     let mut store = MemStore::new();
 
-    let vals_hash = Hash::from_hex_upper(Algorithm::Sha256, SUBJECTIVE_VALS_HASH_HEX).unwrap();
+    let vals_hash =
+        Hash::from_hex_upper(hash::Algorithm::Sha256, SUBJECTIVE_VALS_HASH_HEX).unwrap();
 
     validate_and_init_subjective_state(
         Height::from(SUBJECTIVE_HEIGHT),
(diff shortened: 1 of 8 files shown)


tendermint-lite/src/main.rs
```diff
@@ -0,0 +1,134 @@
+use tendermint::hash::Algorithm;
```
tarcieri: Perhaps it would be less ambiguous to import `tendermint::hash` and reference `Algorithm` as `hash::Algorithm`?

tendermint/src/rpc/endpoint/commit.rs
```diff
@@ -12,9 +12,11 @@ pub struct Request {
 impl Request {
     /// Create a new request for commit info about a particular block
     pub fn new(height: block::Height) -> Self {
-        Self {
-            height: Some(height),
+        let mut height_opt = Some(height);
+        if height.value() == 0 {
+            height_opt = None
         }
+        Self { height: height_opt }
```
tarcieri: I'd probably write this like:

```rust
Self {
    height: match height {
        0 => None,
        _ => Some(height)
    }
}
```

Alternatively perhaps define a method on `Height` for this? It seems useful.

tendermint-lite/src/main.rs
xla: Addressed in 712e673

@tarcieri Is it necessary to expose this param at all at this point if the enum has only one variant?

tarcieri: I'd say it's both more explicit and good futureproofing in the event you do ever want to support additional hash algorithms. I'll cite `git` as a project which didn't do this which also still hasn't managed to migrate to a new hash algorithm.

That said I doubt SHA-256 will ever be catastrophically broken.

tendermint/src/rpc/endpoint/commit.rs
xla: This was a shortcut to get the integration to work. There ought to be a more explicit way to request the latest height that depending on the implicit menaing of `0`.

xla: Addressed in 7477c9d

ebuchman: Opened an issue for this https://github.com/interchainio/tendermint-rs/issues/118

PR #122, opened 2020-01-03
Explore clippy rules and apply quality of life recommendations

xla:
First of, this is an attempt to codify clippy rules. And find the right amount to be applied for consistent code shape. Least it helps me explore the codebase. Will keep inventory which commits are driven by what rule. We can disregard any if too rigid.

* f4bc257..61dce99 - https://rust-lang.github.io/rust-clippy/master/#module_name_repetitions
* a0c842f - https://rust-lang.github.io/rust-clippy/master/#doc_markdown
* bad770b - https://rust-lang.github.io/rust-clippy/master/index.html#unseparated_literal_suffix
* 63c673c - https://rust-lang.github.io/rust-clippy/master/index.html#pub_enum_variant_names
* 56d1f48 - https://rust-lang.github.io/rust-clippy/master/index.html#print_stdout
* 7bf1093 - https://rust-lang.github.io/rust-clippy/master/index.html#similar_names
* 41cbd02 - https://rust-lang.github.io/rust-clippy/master/index.html#multiple_inherent_impl
* 67f7358 - https://rust-lang.github.io/rust-clippy/master/index.html#use_self 
* 0ee13e2 - https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn
* 9b9e875 - https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation
* 21b2a8f - https://rust-lang.github.io/rust-clippy/master/index.html#indexing_slicing
* 15e37ca - https://rust-lang.github.io/rust-clippy/master/#default_trait_access
* 1144a4f - https://rust-lang.github.io/rust-clippy/master/#default_trait_access
* d7ef18a - https://rust-lang.github.io/rust-clippy/master/#wildcard_enum_match_arm
* 85e57aa - https://rust-lang.github.io/rust-clippy/master/index.html#missing_inline_in_public_items
* ab338d7 - https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate
* 8534547 - https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unwrap_or_else
* bd9b608 - https://rust-lang.github.io/rust-clippy/master/index.html#option_unwrap_used
* 5cfcb37 - https://rust-lang.github.io/rust-clippy/master/#result_unwrap_used
* 5cfcb37 - https://rust-lang.github.io/rust-clippy/master/index.html#unreachable
* 3dbd0d8..dd405fa - https://rust-lang.github.io/rust-clippy/master/index.html#panic
* e90bc89 - https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
* 2b9b699 - https://rust-lang.github.io/rust-clippy/master/index.html#explicit_iter_loop
* 99e323c - https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
* 7750986 - https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_wrap
* 3ebb516 - https://rust-lang.github.io/rust-clippy/master/index.html#cargo_common_metadata

***

* [ ] Wrote tests
* [ ] Updated CHANGES.md


tendermint/src/amino_types.rs
```diff
@@ -3,27 +3,31 @@
...
+mod ed25519;
+mod message;
+mod ping;
+mod proposal;
+mod remote_error;
+mod signature;
+mod time;
+mod validate;
+mod version;
+mod vote;
```
tarcieri: The disadvantage of relying on re-exports of types defined in non-pub modules is rustc will refer to those types by their fully qualified name, which leaves someone trying to debug in the bizarre position that the compiler referring to a path which cannot be accessed and does not exist in rustdoc.

That said, I do sometimes do this myself, but it's usually when I'm trying to create a crate where the code is factored into internal modules but all types are exported at the toplevel (and when I do that, it's generally because the external API surface of the crate is small to begin with).

The other disadvantage is this approach eliminates the ability to have module-level documentation visible in the rustdoc.

tendermint/src/amino_types/message.rs
```diff
@@ -3,7 +3,7 @@ use std::convert::TryInto;
 
 /// Extend the original prost::Message trait with a few helper functions in order to
 /// reduce boiler-plate code (and without modifying the prost-amino dependency).
-pub trait AminoMessage: prost_amino::Message {
+pub trait Message: prost_amino::Message {
```
tarcieri: 👍I'm a fan of eliminating module (and crate) names from type names. Some relevant API guidelines and issues:

- [RFC#356: No module prefixes](https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md)
- https://github.com/rust-lang/api-guidelines/issues/66

Importing modules rather than individual types and referring to types by their module names can also help cut down on the number of imports.

tendermint/src/lite/types.rs
```diff
@@ -7,14 +7,14 @@ use crate::Hash;
...
-/// TrustedState stores the latest state trusted by a lite client,
+/// `TrustedState` stores the latest state trusted by a lite client,
 /// including the last header (at height h-1) and the validator set
 /// (at height h) to use to verify the next header.
 pub trait TrustedState {
     type LastHeader: SignedHeader;
     type ValidatorSet: ValidatorSet;
 
-    /// Initialize the TrustedState with the given signed header and validator set.
+    /// Initialize the `TrustedState` with the given signed header and validator set.
```
tarcieri: I'd suggest linkifying any types which are in-scope thusly:

```suggestion
    /// Initialize the [`TrustedState`] with the given signed header and validator set.
```

On nightly (which is used by https://docs.rs) these will generate internal links within the documentation.

tendermint/src/amino_types.rs
xla: What would you propose generally? I think `amino_types` module organisation is ripe for a review anyway.

tarcieri: Re-export them, but leave the containing modules `pub` if they're part of the public API

tendermint/src/block/height.rs
```diff
@@ -69,12 +69,6 @@ impl From<Height> for u64 {
     }
 }
 
-impl From<Height> for i64 {
-    fn from(height: Height) -> Self {
-        height.0 as Self
-    }
-}
```
Liamsi: Did we verify this isn't used in the KMS for instance?

tendermint/src/amino_types/ed25519.rs
```diff
@@ -86,10 +89,8 @@ mod tests {
 
         assert_eq!(got, want);
 
-        match PubKeyRequest::decode(want.as_ref()) {
-            Ok(have) => assert_eq!(have, msg),
-            Err(err) => panic!(err.to_string()),
-        }
+        let have = PubKeyRequest::decode(want.as_ref()).unwrap();
```
Liamsi: Didn't 2 other commits turn all option/result `unwrap`s into `expect`s?

tendermint/src/amino_types/ed25519.rs
```diff
@@ -128,10 +129,8 @@ mod tests {
 
         assert_eq!(got, encoded);
 
-        match PubKeyResponse::decode(encoded.as_ref()) {
-            Ok(have) => assert_eq!(have, msg),
-            Err(err) => panic!(err),
-        }
+        let have = PubKeyResponse::decode(encoded.as_ref()).unwrap();
```
Liamsi: same as https://github.com/interchainio/tendermint-rs/pull/122/files#r369391322

tendermint/src/lite/types.rs
```diff
@@ -101,13 +101,13 @@ pub trait Commit {
...
 
-/// TrustThreshold defines what fraction of the total voting power of a known
+/// [`TrustThreshold`] defines what fraction of the total voting power of a known
 /// and trusted validator set is sufficient for a commit to be
 /// accepted going forward.
 /// The default implementation returns true, iff at least a third of the trusted
 /// voting power signed (in other words at least one honest validator signed).
 /// Some clients might require more than +1/3 and can implement their own
-/// TrustLevel which can be passed into all relevant methods.
+/// [`TrustLevel`] which can be passed into all relevant methods.
```
Liamsi: Looks like we forgot to rename `TrustLevel` to `TrustThreshold` here. Running `cargo +nightly doc --open --no-deps` will error now with something like 
```
`[TrustLevel]` cannot be resolved, ignoring it.
```
because of  a `#[deny(warnings)]`. 

Does this mean building for https://docs.rs/ won't work now?

tendermint/src/lite/verifier.rs
```diff
@@ -97,8 +97,8 @@ where
 
 /// Verify that +1/3 of the given validator set signed this commit.
 /// NOTE the given validators do not necessarily correspond to the validator set for this commit,
-/// but there may be some intersection. The trust_level parameter allows clients to require more
-/// than +1/3 by implementing the TrustLevel trait accordingly.
+/// but there may be some intersection. The `trust_level` parameter allows clients to require more
+/// than +1/3 by implementing the `TrustLevel` trait accordingly.
```
Liamsi: Consistent would be ```[`TrustThreshold `]```

tendermint/src/lite/types.rs
tarcieri: I'm not sure if docs.rs will break if you have intra-doc resolution failures and deny warnings but possibly.

At any rate, it's good to capture and correct these. Running `cargo test --doc` on nightly will catch them.

tendermint/src/block/height.rs
ebuchman: Does it matter? Isn't this commit non-breaking?

tendermint/src/abci/transaction.rs
```diff
@@ -23,6 +23,7 @@ impl Transaction {
         Self(into_vec.into())
     }
 
+    #[allow(clippy::missing_const_for_fn)]
```
ebuchman: Why are there exceptions?

tendermint/src/merkle.rs
```diff
@@ -36,7 +36,13 @@ fn get_split_point(length: usize) -> usize {
         0 => panic!("tree is empty!"),
         1 => panic!("tree has only one element!"),
         2 => 1,
-        _ => length.next_power_of_two() / 2,
+        _ => {
+            let (div, overflow) = length.next_power_of_two().overflowing_div(2);
```
ebuchman: Is this really necessary? `length` and `next_power_of_two` are both `usize` which will never overflow ... Same goes for the wrapping_div above

tendermint/src/timeout.rs
```diff
@@ -32,28 +32,35 @@ impl FromStr for Timeout {
...
         let units = match s.chars().nth(s.len() - 2) {
-            Some('m') => "ms",
-            Some('0'..='9') => "s",
+            Some('m') => Unit::Milliseconds,
+            Some('0'..='9') => Unit::Seconds,
             _ => return Err(err!(ErrorKind::Parse, "invalid units")),
         };
 
-        let numeric_part = s.chars().take(s.len() - units.len()).collect::<String>();
+        let numeric_part = match units {
```
ebuchman: What does this enum have to do with `Rework use of unreachable` ?

Also, can the Unit variants contain their own string representation so we don't need a match here like before?

tendermint/src/block/height.rs
ebuchman: Weird, I saw this comment on a different commit: https://github.com/interchainio/tendermint-rs/pull/122/commits/67f735866d358e025c5c836cc40781e97ac97120#r368718476. 

The correct one is https://github.com/interchainio/tendermint-rs/pull/122/commits/77509865c4281e0f15c5f9be503d2318a991b271#r368718476.

So yes we should verify we can get rid of this

ebuchman:
This is great work but I agree it should be broken up into smaller PRs with loosely themed lints.

Also we need to pay more attention to breaking changes and track them accordingly. Maybe the smaller PRs for this can also be distinguished by whether or not they're breaking.

tendermint/src/abci/code.rs
```diff
@@ -19,17 +19,17 @@ pub enum Code {
 }
 
 impl Default for Code {
-    fn default() -> Code {
```
ebuchman: So in general I think this is probably good, but I wonder if in the case of large `impl` blocks if it would be better to use the explicit type name.

brapse:
@gterzian Could you potentially pick this up and drive it home once #125 is merged int 🙏

xla:
@brapse Currently working on incremental changes on top of latest master to superseed this PR.

brapse:
@xla Can/should we link the new PR to this one and close this?

ebuchman:
Let's track in https://github.com/informalsystems/tendermint-rs/issues/193 and work on it in the follow ups

PR #124, opened 2020-01-09
Refactor lite impls into their own module within tendermint 

Liamsi:
* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/block/signed_header.rs
+++ b/tendermint/src/block/signed_header.rs
@@ -1,8 +1,9 @@
 //! SignedHeader contains commit and and block header.
 //! It is what the rpc endpoint /commit returns and hence can be used by a
 //! light client.
-use crate::{block, hash, lite, lite::Error, validator::Set, vote::SignedVote};
 use serde::{Deserialize, Serialize};
+
+use crate::{block, vote::SignedVote};
 
 /// Signed block headers
 #[derive(Clone, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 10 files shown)


tendermint/src/validator.rs
```diff
@@ -151,11 +131,12 @@ impl From<&Info> for InfoHashable {
...
 
-// returns the bytes to be hashed into the Merkle tree -
-// the leaves of the tree. this is an amino encoding of the
-// pubkey and voting power, so it includes the pubkey's amino prefix.
 impl Info {
-    fn hash_bytes(&self) -> Vec<u8> {
+    /// Returns the bytes to be hashed into the Merkle tree -
+    /// the leaves of the tree. this is an amino encoding of the
+    /// pubkey and voting power, so it includes the pubkey's amino prefix.
+    // TODO only used in lite impl ...
```
Liamsi: Maybe we should remove this helper from this type if it is clear that it will only be used in the light client?

tendermint/src/block/signed_header.rs
```diff
@@ -13,23 +14,11 @@ pub struct SignedHeader {
...
-    }
-}
-
 impl SignedHeader {
-    /// This is a private helper method to iterate over the underlying
+    /// This is a helper method to iterate over the underlying
     /// votes to compute the voting power (see `voting_power_in` below).
-    fn iter(&self) -> Vec<Option<SignedVote>> {
+    // TODO move into lite impl module?
+    pub fn iter(&self) -> Vec<Option<SignedVote>> {
```
Liamsi: Similarly to https://github.com/interchainio/tendermint-rs/pull/124/files#r364984494
We probably won't use this somewhere else than in the `lite::Commit` impl for `SignedHeader`.

tendermint/src/validator.rs
Liamsi: Looking at `InfoHashable` and from what I remember from go tendermint it seems unlikely that this will be used not only in the light client.

tendermint/src/lite_impl.rs
```diff
@@ -0,0 +1,5 @@
+//! Implementation of all light client traits.
+
```
Liamsi: Maybe `lite_impl` isn't the best name?

tendermint/src/lite/types.rs
```diff
@@ -114,7 +114,7 @@ pub trait TrustThreshold {
     }
 }
 
-/// Requester can be used to request `SignedHeaders` and `ValidatorSet`s for a
+/// Requester can be used to request [`SignedHeader`]s and [`ValidatorSet`]s for a
```
ebuchman: What's the story with these brackets?

tendermint/src/lite_impl.rs
ebuchman: Once we move the current `lite` into it's own crate, can we just call this `lite` ?

Liamsi: Yes, thought so too.

tendermint/src/lite/types.rs
Liamsi: Creates links to the types in the docs. Run `cargo +nightly doc --open --no-deps`  :-)

Issue #112, opened 2019-12-29
Light Client Daemon


ebuchman:
Now that the library logic and bisection are nearing completion in #96 and the relevant follow up issues (#110, #111), we should implement the daemon that puts it all together.

This requires:
- initial config with subjective initialization and list of peers
- simple address book to track peers and manage primary/backups
- detection module to make requests from backup peers, search for conflicts, and publish any verified ones
- concrete implementation of the store
- command line tool

To start, let's get a simple command line tool that works in the single peer case with a simple store, and then work on the detection module for detecting conflicts from backup peers.

ebuchman:
The core of this in the form of the "syncing" loop is working in a basic form in #116

ebuchman:
Closing for specific issues:

- initial config with subjective initialization and list of peers - #293  
- simple address book to track peers and manage primary/backups - #293 and #218
- detection module to make requests from backup peers, search for conflicts, and publish any verified ones - #218 and #217 
- concrete implementation of the store - exists now since the refactor, but ensure it works with the tool re #293 
- command line tool - basics exist in light-node but needs eg #293

Issue #120, opened 2019-12-31
Run a tendermint node in CI to test the RPC & light client


ebuchman:
As per some RPC incompatibilities being discovered, namely #88 and #4, we should have the CI spin up a tendermint node, eg. using the latest docker image, and run the currently ignored integration tests against it  (via `cargo test -- --ignored`) - see https://github.com/interchainio/tendermint-rs/issues/88#issuecomment-564680177

Liamsi:
Likewise for the light node.

Basically these 2 things need a tendermint instance to run / test against:
1. Light node:
 - Current light client main binary:
https://github.com/interchainio/tendermint-rs/blob/master/tendermint-lite/src/main.rs
 - And after refactoring (after #125 gets merged):
https://github.com/interchainio/tendermint-rs/blob/9c4fd2fb953aab71d4073c4fac106027073b3b5b/light-node/src/bin/light_node/main.rs
I can probably use the acceptance tests against a tendermint instance instead of calling the main in a script: https://github.com/interchainio/tendermint-rs/blob/9c4fd2fb953aab71d4073c4fac106027073b3b5b/light-node/tests/acceptance.rs

2.  Currently ignored integration tests:
https://github.com/interchainio/tendermint-rs/blob/master/tendermint/tests/integration.rs

greg-szabo:
I also opened an [issue](https://github.com/tendermint/tendermint/issues/4569) in the tendermint-go repository to update the tendermint docker image so we can use it in GitHub Actions. This would make the conifguration simpler and we could get away from the custom tendermint docker image.

greg-szabo:
Opened [PR](https://github.com/tendermint/tendermint/pull/4577) to update the tendermint-go `Dockerfile`.

greg-szabo:
The new `Dockerfile` was merged to `master` in the tendermint repository. This means that changes will be available in the next 0.33 release.
I propose we keep the current custom docker image for 0.32 and by the time we upgrade our code to 0.33, hopefully there will be a new 0.33 release with the changes - and we can move over.

greg-szabo:
I've been trying to run the light_node acceptance tests but I get bisect errors just starting the light_node. I'm going to park working on this while the code (or me) gets fixed.
Running a full node on the dev machine is slightly easier now because you can run a full node on your machine with one command, exactly how the CI does it:
[...]

PR #125, opened 2020-01-11
Abscissify light node

Liamsi:
ref #112

* [ ] Wrote tests
* [ ] Updated CHANGES.md


lite-node/tests/acceptance.rs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// Storing this value as a [`Lazy`] static ensures that all instances of
+/// the runner acquire a mutex when executing commands and inspecting
+/// exit statuses, serializing what would otherwise be multithreaded
+/// invocations as `cargo test` executes tests in parallel by default.
+pub static RUNNER: Lazy<CmdRunner> = Lazy::new(CmdRunner::default);
+
+/// Use `LiteNodeConfig::default()` value if no config or args
+#[test]
+#[ignore]
```
Liamsi: Note: To do proper tests here, we'd need to mock the requester.

lite-node/src/config.rs
```diff
@@ -0,0 +1,56 @@
...
+/// LiteNode Configuration
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct LiteNodeConfig {
+    /// RPC address to request headers and validators from.
+    pub rpc_address: String,
+    /// The duration until we consider a trusted state as expired.
+    pub trusting_period: Duration,
+    /// Subjective initialization.
+    pub subjective_init: SubjectiveInit,
```
Liamsi: We need to figure out a clever way to represent `TrustThreshold` as sth that can be written to a file. It's currently just defined via it's behaviour.

lite-node/src/commands/start.rs
```diff
@@ -0,0 +1,180 @@
...
+                    .last_header()
+                    .header()
+                    .height(),
+                latest_peer_height,
+            );
+
+            let now = &SystemTime::now();
+            lite::verify_and_update_bisection(
+                latest_peer_height,
+                THRESHOLD, // TODO
```
Liamsi: https://github.com/interchainio/tendermint-rs/pull/125/files#r365508008

lite-node/src/commands/start.rs
```diff
@@ -0,0 +1,180 @@
...
+#[derive(Command, Debug, Options)]
+pub struct StartCmd {
+    /// RPC address to request headers and validators from.
+    #[options(free)]
+    rpc_addr: String,
+}
+
+// TODO: this should also somehow be configurable ...
+// we can't simply add this as a field in the config because this either would
+// be a trait (`TrustThreshold`) or immediately and impl thereof (`TrustThresholdOneThird`).
```
Liamsi: https://github.com/interchainio/tendermint-rs/pull/125/files#r365508008

lite-node/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+    unused_qualifications
+)]
+// TODO(ismail): add proper docs and remove this!
+#![allow(missing_docs)]
+
+pub mod application;
+pub mod commands;
+pub mod config;
+pub mod error;
+pub mod prelude;
```
Liamsi: Everything below `prelude` should likely live somewhere else (in a different module)?

tarcieri:
@xla I can give a short rationale:

Abscissa provides solutions for: command-line option parsing, components (i.e. dependency injection), configuration, error handling, logging, secrets management, and terminal interactions.

To the extent it has "a lot of overhead to setup", I'd argue that it has less than wiring all of these things up subsystem-by-subsystem, and that you probably need all of these things.

> could you outline what dimensions should be considered and how abscissa is going to make any other (e.g. structopt, pico-args into function invocation) option non-viable

Abscissa provides the full functionality of `structopt` + `clap`, with arguably a better DSL (provided by [gumdrop](https://github.com/murarth/gumdrop), which is also being used by [str4d](https://github.com/str4d/rage)). However, where `structopt` + `clap` provides one of the high-level features of Abscissa (command-line option parsing), it requires 26 dependencies, roughly half of the ~50 total Abscissa requires to provide 7 different subsystems.

I'll also note that Abscissa is being used by [Zebra](https://github.com/ZcashFoundation/zebra), a clean-room Rust implementation of a Zcash node, as well as [Tendermint KMS](https://github.com/tendermint/kms). As such, we're able to work on common solutions to problems like [how to integrate Abscissa and Tokio](https://github.com/ZcashFoundation/zebra/pull/169).

If you'd like to read more about Abscissa, here's a blog post: https://iqlusion.blog/introducing-abscissa-rust-application-framework

light-node/src/commands/start.rs
```diff
@@ -0,0 +1,180 @@
...
+#[derive(Command, Debug, Options)]
+pub struct StartCmd {
+    /// RPC address to request headers and validators from.
+    #[options(free)]
+    rpc_addr: String,
+}
+
+// TODO: this should also somehow be configurable ...
+// we can't simply add this as a field in the config because this either would
+// be a trait (`TrustThreshold`) or immediately and impl thereof (`TrustThresholdOneThird`).
```
ebuchman: What's the best UX for this? The trust threshold seems most naturally to be a fraction, like `1/3`, `2/5`, etc. So we could just provide an implementation of TrustThreshold that takes a numerator and denominator and applies it accordingly. We can keep TrustThresholdOneThird as a thing but it would just be TrustThresholdFraction::new(1, 3)

Liamsi: >  So we could just provide an implementation of TrustThreshold that takes a numerator and denominator and applies it accordingly.

Yeah, I thought about this too. I think that would be the easiest and general enough!

tendermint-lite/src/main.rs
```diff
@@ -1,135 +0,0 @@
...
-
-/*
- * The following is initialization logic that should have a
- * function in the lite crate like:
- * `subjective_init(height, vals_hash, store, requester) -> Result<(), Error`
- * it would fetch the initial header/vals from the requester and populate a
- * trusted state and store it in the store ...
- * TODO: this should take traits ... but how to deal with the State ?
- * TODO: better name ?
- */
```
Liamsi: @ebuchman @Shivani912 I'll move this into the lite crate in a separate PR. Note: we need to rework the spec to match the initialization via a pair of subjective (height, vals_hash). Currently, it only talks about init. through a header.

light-node/src/commands/start.rs
Liamsi: Refactored here: https://github.com/interchainio/tendermint-rs/pull/132/files#diff-b1b6c7ff77dbf13838f76b8403715185R88-R119

ebuchman:
What's needed to move this into Ready for Review? Would be awesome to get merged so future work (eg fork detection) can be based on it

Liamsi:
I haven't tested it against a running node. I think everything that worked in the previous main.rs is still working as before though (only we have config flags / can have config files now). I'll move this ready to review now.

Liamsi:
Have a bunch of merge conflicts now. Will resolve them after I resolved those in #154.

light-node/src/commands/start.rs
```diff
@@ -0,0 +1,161 @@
...
+    // TODO: check next_vals ...
+
+    let trusted_state = &State::new(&signed_header, &next_vals);
+
+    store.add(trusted_state.to_owned())?;
+
+    Ok(())
+}
+
+fn block_on<F: Future>(future: F) -> F::Output {
```
gterzian: Have you considered not building a new runtime on each call?

You could probably do something like:

```rust
lazy_static! {
    pub static ref RUNTIME: Runtime = { Builder::new()
        .basic_scheduler()
        .enable_all()
        .build()
        .unwrap() };
}

fn block_on<F: Future>(future: F) -> F::Output {
    RUNTIME.block_on(future)
}
```

I can't tell if this is called from multiple threads or not, if so, you could just put a `Mutex` around the `Runtime`.

tarcieri: You might want to check out [`abscissa_tokio`](https://github.com/iqlusioninc/abscissa/tree/develop/tokio) which allows you to use the Tokio runtime as a component (so you can spawn things onto it during application startup if you so desire) and boot the runtime using:

https://docs.rs/abscissa_tokio/0.5.1/abscissa_tokio/#add-tokiocomponent-to-your-abscissa-application

```rust
use crate::application::APPLICATION;

impl Runnable for StartCmd {
   fn run(&self) {
       abscissa_tokio::run(&APPLICATION, async {
           println!("now running inside the Tokio runtime");
       });
   }
}
```

gterzian: It seems the current approach is to block on the future in the current thread, I assume the solution linked to above spawns the future onto Tokio asynchronously instead? It might be useful and it would seem like a change from the current approach.

Liamsi: I think you are right @gterzian. 

Note that currently there is only one call to this particular `block_on` method and runtime. But yes, we need to do this properly throughout the repo. We currently [copy](https://github.com/interchainio/tendermint-rs/pull/116#issue-357829136)&pasted [this](https://github.com/interchainio/tendermint-rs/pull/85#issuecomment-564437616) throughout the repo when async/await was introduced [here](https://github.com/interchainio/tendermint-rs/pull/85).

Currently the [rpc integration tests](https://github.com/interchainio/tendermint-rs/blob/6df2a927623f4c50869249c89a4502f750f246f5/tendermint/tests/integration.rs#L16-L23), above start command and the [requester](https://github.com/interchainio/tendermint-rs/blob/9c4fd2fb953aab71d4073c4fac106027073b3b5b/light-node/src/requester.rs#L51-L58) use this approach. Maybe worth to have a closer look into this in a separate issue, too?

docs/architecture/adr-005-light-client-fork-detection.md
```diff
@@ -45,3 +45,47 @@ one correct full node in order to detect conflicts in a timely fashion. We keep
 this mechanism simple for now, but in the future a more advanced peer discovery
 mechanism may be utilized.
 
+
+#### Sequential Sync
```
brapse: Should this be split into it's own PR?

light-node/src/commands/start.rs
tarcieri: @gterzian the semantics are the same, but yes it would use the Tokio executor.

You can make it synchronous by replacing `block_on` with `await` to wait for the future to complete, otherwise the semantics are the same (except with async/await, you can add async behavior in the future)

docs/architecture/adr-005-light-client-fork-detection.md
Liamsi: Sure! It was a separate PR (see #148) but can definitely be broken out again if it makes sense.

light-node/src/commands/start.rs
gterzian: Yes I think it's worth discussing the overall approach in a separate issue(it shouldn't matter much to be creating a new runtime on each call for now). 

Yes if we call `await` from inside an async context we'll get the equivalent of the current `block_on`, and I think the tricky part is `block_on` is not called from an async context? 

So if we run the entire application on Tokio, we could indeed use `await` instead of `block_on`. 

Probably best to keep the current `block_on`, and then figure out how we want to run the light client.

gterzian: https://github.com/interchainio/tendermint-rs/issues/167

brapse:
Reading through the last comments it seems like there more or less OK with the imperfect state of this PR. It also seems that keeping it un-merged is a liability. Unless there are concrete objections in the next day or so I'm gunna go ahead and merge this.

Liamsi:
Agree with @brapse here. Review and merge as soon as possible makes sense. There is probably a lot followup on. E.g. writing and enabling some acceptance / integration tests that run this against an actual tendermint instance is a major piece that is missing, even for an MVP. But we can followup on this leveraging the outcome of #120.

tarcieri:
FYI, if you do end up merging https://github.com/interchainio/tendermint-rs/pull/171 you'll almost certainly want to use `abscissa_tokio`

Liamsi:
Just merged #171. I'll look into `abscissa_tokio` and resolve merge conflicts here (I plan to do so before end of the week). I hope we can merge this PR (with it's imperfections as Sean said) afterwards.

Liamsi:
Done! @tarcieri @brapse

light-node/src/commands/start.rs
```diff
@@ -0,0 +1,159 @@
...
+pub struct StartCmd {
+    /// RPC address to request headers and validators from.
+    #[options(free)]
+    rpc_addr: String,
+}
+
+impl Runnable for StartCmd {
+    /// Start the application.
+    fn run(&self) {
+        if let Err(err) = abscissa_tokio::run(&APPLICATION, async {
```
gterzian: Perhaps add a comment here about what this does? I assume spawning a task to run the async block? What kind of runtime does it start, threaded or current thread? If threaded, how many? I'd say we just need a basic scheduler for now as in https://docs.rs/tokio/0.1.20/tokio/runtime/current_thread/index.html

tarcieri: The runtime is started as an Abscissa component. You can use whatever runtime you'd like... see the `From<Runtime>` impl on `TokioComponent`:

https://docs.rs/abscissa_tokio/0.5.1/abscissa_tokio/struct.TokioComponent.html

Right now it's using the default, which is equivalent to `tokio::runtime::Runtime::new()`:

https://github.com/interchainio/tendermint-rs/pull/125/files#diff-e4205fba6a4d881725a2a98ea6858b7aR87

By using an Abscissa component to manage the runtime, other components can spawn additional tasks on the reactor or leverage other integrations (e.g. [tracing](https://github.com/tokio-rs/tracing), which Zebra is using extensively)

PR #126, opened 2020-01-19
Update to bytes 0.5 and amino_rs 0.5

zmanian:
Update to bytes_0.5 and amino_rs 0.5 and makes all the amino_rs tokens independent of prost tokens enabling prost and amino_rs to be combined in a dependency.

### DIFF
--- a/tendermint/src/lib.rs
+++ b/tendermint/src/lib.rs
@@ -26,6 +26,7 @@ extern crate prost_amino as prost;
 #[allow(unused_extern_crates)]
 #[macro_use]
 extern crate prost_amino_derive as prost_derive;
+extern crate bytes;
 
 #[macro_use]
 pub mod error;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Pending [0.12.0] (2019-12-11)
+
+Dependencies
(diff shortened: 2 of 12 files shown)


tendermint/src/lib.rs
```diff
@@ -26,6 +26,7 @@ extern crate prost_amino as prost;
 #[allow(unused_extern_crates)]
 #[macro_use]
 extern crate prost_amino_derive as prost_derive;
+extern crate bytes;
```
tarcieri: Are either of these `extern crate`s still needed? I know `prost` was looking directly in the `sysroot` for them before which precluded renaming them, but if everything is namespaced correctly, it seems like it should be possible to get rid of them?

Liamsi:
I assume this was tested locally, as well as running clippy on it? CircleCi doesn't run on forks :-/

tendermint/src/lib.rs
Liamsi: @xla might have addressed some of it here: https://github.com/interchainio/tendermint-rs/pull/122

Liamsi: https://github.com/interchainio/tendermint-rs/pull/127

ebuchman:
> CircleCi doesn't run on forks :-/

Fixed

PR #128, opened 2020-01-19
Remove `extern crate` directives

tarcieri:
They were used as a hack around an old version of `prost`.

Now that `amino_rs` is properly namespaced, we don't need them.

### DIFF
--- a/tendermint/src/lib.rs
+++ b/tendermint/src/lib.rs
@@ -18,15 +18,6 @@
     html_root_url = "https://docs.rs/tendermint/0.11.0"
 )]
 
-// NOTE(EB): can't figure out how to easily remove the extern crate per Rust2018 upgrade ...
-#[allow(unused_extern_crates)]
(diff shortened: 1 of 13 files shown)


tendermint/src/lib.rs
```diff
@@ -18,15 +18,6 @@
...
 
-// NOTE(EB): can't figure out how to easily remove the extern crate per Rust2018 upgrade ...
-#[allow(unused_extern_crates)]
-extern crate prost_amino as prost;
-
-// NOTE(EB): can't figure out how to easily remove the extern crate per Rust2018 upgrade ...
-#[allow(unused_extern_crates)]
-#[macro_use]
-extern crate prost_amino_derive as prost_derive;
-
```
Liamsi: 👍

tarcieri:
Gonna push a Signatory upgrade, otherwise I'd say we're ready to cut a (pre)release

PR #131, opened 2020-01-23
Move Amino type prefixes over from KMS

tarcieri:
I'd like to compute some Amino type prefixes in something other than the KMS (namely an oracle feeder) and unfortunately the code to do so is presently vendored into the KMS, and used to compute these prefixes (with a TODO to move it elsewhere):

https://github.com/tendermint/kms/blob/master/src/rpc.rs#L45

It seems like these should go with the Amino types. Perhaps the `compute_prefix` method should be moved into the `amino_rs` crate.

### DIFF
--- a/tendermint/src/amino_types.rs
+++ b/tendermint/src/amino_types.rs
@@ -17,13 +17,40 @@ pub mod vote;
 
 pub use self::{
     block_id::{BlockId, CanonicalBlockId, CanonicalPartSetHeader, PartsSetHeader},
-    ed25519::{PubKeyRequest, PubKeyResponse, AMINO_NAME as PUBKEY_AMINO_NAME},
-    ping::{PingRequest, PingResponse, AMINO_NAME as PING_AMINO_NAME},
-    proposal::{SignProposalRequest, SignedProposalResponse, AMINO_NAME as PROPOSAL_AMINO_NAME},
+    ed25519::{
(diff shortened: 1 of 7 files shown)


tendermint/src/amino_types.rs
```diff
@@ -17,13 +17,40 @@ pub mod vote;
...
+    vote::{
+        SignVoteRequest, SignedVoteResponse, AMINO_NAME as VOTE_AMINO_NAME,
+        AMINO_PREFIX as VOTE_PREFIX,
+    },
 };
+
+use sha2::{Digest, Sha256};
+
+/// Compute the Amino prefix for the given registered type name
+pub fn compute_prefix(name: &str) -> Vec<u8> {
```
tarcieri: I would like to make use of this directly, which is why I marked it as `pub`.

However, worst case it's simple enough I can probably just copy and paste it.

I think ideally it'd go in `amino-rs`

Liamsi: Agree, this should live in the amino lib instead of in the kms/here. 

Note that there actually is a function that contains computing the prefix and returning it as the 2nd value. 

As far as I remember, it was one of the first things that existed in that amino crate: https://github.com/tendermint/amino_rs/blob/06d64fab61adfed371062700a9b413418ed3763f/prost-amino-derive/src/lib.rs#L583-L604

Not sure why we never made it public/used it directly.

PR #132, opened 2020-01-26
Simplify some light client types and testing

Liamsi:
Main change: make `TrustedState` and `SignedHeader` structs (instead of traits with assoc. types):

- as a result of parametrarizing `TrustedState<C, H>` and `SignedHeader<C, H>`
(instead of trait with assoc. types) a few things have to be (re)parametrarized and
others can be deleted as they are redundant and consumers can use the
param. structs instead of the trait impls (which duplicated code):
  - parametrarize Store<C, H> too
  - delete mocks: MockSignedHeader, MockState
  - delete impl for block::signed_header::SignedHeader
  - delete impl for lite::TrustedState in JSON tests
  - remove `State` and state.rs module from implementation completely
  - adapt params in verifier to match generic params of TrustedState & SignedHeader

- More (minor) changes / improvements:
  - remove dependency on `block::Height` by making `lite::Height` a `u64` type alias
  - differentiate error from `verify_commit_`(`trusting` vs `full`)
  - increase test coverage (slightly only, compare codecovs total below with codecov)
  - simplify test further by removing `votes_len` from `Commit`
  - use `verify_update_single` instead of `verify_single` in tests (unexport the latter) (f2fd6fe55ef9033b2a8e281dc42dc9d364c0327a)

- rename:
  - verify_and_update_single -> verify_single
  - verify_and_update_bisection -> verify_bisection
- remove store from public facing functions

* [x] Wrote tests
* [ ] Updated CHANGES.md


tendermint/Cargo.toml
```diff
@@ -56,3 +56,5 @@ zeroize = { version = "1.1", features = ["zeroize_derive"] }
 [dev-dependencies]
 serde_json = "1"
 tokio = "0.2"
+# Note: used in tests only:
+tendermint-lite = {path = "../tendermint-lite"}
```
Liamsi: This is used in the JSON tests now. They use the mem-store it provides.

tendermint-lite/src/lib.rs
```diff
@@ -1,4 +1,2 @@
 pub mod requester;
-pub mod state;
 pub mod store;
-pub mod threshold;
```
Liamsi: state and threshold modules are not needed anymore as the provided structs `TrustedState<C, H>` and `TrustThresholdFraction` provide exactly the same functionality.

tendermint/src/lite/types.rs
```diff
@@ -107,46 +85,156 @@ pub trait Commit {
...
+    pub fn new(numerator: u64, denominator: u64) -> Self {
+        // TODO: should we make sure we got a proper fraction here?
+        Self {
+            numerator,
+            denominator,
+        }
+    }
+}
+
+// TODO: should this go in the central place all impls live instead? (currently lite_impl)
```
Liamsi: Do we want to keep `types.rs` purely abstract / generic only? If so this should go into a different file/module.

tendermint/src/lite/types.rs
```diff
@@ -107,46 +85,156 @@ pub trait Commit {
...
     /// https://github.com/interchainio/tendermint-rs/issues/118
-    fn get(&self, h: Height) -> Result<&Self::TrustedState, Error>;
+    fn get(&self, h: Height) -> Result<&TrustedState<C, H>, Error>;
+}
+
+/// TrustedState stores the latest state trusted by a lite client,
+/// including the last header (at height h-1) and the validator set
+/// (at height h) to use to verify the next header.
+#[derive(Clone)]
+pub struct TrustedState<C, H>
```
Liamsi: I do remember we've decided to make this (and `SignedHeader `) a struct. I couldn't find an issue where this decision was made. Hence I didn't link to it.

tendermint/src/lite/types.rs
```diff
@@ -243,10 +222,109 @@ pub enum Error {
...
+    #[test]
+    fn trusted_state() {
+        let h = MockHeader {};
+        let c = MockCommit {};
+        let vs = MockValSet {};
+        let sh = SignedHeader::new(c, h);
+        let ts = TrustedState::new(&sh, &vs);
+        assert_eq!(ts.last_header(), &sh);
+        assert_eq!(ts.validators(), &vs);
+    }
```
Liamsi: Added above two tests mostly to see the impact on code coverage: It's a bit annoying that all new lines in `mod tests` also count and hence adding mock types decreases the code coverage despite covering actual (non-test) lines. 

Anyways, guess that's not really important. Will focus on the light-client testing and refactoring.

tendermint/src/lite/verifier.rs
```diff
@@ -260,30 +260,25 @@ where
...
 // see that function's docs.
-fn _verify_and_update_bisection<H, C, L, R, S>(
+fn verify_and_update_bisection_inner<H, C, L, R>(
+    trusted_state: &TrustedState<C, H>,
     untrusted_height: Height,
     trust_threshold: &L,
     req: &R,
-    store: &mut S,
-) -> Result<(), Error>
+) -> Result<TrustedState<C, H>, Error>
```
ancazamfir: imho some explanation in the comments on the new returned value(s) would be useful. Also, I thought this should return a list of trusted states that at the end of the recursion will be added to the store by the caller.

Liamsi: Thanks @ancazamfir. Yeah, I wasn't sure. But yes a list would be exactly the same behaviour as before! Will do. Regarding the comment: this is just a non-public helper method. The public one (`verify_bisection`) has a brief explanation but yes, these could be better too.

Liamsi: Haha, today I learned there is a whole crate dealing with memoization: https://github.com/jaemk/cached
It provides a macro for things like this: https://medium.com/swlh/on-memoization-291fd1dd924
Interesting, but I likely just use a HashSet / HashMap or even a simple Vec (have to think about the latter, as we might request headers twice if there is some overlap?).

ancazamfir: > Haha, today I learned there is a whole crate dealing with memoization: https://github.com/jaemk/cached
It provides a macro for things like this: https://medium.com/swlh/on-memoization-291fd1dd924

Interesting!

> ....or even a simple Vec (have to think about the latter, as we might request headers twice if there is some overlap?).

Vec is nice as it will have the headers in the height order. I don't think we ask for headers twice...or at least we shouldn't.

Liamsi: Made it a HashMap (used more like a HashSet) in 48a7a37. Changing this to a Vec and checking that we don't have double entries in there sounds like a good first test then.

Liamsi: Changed to Vec in ed8c06e. Will add tests in a bit.

tendermint/src/lite/types.rs
```diff
@@ -161,13 +226,230 @@ pub enum Error {
...
-    InvalidCommitLength,
-
-    // TODO(EB): we need to differentiate when this is from
-    // skipping and when it's from verifying !
-    // https://github.com/interchainio/tendermint-rs/issues/119
-    InsufficientVotingPower, /* TODO(Liamsi): change to same name as spec if this changes
-                              * (curently ErrTooMuchChange) */
+    InvalidCommitValue,
+    // commit is not for the header we expected
+    InvalidCommitSignatures,
```
ebuchman: Note this is only used by implementation (ie. expected return in `validate()`)

tendermint/src/lite/verifier.rs
```diff
@@ -330,349 +333,296 @@ where
...
     // valid to skip, but invalid commit. 1 validator.
     #[test]
     fn test_verify_single_skip_1_val_verify() {
-        let ts = &init_state(vec![0]);
+        let ts = &init_trusted_state(vec![0], vec![0], 1);
 
         // 100% overlap, but wrong commit.
         // NOTE: This should be an invalid commit error since there's
-        // a vote from a validator not in the set !
-        // but voting_power_in isn't smart enough to see this ...
```
ebuchman: So this comment is still relevant in another case, for instance `assert_err(ts, vec![1], vec![S0, S1], Error::InvalidCommit)`, but here it's an InvalidCommit because the commit includes a signature from someone not in the validator set (S0). But voting_power_in still can't detect that!

tendermint/src/lite/types.rs
ebuchman: Note this is only used by implementation (ie. expected return in validate()). Maybe it should be distinguished from other errors which `lite` itself will return

tendermint/src/lite/verifier.rs
```diff
@@ -330,349 +333,296 @@ where
...
+        let req = init_requester(vec![vec![0], vec![0], vec![0]]);
+        let sh = req.signed_header(1).expect("first sh not present");
+        let vals = req.validator_set(1).expect("init. valset not present");
+        let ts = &MockState::new(&sh, &vals);
+
+        assert_bisection_ok(&req, &ts, 2, 1);
+    }
+
+    #[test]
+    fn test_verify_bisection_1_val_2_heights() {
```
ebuchman: this can be consolidate with the previous test.

tendermint/Cargo.toml
ebuchman: Do they need to? Can't they just ignore the Store and operate on the returned Vec ?

tendermint/src/lite/types.rs
```diff
@@ -96,57 +66,152 @@ pub trait Commit {
...
-        H: Into<Height>;
+    fn validator_set(&self, h: Height) -> Result<C::ValidatorSet, Error>;
 }
 
 /// This store can be used to store all the headers that have passed basic verification
 /// and that are within the light client's trust period.
-pub trait Store {
-    type TrustedState: TrustedState;
-
+pub trait Store<C, H>
```
ebuchman: Do we even need this trait in here anymore? We don't use it in the `lite` now, users can be free to make any kind of store they like ...

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -82,3 +72,19 @@ impl block::signed_header::SignedHeader {
             .collect()
     }
 }
+
+impl From<block::signed_header::SignedHeader>
+    for lite::types::SignedHeader<block::signed_header::SignedHeader, block::header::Header>
```
ebuchman: can we use `lite::types::SignedHeader<C, H> where C: ..., H: ...` here and below?

tendermint/tests/lite.rs
```diff
@@ -69,68 +61,45 @@ fn header_tests_verify() {
...
-        // also take a store, so we need to mock one ...
-        /*
         let trusting_period: std::time::Duration = tc.initial.clone().trusting_period.into();
-        let now = tc.initial.now;
-        */
+        let tm_now = tc.initial.now;
+        let now = tm_now.to_system_time().unwrap();
+        // Note: Alternative to using mem-store from the light node crate would be mocking
+        // the store here directly.
+        let mut store = store::MemStore::new();
```
ebuchman: We don't really need an externally defined store here, we can just use a cache again.

tendermint/src/lite/types.rs
```diff
@@ -161,13 +226,230 @@ pub enum Error {
...
-    InvalidCommitValue, // commit is not for the header we expected
-    InvalidCommitLength,
-
-    // TODO(EB): we need to differentiate when this is from
-    // skipping and when it's from verifying !
-    // https://github.com/interchainio/tendermint-rs/issues/119
-    InsufficientVotingPower, /* TODO(Liamsi): change to same name as spec if this changes
-                              * (curently ErrTooMuchChange) */
+    InvalidCommitValue,
+    // commit is not for the header we expected
```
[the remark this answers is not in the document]
Liamsi: No, I'm not sure how this happened. Addressed in follow-up PR.

tendermint/src/lite_impl/signed_header.rs
Liamsi: `where` expects traits, so we can say e.g. `where H: block::header::Header`. I can intro a type alias, or an import alias to make this shorter and more readable though.

tendermint/src/lite/verifier.rs
Liamsi: Added back the comment for now which links the tracking issue.

tendermint/src/lite/types.rs
Liamsi: It's gone 🎉

tendermint/src/lite_impl/signed_header.rs
Liamsi: https://github.com/interchainio/tendermint-rs/pull/141/commits/be65629e64b3d7aecf50d94b39147ab9ba61d5ba

PR #134, opened 2020-01-29
Generate code coverage metrics and upload to codecov.io

Liamsi:
- add grcov to circleci and upload coverage metrics to codecov.io

- add cargo audit while at it

Preferably we merge this before other PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   rust/format:
     docker:
-      - image: circleci/rust:buster
+      - image: rust:1
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


marbar3778:
is there a preference to stay with circle? with github actions its a bit nicer since it is all in one place and it has nice support for rust. 

If people want to go to one location for ci and code (github) I can make a follow-up PR to transfer over to GA

Liamsi:
I'm OK with switching to github actions. They look very cool. I didn't look into them that much yet. I mostly chose to not dig deeper as it seemed all the go-repos are still using circle and not really gh-actions? I think circle is mostly used because of performance and configurability.

Would you use this grcov support then? https://github.com/actions-rs/grcov)

Liamsi:
I'll try to figure out why the test coverage reports 0% first :-D If others agree to switch to gh-Actions, I think it would be a good move. Any opinions cc @ebuchman @greg-szabo @xla?

Liamsi:
Since the last commit we the total coverage (finally)... I'll cleanup the config in a bit and then we can merge.

PR #141, opened 2020-02-03
132 followup

Liamsi:
Addresses all comments from #132 (almost all, https://github.com/interchainio/tendermint-rs/pull/132#discussion_r373813746 still needs to be addressed; will be addressed shortly too)

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/lite/types.rs
+++ b/tendermint/src/lite/types.rs
@@ -151,7 +151,7 @@ where
     fn get(&self, h: Height) -> Result<&TrustedState<C, H>, Error>;
 }
 
-/// TrustedState stores the latest state trusted by a lite client,
+/// TrustedState contains a state trusted by a lite client,
 /// including the last header (at height h-1) and the validator set
 /// (at height h) to use to verify the next header.
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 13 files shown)


tendermint/src/lite/types.rs
```diff
@@ -87,19 +87,21 @@ pub trait TrustThreshold: Copy + Clone {
...
 }
 
 impl TrustThresholdFraction {
-    pub fn new(numerator: u64, denominator: u64) -> Self {
-        // TODO: should we make sure we got a proper fraction here?
-        Self {
-            numerator,
-            denominator,
+    pub fn new(numerator: u64, denominator: u64) -> Result<Self, Error> {
+        if numerator <= denominator && denominator > 0 {
```
ebuchman: Actually I wonder if we should bound this below at 1/3 ... it's not really safe with anything less than that

Liamsi:
Thanks a lot for the quick review! I'll merge this as is and will address dealing with the 1/3 bound and https://github.com/interchainio/tendermint-rs/pull/132#discussion_r373813746 in a small separate PR .

PR #143, opened 2020-02-03
Added single-step-skipping JSON tests

Shivani912:
Purpose: To add the remaining single step skipping JSON test cases

Changes:
- Modified the test case structure -> test_name is changed to batch_name
- Diretory structure -> Categorized and split the json files into single_step_sequential and single_step_skipping
- Added tests for single_step_skipping cases


Liamsi:
Can you point me to the exact case in the JSON where the tests pass despite the wrong header hash?

tendermint/tests/support/lite/single_step_sequential/commit_tests.json
```diff
@@ -0,0 +1,2360 @@
...
+ 							"next_validators_hash": "75B9F27F25F3515EF7CA61DD42C272CCC280C537E6A9ED493710EB361CB90B0D",
+ 							"consensus_hash": "048091BC7DDC283F77BFBF91D73C44DA58C3DF8A9CBC867405D8B7F3DAADA22F",
+ 							"app_hash": "6170705F68617368",
+ 							"last_results_hash": "",
+ 							"evidence_hash": "",
+ 							"proposer_address": "01F527D77D3FFCC4FCFF2DDC2952EEA5414F2A33"
+ 						},
+ 						"commit": {
+ 							"block_id": {
+ 								"hash": "2D2D2D2D5468697320697320612033322D6279746520737472696E672D2D2D2D",
```
Shivani912: @Liamsi This is the "wrong header hash" in the commit which is passing this check 
` if header.hash() != commit.header_hash() {
        return Err(Error::InvalidCommitValue);
}`

Liamsi: Awesome, thanks! Will look into it now.

Liamsi: Can I see the code that generated this hash? `2D2D2D2D5468697320697320612033322D6279746520737472696E672D2D2D2D`

Shivani912: Here it is: https://github.com/Shivani912/tendermint/blob/shivani-refactor/types/lite-client/generator/single_step_sequential_cases.go#L120

Liamsi: That's really weird: I've added debug output to the pure `validate` function:
```rust
println!("header.hash()={:?}", header.hash());
println!("commit.header_hash()={:?}", commit.header_hash());
```
which yields:
```
header.hash()=Hash::Sha256(E3D4300A0C730F2912219F257314C4F3CE95E4E0954FB9F5E224B4FD076C51C0)
commit.header_hash()=Hash::Sha256(E3D4300A0C730F2912219F257314C4F3CE95E4E0954FB9F5E224B4FD076C51C0)
```

Liamsi: Wait this test passes locally :-D It's `Case: one lite block, wrong PartsHeader.Total, with error` that is causing problems.

Liamsi: This one is "description": "Case: one lite block, wrong header hash, with error"

tendermint/tests/support/lite/single_step_sequential/commit_tests.json
```diff
@@ -0,0 +1,2360 @@
...
+ 							"voting_power": "50",
+ 							"proposer_priority": "0"
+ 						}
+ 					}
+ 				}
+ 			],
+ 			"expected_output": "error"
+ 		},
+ 		{
+ 			"description": "Case: one lite block, wrong PartsHeader.Total, with error",
```
Liamsi: That's the problematic test!

tendermint/tests/support/lite/single_step_sequential/commit_tests.json
Shivani912: Ah, I got confused from the output of tests. Sorry!

Liamsi:
0a30022 solves the problem with the "Case: one lite block, wrong PartsHeader.Total, with error", next one in that file is "Case: one lite block, less than one-third vals don't sign, no error"

tendermint/tests/support/lite/single_step_sequential/commit_tests.json
```diff
@@ -0,0 +1,2360 @@
...
+ 								"voting_power": "50",
+ 								"proposer_priority": "-50"
+ 							},
+ 							{
+ 								"address": "03A238BCAF7D1626DFE8A4AFB9448D00B7A3D2E2",
+ 								"pub_key": {
+ 									"type": "tendermint/PubKeyEd25519",
+ 									"value": "b6hwk3pjiOTJfLVCcLDA3I3lO71zWJ0VSded5LUl9T0="
+ 								},
+ 								"voting_power": "50",
```
Liamsi: That's the next test that is failing. Quick question: Should we rename this to: "Case: one lite block, more than two-third vals did sign, no error". I was first confused why there is no error if less than 1/3 signed :-D 

This one seems to fail on the newly introduced check in regarding the parts.hash btw.

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -47,6 +47,12 @@ impl lite::Commit for block::signed_header::SignedHeader {
         if self.commit.precommits.len() != vals.validators().len() {
             return Err(lite::Error::InvalidCommitSignatures);
         }
+        if let Some(parts) = &self.commit.block_id.parts {
+            if parts.hash != self.header_hash() {
```
Shivani912: @Liamsi `parts.hash` is the `merkleRoot(partsHeader)` which is different from the `header_hash`. That's why this check is making many other tests to fail too. We want to check that the `commit.BlockID` is equal to each of the `precommit.BlockID`. In reality, there can be precommits for other blocks and that's totally fine. But then, those precommits should not be considered for this block. This means that when we calculate the total voting power i.e. `signed_power`, we do not count voting power of validators who signed these precommits. Maybe this check should go under the `voting_power_in` method(?)

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -47,6 +47,13 @@ impl lite::Commit for block::signed_header::SignedHeader {
         if self.commit.precommits.len() != vals.validators().len() {
             return Err(lite::Error::InvalidCommitSignatures);
         }
+        // TODO: https://github.com/interchainio/tendermint-rs/pull/143/commits/0a30022fa47e909e6c7b20417dd178c8a3b84838#r374958528
+        // if let Some(parts) = &self.commit.block_id.parts {
+        //     if parts.hash != self.header_hash() {
+        //         // TODO(ismail): we need a better error here!
+        //         return Err(lite::Error::InvalidCommit);
+        //     }
+        // }
```
Liamsi: I think it's OK to actually delete this now (instead of commenting it out), format this PR with `cargo fmt --` and merge this as is. I think it would be OK to fix the one failing test in a follow up PR.

tendermint/src/lite_impl/signed_header.rs
Liamsi: I see. Thanks for the clarification! To move forward, I'd suggest to merge this, and open an issue referencing this comment and the failing test.

tendermint/tests/support/lite/single_step_sequential/commit_tests.json
Shivani912: Oh it's "less than one-third don't sign" which is same as "more than two-thirds did sign". That's why it does not error on the go side too.

Liamsi: Me wondering was more about the parts hash. I understand the equiv. of both and suggested to rename the description to the latter "more than two-thirds did sign" as it is less confusing (but maybe it's just me).

Shivani912: Ah okay. I can do that

Shivani912:
> It seems now that some of the existing test files can be deleted? Even the descriptions are the same. I think it's fine to review and remove those unless they contain further test cases which aren't in the new ones. The rest LGTM! 💪

Done!

Liamsi:
Congrats to your 1st contribution to tendermint-rs 🎉 👍

Shivani912:
> Awesome! Thanks @Shivani912!
> 
> One last ask: Can you link to the version of your go-code with which you've generated the JSON files? E.g. just here in the comments; could be via linking a commit (after that do not force push in your branch please). Then someone could reproduce this :-)

Here: https://github.com/Shivani912/tendermint/tree/shivani-refactor/types/lite-client
I'll be pushing some more code soon!

tendermint/src/lite_impl/signed_header.rs
ebuchman: Oh, is this about the bug I raised here: https://github.com/interchainio/tendermint-rs/issues/145#issuecomment-583790278 ? We should be testing (either here or in voting_power_in) that the votes being tallied are for the right block ID

PR #144, opened 2020-02-05
ci: change from circle to github actions

### LINKED ISSUE #135: CI: switch to Github Actions?
[Github Actions](https://github.com/features/actions) is an alternative to circleci. It looks worth exploring! Things like code coverage and uploading to codecov.io looks much simpler for instance (it's a bit like [circleci's](https://circleci.com/orbs/?gclid=EAIaIQobChMIs4aGsKOp5wIVieN3Ch2qBwzxEAAYASAAEgIy4_D_BwE) [orbs](https://circleci.com/orbs/registry/orb/glotrade/rust) but integrated into github; also the snippets are hosted here on this platform)

Discussion started here:
https://github.com/interchainio/tendermint-rs/pull/134#issuecomment-579710445

marbar3778:
added github actions in place of circleci

A repo admin will need to provide codecov with the token via the repo settings

- closes #135 

* [ ] Wrote tests
* [ ] Updated CHANGES.md


marbar3778:
odd github actions are not firing

Liamsi:
> odd github actions are not firing

They are now 👍 The error you are seeing is fixed here: https://github.com/interchainio/tendermint-rs/pull/166

marbar3778:
there are some todo's that can be done prioir to the merge or after:

- [ ] remove circle
- [ ] add codecov token

marbar3778:
why cargo-audit is not firing: 
```
It is recommended to add the paths: section into the workflow file, as it would effectively speed up the CI pipeline, since the audit process will not be performed if no dependencies were changed.
```

Liamsi:
> why cargo-audit is not firing: 
>```
>It is recommended to add the paths: section into the workflow file, as it would effectively speed up >the CI pipeline, since the audit process will not be performed if no dependencies were changed.
>```

Can we do sth to fix this here? Not sure I understand this correctly. Because the path comment sounds like an optimization only (and not like an explanation why audit isn't triggered). But maybe I'm missing sth here.

tarcieri:
Here's my recommendation for how to run `cargo audit`:

https://github.com/iqlusioninc/tmkms/blob/develop/.github/workflows/security_audit.yml

It's set to run on either a change to `Cargo.toml`/`Cargo.lock`, or to run once daily.

That recipe also caches the installed `cargo-audit` binary to ensure subsequent runs are fast because they don't have to compile it each time (at least in the event there was a cache hit). With that, you could potentially run it every time and make it blocking if you like.

brapse:
@greg-szabo will look to merge this towards the end of the week, would be great to have your feedback before then 🙏

.github/workflows/audit.yml
```diff
@@ -0,0 +1,15 @@
+name: Audit-Check
+on:
+  push:
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
```
Liamsi: This could be run as a cron job / e.g. scheduled nightly. See: https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule

I'm wondering if this shouldn't be consistent for informal repos? see https://github.com/informalsystems/ibc-rs/blob/db92c7241ae6ea723bb4fbe1148de71f4f8a9493/.github/workflows/audit.yaml#L8-L9

.github/workflows/audit.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/audit-check@v1
```
Liamsi: Similar question regarding the cache: https://github.com/informalsystems/ibc-rs/blob/db92c7241ae6ea723bb4fbe1148de71f4f8a9493/.github/workflows/audit.yaml#L15-L21

But probably this could be done in a followup PR!

.github/actions-rs/grcov.yml
```diff
@@ -0,0 +1,13 @@
...
+output-type: lcov
+output-file: ./lcov.info
+prefix-dir: /home/user/build/
+# ignore:
+#   - "/*"
+#   - "C:/*"
+#   - "../*"
+# path-mapping:
+#   - "/path1"
+#   - "/path2"
```
Liamsi: Why not delete the commented out parts?

.github/actions-rs/grcov.yml
```diff
@@ -0,0 +1,13 @@
+branch: true
+ignore-not-existing: true
+llvm: true
+output-type: lcov
+output-file: ./lcov.info
+prefix-dir: /home/user/build/
```
Liamsi: Is that a default path from github actions?

.github/actions-rs/grcov.yml
```diff
@@ -0,0 +1,13 @@
+branch: true
```
Liamsi: As a general note on github actions, it would be good to make sure we didn't substantially alter the config and will report very different cov numbers after merging. I doesn't seem so but I didn't look into the type-script code for this action.

.github/workflows/audit.yml
tarcieri: You can do both: https://github.com/iqlusioninc/armistice/blob/develop/.github/workflows/security_audit.yml#L8

If you cache the install of `cargo audit` it doesn't take that long, so might as well (I don't block merging on it though, since it only runs with `Cargo.*` changes

marbar3778: will add cache and cron job like in the armstice repo. will replicate in the ibc repo as well

Liamsi:
No objections from my side. I think this could be merged as is. Would be good if https://github.com/interchainio/tendermint-rs/pull/144#issuecomment-595269744 were addressed sooner than later, too.

greg-szabo:
This looks like a nice start for using GitHub Actions. It achieves the goal of moving away from CircleCI and implements the basic set of tests that we had in CircleCI with less scripting and more modular methods.
I can see improvements in it that I would like to work on in a next PR, like using caching for Rust in general and I'm not really sure why auditing is cached (and then not used). But those are details we can work on after this gets merged.

Let's merge this as soon as possible, so I can put additional work on top of it.

.github/actions-rs/grcov.yml
tarcieri: Just noticed this (too late!)

I'd love to give another shout out to Tarpaulin. It's pretty much the go-to coverage tool for the Rust community, IMO:

https://github.com/actions-rs/tarpaulin

I've personally experienced a lot of false negatives trying grcov in the past. I don't know if they've gotten any better, but if you see things you think should be covered under grcov which it reports aren't... maybe check out tarpaulin.

Here's one of my example configs, FWIW:

https://github.com/RustCrypto/AEADs/blob/master/.github/workflows/workspace.yml#L41

PR #146, opened 2020-02-07
JSON bisection tests

Shivani912:
Added some initial code for JSON bisection tests (WIP)

### DIFF
--- a/tendermint/tests/lite.rs
+++ b/tendermint/tests/lite.rs
@@ -45,9 +45,11 @@ struct TestBisection {
     trust_options: TrustOptions,
     primary: Provider,
     height_to_verify: Height,
+    // TODO: make trust_level deserialize from string instead of integer
+    // TODO: trust_level should go under TrustOptions
     trust_level: TrustThresholdFraction,
     now: Time,
-    expected_output: String,
+    expected_output: Option<String>,
 }
 
 #[derive(Deserialize, Clone, Debug)]
@@ -58,6 +60,7 @@ struct Provider {
 
 #[derive(Deserialize, Clone, Debug)]
 struct TrustOptions {
+    // TODO: add trust_level here
     period: Duration,
     height: Height,
     hash: Hash,
@@ -96,6 +99,13 @@ impl MockRequester {
     fn new(chain_id: String, lite_blocks: Vec<LiteBlock>) -> Self {
         let mut sh_map: HashMap<u64, SignedHeader> = HashMap::new();
         let mut val_map: HashMap<u64, Set> = HashMap::new();
+        let last_block = lite_blocks
+            .get(lite_blocks.len() - 1)
+            .expect("last entry not found");
+        val_map.insert(
+            last_block.signed_header.header.height.increment().value(),
+            last_block.to_owned().next_validator_set,
+        );
         for lite_block in lite_blocks {
             let height = lite_block.signed_header.header.height;
             sh_map.insert(height.into(), lite_block.signed_header);
@@ -197,6 +207,8 @@ fn run_test_cases(cases: TestCases) {
     }
 }
 
+// Link to the commit where the happy_path.json was created:
+// https://github.com/Shivani912/tendermint/commit/89aa17ab9ae0a76941eb15b7957452ec78ce5696
 #[test]
 fn bisection_simple() {
     let case: TestBisection =
@@ -215,7 +227,10 @@ fn run_bisection_test(case: TestBisection) {
     let provider = case.primary;
     let req = MockRequester::new(provider.chain_id, provider.lite_blocks);
 
-    let expected_output = case.expected_output;
+    let expects_err = match &case.expected_output {
+        Some(eo) => eo.eq("error"),
+        None => false,
+    };
 
     let trusted_height = case.trust_options.height.try_into().unwrap();
     let trusted_header = &req
@@ -227,8 +242,6 @@ fn run_bisection_test(case: TestBisection) {
 
     let trusted_state = TrustedState::new(trusted_header, trusted_vals);
 
-    let output: String;
-
     match lite::verify_bisection(
         trusted_state,
         untrusted_height,
@@ -237,15 +250,25 @@ fn run_bisection_test(case: TestBisection) {
         &now.into(),
         &req,
     ) {
-        Ok(_) => {
-            // TODO: should we make some assertions on the returned new_states?
-            output = "no error".to_string();
+        Ok(new_states) => {
+            let untrusted_signed_header = &req
+                .signed_header(untrusted_height)
+                .expect("header at untrusted height not found");
+            let untrusted_next_vals = &req
+                .validator_set(untrusted_height + 1)
+                .expect("val set at untrusted height not found");
+            let expected_state = TrustedState::new(
+                &untrusted_signed_header.to_owned().into(),
+                untrusted_next_vals,
+            );
+            assert_eq!(new_states[new_states.len() - 1], expected_state);
+            assert_eq!(new_states.len(), 2);
+            assert!(!expects_err);
         }
         Err(_) => {
-            output = "error".to_string();
+            assert!(expects_err);
         }
     }
-    assert_eq!(output, expected_output);
 }
 
 impl<'de> Deserialize<'de> for Duration {
--- a/tendermint/tests/lite.rs
+++ b/tendermint/tests/lite.rs
@@ -256,7 +256,7 @@ fn run_bisection_test(case: TestBisection) {
                 .validator_set(untrusted_height + 1)
                 .expect("val set at untrusted height not found");
             let expected_state = TrustedState::new(
-                &untrusted_signed_header.to_owned(), //.into(),
+                &untrusted_signed_header.to_owned(),
                 untrusted_next_vals,
             );
             assert_eq!(new_states[new_states.len() - 1], expected_state);
--- a/tendermint/tests/lite.rs
+++ b/tendermint/tests/lite.rs
@@ -46,7 +46,7 @@ struct TestBisection {
     primary: Provider,
     height_to_verify: Height,
     trust_level: TrustThresholdFraction,
-    now: std::time::SystemTime,
+    now: Time,
     expected_output: String,
 }
 
@@ -235,7 +235,7 @@ fn run_bisection_test(case: TestBisection) {
         untrusted_height,
         trust_threshold,
         &trusting_period.into(),
-        &now,
+        &now.into(),
         &req,
     ) {
         Ok(_) => {
--- a/tendermint/tests/lite.rs
+++ b/tendermint/tests/lite.rs
@@ -99,9 +99,7 @@ impl MockRequester {
     fn new(chain_id: String, lite_blocks: Vec<LiteBlock>) -> Self {
         let mut sh_map: HashMap<u64, SignedHeader> = HashMap::new();
         let mut val_map: HashMap<u64, Set> = HashMap::new();
-        let last_block = lite_blocks
-            .get(lite_blocks.len() - 1)
-            .expect("last entry not found");
+        let last_block = lite_blocks.last().expect("last entry not found");
         val_map.insert(
             last_block.signed_header.header.height.increment().value(),
             last_block.to_owned().next_validator_set,
--- a/tendermint/tests/lite.rs
+++ b/tendermint/tests/lite.rs
@@ -1,10 +1,11 @@
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize};
 use serde_json;
+use std::collections::HashMap;
 use std::convert::TryInto;
-use std::{fs, path::PathBuf, collections::HashMap};
-use tendermint::block::Header;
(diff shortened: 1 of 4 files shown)


Liamsi:
Note (for later): this needs rebasing on current master. The commit 
be6e3ea is 'outdated'.

tendermint/tests/lite.rs
```diff
@@ -43,8 +38,51 @@ struct LiteBlock {
...
+    height_to_verify: i64,
+    trust_level: TrustThresholdFraction,
+    now: std::time::SystemTime,
+    expected_output: String,
+}
+
+#[derive(Deserialize, Clone, Debug)]
+struct TrustOptions {
+    period: std::time::Duration,
+    height: i64,
```
Liamsi: Can the height ever be negative here?

Shivani912: No, it should be u64

tendermint/tests/lite.rs
```diff
@@ -125,6 +198,57 @@ fn run_test_cases(cases: TestCases) {
     }
 }
 
+#[test]
+fn bisection_simple() {
+    let case: TestBisection =
+        serde_json::from_str(&read_json_fixture("many_header_bisection/happy_path")).unwrap();
```
Shivani912: @Liamsi 
TODO: fix needed here. Failing to unwrap with error "missing field `height`" at line 7 col 3

Liamsi: Will fix it. I get a slightly different error (not related to height) though:
```
called `Result::unwrap()` on an `Err` value: Error("parse error: bad encoding", line: 7, column: 3)
thread 'bisection_simple' panicked at 'called `Result::unwrap()` on an `Err` value: Error("parse error: bad encoding", line: 7, column: 3)', src/libcore/result.rs:1165:5
```

Liamsi: The reason was a typo in height: https://github.com/interchainio/tendermint-rs/pull/146/commits/8f78bbd0ce3a9cf2dbcaf97b7503d7abe9c901bb
not sure if this stems from the generator or was changed manually.

Liamsi:
If we leave `"height_to_verify": "11"` we need the validators for height `12` which are not provided in the JSON file 🤔 I'll change it manually to 10.

tendermint/tests/lite.rs
Liamsi: Fixed in 8f78bbd0ce3a9cf2dbcaf97b7503d7abe9c901bb

Liamsi:
As we can see from the output, we do not really "bisect": only height 1 and 10 are requested (because we can skip ahead). We should prob. change this test (or add another one) where we request at least one pivot header too.
Output:
```
Case: Trusted height=1, bisecting to verify height=12, should not expect error
requested signed header for height:1
requested validators for height:2
requested signed header for height:10
requested validators for height:10
requested validators for height:11
```

tendermint/tests/support/lite/many_header_bisection/happy_path.json
```diff
@@ -1,5 +1,5 @@
 {
- 	"description": "Case: Trusted height=1, bisecting to verify height=12, should not expect error",
+ 	"description": "Case: Trusted height=1, bisecting to verify height=11, should not expect error",
```
Liamsi: This should actually be 10 now:
```suggestion
 	"description": "Case: Trusted height=1, bisecting to verify height=10, should not expect error",
```

tendermint/tests/lite.rs
```diff
@@ -125,6 +197,57 @@ fn run_test_cases(cases: TestCases) {
...
```
ebuchman: For each JSON we test, we should link to the Go generator that created it, so we know how it was generated. For now those generators are in Shivanis fork of tendermint Go, so we can reference it there for now, but we should work to get that stuff merged in.

ebuchman: Also does the generator for this have enough validator set changes to require actual bisection (ie. intermediate requests)? We should make sure it does, and figure out a nice way to describe the different amounts of bisection we want to test for.

ebuchman: Doesn't look like it. We're using the `generateNextBlock` to make the blocks:  https://github.com/Shivani912/tendermint/blob/shivani-refactor/types/lite-client/generator/single_step_utils.go#L191. We'll need some way to specify to `generateNextBlock` in the Go generators to update the validator set at certain heights. Perhaps the `privVals` we pass in should be a structure that returns the validator set for each height.

Shivani912:
> If we leave `"height_to_verify": "11"` we need the validators for height `12` which are not explicitly provided in the JSON file.
> 
> They are needed here:
> 
> https://github.com/interchainio/tendermint-rs/blob/994b457a8a8fe4f42b0657cb70f3cf78b2dbe885/tendermint/src/lite/verifier.rs#L306-L307
> 
> Changed it manually to 10 in [bc650f0](https://github.com/interchainio/tendermint-rs/commit/bc650f0fdcdaa449f901ac86c3a3372c4d5a636b).
> 
> Note we can also achieve to go to 11 by adding these two lines to `MockRequester::new`:
> 
> ```rust
> let last_block = lite_blocks.get(lite_blocks.len()-1).expect("last entry not found");
> val_map.insert(last_block.signed_header.header.height.increment().value(), last_block.to_owned().next_validator_set);
> ```

I think it makes sense to look for the next validators at height 11, instead of supplying and additional signed header.

Shivani912:
> As we can see from the output, we do not really "bisect": there is not much change in the validatior set; only height 1 and 10 are requested (because we can skip ahead). We should prob. change this test (or add another one) where we request at least one pivot header too.
> Output:
> 
> ```
> Case: Trusted height=1, bisecting to verify height=11, should not expect error
> requested signed header for height:1
> requested validators for height:2
> requested signed header for height:10
> requested validators for height:10
> requested validators for height:11
> ```

Will be adding more tests for bisection. This was only to start with the simplest possible test

tendermint/tests/lite.rs
```diff
@@ -37,6 +39,76 @@ struct LiteBlock {
     next_validator_set: Set,
 }
 
+#[derive(Deserialize, Clone, Debug)]
+struct TestBisection {
+    description: String,
+    trust_options: TrustOptions,
+    primary: Provider,
+    height_to_verify: Height,
+    trust_level: TrustThresholdFraction,
```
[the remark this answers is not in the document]
Shivani912: I thought the same but I'm using TrustOptions struct from the Go lite client code. Maybe make a TrustParameters struct and put all the information there?

tendermint/tests/lite.rs
```diff
@@ -125,6 +197,57 @@ fn run_test_cases(cases: TestCases) {
...
+    match lite::verify_bisection(
+        trusted_state,
+        untrusted_height,
+        trust_threshold,
+        &trusting_period.into(),
+        &now.into(),
+        &req,
+    ) {
+        Ok(_) => {
+            // TODO: should we make some assertions on the returned new_states?
```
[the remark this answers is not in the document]
Shivani912: Yes, we can do that. My only concern is that since this will test each step of bisection, any changes in the bisection algorithm in future might require updating each test file every time.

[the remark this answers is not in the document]
ebuchman: Fair enough. Then maybe we do a bunch of tests pass/fail where we check that it synced to the final height or not, and then a few tests that also check the intermediate headers, and only those will have to be updated with if the bisection algorithm changes (though at first thought I don't think we're expecting it to change :) )

Liamsi:
I think this is almost good to be merged. The only things that should be addressed before this can be merged:
 - [ ] check at least for final state + assert num of intermediate requests https://github.com/interchainio/tendermint-rs/pull/146#discussion_r376741972 (or similar assertions like in the Rust only tests)
- [ ] link to the commit (ideally) with which the file was generated

Everything else could be addressed in a follow-up PR:
 - make sure there is enough change in the val set s.t. we actually bisect (more tests)
 - test erroring paths too (more tests)

tendermint/tests/lite.rs
Shivani912: All of this addressed in commit 98d168f. Thanks! :)

tendermint/tests/lite.rs
Shivani912: That was a typo in the json tag in generator. Thanks!

tendermint/tests/lite.rs
```diff
@@ -37,6 +39,86 @@ struct LiteBlock {
...
+        Err(Error::RequestFailed)
+    }
+}
+
+impl MockRequester {
+    fn new(chain_id: String, lite_blocks: Vec<LiteBlock>) -> Self {
+        let mut sh_map: HashMap<u64, SignedHeader> = HashMap::new();
+        let mut val_map: HashMap<u64, Set> = HashMap::new();
+        let last_block = lite_blocks
+            .get(lite_blocks.len() - 1)
```
Shivani912: TODO: clippy fails here, suggests to use lite_block.last() instead. But then fails to compile

Liamsi: I'll look into it now.

Liamsi: Huh, this actually works. You can apply the suggestion here, pull locally, apply `cargo fmt --` and we are good.
```suggestion
            .last()
```

tendermint/tests/lite.rs
```diff
@@ -37,6 +39,86 @@ struct LiteBlock {
...
+        println!("couldn't get vals for: {}", &h);
+        Err(Error::RequestFailed)
+    }
+}
+
+impl MockRequester {
+    fn new(chain_id: String, lite_blocks: Vec<LiteBlock>) -> Self {
+        let mut sh_map: HashMap<u64, SignedHeader> = HashMap::new();
+        let mut val_map: HashMap<u64, Set> = HashMap::new();
+        let last_block = lite_blocks
```
Liamsi:
```suggestion
        let last_block = lite_blocks.last().expect("last entry not found");
```

tendermint/tests/lite.rs
Liamsi:
```suggestion
```

tendermint/tests/lite.rs
```diff
@@ -37,6 +39,86 @@ struct LiteBlock {
...
+            .get(lite_blocks.len() - 1)
+            .expect("last entry not found");
```
Liamsi:
```suggestion
```

tendermint/tests/lite.rs
```diff
@@ -125,6 +205,70 @@ fn run_test_cases(cases: TestCases) {
     }
 }
 
+// Link to the commit where the happy_path.json was created:
+// https://github.com/Shivani912/tendermint/commit/89aa17ab9ae0a76941eb15b7957452ec78ce5696
```
Liamsi: Make sure you keep this up-to date whenever you generate a new file.

tendermint/tests/lite.rs
```diff
@@ -125,6 +205,70 @@ fn run_test_cases(cases: TestCases) {
...
+    ) {
+        Ok(new_states) => {
+            let untrusted_signed_header = &req
+                .signed_header(untrusted_height)
+                .expect("header at untrusted height not found");
+            let untrusted_next_vals = &req
+                .validator_set(untrusted_height + 1)
+                .expect("val set at untrusted height not found");
+            let expected_state = TrustedState::new(
+                &untrusted_signed_header.to_owned(), //.into(),
```
Liamsi: What's with the comment here? Looks like this is obsolete now?
```suggestion
                &untrusted_signed_header.to_owned(),
```

tendermint/tests/lite.rs
```diff
@@ -125,6 +205,70 @@ fn run_test_cases(cases: TestCases) {
...
+        &req,
+    ) {
+        Ok(new_states) => {
+            let untrusted_signed_header = &req
+                .signed_header(untrusted_height)
+                .expect("header at untrusted height not found");
+            let untrusted_next_vals = &req
+                .validator_set(untrusted_height + 1)
+                .expect("val set at untrusted height not found");
+            let expected_state = TrustedState::new(
```
Liamsi:
```suggestion
            let expected_state =
```

tendermint/tests/lite.rs
```diff
@@ -125,6 +205,70 @@ fn run_test_cases(cases: TestCases) {
...
+                &untrusted_signed_header.to_owned(),
```
Liamsi:
```suggestion
                TrustedState::new(&untrusted_signed_header.to_owned(), untrusted_next_vals);
```

tendermint/tests/lite.rs
```diff
@@ -125,6 +205,70 @@ fn run_test_cases(cases: TestCases) {
...
+                untrusted_next_vals,
```
Liamsi:
```suggestion
```

tendermint/tests/lite.rs
```diff
@@ -125,6 +205,70 @@ fn run_test_cases(cases: TestCases) {
...
+            );
```
Liamsi:
```suggestion
```

Issue #139, opened 2020-02-02
More data in light client errors


ebuchman:
Follow up to #110 Errors.

- [x] `Error::RequestFailed` should wrap underlying requester errors (https://github.com/interchainio/tendermint-rs/pull/100#discussion_r360341870)
- [x] Other error types should wrap relevant data
- [x] Remove `Error::InvalidSignature`? It's not used directly by the `light` package, only by implementations

Liamsi:
I'll give this a stab shortly.

Liamsi:
Reopened because `InvalidSignature` part still needs addressing.

Liamsi:
The last todo was addressed in #151. This can be closed now.

PR #149, opened 2020-02-11
Light: more data in errors

Liamsi:
This uses the [failure](https://boats.gitlab.io/failure/intro.html) crate to annotate the errors we have with relevant data and to tackle #139.

closes https://github.com/interchainio/tendermint-rs/issues/139

* [ ] Wrote tests
* [ ] Updated CHANGES.md


Liamsi:
This is still a draft PR but I'd appreciate some early feedback here. cc @tarcieri @ebuchman @xla @Shivani912

tendermint/src/lite/errors.rs
```diff
@@ -0,0 +1,72 @@
+//! All error types tied to the light client.
+
+use crate::Hash;
+use failure::*;
+use std::time::{SystemTime, SystemTimeError};
+
```
Liamsi: We could add a struct wrapping the ErrorKind as described here: https://boats.gitlab.io/failure/error-errorkind.html

tarcieri:
I'd consider `failure` to be "soft ecosystem deprecated". It's still maintained, but probably not the best choice for new projects. Note that e.g. `tmkms` used to use it and I ripped it out.

I wrote up some background on various modern replacements here:

https://github.com/iqlusioninc/abscissa/issues/144

Personally these days I use use `thiserror` instead of `failure_derive`, and have replaced `failure` itself with `anomaly`:

https://github.com/iqlusioninc/crates/tree/develop/anomaly

Liamsi:
Thanks Tony! I'll checkout the combination of `thiserror` + `anomaly` and likely change this PR to use both then.

tendermint/src/lite/errors.rs
```diff
@@ -0,0 +1,61 @@
+//! All error types tied to the light client.
+
+use crate::Hash;
+use std::time::{SystemTime, SystemTimeError};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Error)]
+pub enum ErrorKind {
```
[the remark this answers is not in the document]
Liamsi: And remove the type alias `Error` then?

tendermint/src/lite/errors.rs
```diff
@@ -1,72 +1,61 @@
 //! All error types tied to the light client.
```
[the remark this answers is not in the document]
Liamsi: Nope, will rename to error.rs

tendermint/src/lite/errors.rs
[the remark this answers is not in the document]
xla: Which alias you referring to? Maybe I missed it in my review.

tendermint/src/lite/error.rs
```diff
@@ -0,0 +1,73 @@
+//! All error types tied to the light client.
+
+use crate::Hash;
+use anomaly::{BoxError, Context};
+use std::time::{SystemTime, SystemTimeError};
+use thiserror::Error;
+
+/// The main error type verification methods will return.
+/// See [`ErrorKind`] for the different kind of errors.
+pub type Error = anomaly::Error<ErrorKind>;
```
Liamsi: This alias @xla

tendermint/src/lite/errors.rs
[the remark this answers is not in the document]
Liamsi: https://github.com/interchainio/tendermint-rs/pull/149#discussion_r377634242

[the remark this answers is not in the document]
xla: Had an old view. Disregard, as long as consumers of the error mod return `Error` all is well.

tendermint/src/lite/error.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Clone, Debug, Error)]
+pub enum ErrorKind {
+    #[error("old header has expired at {at:?} (now: {now:?})")]
+    Expired { at: SystemTime, now: SystemTime },
+
+    #[error("duration error {:?}", _0)]
+    DurationOutOfRange(#[from] SystemTimeError),
```
xla: This is neat.

tendermint/src/lite/types.rs
```diff
@@ -332,7 +311,10 @@ pub(super) mod mocks {
         fn validate(&self, _vals: &Self::ValidatorSet) -> Result<(), Error> {
             // some implementation specific checks:
             if self.vals.is_empty() || self.hash.algorithm() != Algorithm::Sha256 {
-                return Err(Error::InvalidCommitSignatures);
+                return Err(ErrorKind::InvalidCommitSignatures {
+                    info: "validator set is empty, or, invalida hash algo".to_string(),
```
Liamsi:
```suggestion
                    info: "validator set is empty, or, invalid hash algo".to_string(),
```

tendermint/src/lite/error.rs
```diff
@@ -0,0 +1,73 @@
...
+use anomaly::{BoxError, Context};
+use std::time::{SystemTime, SystemTimeError};
+use thiserror::Error;
+
+/// The main error type verification methods will return.
+/// See [`ErrorKind`] for the different kind of errors.
+pub type Error = anomaly::Error<ErrorKind>;
+
+#[derive(Clone, Debug, Error)]
+pub enum ErrorKind {
```
xla:
```suggestion
pub enum Kind {
```
Prefixes which are equivalent of the mod name are superfluous. This reads nicely: `error::Kind`.

tendermint/src/lite/error.rs
xla:
```suggestion
pub type Error = anomaly::Error<Kind>;
```

tendermint/src/lite/error.rs
```diff
@@ -0,0 +1,73 @@
...
+    }, // commit is not for the header we expected
+
+    #[error("error validating commit signatures: {info}")]
+    InvalidCommitSignatures { info: String }, // Note: this is only used by implementation (ie. expected return in Commit::validate())
+
+    #[error("signed voting power ({signed}) do not account for +2/3 of the total voting power: ({total})")]
+    InvalidCommit { total: u64, signed: u64 },
+
+    #[error("signed voting power ({}) is too small fraction of total voting power: ({}), threshold: {}",
+        .signed, .total, .trust_treshold
```
xla: Why are the fields passed here explicitely, yet not in any of the other cases?

tendermint/src/lite/error.rs
```diff
@@ -0,0 +1,73 @@
...
+    }, // trust threshold (default +1/3) is not met
+
+    #[error("{0:?}")]
+    RequestFailed(String), // Note: we don't want to depend on the tendermint rpc error types, otherwise we should do #[from] Error instead of expecting a string
+
+    #[error("A valid threshold is `1/3 <= threshold <= 1`, got: {got}")]
+    InvalidTrustThreshold { got: String },
+}
+
+impl ErrorKind {
```
xla:
```suggestion
impl Kind {
```

tendermint/src/lite/error.rs
Liamsi: Only because the line was getting too long. There are different way how to access the fields: https://github.com/dtolnay/thiserror#details

tendermint/src/lite/error.rs
Liamsi: Good point! Addressed (needed to change this in other places too).

Liamsi:
Thanks for the review @xla!

tendermint-lite/src/main.rs
```diff
@@ -101,7 +102,7 @@ fn subjective_init(
     vals_hash: Hash,
     store: &mut MemStore,
     req: &RPCRequester,
-) -> Result<(), Error> {
+) -> Result<(), Kind> {
```
ebuchman: Probably clearer to use error::Kind here, no ? Not a big deal

tendermint/src/lite/error.rs
```diff
@@ -0,0 +1,73 @@
+//! All error types tied to the light client.
+
+use crate::Hash;
+use anomaly::{BoxError, Context};
+use std::time::{SystemTime, SystemTimeError};
+use thiserror::Error;
+
+/// The main error type verification methods will return.
+/// See [`ErrorKind`] for the different kind of errors.
+pub type Error = anomaly::Error<Kind>;
```
ebuchman: How is this used? Seems everything is returning Kind

tendermint/src/lite/verifier.rs
```diff
@@ -29,11 +29,15 @@ where
     match now.duration_since(last_header.bft_time().into()) {
         Ok(passed) => {
             if passed > *trusting_period {
-                return Err(Error::Expired);
+                return Err(Kind::Expired {
+                    at: last_header.bft_time().into() + *trusting_period,
```
ebuchman: Would it be clearer to have some variable like `expires_at = ...` for this and then check `if expires_at < now` instead of using `passed` ?

tendermint/src/lite/error.rs
Liamsi: As per #151 everything should be using Error and not Kind anymore. (Can you re-approve btw).

tendermint/src/lite/verifier.rs
Liamsi: Will address in a followup PR 👍

Liamsi: We can do this but then we must additionally check for header_time > now. Good side effect we can simplify  DurationOutOfRange(#[from] SystemTimeError) to simply `DurationOutOfRange` (maybe the name should actually be `HeaderInTheFuture` or sth like this.

Also, shouldn't this be `expires_at <= now` instead of `<`? On equality we have just expired...

Liamsi: See: 1833e77

Issue #150, opened 2020-02-11
Failure crate: "soft ecosystem deprecated"


Liamsi:
- Related discussion: https://github.com/interchainio/tendermint-rs/pull/149#issuecomment-584448738

- Reasoning in abscissa repo: https://github.com/iqlusioninc/abscissa/issues/144

- Good read: https://blog.yoshuawuyts.com/error-handling-survey/

Currently:
https://github.com/interchainio/tendermint-rs/blob/f9287e04ee84e6659ea209529a1bb700ab53e45b/tendermint/Cargo.toml#L36

Used in:
 - [ ] tendermint-rs/error.rs 
 - [ ] tendermint-rs/amino_types/validate.rs
 - [ ] tendermint-rs/rpc/error.rs

We should switch to a combination of thiserror + anonmaly for the whole code base as done in #151 and #149 for the light client module.

Liamsi:
I think we should use `thiserror` to replace the dependency on the failure derives. To have a single error-type for the light client lib (which methods can return instead of an enum variant) `anomaly` looks like a very minimalistic, very good choice.

We could, of course, write a bunch of boilerplate code ourself and not depend on any other lib but this will eat our time for no good reason.

ref related issue: #139.
https://github.com/dtolnay/thiserror
https://docs.rs/anomaly/0.2.0/anomaly/

xla:
> I think we should use `thiserror` to replace the dependency on the failure derives. To have a single error-type for the light client lib (which methods can return instead of an enum variant) `anomaly` looks like a very minimalistic, very good choice.

What does that mean for the error type itself in the light module, will it be modeled like in the anomaly example and carry an enum kind?
> We could, of course, write a bunch of boilerplate code ourself and not depend on any other lib but this will eat our time for no good reason.

What is the typical boilerplate code we would have to maintain? I think it's important to know the cost/downsize.

> ref related issue: #139.
> https://github.com/dtolnay/thiserror
> https://docs.rs/anomaly/

Liamsi:
> What does that mean for the error type itself in the light module, will it be modeled like in the anomaly example and carry an enum kind?

We have an enum with different errors (ErrorKind) anyways because we want to differentiate different errors.

> What is the typical boilerplate code we would have to maintain? I think it's important to know the cost/downsize.

See for example: https://doc.rust-lang.org/src/std/io/error.rs.html#91-174

Liamsi:
[...]

PR #151, opened 2020-02-11
Light errors: Remove `InvalidSignature` and `InvalidCommitSignatures` 

Liamsi:
Followup to #149:
- remove `InvalidSignature` and `InvalidCommitSignatures`
- introduce `ImplementationSpecific` Kind instead
- simplify `RequestFailed` in the same way

closes #149 

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/lite/error.rs
+++ b/tendermint/src/lite/error.rs
@@ -6,43 +6,52 @@ use std::time::{SystemTime, SystemTimeError};
 use thiserror::Error;
 
 /// The main error type verification methods will return.
-/// See [`ErrorKind`] for the different kind of errors.
+/// See [`Kind`] for the different kind of errors.
 pub type Error = anomaly::Error<Kind>;
 
+/// All error kinds related to the light client.
 #[derive(Clone, Debug, Error)]
 pub enum Kind {
+    /// The provided header expired.
     #[error("old header has expired at {at:?} (now: {now:?})")]
     Expired { at: SystemTime, now: SystemTime },
 
+    /// Trusted header is from the future.
     #[error("duration error {:?}", _0)]
     DurationOutOfRange(#[from] SystemTimeError),
 
+    /// Header height smaller than expected.
     #[error("expected height >= {expected} (got: {got})")]
     NonIncreasingHeight { got: u64, expected: u64 },
 
+    /// Invalid validator hash.
     #[error("header's validator hash does not match actual validator hash ({header_val_hash:?}!={val_hash:?})")]
     InvalidValidatorSet {
         header_val_hash: Hash,
         val_hash: Hash,
     },
 
+    /// Invalid next validator hash.
     #[error("header's next validator hash does not match next_val_hash ({header_next_val_hash:?}!={next_val_hash:?})")]
     InvalidNextValidatorSet {
         header_next_val_hash: Hash,
         next_val_hash: Hash,
     },
 
+    /// commit is not for the header we expected.
     #[error(
         "header hash does not match the hash in the commit ({header_hash:?}!={commit_hash:?})"
     )]
     InvalidCommitValue {
         header_hash: Hash,
         commit_hash: Hash,
-    }, // commit is not for the header we expected
+    },
 
+    /// Signed power does not account for +2/3 of total voting power.
     #[error("signed voting power ({signed}) do not account for +2/3 of the total voting power: ({total})")]
     InvalidCommit { total: u64, signed: u64 },
 
+    /// This means the trust threshold (default +1/3) is not met.
     #[error("signed voting power ({}) is too small fraction of total voting power: ({}), threshold: {}",
         .signed, .total, .trust_treshold
     )]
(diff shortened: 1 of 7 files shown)


tendermint/src/lite/error.rs
```diff
@@ -63,6 +57,9 @@ pub enum Kind {
 
     #[error("A valid threshold is `1/3 <= threshold <= 1`, got: {got}")]
     InvalidTrustThreshold { got: String },
+
+    #[error("Implementation specific error: {0}")]
+    ImplementationSpecific(String),
```
Liamsi: This isn't optimal in my eyes. Would be cool if we wouldn't  have to deal with strings here.

xla: This will invite lazy catch all for implementations. What control flow is desired for consumers?

Liamsi: That's a very good point. In all other cases the verifier produces errors and consumers (e.g the light node implementation) can match over those and handle them as they wish. Here, roles are kinda reversed because these are errors that come from an implementation. Maybe these should rather be std errors instead. I'm currently looking into using the `context` method to see if we can decorate  `ImplementationSpecific ` with the original error (which doesn't need to be a String / could be anything which implements `Into<BoxError>`. Then, we have the same flexibility and users can get the source error back (and log it or do what they want with it).

https://github.com/interchainio/tendermint-rs/blob/ca6a17a4a9e4d6c65bdb6a19ccf069237919ad6e/tendermint/src/lite/error.rs#L68-L73

Liamsi: That actually works! Had to modify rpc error to implement std::error:Error and I think we are good now.

tendermint/src/rpc/error.rs
```diff
@@ -95,12 +96,6 @@ impl Display for Error {
     }
 }
 
-impl Fail for Error {
-    fn name(&self) -> Option<&str> {
-        self.code.name()
-    }
-}
-
```
Liamsi: Had to remove this as `std::error::Error` implements this too. I more convinced now that we should switch to thiserror + anomaly everwhere in this repo.
ref  #150

tarcieri: The KMS doesn't use the RPC client (yet) so that should be fine

Liamsi: Cool, thanks!

Liamsi: I'll change the RPC errors to also use thiserror + anomaly in a different PR then.

tendermint/src/lite/error.rs
```diff
@@ -6,63 +6,74 @@ use std::time::{SystemTime, SystemTimeError};
...
     },
 
+    /// Invalid next validator hash.
     #[error("header's next validator hash does not match next_val_hash ({header_next_val_hash:?}!={next_val_hash:?})")]
     InvalidNextValidatorSet {
         header_next_val_hash: Hash,
         next_val_hash: Hash,
     },
 
+    /// commit is not for the header we expected.
```
ebuchman:
```suggestion
    /// Commit is not for the header we expected.
```

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -35,7 +35,13 @@ impl lite::Commit for block::signed_header::SignedHeader {
             let sign_bytes = vote.sign_bytes();
 
             if !val.verify_signature(&sign_bytes, vote.signature()) {
-                return Err(Kind::InvalidSignature.into());
+                return Err(Kind::ImplementationSpecific
+                    .context(format!(
+                        "Couldn't verify signature {:?} with validator {:?}",
```
ebuchman: Should we include the sign bytes in here too ?

Liamsi: Added in 8c99269bec65fcc4dbe9e9134562fba3a5f0cb27

tarcieri: Note that if you do use anomaly, this whole thing can be replaced by:

```rust
use anomaly::fail;
fail!(Kind::ImplementationSpecific, "Couldn't verify signature {:?} with validator {:?} on sign_bytes {:?}", vote.signature(), val, sign_bytes);
```

Liamsi: Aha, good to now, will use `fail!` here and everywhere else then. As always, cool library BTW! Maybe the example doc could be slightly more elaborate.

tendermint/src/lite/error.rs
```diff
@@ -6,63 +6,74 @@ use std::time::{SystemTime, SystemTimeError};
...
-
-    #[error("{0:?}")]
-    RequestFailed(String), // Note: we don't want to depend on the tendermint rpc error types, otherwise we should do #[from] Error instead of expecting a string
+    },
 
+    /// This is returned if an invalid TrustThreshold is created.
     #[error("A valid threshold is `1/3 <= threshold <= 1`, got: {got}")]
     InvalidTrustThreshold { got: String },
+
+    /// Use the [`Kind::context`] method to wrap the underlying error of
```
xla: How does the access of the context work from a callers perspective, given they encountered a wrapped `RequestFailed`? Asking mostly because I'm not familiar with the APIs yet.

tarcieri: It's type erased and stored as a `Box<std::error::Error>`, and available as the [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source).

If you just want a string representation, it impls `Display`, otherwise if you'd like the concrete type again you need to use [`Error::downcast_ref`](https://doc.rust-lang.org/std/error/trait.Error.html#method.downcast_ref)

Liamsi:
> Let's capture it in a follow-up issue.

Thought, that's what https://github.com/interchainio/tendermint-rs/issues/150 is for.

And awesome, thanks for the review! Will merge as is and followup with https://github.com/interchainio/tendermint-rs/pull/151#discussion_r377935655 in a separate PR.

tendermint/src/lite/error.rs
Liamsi: @xla I've added some small tests using source to explore and document how this looks like:
https://github.com/interchainio/tendermint-rs/blob/52e21325983f445fc87bdb7c888e7ae1859275bd/tendermint/src/lite/error.rs#L91-L127

PR #163, opened 2020-02-27
Shivani/detect faulty vals

### LINKED ISSUE #140: Light client voting_power_in can't detect invalid signers
Follow up to one point in #110:

The `voting_power_in` method can't detect signatures from validators not in the current validator set. While this isn't really a big deal, since we can't be fooled into accepting something that wasn't correctly signed, there is no good reason for us to receive a commit with a signature from a validator who's not in the validator set, and thus we should be able to detect it and mark the full node we got it from as faulty.

Shivani912:
closes #140 

* [x] Wrote tests
* [ ] Updated CHANGES.md


tendermint/src/lite/error.rs
```diff
@@ -1,6 +1,6 @@
 //! All error types tied to the light client.
 
-use crate::Hash;
+use crate::{account, Hash};
```
Liamsi: We don't want to introduce dependencies from from the `tendermint` crate into the light module. (the other way around is OK though).

tendermint/src/lite/error.rs
```diff
@@ -78,6 +78,10 @@ pub enum Kind {
     /// the implementation, if any.
     #[error("Implementation specific error")]
     ImplementationSpecific,
+
+    /// This is returned when a validator has signed but isn't a part of the validator set
+    #[error("Found faulty validator: {address}")]
+    FaultyValidator { address: account::Id },
```
Liamsi: A way around this, could be to remove the address / account::Id and annotate the error in the implementation using the `fail!` macro. Or we could make use of `ImplementationSpecific` with the proper annotation instead (as it is in fact implementation specific). Not sure what would be the best approach here. Will `FaultyValidator` be used by the light module itself? If not, that could be seen as indicator that we do not necessarily need to define a particular error kind in the light module (currently still called lite) for this 🤔

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -75,14 +75,19 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
                         }
                     }
+
+                    // returns FaultyValidator error if the signer isn't present in the validator set
+                    if vals.validator(precommit.validator_address) == None {
+                        return Err(Kind::FaultyValidator {
+                            address: precommit.validator_address,
+                        }
+                        .into());
+                    }
```
Liamsi: Nice! 👍

tendermint/src/lite/error.rs
Liamsi: Note that if we decide to keep this error kind here, it should probably also be in the spec (cc @milosevic).

Shivani912: Good point @Liamsi . Should `FaultyValidator` go under Tendermint module, then? Also, this can be converted into a generalized error type `FaultyFullNode` and this error should trigger verification logic to mark this as a "faulty node" in the address book as explained in the [spec](https://github.com/tendermint/spec/tree/master/spec/consensus/light-client#components) (if i'm reading it right). You could also attach a reason with this error, for example in this case it can be - 'the signed header received from this node contains commit from a validator not in the validator set' for more clarity. This way, we also get rid of the extra dependency. The reason why I think we shouldn't use `ImplementationSpecific` here is that this is a serious misbehaviour and must be distinctly identifiable. 

If that's not sth we want to do, I'm okay to use `ImplementationSpecific` but then I'm not sure how updating the address book would work.

Liamsi: Thanks for the explanation! You're right:
> Should FaultyValidator go under Tendermint module, then?

I think that makes the most sense.

Shivani912: Great! Thanks :)

Liamsi: Unless the logic that that needs to handle the error kind `FaultyFullNode` will live inside light module (which it sounds like in the spec/t[he diagram](https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/assets/light-node-image.png)). But then we do not want that additional dependency on `account::Id` (there are probably alternatives around this, e.g. as you mentioned above). If that is the simplest way to unblock you and this PR, we can also keep it here (and just remove the dependency on account::Id). It wouldn't be a biggie to move that error kind around later anyways.

Liamsi: If needed in the verification logic directly, we might introduce a trait for testing equality between peers later (doesn't need to be the concrete account::Id).

tendermint/src/lite/error.rs
```diff
@@ -78,6 +78,10 @@ pub enum Kind {
     /// the implementation, if any.
     #[error("Implementation specific error")]
     ImplementationSpecific,
+
+    /// This is returned when a faulty i.e misbehaving full node is found
+    #[error("Faulty full node: {reason}")]
+    FaultyFullNode { reason: String },
```
Liamsi: I don't think we need the `String` reason here, as anomaly lets you annotate / provide (string) context for errors.

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -75,14 +76,21 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
                     }
+
+                    // returns FaultyFullNode error if it detects a signer isn't present in the validator set
+                    if vals.validator(precommit.validator_address) == None {
+                        let reason = format!(
+                            "Found a faulty signer ({}) not present in the validator set ({})",
+                            precommit.validator_address,
+                            vals.hash()
+                        );
+                        return Err(Kind::FaultyFullNode { reason }.into());
```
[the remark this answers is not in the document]
Shivani912: Thanks @Liamsi

tendermint/tests/lite.rs
```diff
@@ -119,6 +119,8 @@ impl MockRequester {
     }
 }
 
+// Link to the commit that generated below JSON test files:
+// https://github.com/Shivani912/tendermint/commit/f7d16ab59b55a4f1a5cdbfa6b0c24467aa88fdb2
```
Shivani912: @Liamsi Here is the link to the commit that generated the new val_set_tests.json

Shivani912: I added this test that fails with FaultyFullNode error:
https://github.com/interchainio/tendermint-rs/blob/b28b3da6049cb4b0868e3c234f447c22daa2d289/tendermint/tests/support/lite/single_step_sequential/val_set_tests.json#L13078

Here, the input contains a validator set of 3 vals (v1, v2, v3) for header at height 2 with precommits from 3 vals (v1, v2, v4). This makes it pass all other validation tests but fails because there is a vote from val v4 which is not in the val set and thus there is a chance that the full node who sent this data is misbehaving. 

@Liamsi

Shivani912:
> Awesome work! Thanks @Shivani912 👍
> 
> Although a bit redundant, I think it would be good to add a (rust only) unit test for the (implementation specific) `validate` method. This and adding the error to the spec can happen in separate PRs.

Thanks @Liamsi ! I actually pinged @milosevic about this and we're not currently sure on what level of specification this should appear. We probably need some discussions over this. Do we have a spec where these error types are defined?

For rust only test, we need a way to generate votes in the code. I was looking into it, but couldn't find how to produce signatures for votes.

Liamsi:
> and we're not currently sure on what level of specification this should appear. We probably need some discussions over this. 

IMHO wherever we specify detecting (and handling) cases of faulty nodes, we need to mention these errors (prob [here](https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/detection.md)).

> Do we have a spec where these error types are defined?

Not yet AFAIK. The other error kinds also live in the spec though: (e.g see `fatalError` helper function in  https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/verification.md).

Liamsi:
> For rust only test, we need a way to generate votes in the code. I was looking into it, but couldn't find how to produce signatures for votes.

I see. It might not be too difficult to generate a signature for votes (with the same libs / deps we are using for verification) but it looks definitely like more work than I though. Thanks for looking into this  👍

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -53,36 +54,43 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+                                Kind::ImplementationSpecific,
                                 "validator({}) voted for header {}, but current header is {}",
                                 precommit.validator_address,
                                 header_hash,
                                 self.header_hash()
                             );
                         }
                     }
+
+                    // returns FaultyFullNode error if it detects a signer isn't present in the validator set
```
ebuchman: I think this is actually an ImplementationSpecific error, since it depends on Commits containing the validator set addresses. So we don't need a new error type in the light client.

Shivani912: Makes sense @ebuchman. Will address that :)

PR #164, opened 2020-03-02
Make TrustedState serializable, Fix #161

### LINKED ISSUE #161: No serialization support for `TrustedState`
Since `TrustedState` is supposed to be stored, so it'll need serialization. But users can't implement trait for external types, so it's better to builtin?

yihuang:
- Add Serialize/Deserialize instances to TrustedState and SignedHeader
- Change MemStore to use serialization

### DIFF
--- a/tendermint/src/lite/types.rs
+++ b/tendermint/src/lite/types.rs
@@ -144,7 +144,7 @@ where
 /// TrustedState contains a state trusted by a lite client,
 /// including the last header (at height h-1) and the validator set
 /// (at height h) to use to verify the next header.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct TrustedState<C, H>
 where
     H: Header,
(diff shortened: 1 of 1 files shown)


tendermint-lite/src/store.rs
```diff
@@ -25,17 +25,18 @@ impl MemStore {
...
 
-    pub fn get(&self, h: Height) -> Result<&State, Error> {
+    pub fn get(&self, h: Height) -> Result<State, Error> {
         let mut height = h;
         if h == 0 {
             height = self.height
         }
         match self.store.get(&height) {
-            Some(state) => Ok(state),
+            Some(s) => Ok(serde_json::from_str(&s).unwrap()),
```
Liamsi: What is the reason for serializing in the MemStore?

yihuang: It's only to demostrate it's serializable, but I'm ok to remove it.

yihuang:
> Without a good reason, I'd vote to not change the `MemStore`(I don't see why we'd want to store State as JSON in memory). Other than that this LGTM 👍

Done

Liamsi:
ref #161

Issue #170, opened 2020-03-10
Turn `lite::types::Requester` into an async trait


romac:
Requester not being an async trait forces implementations to block on RPC queries, eg. by spawning a new executor, which is undesirable considering that we eventually want the whole codebase to play nice with async/.await, and not unnecessarily block the enclosing task.

This can be accomplished with [`async-trait`]() crate, which turns async fns within traits and their implementations into plain functions returning `Pin<Box<dyn Future<Output = T + ...>>>`.

The obvious drawback of this approach is that these async Fns now allocate memory on the heap. Hopefully, the language will soon stabilize the feature, which will help us get rid of these undesirable heap allocations.

PR #171, opened 2020-03-10
Turn `lite::types::Requester` into an async trait

romac:
Closes: #170

On top of the changes mentioned in the linked issue, `lite::verifier::verify_bisection_inner` had to be manually adapted to return a boxed future because recursive async fn are not currently supported in Rust.

An possible alternative would be to rewrite `verify_bisection_inner` to not be recursive anymore, but that is outside of the scope of this PR.

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint-lite/Cargo.toml
+++ b/tendermint-lite/Cargo.toml
@@ -8,4 +8,6 @@ edition = "2018"
 
 [dependencies]
 tendermint = { path = "../tendermint" }
-tokio = "0.2"
+
+async-trait = "0.1"
+tokio = { version = "0.2", features = ["full"] }
--- a/tendermint-lite/src/main.rs
+++ b/tendermint-lite/src/main.rs
@@ -7,11 +7,9 @@ use tendermint::{block::Height, Hash};
 
 use tendermint_lite::{requester::RPCRequester, store::MemStore};
 
-use core::future::Future;
 use std::time::{Duration, SystemTime};
 use tendermint::lite::error::Error;
 use tendermint_lite::store::State;
-use tokio::runtime::Builder;
 
 // TODO: these should be config/args
 static SUBJECTIVE_HEIGHT: u64 = 1;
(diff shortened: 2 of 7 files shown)


tendermint/src/lite/verifier.rs
```diff
@@ -333,75 +337,81 @@ where
...
+fn verify_bisection_inner<'a, H, C, L, R>(
+    trusted_state: &'a TrustedState<C, H>,
     untrusted_height: Height,
     trust_threshold: L,
-    req: &R,
-    mut cache: &mut Vec<TrustedState<C, H>>,
-) -> Result<TrustedState<C, H>, Error>
+    req: &'a R,
+    mut cache: &'a mut Vec<TrustedState<C, H>>,
+) -> LocalBoxFuture<'a, Result<TrustedState<C, H>, Error>>
```
Liamsi: It is a bit unfortunate that the we had to change the method signature to return a boxed future. But this change is good in the sense that it surfaces that there is still a too tight coupling between verification logic and the requester (or networking / IO).

@brapse also suggested yesterday that we should probably implement the iterative version here too. It seems like this is a good idea for several reasons now.

tendermint-lite/src/main.rs
```diff
@@ -7,35 +7,25 @@ use tendermint::{block::Height, Hash};
...
-        .basic_scheduler()
-        .enable_all()
-        .build()
-        .unwrap()
-        .block_on(future)
-}
-
-fn main() {
+#[tokio::main]
+async fn main() {
```
Liamsi: I can look into incorporating these changes into the abscissa version of this main (in #125) likely with `abscissa_tokio` as @tarcieri suggested.

tendermint-lite/Cargo.toml
```diff
@@ -8,4 +8,6 @@ edition = "2018"
 
 [dependencies]
 tendermint = { path = "../tendermint" }
-tokio = "0.2"
+
+async-trait = "0.1"
+tokio = { version = "0.2", features = ["full"] }
```
Liamsi: Just to double check: do we need _all_ Tokio public API features? Is there a good reason to enable them all anyways?

tendermint/src/lite/verifier.rs
Liamsi: https://github.com/interchainio/tendermint-rs/issues/174

tendermint-lite/Cargo.toml
romac: Yeah sorry, `macros` should be sufficient (as in #173).

tendermint/src/lite/verifier.rs
```diff
@@ -566,8 +579,8 @@ mod tests {
     }
 
     // valid to skip, but invalid commit. 1 validator.
-    #[test]
-    fn test_verify_single_skip_1_val_verify() {
+    #[tokio::test]
+    async fn test_verify_single_skip_1_val_verify() {
```
ebuchman: Do the tests which don't use async functions need to be async?

Liamsi: No, they don't. Maybe this wasn't intentional (only the bisection test use async functions).

Liamsi: https://github.com/informalsystems/tendermint-rs/pull/195

PR #181, opened 2020-03-12
New commit struct!

### LINKED ISSUE #145: SignedHeader impl needs to be updated
The Commit and CommitSig struct in Go code has been updated to [Commit](https://github.com/tendermint/tendermint/blob/master/types/block.go#L556) and [CommitSig](https://github.com/tendermint/tendermint/blob/master/types/block.go#L452) , while we are using the older version which looks like this:
`type Commit struct {
     BlockID BlockID,
     Precommits []*CommitSig
}
`and `type CommitSig Vote`

This makes the current Rust implementation incompatible with the current Go implementation and we cannot validate these SignedHeader.

Shivani912:
closes #145 

NOTE: I've updated the commit struct here but as a result we'll have to give up compatibility with current Gaia (assuming they're still using the older commit struct)

* [ ] Wrote tests
* [ ] Updated CHANGES.md


Shivani912:
@Liamsi There are rpc JSON tests here that are failing because of the new commit struct. I don't have pointers to the code that generated those files. Do you want to take it from here?

Also, would appreciate your early review!

Liamsi:
Thanks @Shivani912! I'll review and look into the failing tests (I'm also not sure from what they where generated; the problem is that a bunch of these aren't up to date even without the commit changes).

Shivani912:
@Liamsi Ahan! Time for more tests then, haha 🙃

tendermint/src/account.rs
```diff
@@ -12,7 +12,8 @@ use subtle::{self, ConstantTimeEq};
 use subtle_encoding::hex;
 
 /// Size of an  account ID in bytes
-const LENGTH: usize = 20;
+// TODO: Is it okay to publicize this??
```
Liamsi: Sure, why not. Especially as it's a const it can't hurt. It only means that this is more exposed users of the API. It was implicitly accessible anyways...

Shivani912:
Note: All the JSON files have been re-generated so they are compatible with the latest changes. Link to generator updated in code!

Liamsi:
It looks like the RPC tests where either copy & pasted from the documentation, or, dumped via a modified fork. I can't find any code that generated the RPC JSON tests either. @tarcieri can you confirm that these files weren't generated with some tool in tendermint/gaiad?

tarcieri:
@Liamsi the JSON used in the original RPC unit tests was taken from a `cosmoshub-2` full node

tendermint/tests/rpc.rs
```diff
@@ -48,28 +48,29 @@ mod endpoints {
...
-
-        assert_eq!(last_commit.as_ref().unwrap().precommits.len(), 2);
-        assert!(last_commit.unwrap().precommits[0]
-            .as_ref()
-            .unwrap()
-            .block_id
-            .is_none());
-    }
+    // NOTE: Since the commit struct changed, the votes i.e. CommitSig no longer contains BlockID
+    // TODO: Do we still need this test?
```
Liamsi: No, we don't need that exact test anymore. But it would we good to translate it into the new semantic. It tested that the blockId is none. Which looks suspiciously to saying `BlockIDFlagAbsent` now. But I'll double check.

Liamsi:
Nice, the tests @greg-szabo added fired 👍

greg-szabo:
I looked through the failing integration tests and found the following work to do so far:
1. abci_info: app_version was missing, I've just added that (it's quite verbose, we might want to talk about it)
2. abci_info: still fails with unknown parsing error, needs more checks
3. block_results: missing fields: TxsResults, BeginBlockEvents, EndBlockEvents, ValidatorUpdates, ConsensusParamUpdates. The old "results" field need to be removed.
4. genesis, evidence Params: missing fields: max_age_num_blocks, max_age_duration. The old field "max_age" has to be removed.

Shivani912:
After a discussion with @greg-szabo, I'm closing this pull request. Since Gaia is using Tendermint v0.32 and we don't want to give up compatibility with it, the master branch will stay as is i.e. compatible with Tendermint v0.32. So, all the v0.33 related upgrades will stay in a separate branch. That means, this branch will be compatible with Tendermint v0.33.

ebuchman:
@Shivani912 we reverted this decision right? We're going to update master to tendermint v0.33 and drop compatibility with current gaia?

Shivani912:
> @Shivani912 we reverted this decision right? We're going to update master to tendermint v0.33 and drop compatibility with current gaia?

Yes, that's right @ebuchman. For a record, the progress is being posted in issue #184

PR #183, opened 2020-03-16
Enable integration tests + GitHub Actions Service for Tendermint node

greg-szabo:
* [X] Referenced an issue explaining the need for the change. Issue: #120 
* [X] Wrote tests
* [X] Updated CHANGES.md

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -47,11 +47,31 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --all --all-targets
+          args: --workspace --all-targets
       - uses: actions-rs/cargo@v1
         with:
           command: test
           args: --all-features --no-fail-fast
+
+  test-integration-ignored:
+    runs-on: ubuntu-latest
+    services:
+      tendermint:
+        image: interchainio/tendermint
+        ports:
+          - 26656:26656
+          - 26657:26657
+          - 26660:26660
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: -p tendermint --test integration --no-fail-fast -- --ignored
 
   test-nightly-coverage:
     runs-on: ubuntu-latest
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 Dependencies
 - Update to bytes `0.5` and amino_rs `0.5`.
 - Tokens for amino_rs are now fully non-conflicting with prost. Allowing both to be used together 
-- Made RPC type values optional for full compatibility with tendermint-go: `abci_info`, `abci_query` [#120]
+- Made RPC type values optional for full compatibility with tendermint-go@v0.32: `abci_info`, `abci_query` [#120]
 
 ## [0.11.0] (2019-12-11)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 Dependencies
 - Update to bytes `0.5` and amino_rs `0.5`.
 - Tokens for amino_rs are now fully non-conflicting with prost. Allowing both to be used together 
-
+- Made RPC type values optional for full compatibility with tendermint-go: `abci_info`, `abci_query` [#120]
 
 ## [0.11.0] (2019-12-11)
 
--- a/tendermint/src/rpc/endpoint/abci_info.rs
+++ b/tendermint/src/rpc/endpoint/abci_info.rs
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
 
 /// ABCI information
 #[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(default)]
 pub struct AbciInfo {
     /// Name of the application
     pub data: String,
 
     /// Version
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
 
     /// Last block height
-    pub last_block_height: block::Height,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub last_block_height: Option<block::Height>,
 
     /// Last app hash for the block
     #[serde(
         serialize_with = "serialize_app_hash",
-        deserialize_with = "parse_app_hash"
+        deserialize_with = "parse_app_hash",
+        skip_serializing_if = "Option::is_none"
     )]
-    pub last_block_app_hash: Hash,
+    pub last_block_app_hash: Option<Hash>,
 }
 
 /// Parse Base64-encoded app hash
-pub(crate) fn parse_app_hash<'de, D>(deserializer: D) -> Result<Hash, D::Error>
+pub(crate) fn parse_app_hash<'de, D>(deserializer: D) -> Result<Option<Hash>, D::Error>
 where
     D: Deserializer<'de>,
 {
     let bytes = base64::decode(String::deserialize(deserializer)?.as_bytes())
         .map_err(|e| D::Error::custom(format!("{}", e)))?;
 
-    Hash::new(hash::Algorithm::Sha256, &bytes).map_err(|e| D::Error::custom(format!("{}", e)))
+    Hash::new(hash::Algorithm::Sha256, &bytes)
+        .map(Some)
+        .map_err(|e| D::Error::custom(format!("{}", e)))
 }
 
 /// Serialize Base64-encoded app hash
-pub(crate) fn serialize_app_hash<S>(hash: &Hash, serializer: S) -> Result<S::Ok, S::Error>
+pub(crate) fn serialize_app_hash<S>(hash: &Option<Hash>, serializer: S) -> Result<S::Ok, S::Error>
 where
     S: Serializer,
 {
-    String::from_utf8(base64::encode(hash.as_bytes()))
+    String::from_utf8(base64::encode(hash.unwrap().as_bytes()))
         .unwrap()
         .serialize(serializer)
+}
+
+impl Default for AbciInfo {
+    fn default() -> Self {
+        AbciInfo {
+            data: "".to_string(),
+            version: None,
+            last_block_height: None,
+            last_block_app_hash: None,
+        }
+    }
 }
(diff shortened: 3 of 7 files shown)


tendermint/src/rpc/endpoint/abci_info.rs
```diff
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
 
 /// ABCI information
 #[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(default)]
 pub struct AbciInfo {
     /// Name of the application
     pub data: String,
 
     /// Version
+    #[serde(skip_serializing_if = "Option::is_none")]
```
Liamsi: Interesting. So this usually would be serialized to `null` but now we skip this completely in the resulting JSON (I didn't check if the go code also omits this but it probably works anyways).

.github/workflows/rust.yml
```diff
@@ -47,12 +47,32 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: -p tendermint --test integration --no-fail-fast -- --ignored
```
Liamsi: Note to myself: this won't run the light client tests yet.

CHANGES.md
```diff
@@ -3,7 +3,7 @@
 Dependencies
 - Update to bytes `0.5` and amino_rs `0.5`.
 - Tokens for amino_rs are now fully non-conflicting with prost. Allowing both to be used together 
-
+- Made RPC type values optional for full compatibility with tendermint-go: `abci_info`, `abci_query` [#120]
```
Liamsi: Maybe add a version number (0.32.x ?).

tendermint/src/rpc/endpoint/abci_info.rs
```diff
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
...
 where
     D: Deserializer<'de>,
 {
     let bytes = base64::decode(String::deserialize(deserializer)?.as_bytes())
         .map_err(|e| D::Error::custom(format!("{}", e)))?;
 
-    Hash::new(hash::Algorithm::Sha256, &bytes).map_err(|e| D::Error::custom(format!("{}", e)))
+    Hash::new(hash::Algorithm::Sha256, &bytes)
+        .map(Some)
+        .map_err(|e| D::Error::custom(format!("{}", e)))
```
Liamsi: Maybe add a little one-liner comment above this code or in the function doc, explaining when it errs, when it returns Some and when None? What seems confusing to me: it never returns None, right? Either it errs or it returns Some(...).

tendermint/src/rpc/endpoint/abci_info.rs
```diff
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
...
 }
 
 /// Serialize Base64-encoded app hash
-pub(crate) fn serialize_app_hash<S>(hash: &Hash, serializer: S) -> Result<S::Ok, S::Error>
+pub(crate) fn serialize_app_hash<S>(hash: &Option<Hash>, serializer: S) -> Result<S::Ok, S::Error>
 where
     S: Serializer,
 {
-    String::from_utf8(base64::encode(hash.as_bytes()))
+    String::from_utf8(base64::encode(hash.unwrap().as_bytes()))
```
Liamsi: Won't this panic if hash is `None`?

tendermint/src/rpc/endpoint/abci_query.rs
```diff
@@ -10,6 +10,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct Request {
     /// Path to the data
+    #[serde(skip_serializing_if = "Option::is_none")]
     path: Option<Path>,
```
Liamsi: Surprising that this can be empty. But yeah, in JSON everything seems possible :-D

.github/workflows/rust.yml
```diff
@@ -47,12 +47,32 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --all --all-targets
+          args: --workspace --all-targets
```
Liamsi: I just learned `--workspace` is the new `--all` 👍

tendermint/src/rpc/endpoint/abci_info.rs
```diff
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
...
     pub data: String,
 
     /// Version
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
 
     /// Last block height
-    pub last_block_height: block::Height,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub last_block_height: Option<block::Height>,
```
Liamsi: This can only be empty on genesis, right? Maybe add a comment.

tendermint/src/rpc/endpoint/abci_info.rs
```diff
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
...
 
     /// Last app hash for the block
     #[serde(
         serialize_with = "serialize_app_hash",
-        deserialize_with = "parse_app_hash"
+        deserialize_with = "parse_app_hash",
+        skip_serializing_if = "Option::is_none"
     )]
-    pub last_block_app_hash: Hash,
+    pub last_block_app_hash: Option<Hash>,
```
Liamsi: Similar to above: This can only be empty on genesis, right? Maybe add a comment.

tendermint/src/rpc/endpoint/abci_info.rs
```diff
@@ -27,41 +27,58 @@ impl rpc::Response for Response {}
...
+impl Default for AbciInfo {
+    fn default() -> Self {
+        AbciInfo {
+            data: "".to_string(),
+            version: None,
+            last_block_height: None,
+            last_block_app_hash: None,
+        }
+    }
+}
```
Liamsi: Add a comment why `Default` is needed. (My best guess is because is that we want to be able to skip over fields but I might be wrong).

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: Yup, exactly.

.github/workflows/rust.yml
greg-szabo: Yes, I actually didn't find those tests. Also, these tests are not mandatory to merge something in GitHub.

CHANGES.md
greg-szabo: I thought those were universally optional in tendermint-go. I'll check.

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: So, `Hash::new` returns a `Result<Hash,Error>`. The original function directly returned that.
The function now has to return a `Result<Option<Hash>,Error>`. As `Hash:New` cannot return `None`, using the transitive property of `Some`, our function cannot return `None` either.

So we return `Some(Hash)` or `Error`, just as the original function returned `Hash` or `Error`. The reason to make the return into an `Option` is that the whole value in the structure can be omitted.

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: That's an interesting point. The normal execution in `serde` is that if the parameter is missing or None and you have the `Default` trait implemented, then this serializer function is not run. The `Default` implementation takes over and assumes whatever value is defined there.

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: Yes. Normally this is not needed but unfortunately the serialization implementation of `last_block_height` and `last_block_app_hash` is custom, hence they don't have any defaults for the case when they are missing from a deserialization.

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: Apparently that's how it's defined in Tendermint Go and v0.32 doesn't show it at all.

```
$ curl localhost:26657/abci_info
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "response": {
      "data": "{\"size\":0}",
      "version": "0.16.1",
      "app_version": "1"
    }
  }
}
```

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: Nope, abci_info doesn't print this over RPC.

tendermint/src/rpc/endpoint/abci_info.rs
Liamsi: OK, looks like these fields where removed:
image
image
https://docs.tendermint.com/master/rpc/#/ABCI/abci_info

Liamsi: Although the proto definition still contain these fields. Probably the best to keep it the way your dealt with it for now 👍

tendermint/src/rpc/endpoint/abci_info.rs
Liamsi: Yeah, it's probably worth capturing this reason in the code directly.

tendermint/src/rpc/endpoint/abci_info.rs
Liamsi: There is definitely something fishy about this 🤔 I think the method should err if hash is `None` which is possible. At least the type system allows that even if used as intended this won't happen.

greg-szabo: That's why we have the `skip_serializing_if = "Option::is_none"` added. If hash is `None`, `serde` will not call the function.

tendermint/src/rpc/endpoint/abci_query.rs
greg-szabo: It's the Tendermint Go code that's surprising in most of these cases. Some of their definitions don't seem to make sense to me, but if we want to be compatible with a full node, we have to follow them.

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: Yeah. We'll have a few surprises like that with 0.33

tendermint/src/rpc/endpoint/abci_info.rs
greg-szabo: Yup, added it.

Liamsi:
ref: #71

Issue #167, opened 2020-03-09
Light-client concurrency model


gterzian:
Place to discuss "how we want to run" the light-client. 

Currently it seems like we use a single thread, and then use a Tokio basic-scheduler to block on results for futures via 

https://github.com/interchainio/tendermint-rs/blob/5c3aff6ab1835f0e4db6215c9deaad6a12c10062/tendermint-lite/src/main.rs#L22

Discussion started at https://github.com/interchainio/tendermint-rs/pull/125#pullrequestreview-369307277

Liamsi:
ref #172, #171, #173

gterzian:
It looks like a single-threaded tokio runtime is a good idea for now? Then we can use a threaded runtime later if there is a need for parallellism somewhere, and spawn additional tasks, as opposed to using `await` in the "current task" everywhere?

Liamsi:
Sounds reasonable to me. There might be some need for parallellism in what requests headers and what verifies those. These _could_ be separate tasks.

brapse:
I have some suggestions coming by weeks end sorry for the delay! 🙏

tarcieri:
It seems like there's a lot of potential for parallelization. If you'd like some inspiration, check out how [Zebrad implements batch verification](https://www.zfnd.org/blog/futures-batch-verification/) for Ed25519 signatures.

Here's a plot of the performance improvements they saw:

![image](https://user-images.githubusercontent.com/797/76458793-532f2980-6398-11ea-8ad3-0318fd3d62e9.png)

ebuchman:
For the initial version of this ala https://github.com/informalsystems/tendermint-rs/milestone/1 we will do everything synchronously and revisit concurrency and batch verification down the road.

brapse:
Will close for now and create issues for specific component concurrency concerns.

PR #185, opened 2020-03-16
[ADR] Light Client Architecture

brapse:
Architecture refactor the the light client based on various discussions

[Rendered](https://github.com/informalsystems/tendermint-rs/blob/brapse/adr-light-client-concurrency.md/docs/architecture/adr-006-light-client-refactor.md)

Ref. #230, #167

* [ ] Wrote tests
* [ ] Updated CHANGES.md


romac:
Looks good! 👍 

As an aside, Dropbox recently published a post on their experience rewriting their sync engine in Rust, and I found the two points below quite interesting, as they seem to match the conclusions you've drawn yourself from your experiments:

> - Almost all of our code runs on a single thread (the “Control thread”) and uses Rust’s futures library for scheduling many concurrent actions on this single thread. We offload work to other threads only as needed: network IO to an event loop, computationally expensive work like hashing to a thread pool, and filesystem IO to a dedicated thread. This drastically reduces the scope and complexity developers must consider when adding new features.
>
> - The Control thread is designed to be entirely deterministic when its inputs and scheduling decisions are fixed. We use this property to fuzz it with pseudorandom simulation testing. With a seed for our random number generator, we can generate random initial filesystem state, schedules, and system perturbations and let the engine run to completion. Then, if we fail any of our sync correctness checks, we can always reproduce the bug from the original seed. We run millions of scenarios every day in our testing infrastructure.
>
> — https://dropbox.tech/infrastructure/rewriting-the-heart-of-our-sync-engine

docs/architecture/adr-006-light-client-concurrency.md
```diff
@@ -0,0 +1,106 @@
...
+![Concurrency Diagram](assets/light-client-concurrency.png)
+
+This architecture outlines components that have been 
+described in previous ADRs and will only be expanded upon here.
+
+### Peer Exchange
+The light clients requires an expanding set of peers to get headers
+from. The larger and more diverse the set of peers; the better the
+chances that the light client will detect a fork. The peer list can be
+seeded with peers from configuration and will can crawl the network.
```
zmanian: Here is a question. 

Should the light client be running the light client protocol only with 1 peer and only running some kind minimal fork detection protocol with other peers?

I don't think it makes any sense to be running the entire light client protocol on multiple peers.

brapse: By running the entire light client protocol on multiple peers do you mean fetching headers from multiple peers? I think the idea is to get the primary header from a single peer and then asynchronously fetch the same header from multiple peers to perform fork detection. If the headers differ, then the different header will also go through verification. If the header differs and also pass verification, we have a fork on the main chain. Is that what you mean?

zmanian: Yes this is what I'm thinking.

ebuchman:
Couple thoughts from review and discussions:

- `Requester` should be called `Scheduler` for consistency and because it schedules requests, but doesn't perform them itself
- `ValAndHeader` is really `HeaderValsCommit` (we request the whole package), or `HVC`
- the Scheduler is the component that actually performs bisection by scheduling the intermediate requests and by resubmitting previous HVC to the Verifier
- the Scheduler should have some kind of `Sync` event that triggers it to start syncing to a particular height using a given primary - this would be used to start syncing in the first place and to sync to a conflicting header if one is detected in order to verify it (this should reuse the Verifier)
- the Verifier can/should be stateless 
- Verifier should have two kinds of output errors:
  - `ErrData` - where there is some problem validating or verifying the data, indicating a bad peer 
  - `ErrTrust` - where there's not enough trust in the new validator set to skip to it and we need to do bisection
- Detector should likely be a minimal component that simply tell the scheduler to request headers, and checks if the headers match the last trusted one. Note ti doesn't need  HVC, just Headers. So the Scheduler should also have `HeaderReq/HeaderResponse` events along with the `HVCRequest/HVCResponse` so we don't request the extra validator and commit data if we don't need it. If a conflict is detected, the detector tells the scheduler to start bisection for that header with the given secondary peer to verify (via the Verifier) if the conflict is real or its just a bad peer. 
- Given the detectors minimalism, how to justify its independence from the Scheduler? It may not need to be an independent component, but rather just (potentially parameterizable) part of the logic of the Scheduler.

We also discussed mocking functions instead of data - so instead of traits flowing into the Verifier, we can use events that have all the fields we expect plus some closures for any data-structure specific computation we don't want to or can't do upfront (eg voting_power_in(), which verifies signatures ). 

A larger question I have is about how to better visualize the flow of events here, for instance how to handle cases where a component should output multiple events, or where the same event should go to multiple components.

docs/architecture/adr-006-light-client-concurrency.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Requester
+The light client requests validator sets and signed headers from peers.
+The order of data requested from peers is determined by the specific verification
+algorithm being followed. Specifically either sequential or bisection.
+Additionally the Requester is expected to request signed header for a
```
melekes, line 55: _This may lead to mixed concerns between Requester and Detector modules. If you are trying to archive “better isolation of concerns”, it could be beneficial to leave the logic of picking what nodes to ask for headers to Detector module (also, when to ask for headers). This does not necessary mean Detector module would have to store a list of peers internally. It can use some kind of shortcuts (e.g. “all” - all backup nodes, “random” - random backup node, “randomN{N}” - random N backup nodes)_

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,100 @@
...
+independently verifiable components. Communication between components is
+done by a demuxer component which ownes state and dispatches events
+between business logic executing components. Business logic is executed
+as a synchronous deterministic function. The composition of components
+can also be executed as a deterministic function.
+
+![Decomposition Diagram](assets/light-client-decomposition.png)
+
+### State
+The light client maintains state about commit/header/validators abbreviated as LightBlocks. It also has state pertaining to peers and evidence.
```
josef-widder, line 42: LightBlocks will need to maintain ValidatorSet and NextValidatorSet.

ancazamfir:
If I understand correctly there is a single thread of execution, so all blocks shown are functional, correct?
A few comments, high level:
- Looking at Verification Flow diagram, I think it would be easier to read if it was showing each component once. 
- not clear if both Demux and Scheduler are needed. We drew the same conclusion on the blockchain refactor so was wondering. 
- I tried to go through the flow and it is not clear to me what happens if `Verifier` fails with validator set not trusted, in which case more headers should be requested by the Scheduler (?) and verified, vs bad header and/ or commits where the flow is different.
- Why is the header+commit request called "VerificationRequest"? I feel we call too many things "verify".
- In previous implementation, for optimization reasons, there were basically two verification steps:
1. check if validator set of new header should be trusted (requires trusted header and its validator set)
2. fully verify the new header and commits 
Intermediate headers were downloaded while 1 was failing. Then, 2) was done on each header in the sequence of headers. This seems to have been lost here but not sure as the flow is not clear. 

I have many more questions :) is it possible to have a review/ working session for this?

ancazamfir:
Also, why is there a "Relayer" there?

brapse:
Thanks for your thoughts @ancazamfir , I'll try to answer inline ☺️ 
> If I understand correctly there is a single thread of execution, so all blocks shown are functional, correct?

Yes precisely; each flow is a synchronous method call and each delegation to a subcomponent is also a synchronous method call.

> * Looking at Verification Flow diagram, I think it would be easier to read if it was showing each component once.

Do you mean the repeated references to the demuxer in each diagram? Yeah perhaps we can remove those and just emphasise the state changes.

> * not clear if both Demux and Scheduler are needed. We drew the same conclusion on the blockchain refactor so was wondering.

I think in FastSync v2 we concluded that the scheduler didn't need to run in it's own go-routine. We've integrated that learning here and put everything in the same execution thread. What the scheduler gives us here is a component which encapsulates "next request" concerns while the demuxer encapsulate "delegation to multiple component concerns". That's the thinking at least.

> * I tried to go through the flow and it is not clear to me what happens if `Verifier` fails with validator set not trusted, in which case more headers should be requested by the Scheduler (?) and verified, vs bad header and/ or commits where the flow is different.

We spoke about this yesterday with @josef-widder and our thinking has continued to evolve. What we want to do now is to augment the TrustedStore to contain light blocks it's received, verified and failed to verify. The TrustedStore will also contain a mapping of which peer sent which light block in which state. This way, the scheduler can make decision on which light block to request next based on the highest trusted and untrusted light block it has seen.

> * Why is the header+commit request called "VerificationRequest"? I feel we call too many things "verify".

Yes it probably makes sense to just call this LightBlockRequest 🤦 

> * In previous implementation, for optimization reasons, there were basically two verification steps:

Yes that detail is captured in the verifier but not presented in these diagrams.

> 
> I have many more questions :) is it possible to have a review/ working session for this?

Yes for sure. We are trying to get something working by the end of this week but then we can look forward to doing a full review as early as next week.

docs/architecture/adr-006-light-client-refactor.md
melekes: Out of curiosity: is there something wrong with using ValidatorSet (not NextValidatorSet)? _(other than validator set might have changed between ValidatorSet and NextValidatorSet)_

melekes:
Thanks for the update 👍 A few thoughts from me:

1. [State] `lightBlocks[peerID][height]`. Most of the blocks will be the same and we end up storing a lot of duplicates, which is bad if you care about memory consumption. If you don’t, then I guess it’s fine.
2. [State] `detectedHeight` -> `forkDetectedHeight`
    1. Is this a height of the first detected fork?
3. [State] `evidence[peerID][]`. Evidence usually involves 2 peers: primary and one of the secondary nodes.
4. [State] It’s not a `TrustedStore` if you store untrusted light blocks in it. We should probably make a distinction between which blocks are trusted and which aren’t. Also, we probably should not trust a `LightBlock` until fork detection is finished (for that height).
5. [Detection Flow] “update last_detected” I don’t see that field in `TrustedStore`

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+
+    let relayer = Relayer::new(light_client);
+    ...
+
+    // Run (consume master)
+    light_client.run();
+}
+```
+
+For lack of better naming we call this abstraction "Handler". The
```
romac, line 87:
```suggestion
For lack of better naming we call this abstraction "Handle". The
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+Supervisor outlined above is executed by running in it's own thread. But
+before doing so, it spawns a `Hander`. The Handler is clonable and can
```
romac, line 89:
```suggestion
before doing so, it spawns a `Handle`. The Handle is clonable and can
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+ That interface should be free of
+serialization concerns for the user. Under the hood, that facade should
+be facilitating interactions boundaries allowing components to operate
+freely and safely in distinct threads or runtime.
+
+```rust
+fn main() {
+    let mut light_client = Supervisor::new();
+
+    // Get a handler specifically for the relayer
```
romac, line 76:
```suggestion
    // Get a handle specifically for the relayer
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+    let relayer_light_client = light_client.handler();
```
romac, line 77:
```suggestion
    let relayer_light_client = light_client.handle();
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+
+For lack of better naming we call this abstraction "Handler". The
+Supervisor outlined above is executed by running in it's own thread. But
+before doing so, it spawns a `Hander`. The Handler is clonable and can
+be given to any component who wants safe, serialized access to the
+Supervisor.
+
+The implementation is a simple abstraction which facilitates
+communication over channels. Events send across a channel include a
+callback. This callback abstraction allows the outer handler to expose a
```
romac, line 95:
```suggestion
callback. This callback abstraction allows the outer handle to expose a
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+synchronous operation event if the delegation is inherently
+asynchronous. Delegating at this level opens to door to exposing
+synchronous interaction to thread pools and allowing components to
+handle their concurrency optimization internally without burdening the
+component user.
+
+```rust
+pub struct Handler {
```
romac, line 103:
```suggestion
pub struct Handle {
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+    sender: channel::Sender<Event>,
+}
+...
+impl Handler {
```
romac, line 107:
```suggestion
impl Handle {
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+// Supervisor
+pub struct Supervisor {
+    ...
+    sender: channel::Sender<Event>,
+    receiver: channel::Receiver<Event>,
+}
+
+impl Supervisor {
+    ...
+    pub fn handler(&mut self) -> Handler {
```
romac, line 149:
```suggestion
    pub fn handle(&mut self) -> Handle {
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+        let sender = self.sender.clone();
+
+        return Handler::new(sender);
```
romac, line 152:
```suggestion
        return Handle::new(sender);
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
+# ADR 006: Light Client Refactor
+
+## Changelog
+- 2020-03-15: Initial draft
+- 2020-03-23: New decomposition
+- 2020-05-18: Concurrency
+- 2020-06-26: Handler Abstraction
```
romac, line 7:
```suggestion
- 2020-06-26: Handle Abstraction
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+multiple Peers available as Secondaries.
+
+Each lightClient Instance will be configured to perform the light client
+protocol against a single peer. The Instance will be decomposed into
+IO independently verifiable components for:
+
+    * IO: Fetching LightBlocks from the peer
+    * Verifier: Verifying a LightBlock based on a TrustedState
+    * Scheduler: Schedule the next block based on the last TrustedState
+
```
romac, line 47:
```suggestion
    * Clock: Provide the current time

```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,206 @@
...
+Each component will be represented as a trait allowing mock replacements
```
romac, line 48:
```suggestion
In addition, cryptographic operations are also decomposed into three "sub-components":

    * Hasher: For hashing headers
    * CommitValidator: For validating commit integrity
    * VotingPowerCalculator: For tallying the voting power of a validator set in a commit and verifying signatures

Each component and operation will be represented as a trait allowing mock replacements
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+
+Each LightClient Instance will be configured to perform the light client
+protocol against a single peer. The Instance will be decomposed into
+independently verifiable components for:
+
+    * IO: Fetching LightBlocks from the peer
+    * Verifier: Verifying a LightBlock based on a TrustedState
+    * Scheduler: Schedule the next block based on the last TrustedState
+    * Clock: Provide the current time
+    * Clock: Provide the current time
```
romac, line 48:
```suggestion
```

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+```
+## Status
+
+Implemented:
+    * [Verification Predicates](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/predicates.rs)
+    * [Verifier Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/verifier.rs)
+    * [IO Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/io.rs)
+    * [Scheduler Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/scheduler.rs)
+    * [Supervisor](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/src/supervisor.rs)
+    * [Example](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/examples/light_client.rs)
```
xla, line 195:
```suggestion

* [Verification Predicates](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/predicates.rs)
* [Verifier Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/verifier.rs)
* [IO Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/io.rs)
* [Scheduler Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/scheduler.rs)
    * [Supervisor](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/src/supervisor.rs)
    * [Example](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/examples/light_client.rs)
```

This will fix the render of the markdown.

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+fn main() {
+    let mut light_client = Supervisor::new();
+
+    // Get a handle specifically for the relayer
+    let relayer_light_client = light_client.handle();
+
+    let relayer = Relayer::new(light_client);
+    ...
+
+    // Run (consume master)
```
xla, line 90:
```suggestion
    // Run (consume Supervisor)
```

AFAIK that is what is happening by starting the run loop.

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+Supervisor outlined above is executed by running in it's own thread. But
+before doing so, it spawns a `Handle`. The Handle is clonable and can
+be given to any component who wants safe, serialized access to the
+Supervisor.
+
+The implementation is a simple abstraction which facilitates
+communication over channels. Events send across a channel include a
+callback. This callback abstraction allows the outer handle to expose a
+synchronous operation event if the delegation is inherently
+asynchronous. Delegating at this level opens to door to exposing
```
xla, line 105:
```suggestion
asynchronous. Delegating at this level opens the door to exposing
```

romac:
@xla That's a good point! Definitely something we'll need to revisit when we'll want to make things truly concurrent.

docs/architecture/adr-006-light-client-refactor.md
```diff
@@ -0,0 +1,215 @@
...
+## Status
+
+Implemented:
+
+* [Verification Predicates](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/predicates.rs)
+* [Verifier Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/verifier.rs)
+* [IO Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/io.rs)
+* [Scheduler Component](https://github.com/informalsystems/tendermint-rs/blob/e2335c4/light-client/src/components/scheduler.rs)
+    * [Supervisor](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/src/supervisor.rs)
+    * [Example](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/examples/light_client.rs)
```
romac, line 196:
```suggestion
* [Supervisor](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/src/supervisor.rs)
* [Example](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/light-client/examples/light_client.rs)
```

brapse:
@xla I think your right and would look forward to such a simplification but should those changes block the current PR considering it's already implemented. Might be good to get this landed to align some level of documentation with the implementation. What do you think?

xla:
> @xla I think your right and would look forward to such a simplification but should those changes block the current PR considering it's already implemented. Might be good to get this landed to align some level of documentation with the implementation. What do you think?

Yeah for sure! Wasn't sure if this ADR is still worked towards, but the implementation is there. Is there a follow-up issue already? Think a change to the internal Supervisor API could be in a separate ADR accompanied by a change-set implementing it.

brapse:
Went ahead and created a follow up issue #398 , feel free to edit the description to align with your original intentions @xla 🙏

Issue #190, opened 2020-03-23
Require that light client datatypes are serializable


romac:
I would like to add `Serialize` and `Deserialize` bound on the traits defined in the `tendermint::lite::types` module to help get rid of [some very ugly bounds](https://github.com/informalsystems/ibc-rs/blob/96ee56d7b1c731c982f8085b9441e62cce974c00/relayer/relay/src/store.rs#L64) in `ibc-rs`.

Those are required by the current persistent store implementation, which needs to be able to (de)serialize these datatypes with a binary encoding in order to store them in the underlying [`sled`](https://sled.rs) database.

I do not see any reason why datatypes implementing those traits would not be serializable, but if you do please correct me :)

We could work around the need for those trait bounds by extending them with our own traits in the `ibc-rs` crate, but that would only introduce more traits whereas I [feel](https://github.com/informalsystems/tendermint-rs/issues/174#issuecomment-600169273) we should perhaps go the other way.

Liamsi:
That makes absolute sense IMHO. Note that some of the mentioned types which aren't traits but structs currently derive (De)Serialize from serde: e.g. `SignedHeader` and `TrustedState`.

PR #191, opened 2020-03-23
Require that light client datatypes be serializable

romac:
Closes: #190

* [ ] Updated all relevant documentation in docs (N/A)
* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/lite/types.rs
+++ b/tendermint/src/lite/types.rs
@@ -5,7 +5,7 @@ use std::fmt::Debug;
 use std::time::SystemTime;
 
 use async_trait::async_trait;
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use crate::lite::error::{Error, Kind};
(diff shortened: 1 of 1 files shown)


tendermint/src/lite/types.rs
```diff
@@ -180,12 +185,8 @@ where
 }
 
 /// SignedHeader bundles a [`Header`] and a [`Commit`] for convenience.
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)] // NOTE: Copy/Clone/Debug for convenience in testing ...
-pub struct SignedHeader<C, H>
-where
```
romac: I removed the bounds here to avoid having to use the `#[serde(bound = ...)]` trick, but also because it is [considered a best practice](https://github.com/rust-lang/rust-clippy/issues/1689) to not add bounds on structs when one can avoid it.

romac:
@Liamsi Thanks for merging this so promptly :)

PR #195, opened 2020-03-28
make tests sync instead of async where no async function is used

Liamsi:
#171 introduced async tests in the verifier. It seems that also tests that don't call async functions were marked as async (which doesn't harm but makes the tests look as if they were calling async functions). This small PR reverts this.

ref https://github.com/informalsystems/tendermint-rs/pull/171#discussion_r399712319

* [ ] Wrote tests
* [ ] Updated CHANGES.md


romac:
Ah good catch! My bad, sorry, I was a bit too enthusiastic with the async annotation in my previous PR.

PR #188, opened 2020-03-19
Add matching serializer for `app_hash` field of `block::Header` struct

romac:
Closes: #187

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs (N/A)
* [ ] Updated all code comments where relevant (N/A)
* [x] Wrote tests
* [ ] Updated CHANGES.md (N/A)

Issue #101, opened 2019-12-20
Exceptions for the "first block"


ebuchman:
This first block in Tendermint is unique in that all of the `last_xxx` fields are actually empty, while in normal operation they are required to not be empty. On the other hand, some fields can be empty in normal operation (eg. app_hash, data_hash, evidence_hash). Recent PRs like #66 have been dealing with this by making all fields that can be empty Options, but it may be clearer if there is actually an explicit FirstBlock type that doesn't have the empty fields and and we reserve Option only for the fields that can be empty in normal operation. See the discussion in #66.

Another possibility is to clean this up by making a chance in the underlying tendermint protocol, see https://github.com/tendermint/tendermint/issues/4241

tarcieri:
Sorry for waffling, but given the set of tradeoffs here, I'm now wondering if `Hash::Null` might've actually been a good idea. I felt a little dirty introducing it in the first place, but after looking at the `Option`-based alternatives empirically, it seems like it might actually be a fairly good fit for the protocol as it exists today.

ebuchman:
Yes, except we might also need something for BlockID and Commit, which aren't already enums ...

ebuchman:
@tarcieri @yihuang @Liamsi what do we think about making block an enum with a FirstBlock variant that just doesn't have all the fields that would be empty? Would that encode properly? Of course it means every time we handle a block we have to account for both variants, but that might actually be net positive since it's already the case that we'll have to special logic the first block, so this would just reflect it in the type system.

Then we'd only have to consider a few remaining fields that can be empty in normal operation, and for those we either use Option<Hash> like we do now or bring back the Hash::Null variant. @yihuang mentioned the downside of Hash::Null being that we have to consider the null variant every time we use Hash, but I'm not sure I understand, since we're mostly just checking for equality or trying to get the bytes, and we can implement that functionality once for both variants and then never have to worry about them again, unless I'm missing something ...

We might also be able to make Hash::Null go away if we can get Tendermint to use the hash of empty values instead of empty hashes for instance when there's no txs or evidence. Then we'd always have full hashes (except the first block).

ebuchman:
I'm leaning in favour of being explicit with a FirstBlock variant but not using Hash::Null. 

[...]

Issue #88, opened 2019-12-11
rpc: incompatibilities


Liamsi:
I've tested the (ignored) integration tests for rpc against a tendermint 0.32.7 version and the following tests fail:

 `abci_info` and `genesis` 
- genesis fails with: called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("invalid digit found in string at line 11 column 26") }
 - abci_info with: called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("missing field `last_block_height` at line 9 column 5") }

https://github.com/interchainio/tendermint-rs/pull/85#issuecomment-564584641

I've also noted that example responses on https://docs.tendermint.com/master/rpc/#/ have a numeric `id` (every JSON rpc request comes with such an id):
```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": {
````
while on my local `0.32.7` returns a string based id ` "id": "",`. 

We should make sure to support both as the spec says: 
> An identifier established by the Client that MUST contain a String, Number, or NULL value if included. 

https://www.jsonrpc.org/specification#request_object

Compatibility with some tendermint version is also relevant in the context of a light client making requests via rpc.

tarcieri:
@Liamsi can we figure out a way to run the currently `#[ignore]`d integration tests in CI? I think that'd be the ideal way to spot incompatibilities

Liamsi:
Yes, that would be ideal! A naive approach that should work: grab a docker image (with the latest / a pinned tendermint version), pull in tendermint-rs source, spin up a single tendermint node and run the ignored tests against this node (via `cargo test -- --ignored`). Wasn't there a script that done exactly this in the past?

It makes sense to sync with @greg-szabo and @thanethomson on this too. Not likely that I'll work on this this year but I I'll open a PR with a fix for the regressions.

Liamsi:
https://github.com/interchainio/tendermint-rs/pull/169 surfaced new incompatibilities with the latest tendermint version.

Liamsi:
As mentioned above the spec defines:

> An identifier established by the Client that MUST contain a String, Number, or NULL value if included.

https://www.jsonrpc.org/specification#request_object

[...]

PR #196, opened 2020-03-29
Tendermint/v0.33

greg-szabo:
This is a minimum viable product for Tendermint v0.33 compatibility. Lots of improvements can be made (more tests, further optimizations) but most probably those should be added as new issues and PRs.

I'm writing an ADR that should address the decisions made in updating/implementing the RPC structs.

* [X] Wrote tests
* [X] Updated CHANGES.md


tendermint/src/lite_impl/signed_header.rs
```diff
@@ -16,15 +16,11 @@ impl lite::Commit for block::signed_header::SignedHeader {
         // NOTE we don't know the validators that committed this block,
         // so we have to check for each vote if its validator is already known.
         let mut signed_power = 0u64;
-        for vote_opt in &self.iter() {
+        for vote in &self.iter().unwrap() {
```
Liamsi: Can't this panic? Why are there no `None` in the `Option<vote::SignedVote>`s returned by the iter? Ah OK, I see the method signature changed to ` fn iter(&self) -> Result<Vec<vote::SignedVote>, Error>`. Why is the error not handled?

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,68 +48,89 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+
 impl block::signed_header::SignedHeader {
     /// This is a private helper method to iterate over the underlying
     /// votes to compute the voting power (see `voting_power_in` below).
-    fn iter(&self) -> Vec<Option<vote::SignedVote>> {
+    fn iter(&self) -> Result<Vec<vote::SignedVote>, Error> {
         let chain_id = self.header.chain_id.to_string();
-        let mut votes = self.commit.precommits.clone().into_vec();
-        votes
+        // if let Ok(mut votes) = commit_to_votes(self.commit.clone())
```
Liamsi: ?

tendermint/tests/rpc.rs
```diff
@@ -47,29 +47,28 @@ mod endpoints {
...
+    //
+    //     let tendermint::Block { last_commit, .. } = response.block;
+    //
+    //     assert_eq!(last_commit.as_ref().unwrap().precommits.len(), 2);
+    //     assert!(last_commit.unwrap().precommits[0]
+    //         .as_ref()
+    //         .unwrap()
+    //         .block_id
+    //         .is_none());
+    // }
```
Liamsi: Will this be added back in this PR or later?

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,68 +48,89 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+                        validator_index: i as u64,
+                        signature: sig,
+                    };
+                    votes.push(vote);
+                }
+            }
+            None => {
+                fail!(
+                    Kind::ImplementationSpecific,
+                    "validator address missing in commit_sig {:#?}",
```
Liamsi: I'm not familiar with the `#?` formatting directive. Couldn't find anything about it: https://doc.rust-lang.org/std/fmt/#formatting-traits
Did the `#` sneak in accidentally?

tendermint/src/serializers.rs
```diff
@@ -185,3 +188,34 @@ where
...
+        // TODO: how can we avoid rewriting code here?
+        match Id::from_str(&s).map_err(|_| {
+            D::Error::custom(format!(
+                "expected {}-character hex string, got {:?}",
+                LENGTH * 2,
+                s
+            ))
+        }) {
+            Ok(id) => Ok(Option::from(id)),
+            Err(_) => Ok(None),
```
Liamsi: In case there is an error above, we just return `Ok(None)`. That doesn't seem right unless we want to ignore it. But then the return could be changed from `Result<Option<Id>, D::Error>` to `Option<Id>`.

tendermint/src/rpc/id.rs
```diff
@@ -3,9 +3,17 @@
 use getrandom::getrandom;
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, Ord, PartialOrd)]
+#[serde(untagged)]
+pub enum IdType {
+    Num(i64),
+    Str(String),
+    None,
```
Liamsi: Cool! That will make the jsonrpc id field spec compliant (see https://github.com/informalsystems/tendermint-rs/issues/88#issuecomment-599227891). I'd vouch to implement this in a separate PR if possible. Maybe testing it in a similar fashion as @romac did in #188 (see https://github.com/informalsystems/tendermint-rs/pull/188/files#diff-a1644bb72a1c83c54d82738ef71ce095R107).

tendermint/src/rpc/id.rs
```diff
@@ -3,9 +3,17 @@
...
+}
+
 /// JSONRPC ID: request-specific identifier
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-pub struct Id(String);
+pub struct Id(IdType);
```
Liamsi: Instead of the wrapper, couldn't this directly be `pub enum Id`? My understanding is that serde can handle these (didn't try though).

tendermint/src/rpc/id.rs
```diff
@@ -3,9 +3,17 @@
 use getrandom::getrandom;
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, Ord, PartialOrd)]
+#[serde(untagged)]
+pub enum IdType {
```
Liamsi: Does this internal type need to be public? Also, I'm surprised the linter doesn't complain about missing docs here.

tendermint/src/lite_impl/signed_header.rs
greg-szabo: Based on the comment right under it, I guess it can. Maybe iter() should be iterated and unwrapped with check in the loop.

Something like
```
for vote in $self.iter() {
  if vote.is_none() {
    continue
  }
  unwrapped_vote = vote.unwrap()
..
``` ?

File: tendermint/src/lite_impl/signed_header.rs (hunk 2 above)
Comment (greg-szabo) [reply tendermint/src/lite_impl/signed_header.rs]: Seems to me that it was just refactored and maybe that line can go. Shivani?

File: tendermint/tests/rpc.rs (hunk 3 above)
Comment (greg-szabo) [reply tendermint/tests/rpc.rs]: Nice find, I missed that. I'll let Shivani decide. She's working on more tests, that's why I didn't wait with opening this PR. I expect more tests will come either into this PR (bloating it even more) or in new PRs after this has been merged (which seems to cut our work better for us).

File: tendermint/src/serializers.rs (hunk 5 above)
Comment (greg-szabo) [reply tendermint/src/serializers.rs]: I'm going to try to find a way to get away from this custom serialization completely. The whole problem of `Option` is here because we again assume that the value of CommitSig.validator_address can be `null`.

File: tendermint/src/rpc/id.rs (hunk 6 above)
Comment (greg-szabo) [reply tendermint/src/rpc/id.rs]: This is also a requirement for v0.33 compatibility, as it turns out. We agreed before that I'll implement this separately but then all of Shivani's JSON test files need to be updated for compatibility and I wasn't planning to do that in this PR. Tests are still moving around a lot and making all of them comply strictly with a broken implementation just to be able to push this PR through didn't make sense.

The change is in it's own commit, we can cherry-pick it into its own branch and dissect it more. But then this PR will be dependent on that one - OR an additional 10+ files need to be changed in this one to get away from that dependency.

File: tendermint/src/rpc/id.rs (hunk 7 above)
Comment (greg-szabo) [reply tendermint/src/rpc/id.rs]: Maybe. That's a good idea. I'll try it.

File: tendermint/src/lite_impl/signed_header.rs (hunk 1 above)
Comment (Liamsi) [reply tendermint/src/lite_impl/signed_header.rs]: We shouldn't use `unwrap` here but handle the error instead (this could mean to just bubble it up using the `?` operator via `&self.iter()?`; depends what we want to do with it but this would at least not panic)

Comment (Liamsi) [reply tendermint/src/lite_impl/signed_header.rs]: Looks like this error can only occur if there is a `CommitSig` with a missing (`None`), `validator_address`. If the address is in fact not optional, we could get rid of the error altogether.

File: tendermint/src/lite_impl/signed_header.rs [hunk 9]
```diff
@@ -52,68 +48,89 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+            Some(val_addr) => {
+                if let Some(sig) = commit_sig.signature.clone() {
+                    let vote = vote::Vote {
+                        vote_type: vote::Type::Precommit,
+                        height: commit.height,
+                        round: commit.round,
+                        block_id: Option::from(commit.block_id.clone()),
+                        timestamp: commit_sig.timestamp,
+                        validator_address: val_addr,
+                        validator_index: i as u64,
```
Liamsi: Use `TryFrom` and `expect` instead.

tendermint/src/serializers.rs
Liamsi: This is related then: https://github.com/informalsystems/tendermint-rs/pull/196/files#r399823439

tendermint/src/rpc/id.rs
greg-szabo: Yup, you're right, it works. I've submitted it into the PR.

tendermint/src/rpc/id.rs
greg-szabo: This is now obsolete, because IdType was removed and Id was made into an enum.

tendermint/src/block/commit_sig.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct CommitSig {
+    /// Block ID FLag
+    pub block_id_flag: BlockIDFlag,
+
+    /// Validator address
+    #[serde(deserialize_with = "serializers::parse_non_empty_id")]
+    pub validator_address: Option<account::Id>,
+
+    /// Timestamp
+    pub timestamp: Time,
```
ebuchman: I guess this would need an Option too ? I believe if we have BlockIDFlagAbsent, then all these fields are empty https://github.com/tendermint/tendermint/blob/master/types/block.go#L476. Otherwise, they should all be populated. Or do we not need it here because there's a sane default?


In general I think there are a number of cases where a set of fields are either all present or all absent, but if they each individually are encoded as Option, this information isn't clear, since it looks like they can all independently be present or absent. In these cases, maybe it's better to use enums with Present and Absent variants to better indicate this ... related https://github.com/informalsystems/tendermint-rs/issues/101 - but that's for a follow up PR

tendermint/src/evidence.rs
```diff
@@ -92,5 +92,37 @@ pub struct Params {
...
-    pub max_age: u64,
+    pub max_age_num_blocks: u64,
+
+    /// Max age duration
+    pub max_age_duration: Duration,
+}
+
+/// Duration
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Duration(u64);
```
ebuchman: Can we add some comments for why we need this (eg. compared to max_age_num_blocks, which is also just a u64)

tendermint/src/lite_impl/signed_header.rs
ebuchman: The address is optional if the vote is Absent, ie. we have BlockIDFlagAbsent

Shivani912: @Liamsi You're right, the error occurs only when there is a missing `validator_address` with a non-absent block ID flag i.e. not `BlockIDFlagAbsent`. Should we simply `continue` if that happens or return an error from within `voting_power_in` ?

tendermint/src/lite_impl/signed_header.rs
Shivani912: @Liamsi Is there a reason why we shouldn't use `as`?

tendermint/src/lite_impl/signed_header.rs
Shivani912: Oh yeah, I had kept that for a reference and then forgot to delete it. Thanks!

tendermint/tests/rpc.rs
Shivani912: Yes @greg-szabo, I'll let these test-related things come in a separate PR.

tendermint/src/evidence.rs
Shivani912: @ebuchman I've addressed this [here](https://github.com/informalsystems/tendermint-rs/pull/196/commits/f0048c58f3764a3a29beec4cba0c81b43cd14f84#diff-3b2ccc4b9e2bf7cc4225b022246b6642R101)

tendermint/src/lite_impl/signed_header.rs
Liamsi: TryFrom would checks for explicit bounds while `as` does the conversion if one types fits into the other.

tendermint/src/rpc/id.rs
Liamsi: Nice one :-)

tendermint/src/rpc/id.rs
Liamsi: > This is also a requirement for v0.33 compatibility, as it turns out.

That this can deal with all 3 cases (number, string, null) now? OK.

I would still have been in favour of implementing this in a separate PR. Maybe even against this branch (not against master). Just because it is a clear small chunk of work that can be reviewed on it's own (now it's already in two commits). Next time :-)

tendermint/src/rpc/id.rs
```diff
@@ -18,12 +26,16 @@ impl Id {
...
-        Id(uuid.to_string())
+        Id::Str(uuid.to_string())
     }
 }
 
 impl AsRef<str> for Id {
     fn as_ref(&self) -> &str {
-        self.0.as_ref()
+        match self {
+            Id::Num(_) => "",
```
Liamsi: So if this is a number `as_ref` is supposed to return an empty string? Is this used somewhere? Maybe add a TODO here (or capture this in an issue, e.g #88 or a dedicated one).

tendermint/src/lite_impl/signed_header.rs
Liamsi: Didn't think this through but sounds like an error to me. But I'd double check with the spec and maybe how the current go code deals with it. Can think about it more tmrw.

tendermint/src/rpc/id.rs
```diff
@@ -18,12 +26,16 @@ impl Id {
             .set_version(uuid::Version::Random)
             .build();
 
-        Id(uuid.to_string())
+        Id::Str(uuid.to_string())
     }
 }
 
 impl AsRef<str> for Id {
```
Liamsi: Currently, this isn't used anywhere 🤔

tendermint/src/lite_impl/signed_header.rs
Liamsi: I think if this is not `BlockIDFlagAbsent` it is an `error` if the `validator_address` is missing (and we should not simply continue).

Liamsi: Thinking about this further, maybe we should rename that method (iter) to be more explicit about what it does (it does some validation, extracts and returns valid SignedVotes from the SignedHeader, or, returns an error). Keeping the current name and flow is quite unidiomatic: this method doesn't implement an iterator (anymore).

Shivani912: It's very unlike that we receive a commit with a missing `validator_address` and yeah i guess this can be an `ImplementationSpecific` error? 

Hmm, the method converts `CommitSigs` to `SignedVotes` with some validation done there  and returns a `Vec<SignedVote>` (when Result is Ok). Should we call it something like `signed_votes_iter`?

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,68 +48,89 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+        for commit_sig in self.commit.signatures.iter() {
+            // returns FaultyFullNode error if it detects a signer that is not present in the validator set
+            if let Some(val_addr) = commit_sig.validator_address {
+                if vals.validator(val_addr) == None {
+                    let reason = format!(
+                        "Found a faulty signer ({}) not present in the validator set ({})",
+                        val_addr,
+                        vals.hash()
+                    );
+                    fail!(Kind::FaultyFullNode, reason);
```
Liamsi: The var `reason` can be deleted and this can be simplified to:
```suggestion
                    fail!(
                        Kind::FaultyFullNode,
                        "Found a faulty signer ({}) not present in the validator set ({})",
                        val_addr,
                        vals.hash()
                    );
```

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,68 +48,89 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
-                    }
+        for commit_sig in self.commit.signatures.iter() {
+            // returns FaultyFullNode error if it detects a signer that is not present in the validator set
+            if let Some(val_addr) = commit_sig.validator_address {
+                if vals.validator(val_addr) == None {
+                    let reason = format!(
+                        "Found a faulty signer ({}) not present in the validator set ({})",
+                        val_addr,
+                        vals.hash()
+                    );
```
Liamsi:
```suggestion
```

tendermint/src/lite_impl/signed_header.rs
Liamsi: What would e best: there is already a `validate` method that iterates over the commit's signatures. We can extend that loop to check for one sig with `!commit_sig.is_absent()` there is a `None` `commit_sig.validator_address` (the same validation that currently happens in `commit_to_votes` / `iter`).

We can rename `iter` to `signed_votes`.  We can collapse `commit_to_votes` and make it return `Vec<vote::Vote>` and not a result.

Liamsi: And yes, the error would be `ImplementationSepcific`

Liamsi: I'd like to refactor this in a PR against this branch if that's OK for everyone.

tendermint/src/rpc/id.rs
greg-szabo: I can still break it out into its own thing. :)

Liamsi: I'll let you decide. Looks like the id isn't used anywhere currently. At this stage maybe capturing what needs to be followed up with in an issue is less work then reverting, cherrypicking and opening a separate PR for it.

tendermint/src/rpc/id.rs
greg-szabo: I believe so and I'd like to get rid of it. AsRef-ing a string (previous iteration) made sense, AsRef-ing this enum doesn't really.
I'm going to delete it and see if tests still succeed.

tendermint/src/rpc/id.rs
greg-szabo: I think, this AsRef doesn't make much sense with the `enum`. AsRef trait has some interesting requirements: [doc](https://doc.rust-lang.org/std/convert/trait.AsRef.html)
- This trait must not fail. So, I need to return something.
- Used to do a cheap reference-to-reference conversion. So, I shouldn't use new memory.

This means, I can't do much with a number. The reutrn type is `str`, which means it would make sense to convert the number into its string representation but then it shouldn't be done in AsRef because it requires new memory. The documentation suggests: If you need to do a costly conversion it is better to implement From with type &T or write a custom function.

So, we should just get rid of it.

greg-szabo: (actually, maybe using a generic helps...)

tendermint/src/rpc/id.rs
greg-szabo: Yeah, we didn't use it. Tests and build works.

tendermint/src/rpc/id.rs
greg-szabo: Got rid of it.

tendermint/src/block/commit_sig.rs
Liamsi: Looks like the go-code sends a zero time. Meaning sec and ms are 0, or more accurately: `0001-01-01T00:00:00Z` is the JSON we'll see for a `BlockIDFlagAbsent` commit:
```go
csa := NewCommitSigAbsent()
b, _ := cdc.MarshalJSON(&csa)
fmt.Println(string(b))
```
yields:  `{"block_id_flag":1,"validator_address":"","timestamp":"0001-01-01T00:00:00Z","signature":null}`

tendermint/src/rpc/id.rs
greg-szabo: Submitted PR #202 against master that makes this happen separately, as you asked. We'll have to make sure to resolve the CHANGES.md conflict here, but that's about it. (The rest were cherry-picked.)

tendermint/src/lite_impl/signed_header.rs
greg-szabo: Here's the PR for that: #199 (https://github.com/informalsystems/tendermint-rs/pull/199)

tendermint/src/serializers.rs
greg-szabo: I've been trying to figure out better ways of handling serialization - especially with the input from Ethan et al. I think we should wrap up this PR as soon as possible without adding even more features and break out any finds like this into their own issue. Some of the solutions I found would require the refactoring of the serializers library, which will lead to a cleaner code, but it would add more complexity on top of this PR.

Also, getting away from custom serialization is a tall order, but we can put some of the complexity into the serializers so the user of the structs can use them without any precursor.

ancazamfir:
Could https://github.com/informalsystems/tendermint-rs/issues/198 be fixed part of this?

Liamsi:
We should merge in `master`. (BTW what is the story of the several `Merge branch 'tendermint/v0.33' of github.com:informalsystems/tendermint-rs into tendermint/v0.33`commits)?

Liamsi:
> Could #198 be fixed part of this?

https://github.com/informalsystems/tendermint-rs/pull/206

greg-szabo:
Removed WIP, let's see if everything is in place for merge. I'd consider @Liamsi or @ebuchman as the best people to approve this.

Liamsi:
I'll go through the changes again and collect everything in a followup issue. I think we can merge this later today.

romac:
FYI there are missing serializers for `TrustThresholdFraction`. I just opened a PR to fix those at https://github.com/informalsystems/tendermint-rs/pull/210.

Also, I just ran the relayer (which depends on this branch) against a Tendermint 0.33.3 node and got an RPC error saying `missing field `height` at line 59 column 7`. I suspect it comes from either  `tendermint::rpc::Client::latest_commit` or `tendermint::rpc::Client::commit`.

ancazamfir:
I tried with gaia nodes running IBC and get `Some("missing field `num_txs` at line 30 column 7")`, this is with `tendermint v0.33.2`

tendermint/src/block/commit_sig.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use crate::serializers;
+use crate::{account, Signature, Time};
+use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
+
+/// BlockIDFlag is used to indicate the validator has voted either for nil, a particular BlockID or was absent.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum BlockIDFlag {
+    /// BlockIDFlagAbsent - no vote was received from a validator.
+    BlockIDFlagAbsent = 1,
```
melekes, line 11: does it make sense to drop prefix?

tendermint/src/block/commit_sig.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    /// Serialize this type as a byte
+    pub fn to_u8(self) -> u8 {
+        self as u8
+    }
+
+    /// Serialize this type as a 32-bit unsigned integer
+    pub fn to_u32(self) -> u32 {
```
melekes, line 35: why would someone need that?

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,42 +45,109 @@ impl lite::Commit for block::signed_header::SignedHeader {
     }
 
     fn validate(&self, vals: &Self::ValidatorSet) -> Result<(), Error> {
-        if self.commit.precommits.len() != vals.validators().len() {
+        // TODO: self.commit.block_id cannot be zero in the same way as in go
```
melekes, line 48: only the very first block (height: 1) is allowed (actually required) to have 0 signatures

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,42 +45,109 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+        // clarify if this another encoding related issue
+        if self.commit.signatures.len() == 0 {
+            fail!(Kind::ImplementationSpecific, "no signatures for commit");
+        }
+        if self.commit.signatures.len() != vals.validators().len() {
             fail!(
                 Kind::ImplementationSpecific,
                 "pre-commit length: {} doesn't match validator length: {}",
```
melekes, line 56:
```suggestion
                "commit signatures count: {} doesn't match validators count: {}",
```

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -52,42 +45,109 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+                    );
+                }
+                if self.signature.is_none() {
+                    fail!(
+                        Kind::ImplementationSpecific,
+                        "missing signature for non-absent CommitSig {:#?}",
+                        self
+                    );
+                }
+                // TODO: this last check is only necessary if we do full verification (2/3) but the
```
melekes, line 136: seems to me this should not be the concern of Commit to validate if it's a part of some validator set, but it should be rather the job of external object, who's probably holding a validator set to make sure commit signatures match validator set. that's how it's done in Go. It's interesting that you've decided to revert this relationship.

tarcieri:
> Probably good to create a 0.32 based release from master first!

I think it'd be good to cut a final `tendermint-rs` v0.12 release (the current release is `v0.12.0-rc0`) then merge this.

Happy to help with that.

Liamsi:
Merging this now. Whenever we want, we can now release this as `0.14.0` (a 0.12.0 and a 0.13.0 were released in the meantime). Will add the adding changelog entries to  #211.

tendermint/src/block/commit_sig.rs
greg-szabo: Opened issue #259 for this.

tendermint/src/block/commit_sig.rs
Liamsi: was removed;  using `#[repr(u8)]` now

PR #199, opened 2020-03-31
commit followup

Liamsi:
* [ ] Wrote tests
* [ ] Updated CHANGES.md


tendermint/src/lite_impl/signed_header.rs
```diff
@@ -58,69 +59,107 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
-            }
+            commit_sig.validate(vals)?;
         }
 
         Ok(())
     }
 }
 
-fn commit_to_votes(commit: block::Commit) -> Result<Vec<vote::Vote>, Error> {
+// this private helper function does *not* to any validation but extracts
```
Shivani912, line 69:
```suggestion
// this private helper function does *not* do any validation but extracts
```

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -48,6 +44,11 @@ impl lite::Commit for block::signed_header::SignedHeader {
     }
 
     fn validate(&self, vals: &Self::ValidatorSet) -> Result<(), Error> {
+        // TODO: self.commit.block_id cannot be zero in the same way as in go
```
Shivani912, line 47: oh, in which case do we find `self.commit.block_id` to be zero in Go?

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -58,69 +59,107 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+                };
+                votes.push(vote);
+            }
+        }
+    }
+    votes
+}
+
+// TODO: consider moving this into commit_sig.rs instead and making it pub
+impl block::commit_sig::CommitSig {
```
Shivani912, line 98: yes, I think this should be moved to commit_sig.rs

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -58,69 +59,107 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+        if let Some(val_addr) = commit_sig.validator_address {
+            if let Some(sig) = commit_sig.signature.clone() {
+                let vote = vote::Vote {
+                    vote_type: vote::Type::Precommit,
+                    height: commit.height,
+                    round: commit.round,
+                    block_id: Option::from(commit.block_id.clone()),
+                    timestamp: commit_sig.timestamp,
+                    validator_address: val_addr,
+                    validator_index: i as u64,
```
Shivani912, line 87:
```suggestion
                    validator_index: u64::try_from(i).expect("usize to u64 conversion failed for val index"),
```

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -58,69 +59,107 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
+                        "missing signature for non-absent CommitSig {:#?}",
+                        self
+                    );
+                }
+                // TODO: this last check is only necessary if we do full verification (2/3) but the
+                // above checks should actually happen always (even if we skip forward)
+                //
+                // returns ImplementationSpecific error if it detects a signer
+                // that is not present in the validator set:
+                if let Some(val_addr) = self.validator_address {
```
Shivani912, line 139: Maybe move this check into a separate function which is only called in case of full verification?

tendermint/src/lite_impl/signed_header.rs
Liamsi: I think I meant, we validate if it is zero the same way as in golang 
https://github.com/tendermint/tendermint/blob/ef56e6661121a7f8d054868689707cd817f27b24/types/block.go#L689-L691
because in the Struct here it is not an `Option`.

tendermint/src/lite_impl/signed_header.rs
Liamsi: Will do in a followup PR.

tendermint/src/lite_impl/signed_header.rs
Liamsi: Will do in a followup PR.

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -58,69 +59,107 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
                 }
+                if self.signature.is_some() {
+                    fail!(
+                        Kind::ImplementationSpecific,
+                        "signature is present for absent CommitSig {:#?}",
+                        self
+                    );
+                }
+                // TODO: deal with Time
+                // see https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r401027989
```
Liamsi, line 117: Will do in a followup PR

Liamsi: resolved in #277

PR #204, opened 2020-04-06
sketching

brapse:
This is an experiment in refactoring the light client to  simplify the trait/type dependencies of core verification logic/testing. 

We have a few problems that we are trying to solve here. Right now those problems are kind of tangled up together so we are trying to understand them better through experimentation.

One problem we have is that  we have to keep specifying trait dependencies everywhere because we want to generalize over the complex crypto functions in testing. Instead it might be better to have simpler types and then break out crypto function into dependencies that we inject where needed.

Another problem is that we have a large test harness used to construct the perfect set of complex data relations used to exercise predicate logic. It might be better to test predicates independently and express core verification as the composition of predicates which can be mocked out independently to test complex scenarios.  

This code is pseudo code and this branch will probably never be merged. The branch is simply used to illustrate a potential approach to solving the problems mentioned above.

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/lite/predicates.rs
+++ b/tendermint/src/lite/predicates.rs
@@ -1,0 +1,162 @@
+/// Sketch out the pseudo code for a light client
+/// That integrates the learning from the last iteration.
+/// What we want:
+/// + Simple light client specific types, no crypto
+/// + Crypto can abstracted into traits which implement crypto specific functions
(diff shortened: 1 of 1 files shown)


romac:
This PR can be closed since the ideas sketched here have been implemented in the [light client refactoring](https://github.com/informalsystems/tendermint-rs/pull/237).

PR #205, opened 2020-04-06
Predicate library

romac:
This PR bundles the predicate library within the repo via [`git-subtree`](https://manpages.debian.org/testing/git-man/git-subtree.1.en.html), and iterates on it.

## TODO

- [x] Figure out a way for closure-based predicate to borrow the data they close over
- [x] Figure out how to express the combinators so that they work over predicates which borrow data
- [x] Add ability to convert predicates to assertions
- [ ] Hide all trait bounds behind common interface
- [ ] Add more boolean combinators (`==`, `>`, `>=`, `<=`, `<=>`, `xor`, etc.)
- [ ] Figure out how to recover tagging for predicates which borrow data

---

* [ ] Wrote tests
* [ ] Updated CHANGES.md


melekes:
Is there an issue where one can read more about the motivation behind creating this library?

melekes:
Have you thought about using conditional compilation to enable/disable certain matchers and using traits to allow mock types? _(writing here since there's no issue)_

```rust
 //////////////////////////////////////////////////////////////////////////////
    // ensure the header validator hashes match the given validators
    #[cfg(validator_hash_matches=true)]
    if header.validators_hash() != vals.hash() {
        return Err(Kind::InvalidValidatorSet {
            header_val_hash: header.validators_hash(),
            val_hash: vals.hash(),
        }
        .into());
    }
    #[cfg(next_validator_hash_matches=true)]
    if header.next_validators_hash() != next_vals.hash() {
        return Err(Kind::InvalidNextValidatorSet {
            header_next_val_hash: header.next_validators_hash(),
            next_val_hash: next_vals.hash(),
        }
        .into());
    }
    //////////////////////////////////////////////////////////////////////////////
```

pred/examples/light_client.rs
```diff
@@ -0,0 +1,470 @@
...
+    commit: &Commit,
+    validators: &ValidatorSet,
+    trust_level: &TrustLevel,
+    calculator: &impl VotingPowerCalculator,
+) -> bool {
+    let total_power = calculator.total_power_of(validators);
+    let voting_power = calculator.voting_power_in(commit, validators);
+
+    if let (Ok(total_power), Ok(voting_power)) = (total_power, voting_power) {
+        // XXX: Maybe trust_level doesn't need a very sophisticated type
```
melekes, line 223: yeah, in Rust it can simply be a tuple imho

pred/examples/light_client.rs
```diff
@@ -0,0 +1,470 @@
...
+}
+
+#[derive(Clone, Debug, Display)]
+#[display(fmt = "{:?}", self)]
+struct Header {
+    height: Height,
+    bft_time: SystemTime,
+    validator_set_hash: Hash,
+    next_validator_set_hash: Hash,
+    hash: Hash, // What if we don't have this
```
melekes, line 40: header is invalid?

pred/examples/light_client.rs
```diff
@@ -0,0 +1,470 @@
...
+#[derive(Clone, Debug, Display)]
+#[display(fmt = "{:?}", self)]
+struct TrustedState {
+    header: Header,
+    validators: ValidatorSet,
+}
+
+// Crypto function traits allowing mocking out during testing
+trait VotingPowerCalculator: Sized {
+    // What kind of errors should we be reporting here?
```
melekes, line 80: not enough voting power signed this commit

romac:
We had a call today with @brapse and @melekes, and came to the conclusion that this library brings in too much code and complexity in exchange for very little reward, ie. only introspection.

We thus decided to drop it, and replace it with plain functions wrapped in a trait to allow for mocking purposes. We'll keep the predicates themselves, as well as the three traits which encapsulate crypto operations.

PR #208, opened 2020-04-08
Fastsync English spec for review

josef-widder:


docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
+# Fastsync
+
+<!--
```
milosevic, line 3: Maybe we should remove these comments from the template.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+and the corresponding application state that is slightly outdated. It
+then queries its peers for the blocks that were decided on by the
+Tendermint blockchain during the period the full node was
+disconnected. After receiving these blocks, it executes the
+transactions in the blocks in order to catch-up to the current height
+of the blockchain and the corresponding application state.
+
+In practice it is sufficient to catch-up only close to the current
+height: The Tendermint consensus reactor implements a similar
+functionality and can synchronize a node that is approximately 100
```
milosevic, line 49: I don't remember seeing code that does this in consensus reactor. @ebuchman @melekes Can someone confirm this? Ideally we should also add a reference to the corresponding code here.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+then there exists a subset *CorrV*
+of *h.NextValidators*, such that:
+  - *TotalVotingPower(CorrV) > 2/3
+    TotalVotingPower(h.NextValidators)*; 
+  - For every validator pair *(n,p)* in *CorrV*, it holds *correctUntil(n,
+    h.Time + trustingPeriod)*.
+
+#### **[TMBC-CorrFull]**: 
+Every correct full node locally stores a prefix of the
+current list of headers from [**[TMBC-SEQ]**](TMBC-SEQ-link).
```
andrey-kuprianov, line 124:
```suggestion
current list of headers from [**[TMBC-SEQ]**][TMBC-SEQ-link].
```
the link is broken

docs/spec/fastsync/fastsync.md
melekes: Consensus reactor can catch up, but it's very slow comparing to fast sync & state sync.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+
+####  **[FS-ISSUE-NON-TERM]**:
+
+Due to [**[FS-ISSUE-KILL]**](#fs-issue-kill), from some point on, only
+faulty peers may be in *peerIDs*. They can thus control at which rate
+*Fastsync* gets blocks. If the timeout duration from [FS-V2-TIMEOUT]
+is greater than the time it takes to add a block to the blockchain
+(LTIME in [**[TMBC-SEQ-APPEND-E]**][TMBC-SEQ-APPEND-E-link]), the
+protocol may never terminate and thus violate [FS-VC-TERM].  This
+scenario is even possible if a correct peer is always in *peerIDs*,
```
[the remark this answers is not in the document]
ebuchman: Slash isn't it possible with just one correct peer if we manage to always eventually request pairs of sequential blocks from that correct peer?

ebuchman:
Other high level thoughts I'll try to summarize in advance of a discussion:

- Is there convention on the definition names over what gets upper cased and what doesnt? Eg. SEQ vs Auth ...
- Can we include some kind of legend to make the short forms more understandable, ie:
  - FM = Fault Model
  - A = Assumption
  - FS = Fast Sync
  - etc ...
- Without the context of the VDD document, or even with it, some of the titles/headings are a bit confusing. For instance the Sequential vs Distributed problem statement and other document organization. We can do a better job with a table of contents and description of how the doc is laid out and how to follow it.
- In our contract we distinguish between the "core" and "peer" state machines of a reactor. This is most notable in the consensus reactor, but we should make some effort to distinguish them here as well. For the fast sync, it's practically all "peer" state machine, though the verifycommit/execute logic would be the "core" state machine.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+nodes. That is, faulty nodes are limited in that they cannot forge
+messages.
+
+<!-- The authenticated Byzantine model assumes [TMBC-Sign-NoForge] and -->
+<!-- [TMBC-FaultyFull], that is, faulty nodes are limited in that they -->
+<!-- cannot forge messages [TMBC-Sign-NoForge]. -->
+
+#### **[TMBC-VALIDATOR-Set]**:
+
+A *validator set* is a set of validator pairs. For a validator set
```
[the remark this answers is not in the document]
josef-widder: The problem was, that in the existing documentation, "validator" refers to both a data structure and a full node. I added the definition. Validator pair refers to the data structure (Address + voting power) and validator node is the full node.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+
+## Sequential Problem statement
+
+<!--
+> should be English and precise. will be accompanied with a TLA spec.
+---->
+
+*Fastsync* gets as input a block of height *h* and the corresponding
+application state *s* that corresponds to the block and state of that
+height of the blockchain [**[TMBC-SEQ]**][TMBC-SEQ-link], and produces
```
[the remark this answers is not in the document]
josef-widder: I guess it is not. I removed the reference.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+as output (i) a list *L* of blocks starting at height *h* to some height
+*terminationHeight*, and (ii) the application state when applying the
+transaction of the list *L* to *s*. Fastsync has to satisfy the following
+properties [FS-Seq-?]:
```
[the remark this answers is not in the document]
josef-widder: I wanted to use '?' as wildcard but I guess it is not clear. Any idea how to write that?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+
+
+#### **[FS-Seq-Live]**: 
+*Fastsync* eventually terminates.
+
+ 
+#### **[FS-Seq-Term]**:
+Let *bh* be the height of the blockchain at the time *Fastsync* starts.
+When *Fastsync* terminates, it outputs a list of all blocks from
+height *h* to some height *terminationHeight >= bh - 1*,
```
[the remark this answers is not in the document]
josef-widder: Yes. Good point. I added a clarification right there.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+[**[TMBC-SEQ]**][TMBC-SEQ-link].
```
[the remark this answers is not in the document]
josef-widder: I agree. It is not so clear (and not consistently throughout the document). I removed the references where they were unclear.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+safety and liveness properties below:
+
+
+#### **[FS-SOME-CORR-PEER]**:
+Initially, the set *peerIDs* contains at least one correct full node.
+
+> While in principle the above condition can be part of a sufficient
+> condition to solve [FS-Seq-Live] and
+> [FS-Seq-Term] and [FS-Seq-Term-SYNC], we will see below that the
+> current implementation of Fastsync (V2) requires the following (much
```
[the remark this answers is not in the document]
josef-widder: As far as I understand the problem (or a similar one) also is present at the other implementations. We should discuss what precisely we can put here.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+- Error condition
+    - fails if precondition is violated
+----
+
+```go
+func OnStatusResponse(addr Address, ht int64)
+```
+- Comment
+    - *ht* is a height
+	- peers can provide the status without being called
```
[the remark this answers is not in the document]
josef-widder: I agree. I aimed at a more uniform terminology with e.g. the light client specs, and at some point I realized I needed that particular addition. The nice thing about functions is that I can describe expected pre- and postconditions very nicely. Let's discuss how to express what we need.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+
+```go
+func Execute()
+```
+- Comments
+    - none
+- Expected precondition
+	- application state is the one of the blockchain at height
+      *height - 1*
+	- **[FS-V2-Verif]** for any two blocks *a* and *b* from
```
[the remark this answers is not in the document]
josef-widder: No, I want this particular condition to be labeled for later reference.

docs/spec/fastsync/fastsync.md
josef-widder: Ethan changed it to 10 already.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+- Comment
+    - *ht* is a height
+	- peers can provide the status without being called
+- Expected precondition
+    - *peerHeights(addr) <= ht*
+- Expected postcondition
+    - *peerHeights(addr) = ht*
+	- *TargetHeight* is updated
+- Error condition
+    - if precondition is violated: *addr* not in *peerIDs* (that is,
```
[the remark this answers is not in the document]
josef-widder: It means the latter, that is, "if it returns an error, then addr will be removed from peerIDs"

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+- Comment
+    - if after adding block *b*, blocks of heights *height* and
+      *height + 1* are in *blockstore*, then `Execute` is called
+- Expected precondition
+    - *pendingblocks(b.Height) = addr*
+	- *b* satisfies basic soundness  
+- Expected postcondition
+    - if function `Execute` has been executed without error or was not executed:
+        - *receivedBlocks(b.Height) = addr*
+	    - *blockstore(b.Height) = b*
```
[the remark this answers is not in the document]
josef-widder: Yes, this just describes bookkeeping that is in addition to what is done within Execute.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+#### **[NewFS-A-INIT]**:
+- *startBlock* is from the blockchain, and within *trustingPeriod*
+(possible with some extra margin to ensure termination before
+*trustingPeriod* expired)
+- *startState* is the application state of the blockchain at Height
+  *startBlock.Height*.
+- *startHeight = startBlock.Height*
+
+#### Additional Variables
+- *trustedBlockstore*: stores for each height greater than or equal to
```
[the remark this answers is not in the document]
josef-widder: Do you mean to rename it?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+		 - have more than 2/3 of the voting power in VS
+- Expected postcondition
+    - returns *true* if precondition holds, and *false* otherwise
+- Error condition
+    - none 
+----
+
+
+```go
+func SequentialVerify {
```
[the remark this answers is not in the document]
josef-widder: We can discuss in what detail we want to describe possible fixes.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+		}
+}
+```
+- Comments
+    - none
+- Expected precondition
+	- [NewFS-VAR-TRUST-INV]
+- Expected postcondition
+    - [NewFS-VAR-TRUST-INV]
+	- there is no block *bnew* with *bnew.Height = height + 1* in
```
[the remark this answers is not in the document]
josef-widder: This says that you don't stop as long as there is another block in blockstore, or in other words, you iterate to the maximal possible height at that moment.

josef-widder:
> * Is there convention on the definition names over what gets upper cased and what doesnt? Eg. SEQ vs Auth ...

No. We will fix a convention soon. For now, I changed all to upper case.

> * Can we include some kind of legend to make the short forms more understandable, ie:
>   
>   * FM = Fault Model
>   * A = Assumption
>   * FS = Fast Sync
>   * etc ...

I added the most frequently used ones

> * Without the context of the VDD document, or even with it, some of the titles/headings are a bit confusing. For instance the Sequential vs Distributed problem statement and other document organization. We can do a better job with a table of contents and description of how the doc is laid out and how to follow it.

I added added a bullet list at the top of the document with pointers to the sections. Please, let me know what you think!

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+> on the time interval between the times at which two
+> successor blocks are added. 
+
+#### **[TMBC-SEQ-APPEND-E]**: 
+If a header is appended at time *t* then no additional header will be
+appended before time *t + ETIME*.
+
+
+#### **[TMBC-AUTH-BYZ]**:
+The authenticated Byzantine model assumes that no node (faulty or
```
milosevic, line 150:
```suggestion
We assume the authenticated Byzantine model in which no node (faulty or
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+correct) may break digital signatures, but otherwise, no additional
+assumption is made about the internal behavior of faulty full
```
milosevic, line 152: Full node is not defined at this point in time. Maybe it's sufficient to say here just faulty nodes.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+nodes. That is, faulty nodes are only limited in that they cannot forge
+messages.
+
+<!-- The authenticated Byzantine model assumes [TMBC-Sign-NoForge] and -->
+<!-- [TMBC-FaultyFull], that is, faulty nodes are limited in that they -->
+<!-- cannot forge messages [TMBC-Sign-NoForge]. -->
+
+
+> We observed that in the existing documentation the term
```
milosevic, line 161: What documentation? Tendermint documentation?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+> the context.
+
+
+#### **[TMBC-VALIDATOR-PAIR]**:
+
+Given a full node, a 
+*validator pair* is a pair *(address, voting_power)*, where 
+  - *address* is the address (public key) of a full node, 
+  - *voting_power* is an integer (representing the full node's
+  voting power in a certain consensus instance).
```
milosevic, line 175:
```suggestion
  voting power in a given consensus instance).
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+  
+> In the Golang implementation the data type for *validator
+> pair* is called `Validator`
```
milosevic, line 178:
```suggestion
> pair* is called `Validator`.
```

docs/spec/fastsync/fastsync.md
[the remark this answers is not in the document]
konnov: Once we have traceability identifiers around, patterns like `?` will be unparseable by the tool.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+#### **[TMBC-CORRECT]**:
+We define a predicate *correctUntil(n, t)*, where *n* is a node and *t* is a 
+time point. 
+The predicate *correctUntil(n, t)* is true if and only if the node *n* 
+follows all the protocols (at least) until time *t*.
+
+#### **[TMBC-TIME-PARAMS]**:
+A blockchain has the following configuration parameters:
+ - *unbondingPeriod*: a time duration.
```
milosevic, line 194: It seems that we don't use unbonding period term.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+#### **[TMBC-CORR-FULL]**: 
+Every correct full node locally stores a prefix of the
+current list of headers from [**[TMBC-SEQ]**][TMBC-SEQ-link].
+
+
+
+
+
+## Informal Problem statement
```
milosevic, line 214:
```suggestion
## Fastsync Informal Problem statement
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+
+A full node has as input a block of the blockchain at height *h* and
+the corresponding application state (or the prefix of the current
+blockchain until height *h*). It has access to a set *peerIDs* of full
+nodes called *peers* that it knows of.  The full node uses the peers
```
milosevic, line 220: Full node is never defined if I am not wrong.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+
+
+*Fastsync* gets as input a block of height *h* and the corresponding
+application state *s* that corresponds to the block and state of that
+height of the blockchain, and produces
+as output (i) a list *L* of blocks starting at height *h* to some height
+*terminationHeight*, and (ii) the application state when applying the
+transactions of the list *L* to *s*. Fastsync has to satisfy the following
+properties [FS-SEQ-?]:
```
milosevic, line 236:
```suggestion
properties:
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+In this document we quite extensively use tags in order to be able to
+reference assumptions, invariants, etc. in future communication. In
+these tags we frequently use the following short forms:
+
+- TMBC: Tendermint blockchain
+- SEQ: for sequential specifications
+- FS: Fastsync
+- LIVE: liveness
+- INV: invariant
+- TERM: termination
```
milosevic, line 90: What is difference between liveness and termination? I would expect that it is the same and that one abbreviation is sufficient. On the other side we don't have abbreviation for safety properties.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+#### **[FS-SEQ-LIVE]**: 
+*Fastsync* eventually terminates.
+
+ 
+#### **[FS-SEQ-TERM]**:
+Let *bh* be the height of the blockchain at the time *Fastsync*
+starts. By assumption we have *bh >= h*.
+When *Fastsync* terminates, it outputs a list of all blocks from
+height *h* to some height *terminationHeight >= bh - 1*.
```
milosevic, line 247: I don't get this property. Not sure why you need to use both bh and h as starting height.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+> The above property is independent of how many blocks are added to the
+> blockchain while Fastsync is running. It links the target height to the
+> initial state. If Fastsync has to catch-up many blocks, it would be
+> better to link the target height to a time close to the
+> termination. This is capture by the following specification:
```
milosevic, line 253:
```suggestion
> termination. This is captured by the following specification:
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+*terminationHeight*, and (ii) the application state when applying the
+transactions of the list *L* to *s*. Fastsync has to satisfy the following
+properties [FS-SEQ-?]:
+
+
+#### **[FS-SEQ-LIVE]**: 
+*Fastsync* eventually terminates.
+
+ 
+#### **[FS-SEQ-TERM]**:
```
milosevic, line 243: This is safety property, so the tag could probably be FS-SAFE-START.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+When *Fastsync* terminates, it outputs a list of all blocks from
+height *h* to some height *terminationHeight >= bh - 1*.
+
+> The above property is independent of how many blocks are added to the
+> blockchain while Fastsync is running. It links the target height to the
+> initial state. If Fastsync has to catch-up many blocks, it would be
+> better to link the target height to a time close to the
+> termination. This is capture by the following specification:
+
+#### **[FS-SEQ-TERM-SYNC]**:
```
milosevic, line 255: FS-SAFE-SYNC?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+terminates, it outputs a list of all blocks from height *h* to some
+height *terminationHeight >= eh - D*.
+
+
+#### **[FS-SEQ-INV]**:
+Upon termination, the application state is the one that corresponds to
+the blockchain at height *terminationHeight*.
+
+> As the block of height *h + 1* is needed to verify the block of
+> height *h* we highlight th following clarification of the
```
milosevic, line 267:
```suggestion
> height *h* we highlight the following clarification of the
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+> termination height.
+
+
+#### **[FS-SEQ-HEIGHT]**: 
+The returned value *terminationHeight* is the height of the block with the largest
+height that could be verified. In order to do so, *Fastsync* needs the
+Commit of the block at height  *terminationHeight + 1* in the blockchain.
```
milosevic, line 274:
```suggestion the block at height  *terminationHeight + 1* in the blockchain.
```

milosevic, line 274:
```suggestion
the block at height  *terminationHeight + 1* in the blockchain.
```

docs/spec/fastsync/fastsync.md
josef-widder: As you noted below, TERM was mostly addressing safety properties upon termination. I will remove TERM here, as we have changed the tags

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+## Environment/Assumptions/Incentives
+
+
+#### **[FS-A-NODE]**:
+We consider a node *FS* that performs *Fastsync*.
+
+#### **[FS-A-PEER-IDS]**:
+*FS* has access to a set *peerIDs* of IDs (public keys) of peers (full
+     nodes). During the execution of *Fastsync*, another protocol (outside
```
milosevic, line 286:
```suggestion
     . During the execution of *Fastsync*, another protocol (outside
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+# Part II - Protocol view
+
+## Environment/Assumptions/Incentives
+
+
+#### **[FS-A-NODE]**:
+We consider a node *FS* that performs *Fastsync*.
+
+#### **[FS-A-PEER-IDS]**:
+*FS* has access to a set *peerIDs* of IDs (public keys) of peers (full
```
milosevic, line 285:
```suggestion
*FS* has access to a set *peerIDs* of IDs (public keys) of peers
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+
+
+#### **[FS-A-PEER]**:
+Peers can be faulty, and we do not make any assumptions about the number or
+ratio of correct/faulty nodes. Faulty processes may be Byzantine
+according to [**[TMBC-AUTH-BYZ]**][TMBC-Auth-Byz-link].
+
+#### **[FS-A-VAL]**:
+The system satisfies [**[TMBC-AUTH-BYZ]**][TMBC-Auth-Byz-link] and [**[TMBC-FM-2THIRDS]**][TMBC-FM-2THIRDS-link]. Thus, there is a
```
milosevic, line 297: This is not really clear in my view. We have a node and peers, and I guess that with this property we want to make a link to validators and Tendermint blockchain. I am not sure what is the easiest way to introduce this without going into details of explaining validators and difference between validators and full nodes. From the fast sync view point the whole blockchain business can be seen as application specific stuff, and it is only relevant as part of block validation in the proposed solution.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+terminate successfully, and satisfy the requirements of the sequential
+problem statement:
+
+#### **[FS-DISTR-TERM]**:
+*Fastsync* eventually terminates: it either *terminates successfully* or it  *terminates with failure*.
+
+
+#### Remote Functions
+
+The Tendermint Full Node exposes the following functions over
```
milosevic, line 334: We haven't defined what is a Tendermint full node. Can we just say that peers expose the following functions over RPC?

docs/spec/fastsync/fastsync.md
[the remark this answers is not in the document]
milosevic: It is also true for the past versions.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+Initially, the set *peerIDs* contains at least one correct full node.
+
+> While in principle the above condition can be part of a sufficient
+> condition to solve [FS-SEQ-LIVE] and
+> [FS-SEQ-TERM] and [FS-SEQ-TERM-SYNC], we will see below that the
+> current implementation of Fastsync (V2) requires the following (much
+> stronger) requirement
+
+#### **[FS-ALL-CORR-PEER]**:
+At all times, the set *peerIDs* contains only correct full nodes.
```
milosevic, line 398: We need to check if some properties of v2 hold in the crash tolerant model.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+> safety specifications / invariants in English 
+---->
+
+
+#### **[FS-VC-ALL-CORR-NONABORT]**:
+Under [FS-ALL-CORR-PEER], *Fastsync* never terminates with failure.
+
+
+#### **[FS-VC-STATE-INV]**:
+If *FastSync* terminates successfully at height *terminationHeight*, then the
```
milosevic, line 414: Isn't this always the case? Difference between success and failure in my view is only with respect to the quality of the result, i.e, about number of synced blocks.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+application state is the one that corresponds to the blockchain at
+height *terminationHeight*.
+
+#### **[FS-VC-BLOCKS-INV]**:
+If *FastSync* terminates successfully at height *terminationHeight*, then the
```
milosevic, line 419: The same comment as above.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+> cannot be ensured in an unreliable distributed setting. We consider
+> the following relaxation. (Which is typically sufficient for
+> Tendermint, as the consensus reactor then synchronizes from that
+> height.)
+
+#### **[FS-VC-CORR-INV]**:
+Under [FS-ALL-CORR-PEER], let *maxh* be the maximum 
+height of a correct peer [**[TMBC-CORR-FULL]**][TMBC-CORR-FULL-link]
+in *peerIDs* at the time *Fastsync* starts. If *FastSync* terminates
+successfully, it is at some height *terminationHeight >= maxh - 1*.
```
milosevic, line 437: I might be wrong but I haven't seen definition of termination height. Is that the height of the last verified and executed block?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+
+> The above property is independent of how many blocks are added to the
```
milosevic, line 439: It is also independent of addition of new peers. As peers can be removed by external component (not just added), we probably need additional assumption that a correct peer is not removed from the peerIds by the external component.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1164 @@
...
+> target height to a time close to the termination. This is captured by
+> the following specification:
+
+
+#### **[FS-VC-CORR-INV-SYNC]**:
+Under [FS-ALL-CORR-PEER], there exists a constant time interval *TD*, such
+that if *term* is the time *Fastsync* terminates and
+*maxh* be the maximum height of a correct peer
+[**[TMBC-CORR-FULL]**][TMBC-CORR-FULL-link] in *peerIDs* at the time
+*term - TD*, then if *FastSync* terminates successfully, it is at
```
milosevic, line 452: Sequential property (FS-SEQ-TERM-SYNC) refers to a logical diff (parameter D that is expressed as lag in number of heights), while here we introduce something similar but not the same with time parameter TD. Can we somehow align this better or it is fine as sequential spec is defined without reference to system parameters.

docs/spec/fastsync/fastsync.md
josef-widder: you are wrong ;-) we have the following:
#### **[FS-SEQ-HEIGHT]**: 
The returned value *terminationHeight* is the height of the block with the largest
height that could be verified. In order to do so, *Fastsync* needs the
block at height  *terminationHeight + 1* of the blockchain.

docs/spec/fastsync/fastsync.md
josef-widder: If we want to say what blocks to accept and execute, we have to state that there is a reference block chain that has these blocks. It is not clear how to specify reasonable invariants without having that reference. It is the validity property of Fastsync.   
Later, in our suggestions to use the lightclient-like verification (at the end of the document) we need this as it is based on the tendermint failure model.

docs/spec/fastsync/fastsync.md
josef-widder: I removed them in a previous revision

docs/spec/fastsync/fastsync.md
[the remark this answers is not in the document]
josef-widder: Zarko suggested to remove that tag

docs/spec/fastsync/fastsync.md
josef-widder: yes.

docs/spec/fastsync/fastsync.md
josef-widder: I think we should still use it as people know the term?

docs/spec/fastsync/fastsync.md
josef-widder: I wanted to recall *terminationheight* here, but we can do without, I removed it.

docs/spec/fastsync/fastsync.md
josef-widder: removed it

docs/spec/fastsync/fastsync.md
milosevic: Ok, fair enough.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+#### Liveness
+
+<!--
+> liveness specifications in English. Possibly with timing/fairness requirements:
+e.g., if the component is connected to a correct full node and communication is
+reliable and timely, then something good happens eventually. 
+---->
+
+#### **[FS-VC-ALL-CORR-TERM]**:
+Under [FS-ALL-CORR-PEER], *Fastsync* eventually terminates successfully.
```
milosevic, line 486: Where do we define what we mean by termination with failure?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+- *peerIDs*: peer addresses [FS-A-PEER-IDS](#fs-a-peer-ids)
+- *peerHeights*: stores for each peer the height it reported. initially 0
+- *pendingBlocks*: stores for each height which peer was
+  queried. initially nil for each height
+- *receivedBlocks*: stores for each height which peer returned
+  it. initially nil
+- *blockstore*: stores for each height greater than
+    *startBlock.Height*, the block of that height. initially nil for
+    all heights
+- *peerTimeStamp*: stores for each peer the last time a block was received
```
milosevic, line 535: There is also pendingTime variable in the code where time of the block request is stored. It is used as part of tryPrune function but only for the peer that is responsible for current height.

docs/spec/fastsync/fastsync.md
josef-widder: In the following. Is this OK?

#### **[FS-V2-TIMEOUT]**:

*Fastsync V2* starts a timeout whenever a block is
executed (that is, when the height is incremented). If the timeout expires
before the next block is executed, *Fastsync* terminates.
We say that if *peerIDs* is empty upon termination, then *Fastsync* terminates
with failure, otherwise it terminates successfully.

docs/spec/fastsync/fastsync.md
josef-widder: I added at this point

- *pendingTime*: stores for each peer the last time a block was requested

and later, I updated:

#### **[FS-V2-PEER-REMOVE]**:
Periodically, *peerTimeStamp* and *peerRate* and *pendingTime* are
analyzed. 
If a peer *p*
has not provided a block recently (check of *peerTimeStamp[p]*) or it
has not provided sufficiently many data (check of *peerRate[p]*), then
*p* is removed from *peerIDs*. In addition, *pendingTime* is used to
estimate whether the peer that is responsible for current height is
still connected.

docs/spec/fastsync/fastsync.md
josef-widder: I agree the issue here is not explicitly discussed. The problem is that in the case the peers are themselves synchronized to the blockchain (they are at the current height), then *TD* together with LTIME in [**[TMBC-SEQ-APPEND-E]**][TMBC-SEQ-APPEND-E-link]  can be used to bound $D$. If the peers themselves lag behind and catch-up during the *TD* time, we cannon bound anything. Perhaps it is enough to add an explanation at this point to make it clear. What do you think?

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+---->
+
+### Outline
+
+<!--
+> Describe solution (in English), decomposition into functions, where communication to other components happens.
+---->
+
+The protocol is described in terms of functions that are triggered by
+(external) events. The implementation uses a scheduler and a
```
milosevic, line 600: I am not sure that this sentence is correct. Maybe we just need to explain that some messages are sent periodically (statusRequest) and some actions that lead to sending messages are also triggered periodically. We just need to explain basically execution model.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+  interval greater than *2 Delta*) queries all peers from *peerIDs*
+  for their current height [TMBC-CORR-FULL]. It does so
+  by calling `Status(n)` remotely on all peers *n*.
+  
+- `CreateRequest`: regularly checks whether certain blocks have no
+  open request. If a block does not have an open request, it requests
+  one from a peer. It does so by calling `Block(n,h)` remotely on one
+  peer *n* for a missing height *h*.
+  
+> We have left the strategy how peers are selected unspecified, and
```
milosevic, line 614: Maybe we should explain the strategy used as it is part of the solution. It basically selects a peer with the minimum number of pending requests that can serve given height h (peerHeight >= h).

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+
+#### **[FS-V2-PEER-REMOVE]**:
+Periodically, *peerTimeStamp* and *peerRate* are analyzed. If a peer *p*
+has not provided a block recently (check of *peerTimeStamp[p]*) or it
+has not provided sufficiently many data (check of *peerRate[p]*), then
+*p* is removed from *peerIDs*.
+  
+#### **[FS-V2-TIMEOUT]**:
+
+*Fastsync V2* starts a timeout whenever a block is
```
milosevic, line 648: To be precise, implementation is slightly different. Node remember the last point in time when a block has been executed. Then there is an action that is triggered periodically and if no progress has been made for at least syncTimeout, i.e., (now - lastUpdate > syncTimeout) then fastSync terminates.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+executed (that is, when the height is incremented). If the timeout expires
+before the next block is executed, *Fastsync* terminates.
+We say that if *peerIDs* is empty upon termination, then *Fastsync* terminates
```
milosevic, line 651: Not sure about this. Note that you can terminate in this case with a peerset not empty, and I am not sure if it is still considered success. My understanding was that success is in case we sync up to maxHeight and failure is when we terminate because of this timeout.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+- Expected precondition
+    - *peerHeights(addr) <= ht*
+- Expected postcondition
+    - *peerHeights(addr) = ht*
+	- *TargetHeight* is updated
+- Error condition
+    - if precondition is violated: *addr* not in *peerIDs* (that is,
+      *addr* is removed from *peerIDs*)
+- Timeout condition
+    - if `OnStatusResponse(addr, ht)` was not invoked within *2 Delta* after
```
milosevic, line 696: Note that in V2 this is not the case, i.e., it is not tracked when statusResponse arrived. Therefore, peers are not removed upon late statusReply.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+```go
+func CreateRequest
+```
+- Expected precondition
+    - *height < TargetHeight*
+	- *peerIDs* nonempty
+- Expected postcondition
+    - Function `Block` is called remotely at a peer *addr* in peerIDs 
+	  for a missing height *h*  
+	  *Remark:* different implementations may have different
```
milosevic, line 710: As this spec is about v2, we should probably details strategy implemented in v2.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+      strategies to balance the load over the peers
+    - *pendingblocks(h) = addr*
```
milosevic, line 712: It also remember time at which request is made, but don't know if we want to enter that level of details as you capture it at more high level above.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+----
+
+
+```go
+func OnBlockResponse(addr Address, b Block)
+```
+- Comment
+    - if after adding block *b*, blocks of heights *height* and
+      *height + 1* are in *blockstore*, then `Execute` is called
```
milosevic, line 721: Execute is in v2 triggered periodically, i.e., it is not triggered by block reception.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+- Expected precondition
+    - *pendingblocks(b.Height) = addr*
+	- *b* satisfies basic soundness
```
milosevic, line 724: blockstore(b.Height) = nil (empty)

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+      *height + 1* are in *blockstore*, then `Execute` is called
+- Expected precondition
+    - *pendingblocks(b.Height) = addr*
+	- *b* satisfies basic soundness  
+- Expected postcondition
+    - if function `Execute` has been executed without error or was not
+      executed:
+        - *receivedBlocks(b.Height) = addr*
+        - *blockstore(b.Height) = b*
+        - *peerTimeStamp[addr]* is set to a time between invocation and
```
milosevic, line 730: peerTimestamp[addr] is set to the current time in v2, but maybe you want to capture here more high level concept.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+          return of the function.
+        - *peerRate[addr]* is updated according to size of received
+          block
```
milosevic, line 733:
```suggestion
          block and time it has passed between current time and last block received from this peer (addr)
```

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+- Error condition
+    - if precondition is violated: *addr* not in *peerIDs*; reset
```
milosevic, line 735: Removing a peer triggers cleaning data structure such that all pending/received but not processed requests/blocks are removed.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+	*pendingblocks(b.Height)* to nil;
+- Timeout condition
+    - if `OnBlockResponse(addr, b)` was not invoked within *2 Delta* after
```
milosevic, line 738: In v2, there is no such check. Removal happens periodically but only for the current height this kind of check is implemented. For other peers, it looks at peer rate and difference between current time and last block reception.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+----
+
+
+
+```go
+func Execute()
+```
+- Comments
+    - none
+- Expected precondition
```
milosevic, line 749: Precondition can be that blocks for height and height+1 are in the blockStore, and that VerifyCommit (a,b.Commit) = true. Postcondition is that block at height is executed, state updated and height incremented. There are two error cases: 1) blocks are not present, do nothing, or 2) blocks are present but VerifyCommit (a,b.Commit) = false. Then both blocks are removed and the corresponding peers are removed.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+
+## Analysis of Fastsync V2
+
+####  **[FS-ISSUE-KILL]**:
+If two blocks are not matching [FS-V2-Verif], `Execute` dismisses both
+blocks and removes the peers that provided these blocks from
+*peerIDs*. If block *a* was correct and provided by a correct peer *p*,
+and block b was faulty and provided by a faulty peer, the protocol
+- removes the correct peer *p*, although it might be useful to
+  download blocks from it in the future
```
milosevic, line 773: Maybe it's better to link it with failed invariant (we don't download all blocks from correct peers).

docs/spec/fastsync/fastsync.md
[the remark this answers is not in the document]
milosevic: I guess your comment applies to the case above: we could ask always consecutive blocks from a peer, but it would lead to waste of resources as you would need for every block to download two blocks (in case requests are made to different peers). More efficient solution would be asking block and commit from a single peer, but that would require Tendermint API.

docs/spec/fastsync/fastsync.md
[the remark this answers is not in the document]
milosevic: In my view, it is sufficient to describe the fix at the high level, as done above, so we don't need to enter the details of the pseudocode.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1101 @@
...
+```go
+func SequentialVerify {
+	while (true) {
+		b1 = blockstore[height];
+		b2 = blockstore[height+1];
+		if b1 == nil or b2 == nil {
+			exit;
+		}
+		if ValidCommit(trustedBlockstore[height - 1].NextValidators, b2.commit) {
+			// we trust b2
```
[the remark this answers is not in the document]
milosevic: This follows from the definition above.

docs/spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1163 @@
...
+   height is increased (a).
+	 
+8. By 2. and 7., faulty peers can delay incrementing the height at
+   most *fmax* times, where each time "costs" *2 Delta* seconds. We
+   have additional *2 Delta* initial offset (3a) plus *2 Delta* to get
+   all missing blocks after the last fault showed itself. (This
+   assumes that an arbitrary number of blocks can be obtained and
+   checked within one round-trip 2 Delta; which either needs
+   conservative estimation of Delta, or a more refined analysis). Thus
+   we reach the *targetHeight* and terminate by time *end*.
```
milosevic, line 1074: I like this analysis. This was my intuition, but I wasn't able to formalise it. I am just wondering if faulty peers sending statusResponse for nonexistent blocks can prevent us from termination. I am wondering also if the fact that correct peers are not fully up to date with the blockchain matters. It seems that what this proposal ensures is that once we hit max height of correct peers, if faulty peers advertise higher height they will eventually be removed as *2 Delta < ETIME*.

docs/spec/fastsync/fastsync.md
josef-widder: At some point we said we want to keep this underspecified on purpose, now it seems that this and the other comments target at specifying V2 more precisely. We should fix what level of abstraction we are aiming for.

docs/spec/fastsync/fastsync.md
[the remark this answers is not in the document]
ebuchman: Ah, I'm used to * as a wildcard, but in any case think it can just be removed here

docs/spec/fastsync/fastsync.md
josef-widder: I added the clarification for V2 to the comment.

docs/spec/fastsync/fastsync.md
josef-widder: We can describe it more precisely, but I would consider this an implementation detail of how it is checked whether a timeout occurred.

docs/spec/fastsync/fastsync.md
josef-widder: I thought that in the synchronous system your scenario is not possible. But I guess I didn't make explicit that the timeout must be sufficiently large.

PR #213, opened 2020-04-11
Update secp address hash

austinabell:
I just manually created a test vector since I couldn't use the go tendermint vector (I didn't look into what the extra address bytes were there and assumed that's handled somewhere else here)

Want me to add this in CHANGES.md?

Edit: Everything passes on stable and nightly toolchain locally, and I can't see why CI failed for some reason

* [x] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -55,6 +55,7 @@ thiserror = "1"
 toml = { version = "0.5" }
 uuid = { version = "0.8", default-features = false }
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
+ripemd160 = "0.8.0"
 
 [dev-dependencies]
 serde_json = "1"
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -55,7 +55,7 @@ thiserror = "1"
 toml = { version = "0.5" }
 uuid = { version = "0.8", default-features = false }
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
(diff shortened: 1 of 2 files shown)


tendermint/src/account.rs
```diff
@@ -58,12 +59,13 @@ impl Debug for Id {
     }
 }
 
-// TODO: should be RIPEMD160(SHA256(pk))
+// RIPEMD160(SHA256(pk))
```
Liamsi, line 62: https://github.com/tendermint/tendermint/blob/ef56e6661121a7f8d054868689707cd817f27b24/crypto/secp256k1/secp256k1.go#L141-L150

Liamsi:
> Everything passes on stable and nightly toolchain locally, and I can't see why CI failed for some reason

I restarted the workflow and it doesn't even start. Looks like a GH issue.

tendermint/Cargo.toml
```diff
@@ -55,6 +55,7 @@ thiserror = "1"
 toml = { version = "0.5" }
 uuid = { version = "0.8", default-features = false }
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
+ripemd160 = "0.8.0"
```
Liamsi, line 58: This should probably be changed to
```suggestion
ripemd160 = "0.8"
```

austinabell: assumed because others in file were only specified up to minor version that it was the preference, updated

PR #221, opened 2020-04-16
Add NoBlockResponse

milosevic:
* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/docs/spec/fastsync/fastsync.tla
+++ b/docs/spec/fastsync/fastsync.tla
@@ -32,12 +32,13 @@ As part of the fast sync protocol a node and the peers exchange the following me
 - StatusResponse
 - BlockRequest
 - BlockResponse
+- NoBlockResponse.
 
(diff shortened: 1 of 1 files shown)


konnov:
How about the following scenario: A faulty peer can claim to have a large height and then send `NoBlockResponse` as a reply to every block. Then the protocol will finish by timeout, without updating its list of blocks.

ebuchman:
> How about the following scenario: A faulty peer can claim to have a large height and then send NoBlockResponse as a reply to every block. Then the protocol will finish by timeout, without updating its list of blocks.

What's the problem here? A single NoBlockResponse gets you removed - its functionally the same as not sending the block at all and causing a peer timeout.

konnov:
> A single NoBlockResponse gets you removed - its functionally the same as not sending the block at all and causing a peer timeout.

I missed that part in the diff. Then it should be fine!

PR #225, opened 2020-04-17
Add subscriptions support

zmanian:
To do
* [x] integration tests
* [x] Typed events 

* [x] Wrote tests
* [ ] Updated CHANGES.md


zmanian:
Would appreciate some initial review here.

tendermint/src/rpc/client.rs
```diff
@@ -182,10 +183,44 @@ impl Client {
...
+        ws.subscribe(query).await?;
+
+        self.ws = Some(ws);
+
+        Ok(())
+    }
+
+    //TODO Have a query type instead of a string
+    /// Subscribe to the Events Websocket with a query string for example "tm.event = 'NewBlock'"
+    pub async fn get_event(&mut self) -> Result<subscribe::Event, Box<dyn std::error::Error>> {
```
Liamsi, line 216: Maybe we can remove the `Box<dyn std::error::Error>` altogether by implementing `impl From<Box<dyn std::error::Error>> for rpc::Error`. Then this method can return `rpc::Error` instead of `Box<dyn std::error::Error>` (consistent with the other methods).

tendermint/src/rpc/client.rs
```diff
@@ -182,10 +183,44 @@ impl Client {
...
+                    Error::invalid_params(&format!("invalid RPC address: {:?}", other)).into();
+
+                return Err(err);
+            }
+        };
+        let mut ws =
+            subscribe::WebSocketEvents::websocket(&format!("ws://{}:{}", host, port)).await?;
+        ws.subscribe(query).await?;
+
+        self.ws = Some(ws);
```
Liamsi, line 209: I didn't look into how the underlying library works (async_tungstenite) but doesn't each subscribe call overwrite the `WebSocketEvents` in the client? Meaning the client looses access to previous `WebSocketStream`s?

tendermint/src/rpc/client.rs
```diff
@@ -16,12 +16,13 @@ use hyper::header;
 pub struct Client {
     /// Address of the RPC server
     address: net::Address,
+    ws: Option<rpc::endpoint::subscribe::WebSocketEvents>,
```
Liamsi, line 19: not sure but shouldn't this be a list of such events? See also the comment in `pub async fn subscribe` about this being overwritten for each `subscribe` call.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,77 @@
...
+use tokio::net::TcpStream;
+
+/// Event Listener over webocket. https://docs.tendermint.com/master/rpc/#/Websocket/subscribe
+pub struct EventListener {
+    socket: async_tungstenite::WebSocketStream<TokioAdapter<TcpStream>>,
+}
+
+impl EventListener {
+    /// Constructor for event listener
+    pub async fn connect(
```
brapse, line 19: Why not `new`?

brapse:
Very cool. Might be helpful to have an issue or ADR describing the approach for context during review 🙏 .

Liamsi:
Changed the base of this PR to `master` as the #196 (`tendermint/v0.33`) was merged.

tendermint/src/rpc/event_listener.rs
romac: To me `connect` is fine here, since the method actually initializes a connection by calling `connect_async`.
In general I feel like `new` should be reserved for non-`async` constructors.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,77 @@
...
+        let (ws_stream, _unused_tls_stream) =
+            connect_async(&format!("ws://{}:{}/websocket", host, port)).await?;
+        Ok(EventListener { socket: ws_stream })
+    }
+    //TODO Have a query type instead of a string
+    /// Subscribe to event query stream over the websocket
+    pub async fn subscribe(&mut self, query: String) -> Result<(), RPCError> {
+        self.socket
+            .send(Message::text(
+                subscribe::Request::new(query.to_owned()).into_json(),
```
melekes, line 40: _newbie question:_ isn't query already has ownership over data? why call `to_owned()`

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,77 @@
...
+            ))
+            .await?;
+        Ok(())
+    }
+
+    /// Subscribe to the Events Websocket with a query string for example "tm.event = 'NewBlock'"
```
melekes, line 46: this does not match `get_event` fn below

tendermint/src/rpc/event_listener.rs
Liamsi: Yeah, you are right there is no need for calling `to_owned` here with that method signature (with `query: String` as param). Zaki probably had a reference based signature (e.g. `query: &str`) first and forgot to remove the (now obsolete) `to_owned`.

tony-iqlusion: Did clippy not catch that?

brapse:
Still missing a bit of context here but seems like important work. What is the the trajectory for getting this reviewed and merged?

zmanian:
Right now this is going to power the back for the Game of Zones scoreboard.

Launching that this weekend. i'm going to try and tidy things up and write an ADR for this.

Right now just trying to get sagan into production to monitor ~180 chains simultaneously😓

tendermint/tests/integration.rs
```diff
@@ -154,15 +154,15 @@ mod rpc {
...
+        // let _ = client.subscribe(&"tm.event='NewBlock'".to_owned()).await.unwrap();
 
         // Collect and throw away the response to subscribe
         let _ = client.get_event().await.unwrap();
 
         // Loop here is helpful when debuging parsing of JSON events
         // loop{
         let resp = client.get_event().await.unwrap();
-        dbg!(&resp);
+        debug!(&resp);
```
Liamsi, line 165: This actually should be `dbg!`: https://doc.rust-lang.org/std/macro.dbg.html

Liamsi: And there is still one place with the type `JsonRPCTransctionResult` in this file below.

tendermint/src/rpc/error.rs
```diff
@@ -179,6 +195,7 @@ impl From<Code> for i32 {
             Code::InvalidParams => -32602,
             Code::InternalError => -32603,
             Code::ServerError => -32000,
+            Code::WebSocketError => 0,
```
Liamsi, line 198: This is the same as HttpError 🤔 And in `impl From<i32> for Code` it will also be interpreted as such. Is that on purpose?

tendermint/tests/integration.rs
```diff
@@ -143,4 +144,30 @@ mod rpc {
...
+
+        // Loop here is helpful when debuging parsing of JSON events
+        // loop{
+        let resp = client.get_event().await.unwrap();
+        dbg!(&resp);
+        // }
+        match resp {
+            Event::GenericStringEvent { data: _ } => assert!(false),
+            Event::GenericJSONEvent { data: _ } => assert!(true),
+            Event::JsonRPCTransactionResult { data: _ } => assert!(true),
```
Liamsi, line 170: So this panics when we get a `Event::GenericStringEvent` but passes in all other cases?

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,166 @@
...
+                Ok(data) => Ok(Event::GenericJSONEvent { data }),
+                Err(_) => Ok(Event::GenericStringEvent {
+                    data: msg.to_string(),
+                }),
+            },
+        }
+    }
+}
+
+//TODO more event types
```
Liamsi, line 70: This can be deleted right?

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,166 @@
...
+    },
+}
+
+/// Standard JSON RPC Wrapper
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct JSONRPC {
+    jsonrpc: String,
+    id: String,
+    result: RPCResult,
+}
```
Liamsi, line 98: This kinda duplicates the existing JSONRPC wrapper:
https://github.com/informalsystems/tendermint-rs/blob/9fc495a88803076b177ec90b8bb0a0b6d3102f47/tendermint/src/rpc/response.rs#L23-L37

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,166 @@
...
+    key: String,
+    value: String,
+}
+
+impl JSONRPC {
+    /// Extract events from TXEvent if event matches are type query
+    pub fn extract_events(
+        &self,
+        action_query: &str,
+    ) -> Result<std::collections::HashMap<String, Vec<String>>, &'static str> {
```
Liamsi, line 156: I think this should rather return a `Box<dyn stdError>` instead of `&'static str` or a concrete error type?

tendermint/src/rpc/endpoint/subscribe.rs
```diff
@@ -0,0 +1,43 @@
...
+impl rpc::Response for Response {
+    /// We throw away response data JSON string so swallow errors and return the empty Response
+    fn from_string(_response: impl AsRef<[u8]>) -> Result<Self, rpc::Error> {
+        Ok(Response {})
+    }
+
+    /// We throw away responses in `subscribe` to swallow errors from the `io::Reader` and provide
+    /// the Response
+    fn from_reader(_reader: impl Read) -> Result<Self, rpc::Error> {
+        Ok(Response {})
```
Liamsi, line 41: Is this also supposed to be empty?

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+//TODO more event types
+/// The Event enum is typed events emitted by the Websockets
+#[derive(Debug, Clone)]
+pub enum Event {
+    /// The result of the ABCI app processing a transaction, serialized as JSON RPC response
+    JsonRPCTransactionResult {
+        /// the tx result data
+        data: Box<JSONRPC>,
```
Liamsi, line 77: Does this really need to contain the wrapped jsonrpc result?

tendermint/tests/integration.rs
```diff
@@ -143,4 +144,31 @@ mod rpc {
...
+
+        // Loop here is helpful when debugging parsing of JSON events
+        // loop{
+        let resp = client.get_event().await.unwrap();
+        dbg!(&resp);
+        // }
+        match resp {
+            Event::GenericStringEvent { data: _ } => panic!(
+                "Expected GenericJSONEvent or JsonRPCTransactionResult, but got GenericStringEvent"
+            ),
```
Liamsi, line 170: I do not fully understand why this is supposed to panic here? What is the purpose of `GenericStringEvent` if the integration tests only test that we panic if we see it? (previously was `assert!(false)` here).

tendermint/src/rpc/event_listener.rs
ancazamfir: `data: JSONRPC,` was causing this clippy error: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

tendermint/src/rpc/endpoint/subscribe.rs
Liamsi: Ah Ok, I see: on subscribing, we really just care about if there was an error or not. But here we even ignore any errors. Something seems dodgy here.

tendermint/tests/integration.rs
Liamsi: note that this integration test fails:
```
failures:

---- rpc::event_subscription stdout ----
[tendermint/tests/integration.rs:165] &resp = GenericStringEvent {
    data: "",
}
thread 'rpc::event_subscription' panicked at 'Expected GenericJSONEvent or JsonRPCTransactionResult, but got GenericStringEvent', tendermint/tests/integration.rs:168:54
```
(and it also failed before I simplified the match arm and the panic instead of `assert!(false)`)

tendermint/src/rpc/event_listener.rs
Liamsi: That's a neat trick, but I was wondering if data shouldn't rather be the events instead of the whole wrapped json result (independent of boxing or not).

tendermint/src/rpc/endpoint/subscribe.rs
zmanian: Add a comment to explain

tendermint/src/rpc/error.rs
zmanian: I don't understand the Error code system. It is similar to an HTTP error?

tendermint/src/rpc/endpoint/subscribe.rs
```diff
@@ -29,6 +29,9 @@ impl rpc::Request for Request {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Response {}
 
+/// Subcribe is weird RPC endpoint. It's only meaningful at websocket response and there isn't a
+/// synchronous reponse offered. It there is an error it's asynchronous and we don't try and stich
+/// the async response backer together with the request.
```
Liamsi, line 34:
```suggestion
/// the async response back together with the request.
```
So that's why you can also ignore the error here: https://github.com/informalsystems/tendermint-rs/pull/225#discussion_r429497738
Interesting. The documentation isn't as clear as it should be https://docs.tendermint.com/master/rpc/#/Websocket/subscribe

Thanks for clarifying this!

tendermint/src/rpc/endpoint/subscribe.rs
Liamsi: Resolved.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,292 @@
...
+pub struct Version {
+    block: String,
+    app: String,
+}
+///Perevious Commit
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct LastCommit {
+    height: String,
+    round: String,
+    block_id: BlockId,
```
Liamsi, line 255: These duplicate existing types. I think for the sake of getting this merged in quickly, we can leave it as is, but this code should certainly reuse the existing types as much as possible.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,292 @@
...
+pub struct ResultEndBlock {
+    validator_updates: Vec<Option<serde_json::Value>>,
+}
+
+impl JSONRpcTxResult {
+    /// Extract events from TXEvent if event matches are type query
+    pub fn extract_events(
+        &self,
+        action_query: &str,
+    ) -> Result<HashMap<String, Vec<String>>, &'static str> {
```
Liamsi, line 282: I'll make this a 
```suggestion
    ) -> Result<HashMap<String, Vec<String>>, Box<dyn std::Error>> {
```
tomorrow. (Do not commit yet).

zmanian: Yeah an actual error type or Box Dyn is one of the last things that needs to be done

tendermint/src/rpc/event_listener.rs
zmanian: Oh DRYing these should be fast

tendermint/src/rpc/error.rs
Liamsi: The error codes are transferred over the wire to signal a certain error type occured. Strings aren't really helpful here and integers work for every language and over JSON. As far as I can see, there is no special WebSocketError on the tendermint go side: https://github.com/tendermint/tendermint/blob/976eebe281685cd7982e03a24ce06ab3ff0561e8/rpc/jsonrpc/types/types.go#L223-L241

tendermint/src/rpc/event_listener.rs
zmanian: I just pushed DRYing the ones that seemed obvious. It looks like some of the response structure here is different enough from the BlockResponse that it requires a new struct.

tendermint/src/rpc/event_listener.rs
zmanian: Did you want to sort this out? i'm down for a break now

tendermint/src/rpc/event_listener.rs
zmanian: This represents an interesting question. Does the consumer want the raw data, the RPC response, just the Event? 

I guess my POV was provide everything and let the consumer figure it out.

tendermint/src/rpc/event_listener.rs
Liamsi: Looks good to me 👍

tendermint/src/rpc/event_listener.rs
Liamsi: done in  668696d

tendermint/src/rpc/event_listener.rs
```diff
@@ -243,7 +243,7 @@ impl JsonRPCTransactionResult {
                 Err("Incorrect Event Type".into())
             }
 
-            None => Err("No result data found".into()),
+            None => Err("No result dcaata found".into()),
```
Liamsi, line 246:
```suggestion
            None => Err("No result data found".into()),
```

tendermint/src/rpc/event_listener.rs
Liamsi: This was addressed in e1549e83e71bee7d846011aef483b895f3a2d403

tendermint/src/rpc/event_listener.rs
Liamsi: I further removed Block, BlockData, Evidence, Parts in 3269d58

zmanian:
Updated the public API to only contain the result body instead of JSON RPC wrapper

tendermint/src/rpc/event_listener.rs
```diff
@@ -73,9 +73,23 @@ impl EventListener {
             .await
             .ok_or_else(|| RPCError::websocket_error("web socket closed"))??;
         match serde_json::from_str::<JsonRPCBlockResult>(&msg.to_string()) {
-            Ok(data) => Ok(Event::JsonRPCBlockResult(data)),
+            Ok(data) => {
+                if let Some(data) = data.0.result {
+                    return Ok(Event::JsonRPCBlockResult(data));
```
Liamsi: It easier to see that this contains jsonrpcWrapper.result:
```suggestion
                if let Some(result) = data.0.result {
                    return Ok(Event::JsonRPCBlockResult(result));
```

tendermint/src/rpc/event_listener.rs
```diff
@@ -73,9 +73,23 @@ impl EventListener {
...
+                } else {
+                    // The Websocket should never send an empty block
+                    panic!("Websocket sent us an empty block")
```
Liamsi, line 83: I think we can leave this as is but I'm a bit worried that someone could send a non-valid response and crash the process (even if tendermint wouldn't send an empty block here). As far as I see the connection isn't authenticated so this should be fairly easy.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,235 @@
...
+                        Ok(Event::JsonRPCTransactionResult(Box::new(tx_result)))
+                    } else {
+                        // The Websocket should never send an empty transaction
+                        panic!("Websocket sent us an empty transaction")
+                    }
+                }
+                Err(_) => match serde_json::from_str::<serde_json::Value>(&msg.to_string()) {
+                    Ok(json_value) => Ok(Event::GenericJSONEvent(json_value)),
+                    Err(_) => Ok(Event::GenericStringEvent(msg.to_string())),
+                },
```
Liamsi, line 98: I'm pretty sure this can be simplified in a way that serde returns the correct `Event` enum variant and you don't need matching here again. Non-blocking for merging this PR though.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Get the next event from the websocket
+    pub async fn get_event(&mut self) -> Result<Event, RPCError> {
+        let msg = self
+            .socket
+            .next()
+            .await
+            .ok_or_else(|| RPCError::websocket_error("web socket closed"))??;
+        match serde_json::from_str::<JsonRPCBlockResult>(&msg.to_string()) {
+            Ok(data) => {
+                if let Some(block_result) = data.0.result {
```
Liamsi, line 79: Why not use the methods to access these fields? (`into_result()`)

Liamsi: This would also account for malformatted responses (e.g. empty ones ...) and remove the panic below too.
I'll push sth. shortly.

Liamsi: b11ea9a

tendermint/src/rpc/event_listener.rs
Liamsi: b11ea9a

Liamsi:
Any idea why the integration test receives a `GenericJSONEvent` (which basically looks like a `JsonRPCBlockResult`)?
https://github.com/informalsystems/tendermint-rs/pull/225/checks?check_run_id=701830448#step:5:266

You were saying that this does not fail against your local gaia build? Which (tendermint) version does that use?

Liamsi:
I'll try to figure out what is going on with the integration test this afternoon and we should be ready to merge.

zmanian:
So the integration test fails if Serde fails to deserialize the Block event as a expected type and instead falls through to a Generic JSON event.

zmanian:
The GOZ code is running 
```
gaiad tendermint version
tendermint: 0.33.4
abci: 0.16.2
blockprotocol: 10
p2pprotocol: 7

``` 

and the integration test appears to pass with this version.

Liamsi:
> So the integration test fails if Serde fails to deserialize the Block event as a expected type and instead falls through to a Generic JSON event.

Exactly. The reason why it's falling through is `Error("missing field `events`", line: 61, column: 32)`. But interestingly, the JSON payload has that field 🤔 

I doubt this has anything todo with different tendermint versions.

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,241 @@
...
+        match serde_json::from_str::<JsonRPCBlockResult>(&msg.to_string()) {
+            Ok(data) => {
+                let block_result = data.into_result()?;
+                Ok(Event::JsonRPCBlockResult(Box::new(block_result)))
+            }
+            Err(e) => {
+                dbg!("Error:");
+                dbg!(e);
+                dbg!("msg:");
+                dbg!(&msg.to_string());
```
Liamsi, line 87: These lines should be removed before merging.
```suggestion
```

tendermint/src/rpc/event_listener.rs
```diff
@@ -0,0 +1,241 @@
...
+                match serde_json::from_str::<JsonRPCTransactionResult>(&msg.to_string()) {
+                    Ok(data) => {
+                        let tx_result = data.into_result()?;
+                        Ok(Event::JsonRPCTransactionResult(Box::new(tx_result)))
+                    }
+                    Err(_) => match serde_json::from_str::<serde_json::Value>(&msg.to_string()) {
+                        Ok(json_value) => Ok(Event::GenericJSONEvent(json_value)),
+                        Err(_) => Ok(Event::GenericStringEvent(msg.to_string())),
+                    },
+                }
```
Liamsi, line 97: I think ideally, we would not match case a lot here but instead of jsonrpc wrapping `JsonRPCBlockResult`, `JsonRPCTransactionResult` separately, we should wrap the whole `Event` enum and make it `impl response::Response`. serde can very well deal with [enums](https://serde.rs/enum-representations.html) so there is no need to try each variant separately. No need to block this PR on this though.

PR #236, opened 2020-04-23
Fix block_id of nil vote

yihuang:
The block_id of nil vote should be `None`, otherwise, the signature verify fails.

### DIFF
--- a/tendermint/src/lite_impl/signed_header.rs
+++ b/tendermint/src/lite_impl/signed_header.rs
@@ -98,6 +98,7 @@ fn non_absent_votes(commit: &block::Commit) -> Vec<vote::Vote> {
         let extracted_validator_address;
         let extracted_timestamp;
         let extracted_signature;
+        let block_id;
         match commit_sig {
             CommitSig::BlockIDFlagAbsent { .. } => continue,
             CommitSig::BlockIDFlagCommit {
@@ -108,6 +109,7 @@ fn non_absent_votes(commit: &block::Commit) -> Vec<vote::Vote> {
                 extracted_validator_address = validator_address;
                 extracted_timestamp = timestamp;
                 extracted_signature = signature;
+                block_id = Some(commit.block_id.clone());
             }
             CommitSig::BlockIDFlagNil {
                 validator_address,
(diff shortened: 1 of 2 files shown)


tendermint/src/lite_impl/signed_header.rs
```diff
@@ -82,7 +82,11 @@ fn non_absent_votes(commit: &block::Commit) -> Vec<vote::Vote> {
                     vote_type: vote::Type::Precommit,
                     height: commit.height,
                     round: commit.round,
-                    block_id: Option::from(commit.block_id.clone()),
+                    block_id: if commit_sig.block_id_flag == BlockIDFlag::BlockIDFlagCommit {
+                        Some(commit.block_id.clone())
+                    } else {
+                        None
+                    },
```
Liamsi, line 89: 👍

Liamsi:
Please base this PR against master: https://github.com/informalsystems/tendermint-rs/pull/196 was merged.

tendermint/src/lite_impl/signed_header.rs
Liamsi, line 89: Slightly more idiomatic would be:
```suggestion
                    block_id: match commit_sig.block_id_flag {
                        BlockIDFlag::BlockIDFlagCommit => Option::from(commit.block_id.clone()),
                        _ => None,
                    },
```

yihuang:
> Would you mind adding a test that would have failed without these changes @yihuang? That would be awesome! Otherwise, I'll do it in a followup issue.

I guess we need to add a nil vote in those JSON test cases, but I don't know how do you produce those JSON files? Do you use some tools?

Liamsi:
> I guess we need to add a nil vote in those JSON test cases, but I don't know how do you produce those JSON files? Do you use some tools?

@Shivani912 do you mind to chime in here?

Shivani912:
> > I guess we need to add a nil vote in those JSON test cases, but I don't know how do you produce those JSON files? Do you use some tools?
> 
> @Shivani912 do you mind to chime in here?

I can add that test here :)

yihuang:
Rebased on current master, and fixed the unit test skipping logic, which ignored many test cases.

tendermint/tests/lite.rs
```diff
@@ -237,7 +237,7 @@ fn run_single_step_tests(dir: &str) {
             .any(|failing_case| fp_str.ends_with(failing_case))
         {
             println!("Skipping JSON test: {}", fp_str);
-            return;
+            continue;
```
Liamsi, line 240: Ouch 🤦 thanks!

Issue #230, opened 2020-04-21
[light-client] Light Client Runtime


brapse:
The light client currently exposes the function `verify_bisection` to verify new header and update the trusted state according to the [verificaiton spec](https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/verification.md). Users such as the [Relayer](https://github.com/informalsystems/ibc-rs/blob/master/relayer/cli/src/commands/start.rs#L90) execute the `verify_bisection` function in a loop to maintain the latest trusted state. 

It would be better for the for light client users to have a consistent runtime and be exposed to light client state changes via events. This way enrichments to the light client protocol (for instance the addition of fork detection) could translate simply to the consumption of new events without changes to the overall runtime behaviour.

A potential design has already been [sketched out](https://github.com/brapse/light-spike/blob/master/src/main.rs#L100) and just needs to be integrated. This work is related to #229 and together will be the natural evolution of the ideas suggested in [ADR-006](https://github.com/informalsystems/tendermint-rs/blob/eab42ec28339a482627e4f00bd583e13bad82c3b/docs/architecture/adr-006-light-client-concurrency.md).

Issue #229, opened 2020-04-21
[light-client] Core verification as predicates


brapse:
The light client core verification [spec](https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/verification.md) Specifies the logic in which a header and validator set is verified by a trusted state. The current implementation depends on various types (Commit, ValidatorSet, Header, etc) that are abstracted over using traits for testing. Mock implementation of those traits are used avoid complex cryptographic functions that obscure the core predicates being tested.
The the current formulation has two problems:
1. It requires trait bounds be used everywhere 
2. It requires the construction of mock types to trigger predicates instead of just mocking out specific predicates for testing.

The goal then is to simplify the types used for core verification in two ways:

It would make sense to change the interface of the core verification to be specified as predicates instead of types. Predicates could then be mocked out with fixed point true or fixed point false to test complex verification sequences. 

Cryptographic functions should be taken out of core types put into separate `Calculator` objects which expose interfaces which can be mocked. The core verification can then use dependency injection of `Calculator` objects to keep cryptographic primitives out of the light client package.

The combination of these two changes will allow depending on simple types instead of abstract trait implementation and significantly reduce the amount type specification.

Current implementation underway: #225.

Issue #174, opened 2020-03-11
Light client: iterative / non-recursive bisection


Liamsi:
We should implement a non-recursive version of the bisection. 

The current recursive bisection logic in the (light client) verifier module was a good start towards an MPV light node but #171 surfaced that we should probably change this into a non-recursive method rather sooner than later. This could be part of general overhaul of the currentl light module that further decouples and separates concerns.

gterzian:
Ok, maybe that's a good one for me to look into, to get started on something...

gterzian:
I'm thinking of doing this as an actual `Stream`, which would produce one `TrustedState` at a time and could be used in iteration. 

cc @brapse It was mentioned you had been thinking about this and had some ideas?

In any case, I should have something to look at fairly quickly. 

Another way to do this would be to have a separate "verification" task(which doesn't mean it would actually have to run in parallel), which could write one `TrustedState` at a time into a channel. 

It's kind of the same thing, implementing a `Stream` vs having a separate task write into a channel(which would implement `Stream`)...

gterzian:
Ok the WIP is over at https://github.com/interchainio/tendermint-rs/pull/180

So this was actually much more complicated than I thought so far, for a couple of reasons:

1. Too much use of generics(in my opinion). 
2. Too much use of generic parameters, as opposed to trait objects(in my opinion).

So for example having `Header`, `Commit`, `ValidatorSet` as traits, as opposed to simple structs, seems unnecessarily complicated. It makes it harder to refactor something because not all traits are boxable. Fore example `ValidatorSet` is not, I'm not even sure why(maybe because of the `Clone` bound?), but it's one of those problems that are best avoided rather than solved. 

It also seems easy enough for the `RPCRequester` to simply return a `Set`, since it imports stuff from `tendermint` already(I guess it does make sense for tendermint to be using a `Requester` trait, and that trait can simply return a struct from within `tendermint` I would say). 

In terms of moving code around, it would be much more convenient to use structs at this stage, and only add traits later to add boundaries. Right now the complexity cost seems not worth it.  

Also, and this is perhaps personal, I much rather like to think as traits as objects, and pass them around as `Box<dyn Trait>`, as opposed to having all those parametrized functions. At the end of the day those functions use the traits passed to them mostly as objects it seems. But just passing a struct when you can is off-course superior than a trait.  A function with a signature like `verify_bisection_inner<'a, H, C, L, R>` seems like a bit too much...

[...]

PR #237, opened 2020-04-24
Light Client refactoring

romac:
Closes: #230 
Closes: #229 
Closes: #174

Here's take 2 of the light client spike.

This version follows pretty much the same decomposition as in the ADR (minus the Demuxer for lack of time), and I find It much more readable and understandable than the former one. It also drops the predicate library in favor of plain functions returning Result<(), Error> packed in a trait.

The big downside currently is that the components are now coupled (ie. the Scheduler has a reference to the Verifier and the Rpc components).

I haven’t found a way to decouple them without having to extract some of the logic into what would be the Demuxer router/event-loop.

I believe that the only way to get both decoupling + simple control-flow would be to go with async/await and have components communicate via channels with the Demuxer, and wait for specific responses. I will try to sketch something about that this weekend so that we can discuss both versions on Monday.

---

* [ ] Wrote tests
* [ ] Updated CHANGES.md


brapse:
Just to ensure that I understand what we are trying to solve here. The previous [light-spike](https://github.com/brapse/light-spike/blob/master/src/main.rs#L70) defined the light-client process as an iterator where each iteration would handle a single event processed by a single component and multiple iterations were necessary to complete a “flow”. This granular process made it very testable but hard to read as your brain had to keep track of multiple iteration to see the whole flow.

This is definitely a valid criticism. We defined it inductively which can be a challenge to read. If it takes three iterations through three different components to get anything meaningful done then we constantly have three things in our head.

In this version we get rid of the iterative nature and just call a function which passes control through the different components. It seems similar although clearer than the current version on master. As you mentioned, we keep the components coupled here which I would say is a significant downside.

I don’t know if async would help here or if we need more complex channel communication. Instead I think we just need to solve the granularity problem directly. With a demuxer (as outlined in ADR-006)[https://github.com/informalsystems/tendermint-rs/blob/brapse/adr-light-client-concurrency.md/docs/architecture/adr-006-light-client-refactor.md]  which handles state as well routing control between components. This way we can have clear separation of concerns and execute entire flows as a single synchronous function calls. 

image

I hope this approach might be the best of both worlds in terms of testability and readability while using minimal language features.

romac:
Here's take 3: https://github.com/informalsystems/tendermint-rs/tree/romac/light-spike-chan/light-spike

This version uses async channels to decouple components, but can still be run synchronously on a single thread.

Doing this gave me an idea for decoupling components without relying on async-await, so expect yet another take on this soon.

romac:
Take 4: https://github.com/informalsystems/tendermint-rs/tree/romac/light-spike-sync-decouple/light-spike

This version removes all async/await from take 3, and replaces the `BiChan` with a `Router` trait which can then be passed to the various components for them to call when they need to query other components. This trait is implemented for the Demuxer, but could be implemented for a standalone struct for the purpose of mocking/testing/etc.

romac:
Just updated this PR to [take 4](https://github.com/informalsystems/tendermint-rs/pull/237#issuecomment-620201348).

light-spike/src/components/demuxer.rs
```diff
@@ -0,0 +1,245 @@
...
+            } else {
+                self.state.untrusted_store_writer.add(current_block.clone());
+            }
+
+            self.state.trace_block(target_height, current_block.height);
+
+            let schedule = self.schedule(&current_block, &trusted_state, verification_result);
+            dbg!(&schedule);
+
+            match schedule {
```
brapse, line 140: Here it seems like we are are sending invalid block to the scheduler just to them to come back as scheduler events. For instance `VerifierOutput::Invalid(e)` maps to `SchedulerOutput::InvalidLightBlock(e)` but really the validity of the block should be in the verifier domain and not the scheduler domain. Would it make sense in this case to just match the verifier output directly here? Something like:


```rust
 match verification_result {
    VerifierOutput::Success => {
        self.state.trusted_store_writer.add(current_block.clone());
        continue
    },
    VerifierOutput::Invalid(e) => {
        self.state.untrusted_store_writer.add(current_block.clone());
        bail!(ErrorKind::InvalidLightBlock(e));
    },
    VerifierOutput::NotEnoughTrust => {
        self.state.untrusted_store_writer.add(current_block.clone());
        let schedule = self.schedule(&current_block,
            &trusted_state,
            verification_result);

        dbg!(&schedule);

        match schedule {
            SchedulerOutput::NextHeight(height) if height <= trusted_state.height => {
                bail!(ErrorKind::BisectionFailed(target_height, height));
            }
            SchedulerOutput::NextHeight(height) => {
                postcondition!(contracts::schedule::postcondition(
                    &trusted_state,
                    target_height,
                    height,
                    &self.state.trusted_store_reader,
                    &self.state.untrusted_store_reader
                ));

                next_height = height;
            }
        }
    }
}
```

Maybe there is a bit more nesting but it keeps the logic somehow together and decouples the scheduler and verifier domains. Just an idea 🤔

melekes:
Should we close this in favor of `romac/light-spike-sync-decouple` branch? Maybe also open a draft PR so it's easy to review?

romac:
## TODO before merging

- [ ] Fix merge conflicts
- [ ] Adapt to new JSON tests structure (#273)
- [ ] Fix performance issue with Sled store
- [ ] Add light client CLI example

## Follow-up

- [ ] Write documentation
- [ ] Tag code as per the traceability proposal

light-client/src/components/io.rs
```diff
@@ -0,0 +1,111 @@
...
+    F: FnMut(PeerId, Height) -> Result<LightBlock, IoError>,
+{
+    fn fetch_light_block(&mut self, peer: PeerId, height: Height) -> Result<LightBlock, IoError> {
+        self(peer, height)
+    }
+}
+
+pub struct ProdIo {
+    rpc_clients: HashMap<PeerId, rpc::Client>,
+    peer_map: HashMap<PeerId, tendermint::net::Address>,
```
brapse, line 34: Nice to have the abstraction of peerIDs here from the onset 👍

light-client/src/components/scheduler.rs
```diff
@@ -0,0 +1,80 @@
...
+        target_height: Height,
+    ) -> Height {
+        self(light_store, next_height, target_height)
+    }
+}
+
+#[pre(light_store.latest(VerifiedStatus::Verified).is_some())]
+#[post(valid_schedule(ret, target_height, next_height, light_store))]
+pub fn schedule(
+    light_store: &dyn LightStore,
```
brapse, line 34: Why do we need to pass the whole store if all we need is the latest trusted height?

romac: This is to match the spec, as @josef-widder suggested the scheduler could perform some optimizations as to what height to verify next based on what headers of nearby heights are already available in the store, but this is not implemented yet.

light-client/src/tests.rs
```diff
@@ -66,3 +66,77 @@ impl Trusted {
         }
     }
 }
+
+// -----------------------------------------------------------------------------
+// Everything below is a temporary workaround for the lack of `provider` field
+// in the light blocks serialized in the JSON fixtures.
+// -----------------------------------------------------------------------------
```
Liamsi, line 73: cc @greg-szabo @Shivani912  this is relevant to testing / serialization

light-client/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+derive_more = "0.99.5"
+serde = "1.0.106"
+serde_derive = "1.0.106"
+thiserror = "1.0.15"
+futures = "0.3.4"
+tokio = "0.2.20"
+prost-amino = "0.5.0"
+contracts = "0.4.0"
+sled = "0.31.0"
+serde_cbor = "0.11.1"
```
Liamsi, line 20: Yet another binary encoding 😱 😄 I assume this is the most reasonable choice to use in combination with sled?

light-client/src/components/scheduler.rs
```diff
@@ -0,0 +1,80 @@
+use crate::prelude::*;
+use contracts::*;
+
+#[contract_trait]
+pub trait Scheduler {
+    #[pre(light_store.latest(VerifiedStatus::Verified).is_some())]
+    #[post(valid_schedule(ret, target_height, next_height, light_store))]
```
Liamsi, line 7: Cool, I didn't know about the contract trait. This looks really neat! 👍

light-client/src/light_client.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// - If the precondition is violated [LVC-PRE-TP.1]
+    /// - If the core verification loop invariant is violated [LCV-INV-TP.1]
+    /// - If verification of a light block fails
+    /// - If it cannot fetch a block from the blockchain
+    #[post(
+        ret.is_ok() ==> trusted_store_contains_block_at_target_height(
+            self.state.light_store.as_ref(),
+            target_height,
+        )
+    )]
```
Liamsi, line 121: ❤️

light-client/src/macros.rs
```diff
@@ -0,0 +1,17 @@
...
+
+/// Ensure a condition holds, returning an error if it doesn't (ala `assert`).
+#[macro_export]
+macro_rules! ensure {
+    ($cond:expr, $kind:expr) => {
+        if !($cond) {
+            return Err($kind.into());
+        }
+    };
+}
```
Liamsi, line 17: Just noting, anomaly, which seems to be already used throughout this module, provides almost the same macros:
https://github.com/iqlusioninc/crates/blob/develop/anomaly/src/macros.rs#L34-L44

Maybe there are good reasons to define your own here. Not sure.

light-client/Cargo.toml
romac: I have to admit I didn't give it too much thought. I initially considered just serializing keys and values to JSON and then to bytes, but that seemed a bit wasteful, so I figured using a dedicated binary encoding was better. In the end, aside from the extra dependencies, the choice of encoding does not matter since it is internal to the light store and specific to the choice of `sled` as a database. But I'd be happy to discuss alternatives, either for the binary encoding, or for `sled` :)

light-client/src/macros.rs
romac: Yeah I saw those, but since we are using `thiserror` we do not need to specify an additional message for the error, as it will pick up the one defined in the `ErrorKind` Display implementation via the `#[error]` attribute. But if you see valuable information we could append to the error message via this extra argument in the original `anomaly` macros, let me know and I'd gladly update the code and get rid of the custom macros.

light-client/src/tests.rs
Shivani912: This seems to be implementation specific to me. Go code has a different approach to identify a node as a primary/secondary provider and does not deal with addresses. But also, this approach looks cleaner and reduces a lot of data from test files. Will need some more work around testing of Go code though. Need more thoughts here! @greg-szabo @Liamsi @melekes

light-client/Cargo.toml
ebuchman: Not a big deal but couldnt prost work here for proto3?

Issue #247, opened 2020-05-02
Serialization refactor


greg-szabo:
The `serializers` library started to become a bit unruly:
* The naming convention is not standardized
* Deserializer functions for `Option<T>` versions of different types were created when not necessary

The serializers library should focus on implementing custom ser/de for primitives and complex types that are not implemented by the source code. Things like `i64` to `String` or `time::Duration`` to a String that holds the duration in nanoseconds.
These should be implemented as functions that can be used in annotation (`with=` or `deserialize_with=`.)

Custom implementation for custom types should reside with the type and not in the serializers library. (For example `CommitSig`.)

Together with putting everything in its place, the custom types that need deserialization should deserialize into a type that's easy to use for the developer. The Go types are not necessarily the best mirrored in Rust. Find the Rust-native implementations and make sure that custom serialization can make it compatible with the specification. (Try to separate the issue of serialization from the daily use of the type.)

greg-szabo:
I'm breaking up PR #248 into multiple, easier-to-handle PRs.

Liamsi:
@greg-szabo I think all issues mentioned in the opening comment are addressed for now? And there is also a high-level serialization issue about the different kinds of serialization and overall structure #197. Feel free to re-open if I'm missing sth here.

Issue #246, opened 2020-05-02
commit_sig absent blockID fails validation (if validator address is present)


greg-szabo:
### Bug description
[ADR-025](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-025-commit.md) decision states:
```
Absent Votes: we include absent votes explicitly with no Signature or Timestamp but with the ValidatorAddress.
```

But the current Rust implementation of commit_sig defines a [validate function that fails](https://github.com/informalsystems/tendermint-rs/blob/2f29d7af242662567355c0f3cd6790598865b833/tendermint/src/lite_impl/signed_header.rs#L107
) with error if the BlockID is "FlagAbsent" and a validator address is present.

Although the neccessity of the validator address is questioned in several places, the current decision is to include it in CommitSigs. So, the Rust implementation should follow that.

### Proposed solution
Rewrite the CommitSig in a simpler form (possibly as an enum) and abstract away the serialization and type validation. (I'm planning to give an example of that soon.)

Short term fix: remove the check in the `validate` function.

greg-szabo:
Here's a passing test to show the issue:

```
#[test]
fn deserialize_commit_sig_absent_vote() {
    let incoming = r#"
    {
        "block_id_flag": 1,
        "validator_address": "",
        "timestamp": "1970-01-01T00:00:00Z",
        "signature": null
    }
    "#;

    let result: CommitSig = serde_json::from_str(&incoming).unwrap();

    assert_eq!(result.block_id_flag, tendermint::block::commit_sig::BlockIDFlag::BlockIDFlagAbsent);
    assert_eq!(result.validator_address, None);
    assert_eq!(result.timestamp, tendermint::time::Time::unix_epoch());
    assert_eq!(result.signature, None);
}
```
Validator address should not be empty and timestamp should be null or omitted.

Liamsi:
> But the current Rust implementation of commit_sig defines a validate function that fails with error if the BlockID is "FlagAbsent" and a validator address is present.
>
> Although the neccessity of the validator address is questioned in several places, the current decision is to include it in CommitSigs. So, the Rust implementation should follow that.

Note that while the ADR clearly states the val addr should be included as you say, the go implementation currently does this:
https://github.com/tendermint/tendermint/blob/e4427b72928e5a8066f75cdc98e3d84a038cc17b/types/block.go#L539-L548

[...]

PR #248, opened 2020-05-02
Serialization improvements

greg-szabo:
#247 - Serialization refactor
#246 - CommitSig bug fix

* [X] Wrote tests
* [X] Updated CHANGES.md


tendermint/src/block/commit_sig.rs
```diff
@@ -1,78 +1,182 @@
...
+    },
+    /// voted for nil.
+    BlockIDFlagNil {
+        /// Validator address
+        validator_address: account::Id,
+        /// Timestamp of vote
+        timestamp: Time,
+        /// Signature of vote
+        signature: Signature,
+    },
```
Liamsi, line 34: Representing this as an enum is a nice idea 👍

tendermint/src/block/commit_sig.rs
```diff
@@ -1,78 +1,182 @@
...
+                                Err(A::Error::custom(format!(
+                                    "signature is present for BlockIDFlagAbsent CommitSig {:?}",
+                                    incoming.signature.unwrap()
+                                )))
+                            } else {
+                                Ok(CommitSig::BlockIDFlagAbsent {
+                                    validator_address: incoming.validator_address,
+                                })
+                            }
+                        }
```
Liamsi, line 129: Could also be written as:
```suggestion
                       if incoming.timestamp.is_some() {
                            Err(A::Error::custom(format!(
                                "timestamp is present for BlockIDFlagAbsent CommitSig {}",
                                incoming.timestamp.unwrap()
                            )))
                        }
                        if incoming.signature.is_some() {
                            Err(A::Error::custom(format!(
                                "signature is present for BlockIDFlagAbsent CommitSig {:?}",
                                incoming.signature.unwrap()
                            )))
                        }
                        Ok(CommitSig::BlockIDFlagAbsent {
                            validator_address: incoming.validator_address,
                        })
```

tendermint/src/block/commit_sig.rs
```diff
@@ -1,78 +1,182 @@
...
+                    // Error: unknown CommitSig type
+                    e => Err(A::Error::custom(format!(
+                        "unknown BlockIdFlag value in CommitSig {}",
+                        e
+                    ))),
+                }
+            }
+        }
+
+        deserializer.deserialize_map(CommitSigVisitor)
```
Liamsi, line 180: While the enum idea is cool, it seems to come with the cost of a rather complicated `Deserialize` impl for `commitSig`.

tendermint/src/block/commit_sig.rs
greg-szabo: If Clippy says so, it needs to happen :)

tendermint/src/block/commit_sig.rs
greg-szabo: Yes and that's the tradeoff @brapse was going to go for in an earlier discussion. (Feel free to engage Sean.) The idea is that serialization is not actually the core product here and during the usual development it's just annoying to do something some particular way just because serialization says so.
This approach puts the logic of making sure that the incoming data is deserialized properly (with some basic validation) and the outgoing data is serialized properly in the serialization "layer" and the internal workings of the application can become completely independent from it. For example it can make use of the additional features of Rust better.
Just to highlight it even more: if/when the Tendermint data model changes because the Go + Rust teams decide on some other implementation, the changes can be implemented in this layer without affecting the application.

greg-szabo: I have to admit there is _one_ not-completely-Rust-native-best-practice feature in the code and it annoys me slightly. I deserialize the incoming CommitSig into an internal CommitSigSpec structure that is defined in ADR-025. I could have gone hard-core and implement complete deserialization while parsing the incoming text. But that would be even more convoluted and harder to change down the road. I feel this is a good compromise that allows us to easily update the deserialization code, if a new ADR comes by.

greg-szabo:
I'm going to look into the clippy messages and fix them. While I do that I also wanted to call out for @Shivani912 and ask her about some of the test-stable tests that are failing. (single_skip_commit_tests_verify and commit_tests_verify). I feel they are failing because the used JSON input is not compliant with the slightly stricter implementation this new CommitSig has: according to ADR-025, if a signature has block_id_flag = 1 (no vote cast), then there should be no signature and timestamp fields. Some of the JSON files have them, hence the above two tests fail.

Shivani, I'm happy to fix those in the JSON files, but I'm worried that I'm not editing the source of truth. Are you still copying these files from your own repository? Is it ok if I update them here?

tendermint/tests/integration.rs
```diff
@@ -34,7 +34,7 @@ mod rpc {
     async fn abci_info() {
         let abci_info = localhost_rpc_client().abci_info().await.unwrap();
 
-        assert_eq!(&abci_info.version, "0.16.2");
+        assert_eq!(&abci_info.version, "0.17.0");
```
Liamsi, line 37: Good, this will also close: https://github.com/informalsystems/tendermint-rs/issues/249

tendermint/src/block/commit_sig.rs
Liamsi: > Yes and that's the tradeoff @brapse was going to go for in an earlier discussion. (Feel free to engage Sean.) The idea is that serialization is not actually the core product here and during the usual development it's just annoying to do something some particular way just because serialization says so.

That is absolutely reasonable. Serialization isn't the core product and we shouldn't build our types around it. All in favour of decoupling serialization concerns and core types.

What I was referring to was that the seemingly beautiful change made the part where the serialization actually happens difficult to read and quite large. A simpler way might be to have a serialization type and a `From` / `Into` for that and the actual type. This is very similar to what you are already doing with the `CommitSigSpec` type but simpler to parse for a human :-)

Liamsi:
Just a remark regarding the JSON tests: some might fail because now basic validation is done while decoding the type (e.g. the relation between BlockIdFlag and absence or presence of certain fields). 

This is a good move (as we can fail earlier on invalid input) but might require changes in the JSON test because that would mean that some test-files won't parse anymore. Currently, they would parse correctly but fail later (when validation is done). This is a sign that the JSON tests might be to narrowly built around the golang version: there the JSON in a few cases might parse correctly and the validation would fail later (and this is implicitly expected as an error in the JSON test description).

Shivani912:
> I'm going to look into the clippy messages and fix them. While I do that I also wanted to call out for @Shivani912 and ask her about some of the test-stable tests that are failing. (single_skip_commit_tests_verify and commit_tests_verify). I feel they are failing because the used JSON input is not compliant with the slightly stricter implementation this new CommitSig has: according to ADR-025, if a signature has block_id_flag = 1 (no vote cast), then there should be no signature and timestamp fields. Some of the JSON files have them, hence the above two tests fail.
> 
> Shivani, I'm happy to fix those in the JSON files, but I'm worried that I'm not editing the source of truth. Are you still copying these files from your own repository? Is it ok if I update them here?

I'll fix this on the generator code and update the tests in this repo. But, looks like this should be reported for the the Go code. A `CommitSig` with `BlockIDFlagAbsent` is expected to have all other fields "empty" including the `ValidatorAddress` field (this check [here](https://github.com/tendermint/tendermint/blob/b7c2d7a977674e3eb62aaf4f7f5ef35f664cb091/types/block.go#L539)). 

Thanks for finding this @greg-szabo !

greg-szabo:
Yes, I'm going to report that on the next Tendermint call. The Go code doesn't comply with the ADR.

tendermint/src/block/commit_sig.rs
greg-szabo: Done and done: Today's commit moved things into serde's "from" annotation. So now, the code looks a bit cleaner, although it's not less code:
In the serializers, a new RawCommitSig type was introduced that deserializes from JSON and it looks very much like the definition in the ADR (including the enum for BlockIDFlags). Deserialization automatically invokes basic type-checking of the struct (like an u8 has to be an u8).
The Rust-native CommitSig type converts from this RawCommitSig type through it's "From" trait which holds the validation for the struct. This logic is more about "if this variable is set, than that other variable also has to have a value".

So the deserializer does type-check and the From trait does validation. My only grief is that type-checking happens in the serializers library while validation happens in the struct's library (in this case at CommitSig). Interestingly, this is similar to how it used to be, so I'm guessing the developers will be fine with it.

Liamsi: If the code looks cleaner that is already worth it 👍  And I think it is OK for the core type to defines when it's valid (s.t. you can only construct a valid type).

Liamsi:
A general remark: I think this PR could have better been 2, 3 or even 4 PRs: 
 1) CommitSig bug fix
 2) serilization refactor:
 2.1) smaller repetitive changes like cleaning up all those `#[serde()]` annotations
 2.2) introduction of that `ByteStringType`
 2.3) spliiting up existing serializers into seperate modules/files

Liamsi:
The failing tests (on stable)  is because https://github.com/informalsystems/tendermint-rs/pull/248#issuecomment-623036274
Previously this would parse but return an error on validation, now this doesn't parse anymore (and because of the panic none of the tests in that json file are run):
```
thread 'commit_tests_verify' panicked at 'called `Result::unwrap()` on an `Err` value: Error("timestamp is present for BlockIDFlagAbsent CommitSig", line: 1343, column: 16)', tendermint/tests/lite.rs:138:9
```

.github/workflows/rust.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           args: --all-features --no-fail-fast
         env:
           CARGO_INCREMENTAL: "0"
-          RUSTFLAGS: "-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Zno-landing-pads"
+          RUSTFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Cpanic=unwind -Zpanic_abort_tests'
```
Liamsi, line 97: Why not fix this on master, get that PR merged in quickly and merge master into this PR? It would benefit all other opened PRs in the meantime.

greg-szabo: I'm a "one-branch" guy :D
Sure, that makes sense.

Shivani912:
@greg-szabo  Another note on `CommitSig` with `BlockIDFlagAbsent`:

An empty `timestamp` field (i.e. of type time.Time) in Tendermint Go is by default "zero" time i.e. "0001-01-01T00:00:00Z". This means that in case of an empty CommitSig, there will always be `timestamp` field with this zero time, unless they change the way it serializes.

greg-szabo:
I thought they'd serialize it as a `null`... They do that ll the time. :)

greg-szabo:
Regarding managing the PR: I was thinking on doing that but all the PRs would have depended on each other (or at least in some cases it would have). Is that ok?

Also, the work is not done yet, I just chose a bigger milestone instead of small changes (which I'll try next time). There is a "huge" TODO in the middle of the serializers code that will refactor at least the Header struct and possibly the Block too.

Liamsi:
> I thought they'd serialize it as a null... They do that all the time. :)

https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r401027989
I quote @ebuchman here (again): 
> Serialization issues have been a constant plague upon us.

Liamsi:
> Is that ok?

OK, but please do not add more to it at least:  

> Also, the work is not done yet, I just chose a bigger milestone instead of small changes (which I'll try next time). There is a "huge" TODO in the middle of the serializers code that will refactor at least the Header struct and possibly the Block too.

It quickly becomes painful to review and also the quality of the review decreases. Some of the changes do not depend on each other, e.g.: https://github.com/informalsystems/tendermint-rs/pull/248#discussion_r419779989
and 2.1) and the first half of 2.3) in https://github.com/informalsystems/tendermint-rs/pull/248#issuecomment-623749432. Also some of the changes could be done incrementally.

tendermint/src/lite_impl/signed_header.rs
```diff
@@ -72,87 +83,43 @@ impl lite::Commit for block::signed_header::SignedHeader {
...
-                }
-                if self.signature.is_some() {
-                    fail!(
-                        Kind::ImplementationSpecific,
-                        "signature is present for absent CommitSig {:#?}",
-                        self
-                    );
-                }
-                // TODO: deal with Time
-                // see https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r401027989
```
Liamsi, line 119: There was also a reminder about the time peculiarities :-)

tendermint/src/block/commit_sig.rs
```diff
@@ -1,78 +1,100 @@
...
-    /// Deserialize this type from a byte
-    pub fn from_u8(byte: u8) -> Option<BlockIDFlag> {
-        match byte {
-            1 => Some(BlockIDFlag::BlockIDFlagAbsent),
-            2 => Some(BlockIDFlag::BlockIDFlagCommit),
-            3 => Some(BlockIDFlag::BlockIDFlagNil),
-            _ => None,
+/// CommitSig implementation
+impl CommitSig {
+    /// Helper: Extract validator address, since it's always present (according to ADR-025)
```
Liamsi, line 41: We really should clarify this first: https://github.com/informalsystems/tendermint-rs/issues/246#issuecomment-622927309

I can very well be that adr isn't representing the full decision anymore, or, that it is slightly confusing nil and absent votes. Also: https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r399844023

tendermint/src/lite_impl/signed_header.rs
greg-szabo: Oh sh*t, I completely missed that. :( Thanks for the thorough review!

Should I add the fix on this PR?

The fix as I see is accepting Nil commits with timestamp and checkinf if the timestamp is 0. If yes, all's well (but don't put the timestamp in the enum), if no, raise an error.

yihuang:
I suggest adding symmetric assertions to the unit tests, something like:
```
serde_json::from_str(&serde_json::to_string(&obj).unwrap()).expect("Should be decodable");
``` 
Or even better the decoded one is the same as original one if it implements `Eq`:
```
assert_eq!(serde_json::from_str(&serde_json::to_string(&obj).unwrap()).unwrap(), obj);

Liamsi:
This is a great suggestion @yihuang. There is a little function added by @romac that can help here:
https://github.com/informalsystems/tendermint-rs/blob/7df19f85bcd364ee095120c58c0e88f049e500c5/tendermint/src/test.rs#L4-L10

tarcieri:
An interesting way to do those "round trip" tests is with a tool like [`proptest`](https://altsysrq.github.io/proptest-book/intro.html)

tendermint/src/block/commit_sig.rs
melekes: https://github.com/tendermint/tendermint/issues/4797#issuecomment-625069649

greg-szabo:
Ok, so I've broken out this PR into three smaller PRs as we discussed with Ismail. This one will be closed soon, I just want to save the additional things that came up into the issue (#247) until they are implemented/decided on.

tendermint/src/lite_impl/signed_header.rs
greg-szabo: Opened issue #259 about this.

tendermint/src/block/commit_sig.rs
greg-szabo: Created tracking issue #260 about this.

greg-szabo:
All right. Closing this PR.

* PR was broken up into three smaller PRs: #255, #257, #258.
* Additional ideas were broken out into new issues: #259, #260, #261

Issue #249, opened 2020-05-02
ABCI compatibility with 0.17.0


Liamsi:
Note that integration tests on master are failing again. Now with:
```
---- rpc::abci_info stdout ----
thread 'rpc::abci_info' panicked at 'assertion failed: `(left == right)`
  left: `"0.17.0"`,
 right: `"0.16.2"`', tendermint/tests/integration.rs:37:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

What is odd about this is that this fails although there was no new release yet (this is from pending):
> [abci] #4704 Add ABCI methods ListSnapshots, LoadSnapshotChunk, OfferSnapshot, and ApplySnapshotChunk for state sync snapshots. ABCIVersion bumped to 0.17.0.

https://github.com/tendermint/tendermint/blob/master/CHANGELOG_PENDING.md#v0335

While it is cool to be warned early on about changes to come it would also be good if there was a mechanism to stick to a particular tendermint release (e.g. currently we are aiming for compatibility with 0.33).

related: 
- #233 (https://github.com/informalsystems/tendermint-rs/pull/238#issuecomment-622940124)
- #29 (particularly: #29 (comment))
- #197

Liamsi:
Closing this for now. The current version bump was fixed by @greg-szabo in #256 
The other concerns around this (how do we keep up to date with abci changes and do we want to be warned about changes before releases and should we maybe fix integration tests to a particular version), will bubble up again soon enough.

PR #255, opened 2020-05-07
Serialization refactor step 1

greg-szabo:
- Refactor the serializer library. Move all functions into modules and rewrite the annotation to use `with`. (Looks nicer.)

This is a first step in refactoring serialization outloined in issue #247  .

* [X] Wrote tests
* [X] Updated CHANGES.md

Soft-depends on Issue #249 / PR #256 for integration test success.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,10 @@
 ## Pending
 
 Testing:
-- [#249] Updated abci_info test to 0.17.0
+- Updated abci_info test to 0.17.0 ([#249])
 
 Serialization:
-- [#248] Refactor serializers library to use modules and give a nicer annotation to structs.
+- Refactor serializers library to use modules and give a nicer annotation to structs. ([#248])
 
 ## [0.13.0] (2020-04-20)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,10 @@
 ## Pending
 
 Testing:
-- Updated abci_info test to 0.17.0 ([#249])
+- Updated abci_info test to 0.17.0 ([#249](https://github.com/informalsystems/tendermint-rs/issues/249))
 
 Serialization:
-- Refactor serializers library to use modules and give a nicer annotation to structs. ([#248])
+- Refactor serializers library to use modules and give a nicer annotation to structs. ([#248](https://github.com/informalsystems/tendermint-rs/issues/248))
 
 ## [0.13.0] (2020-04-20)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## Pending
+
+Serialization:
+- [#248] Refactor serializers library to use modules and give a nicer annotation to structs.
+
 ## [0.13.0] (2020-04-20)
 
 Dependencies:
--- a/tendermint/src/lib.rs
+++ b/tendermint/src/lib.rs
@@ -39,7 +39,7 @@ pub mod node;
 pub mod private_key;
 pub mod public_key;
 pub mod rpc;
-mod serializers;
+pub mod serializers;
 pub mod signature;
 pub mod time;
 mod timeout;
--- a/tendermint/src/rpc/endpoint/net_info.rs
+++ b/tendermint/src/rpc/endpoint/net_info.rs
@@ -30,7 +30,7 @@ pub struct Response {
     pub listeners: Vec<Listener>,
 
     /// Number of connected peers
-    #[serde(with = "serializers::primitives::string")]
+    #[serde(with = "serializers::from_str")]
     pub n_peers: u64,
 
     /// Peer information
@@ -69,7 +69,7 @@ pub struct PeerInfo {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct ConnectionStatus {
     /// Duration of this connection
-    #[serde(rename = "Duration", with = "serializers::timeduration::string")]
+    #[serde(rename = "Duration", with = "serializers::time_duration")]
     pub duration: Duration,
 
     /// Send monitor
@@ -97,43 +97,43 @@ pub struct Monitor {
     pub start: Time,
 
     /// Duration of this monitor
-    #[serde(rename = "Duration", with = "serializers::timeduration::string")]
+    #[serde(rename = "Duration", with = "serializers::time_duration")]
     pub duration: Duration,
 
     /// Idle duration for this monitor
-    #[serde(rename = "Idle", with = "serializers::timeduration::string")]
+    #[serde(rename = "Idle", with = "serializers::time_duration")]
     pub idle: Duration,
 
     /// Bytes
-    #[serde(rename = "Bytes", with = "serializers::primitives::string")]
+    #[serde(rename = "Bytes", with = "serializers::from_str")]
     bytes: u64,
 
     /// Samples
-    #[serde(rename = "Samples", with = "serializers::primitives::string")]
+    #[serde(rename = "Samples", with = "serializers::from_str")]
     samples: u64,
 
     /// Instant rate
-    #[serde(rename = "InstRate", with = "serializers::primitives::string")]
+    #[serde(rename = "InstRate", with = "serializers::from_str")]
     inst_rate: u64,
 
     /// Current rate
-    #[serde(rename = "CurRate", with = "serializers::primitives::string")]
+    #[serde(rename = "CurRate", with = "serializers::from_str")]
     cur_rate: u64,
 
     /// Average rate
-    #[serde(rename = "AvgRate", with = "serializers::primitives::string")]
+    #[serde(rename = "AvgRate", with = "serializers::from_str")]
     avg_rate: u64,
 
     /// Peak rate
-    #[serde(rename = "PeakRate", with = "serializers::primitives::string")]
+    #[serde(rename = "PeakRate", with = "serializers::from_str")]
     peak_rate: u64,
 
     /// Bytes remaining
-    #[serde(rename = "BytesRem", with = "serializers::primitives::string")]
+    #[serde(rename = "BytesRem", with = "serializers::from_str")]
     bytes_rem: u64,
 
     /// Time remaining
-    #[serde(rename = "TimeRem", with = "serializers::primitives::string")]
+    #[serde(rename = "TimeRem", with = "serializers::from_str")]
     time_rem: u64,
 
     /// Progress
--- a/tendermint/src/rpc/endpoint/net_info.rs
+++ b/tendermint/src/rpc/endpoint/net_info.rs
@@ -30,10 +30,7 @@ pub struct Response {
     pub listeners: Vec<Listener>,
 
     /// Number of connected peers
-    #[serde(
-        serialize_with = "serializers::serialize_u64",
-        deserialize_with = "serializers::parse_u64"
-    )]
+    #[serde(with = "serializers::primitives::string")]
     pub n_peers: u64,
 
     /// Peer information
@@ -72,11 +69,7 @@ pub struct PeerInfo {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct ConnectionStatus {
     /// Duration of this connection
-    #[serde(
-        rename = "Duration",
-        serialize_with = "serializers::serialize_duration",
-        deserialize_with = "serializers::parse_duration"
-    )]
+    #[serde(rename = "Duration", with = "serializers::timeduration::string")]
     pub duration: Duration,
 
     /// Send monitor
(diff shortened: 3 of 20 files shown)


greg-szabo:
Step 3 does that but unfortunately the original work had dependencies which would've made this into the "big PR" again, if I implemented it here.

tendermint/src/serializers.rs
```diff
@@ -1,71 +1,199 @@
...
-pub(crate) fn parse_i64<'de, D>(deserializer: D) -> Result<i64, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    String::deserialize(deserializer)?
-        .parse::<i64>()
-        .map_err(|e| D::Error::custom(format!("{}", e)))
-}
+/// Serialize/deserialize primitive types (i64, u64, etc)
+pub mod primitives {
```
Liamsi, line 44: Isn't `primitives` misleading? Although this is currently only used for primitive types, this is actually more general: `serializers::primitives::string` can be used for all types `T` that impl `FromStr` as far as I understand.

As this doesn't contain any submodules we could "flatten" this mod `serializers::string::from_str`?

tendermint/src/serializers.rs
```diff
@@ -1,71 +1,199 @@
...
+        pub fn deserialize<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+        where
+            D: Deserializer<'de>,
+            T: std::str::FromStr,
+            <T as std::str::FromStr>::Err: std::fmt::Display,
+        {
+            String::deserialize(deserializer)?
+                .parse::<T>()
+                .map_err(|e| D::Error::custom(format!("{}", e)))
+        }
```
Liamsi, line 60: Is it only me or was it surprising to you as well that serder doesn't provide this?

CHANGES.md
```diff
@@ -1,7 +1,11 @@
-# Pending
+## Pending
 
+Testing:
 - [#249] Updated abci_info test to 0.17.0
 
+Serialization:
+- [#248] Refactor serializers library to use modules and give a nicer annotation to structs.
```
Liamsi, line 7: Add link like in:

https://github.com/informalsystems/tendermint-rs/pull/255/files#diff-8b1c3fd0d4a6765c16dfd18509182f9dR11-R16

tendermint/src/lib.rs
```diff
@@ -39,7 +39,7 @@ pub mod node;
 pub mod private_key;
 pub mod public_key;
 pub mod rpc;
-mod serializers;
+pub mod serializers;
```
Liamsi, line 42: Do we want to export our serializers? IMO it's better to keep it unexported if possible.

tendermint/src/rpc/endpoint/net_info.rs
```diff
@@ -104,83 +97,43 @@ pub struct Monitor {
     pub start: Time,
 
     /// Duration of this monitor
-    #[serde(
-        rename = "Duration",
-        serialize_with = "serializers::serialize_duration",
-        deserialize_with = "serializers::parse_duration"
-    )]
+    #[serde(rename = "Duration", with = "serializers::timeduration::string")]
```
Liamsi, line 100: Do we anticipate other ways to encode/decode `std::time::Duration`? If not, the extra string sub-module isn't necessary I think (and this could be `serializers::timeduration`).

tendermint/src/serializers.rs
```diff
@@ -1,71 +1,199 @@
...
+    }
 }
 
-/// Parse `Duration` from a JSON string containing a nanosecond count
-pub(crate) fn parse_duration<'de, D>(deserializer: D) -> Result<Duration, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    // TODO(tarcieri): handle 64-bit overflow?
```
Liamsi, line 53: Is this handled by the new code or is nothing to be concerned about?

tendermint/tests/serialization.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[serde(with = "serializers::bytes::hexstring")]
+    myhexbytes: Vec<u8>,
+
+    #[serde(with = "serializers::bytes::base64string")]
+    mybase64bytes: Vec<u8>,
+
+    #[serde(with = "serializers::bytes::string")]
+    stringifiedbytes: Vec<u8>,
+}
+
```
Liamsi, line 33: Do you want merge all the  `_from_` and `_into_` test pairs into a single test respectively? e.g. using the roundtrip function in this repo (https://github.com/informalsystems/tendermint-rs/pull/248#issuecomment-624605718)? This would ensure we can go from json -> struct -> json and all matches expectations.

tendermint/src/serializers.rs
```diff
@@ -1,71 +1,199 @@
...
+//! Available serializers:
+//! i64                  <-> string:               #[serde(with="serializers::primitives::string")]
+//! u64                  <-> string:               #[serde(with="serializers::primitives::string")]
+//! std::time::Dureation <-> nanoseconds as string #[serde(with="serializers::timeduration::string")]
+//! Vec<u8>              <-> HexString:            #[serde(with="serializers::bytes::hexstring")]
+//! Vec<u8>              <-> Base64String:         #[serde(with="serializers::bytes::base64string")]
+//! Vec<u8>              <-> String:               #[serde(with="serializers::bytes::string")]
+//!
+//! Any type that has the "FromStr" trait can be serialized into a string with serializers::primitives::string.
+//!
```
Liamsi, line 36: Thanks for the extensive and well written documentation! 👍

tendermint/src/serializers.rs
greg-szabo: In a later step I updated the documentation to say "you can use this for anything that implements FromStr". So, you are right in your assumption.
I wanted to keep it embedded like this for convenience reasons for the developer: if you check the documentation, all built-in types are deserialized using "serializers::<Rust_nickname>::<JSON_object type>", or "serialize from A to B". If you flatten the modules (the modules don't really add any other additional value), it will work, but you lose this consistency.

I leave it up to you. If you say flatten it, I'll do it.

tendermint/src/serializers.rs
greg-szabo: I was lucky, because I got to read Tony's code first. So I was prepared for it and I just chalked it up as yet another Rust curiosity. :)

tendermint/src/lib.rs
greg-szabo: It's exported so the tests, in a completely different package can access them. I wasn't going to originally, but that's how our testing is arranged.

tendermint/src/rpc/endpoint/net_info.rs
greg-szabo: I have no idea, that's why I just kept the above mentioned "serialize from A to B" type of notion. Again, adding more modules just helps with the `::` annotation, it doesn't serve any other purpose. I'd say, if we have a list of annotation names you'd like to see, I'll just rename everything to it.

tendermint/src/serializers.rs
Liamsi: hmm, that makes sense. For someone who usually does trust code more than documentation, primitives is still a bit misleading though (as you can use it on any type basically). If you really want to keep the module structure as is, then maybe `serializers::from_str::string` is still similar enough to: `"serializers::<Rust_nickname>::<JSON_object type>"``?

tendermint/src/lib.rs
Liamsi: We are basically committing to a public API here (which I don't think we should in this case). We should keep that surface small (as devs using this library will rely on it). 

Either we need to limit the visibility to where it is necessary. Alternatively the tests for a particular type could go as `mod test` into that type (I think that is preferred for simple unit test). See e.g. how `test_serialization_roundtrip` is used in other places.

tendermint/src/serializers.rs
greg-szabo: It's handled the same way, I think Tony left a comment here because he wasn't sure if it's going to cause a problem. (There's not much info about what he meant here.)

In both implementations, the deserialization fails if the incoming JSON object contains a bigger number that fits `u64`. Since it's in nanoseconds and the max value for u64 is `18446744073709551615`, the max number representable is over 584 years. Since BFT time can be faster than real time, this is not really until 2554, it can happen earlier. If/when it happens, my guess is that the chain halts.

Simple workaround is to use u128 in our code but I have a feeling this will be a problem in the Go code too.

Sounds like low priority for me. How about we open an issue about it?

tendermint/src/rpc/endpoint/net_info.rs
Liamsi: Not sure how I feel introducing a wrapping module just to keep the documentation consistent. Seems like an uncommon pattern to me (but I might be wrong).

tendermint/tests/serialization.rs
greg-szabo: In the case of nullable `Vec<u8>` we actually don't get the same object back if we do a roundtrip. `null -> [] -> ""` is what happens because of how we interpret the field. But the idea is good, I'd like to play around with it.

Shall I do it in this PR or open an issue?

greg-szabo: Note: the nullable `Vec<u8>` is implemented in the next serialization step. So it won't show up here. I'd prefer to rewrite the tests after this PR. I'll open an issue about it.

tendermint/src/serializers.rs
Liamsi: > Sounds like low priority for me. How about we open an issue about it?

OK

tendermint/tests/serialization.rs
Liamsi: Sounds good!

tendermint/src/serializers.rs
greg-szabo: To be honest, it should be something like `serializers::u64::string` and `serializers::i64::string` and so on, but that seemed overkill. Ok, I'll byte, let's rename.

Would you accept `serializers::from_str`? That kind of implies that you need a type that has a FromStr trait and it doesn't make any assumptions on what else that type is. It has one less module also. :)
And I'll rename the Time::Duration one to `serializers::timeduration` or `serializers::time_duration` or similar if you want.

The bytes thing still has multiple options to deserialize from. How about
`serializers::hexstring`
`serializers::base64string`
`seriliazers::bytestring` (that's the regular string to byte array serializer) ?

You'll have to implicitly know that you need to apply these to `Vec<u8>` but it's not bad if you can look at any example code.

tendermint/src/rpc/endpoint/net_info.rs
greg-szabo: I mentioned renaming it together with the other types in a comment above. Let's continue there.

tendermint/tests/serialization.rs
greg-szabo: Opened issue #261 to track implementing test pairs.

tendermint/src/serializers.rs
greg-szabo: Opened issue #262 about this.

tendermint/src/lib.rs
greg-szabo: Ok, so testing needs to move within the serialization library. I can refactor here or in a new PR. Which one would you prefer?

greg-szabo: Actually, since testing changes with the step 2 and step 3 stuff, I'd rather do it as step 4 in a new PR... I just have to make sure I do it before any releases...

Feels crappy to be honest. If we don't want to publicize the serializers, they shouldn't end up on the master branch. Opinion?

CHANGES.md
greg-szabo: The link didn't do anything for me, but I updated it based on previous ChangelLog entries. I hope that's what you had in mind.

tendermint/src/lib.rs
Liamsi: > Ok, so testing needs to move within the serialization library. I can refactor here or in a new PR. Which one would you prefer?

It doesn't really matter as long as we do it eventually.

> Feels crappy to be honest. If we don't want to publicize the serializers, they shouldn't end up on the master branch

Not sure I understand: I mean that we should make helper-functions used to deserialze actual types not public (only in this crate of where necessary). Why wouldn't these end up on the master branch? There are a bunch of private (or crate public structs/methods on master).

Users of tendermint-rs shouldn't need to care about deserialzation details. We should be careful what API we make public (as it is a commitment, people will end up using it and be annoyed if we break the API). 
And from the previous discussion it sounded the main motivation was the way we structured tests (which isn't a good reason to publish something that previously wasn't public).

Does that make sense?

CHANGES.md
Liamsi: Sorry, my bad. I though github will expand the lines but it doesn't. If you expand manually I see these lines highlighted:
image

I'm OK as long as this generates links (and looks ok when rendered). I don't feel strongly about the consistency regarding the style of links.

tendermint/src/serializers.rs
Liamsi: > Would you accept serializers::from_str? 
> serializers::time_duration 
> serializers::hexstring
> serializers::base64string
> seriliazers::bytestring 

Yes, sounds good to me!

tendermint/src/lib.rs
greg-szabo: Sorry, I overthought it. I understand that we should move the testing into the serialization library, so we don't have to publish the serialization library.
I was thinking about implementing this here and now: move the testing into the serialization library in this PR. But the next PR already open is based on this and that would also need an overhaul, up until the third PR is merged.

So instead, I propose that we open an issue that tracks this (make it high priority so I implement it before we cut a new release) and do it there.

greg-szabo: Opened issue #263 to track this.

tendermint/src/serializers.rs
greg-szabo: Uh, one last question: the step 3 PR based on this (well, on step 2, that's based on this step 1) breaks up the serializer.rs file into multiple files to support it better. If I do the rename here, I'm worried it won't go over well in that PR. (The curse of basing work on ongoing work.)

Option1: change it here, let's see the fallout and learn from it
Option2: we keep it as-is here, and the the change in the step 3 PR.

I'm fine doing it either way. What's your opinion?

Liamsi: As it is a simple renaming I can't anticipate a bigger problem when moving these into separate files. As you will have to try it anyways probably better to go with option1 now while it is still a smaller change.

greg-szabo: I can implement it later this afternoon. Don’t wait for it today, sorry. We’ll finish merging this when you’re online next time.

Liamsi: Thinking about it: `from_str` could also be just named `string` if you like (only one way it's `FromStr` the other it requires `fmt::Disply`). And if you want, we can keep the wrapping `bytes` module as it actually bundles something.

Liamsi: Wait, I'll open a quick PR against this branch with the first steps to take this off your plate and get a feel for the implications.

Liamsi: #264 (only string and timeduration)

greg-szabo: Thanks, that was awesome! I've reviewed it and merged.

Liamsi: 😊 Let's merge this and do any followup changes regarding the structure in the other step-PRs. My understanding is that some of the modules will change later anyways.

PR #257, opened 2020-05-07
Serialization refactor step 2

greg-szabo:
Implemented `null`-value deserialization for `Vec<u8>` using Visitor.

This is part 2 of the serialization improvement efforts described in #247 .

* [X] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [X] Updated all code comments where relevant
* [X] Wrote tests
* [X] Updated CHANGES.md

Issue #197, opened 2020-03-29
Serialization Concerns


ebuchman:
Serialization issues have been a constant plague upon us. 

At a high level, virtually all our data structures must serialize to/from both Amino and JSON.

The main data structures target JSON, while copies of the data structures in the amino_types module target Amino.

A lot of the difficulties come from peculiarities in Tendermint JSON encoding, especially around empty/missing/nil values. In particular:
- Sometimes Tendermint JSON uses `null`
- Sometimes it uses default values
- Sometimes it omits values completely (`omitempty`)

In certain cases, multiple fields in a struct are all present or all absent together. We should consider making such types enums with Present and Absent variants - the Present variant requires all the fields (except the genuinely Optional ones), and the Absent variant omits all the fields that are omitted together. This is for instance the case with the Block struct - if it's the first block, all the data pertaining to previous blocks is omitted, but if its any other block, it all must be there #101 .

We don't seem to currently have a coherent/standard way to deal with all this complexity. In some cases we use `Option` in structs. In some cases we use custom serialize/deserialize functions. In some cases we use custom types that implement their own serialize/deserialize methods. It would be worth attempting some kind of inventory of all this special casing and try to achieve more consistency and simplification.

While we are aiming for compliance with Tendermint Core v0.33, there may be cases where its worth recommending a change to Tendermint Core to make this more manageable.

Related:
- https://github.com/informalsystems/tendermint-rs/issues/184#issuecomment-604800369
- https://github.com/informalsystems/tendermint-rs/issues/187

Liamsi:
> Serialization issues have been a constant plague upon us.

So true, independent of this repo 😆

tarcieri:
A common technique I've seen elsewhere for dealing with encoding/decoding of complex types (for example, [in Cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/resolver/encode.rs)) is separating the "domain" types (i.e. the ideal ones you want to use) from the ones which handle encoding, then having `From`/`TryFrom` conversions between the two sets of types.

Liamsi:
In this context: instead of having currently two versions of a type (one for JSON another for amino/binary encoding), we could merge these two into one. 

Pros:
 - easier to maintain (if sth changes only one serialzation type needs to be adapted and not two) 
 - no serde/amino/encoding annotation in "ideal" type used for any core logic
- less duplication

Cons:
 - not as clear separation of domains (one type/module handles JSON and binary encoding) and related to this:
[...]

PR #269, opened 2020-05-13
Serialization refactor step 2

greg-szabo:
Implemented null-value deserialization for Vec<u8> using Visitor.

This is part 2 of the serialization improvement efforts described in #247 .

(This PR was opened after part 1 was merged. It was easier to open this than to keep the original #257)

This PR is ready for review.

* [X] Wrote tests
* [X] Updated CHANGES.md

### DIFF
--- a/tendermint/src/serializers.rs
+++ b/tendermint/src/serializers.rs
@@ -23,7 +23,6 @@
 //!
 //! }
 //! ```
-//! Note: null values will be deserialized into vec![].
 //!
 //! Available serializers:
 //! i64                  <-> string:               #[serde(with="serializers::primitives::string")]
@@ -33,8 +32,9 @@
 //! Vec<u8>              <-> Base64String:         #[serde(with="serializers::bytes::base64string")]
 //! Vec<u8>              <-> String:               #[serde(with="serializers::bytes::string")]
 //!
-//! Any type that has the "FromStr" trait can be serialized into a string with serializers::primitives::string.
-//!
+//! Notes:
+//! * Any type that has the "FromStr" trait can be serialized into a string with serializers::primitives::string.
+//! * serializers::bytes::* deserializes a null value into an empty vec![].
 
 use crate::account::{Id, LENGTH};
 use crate::{block, Hash, Signature};
--- a/tendermint/src/serializers.rs
+++ b/tendermint/src/serializers.rs
@@ -97,73 +97,9 @@ pub mod time_duration {
 
 /// Serialize/deserialize bytes (Vec<u8>) type
 pub mod bytes {
-    use serde::{
-        de::{Error, Visitor},
-        Deserialize, Deserializer,
-    };
-    use std::fmt;
-    use subtle_encoding::{base64, hex};
-
-    /// ByteStringType defines the options what an incoming string can represent.
-    enum ByteStringType {
-        Hex,
-        Base64,
-        Regular,
-    }
-
-    /// The Visitor struct to decode the incoming string.
-    struct BytesVisitor {
-        string_type: ByteStringType,
-    }
-
-    /// The Visitor implementation
-    impl<'de> Visitor<'de> for BytesVisitor {
-        type Value = Vec<u8>;
-
-        /// Description of expected input
-        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-            match self.string_type {
-                ByteStringType::Hex => {
-                    formatter.write_str("Hex-encoded byte-array in a String or null")
-                }
-                ByteStringType::Base64 => {
-                    formatter.write_str("Base64-encoded byte-array in a String or null")
-                }
-                ByteStringType::Regular => formatter.write_str("Byte-array in a String or null"),
-            }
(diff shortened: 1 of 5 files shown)


greg-szabo:
I think this is more to do about the strictness in Rust. When you omit a value in JSON, it defaults to `null`. Rust can deal with omitted values by using the `Default` trait. But if you explicitly set a value to `null` in JSON, that's not strictly a default value in Rust (becuase you can set the default to be, say "42"). Hence it's a value you need to receive and process. By default, `null`s are `None`s in Rust, so with an `Option<>` we would be able to receive and process it. But the whole point for us was to get away from Options, to make life easier for the developer, hence we need a "layer" of code that receives the explicit `null` value and turns it into something else. That's where the Visitor comes in.
If serde wanted to deal with this internally, they would need to assume that a `null` value translates to the default value. Since, even the `Default` trait is optional, I think they wanted to keep this optional too. Maybe an annotation for null values would help, like there is one for default values.

Liamsi:
I might have thought that serde does treat `null` as an empty vector. Apparently it doesn't: https://github.com/serde-rs/serde/issues/836

So the code looks fine to me, I was just wondering if we can do without the extra visitor here. Also I though we'd have a separate serilziation type (which could have an option in the fields and the actual ideal tendermint type that just has a vec and From impls between those types. 

Maybe I'm overthinking all this  😄

greg-szabo:
In the case of `Vec<u8>` it doesn't seem to make sense to turn it into a custom type (which would possibly allow us to use `Option<>` and convert internally, like in CommitSig) because the custom type wouldn't add any value (HexBytes and Base64Bytes are only different on the JSON side, it doesn't make any difference for the developer).
Built-in types can only be extended in the library we use them in (everywhere one-by-one) which would sprinkle serialization code all over the place. Or we write an annotation (like in the PR), which will require the Visitor so it can convert between two different types.

So, with annotating built-in types from weird formats, we might need to use a Visitor (like in this case). For custom types, most probably we won't need it, we'll have an internal struct (with `Option<>` or whatever it needs) and convert from that.

At least that's how I see it. If you think we can do this without the Visitor, point me in the right direction.

melekes:
> I might have thought that serde does treat null as an empty vector. 

Maybe the solution would be to not have explicit `null`s?

Liamsi:
> Maybe the solution would be to not have explicit nulls? Is it possible?

I would like that. But it would require some changes in the go code too and we'd need to communicate this clearly in the spec (nulls are not uncommon in JSON). 
The work in https://github.com/tendermint/tendermint/issues/4828 could be an opportunity to validate this idea.

Liamsi:
> In the case of Vec<u8> it doesn't seem to make sense to turn it into a custom type

I didn't mean Vec<u8> type but the wrapping struct which has `Vec<u8>` as a field. E.g. let's say the Header:

`serilaization::block::Header <---- TryFrom ----> core::block::Header`

The left header can use an Option in the fields or whatever is necessary to correctly deserialize this and the TryFrom handles this accordingly and translates into a real, core value on the right (with absolutely no serde annotation whatsoever). See https://github.com/informalsystems/tendermint-rs/issues/197#issuecomment-606103374

Downside would be a lot of duplication as the types would be almost identical. But it would be really, really simple, too. Also, there would be a cleat separation between serialization types and ideal core types that hold more logic and with which devs would mainly deal with. Devs wouldn't need to care at all about the serialization specifics. The only thing they need to now is that under a `serialization` module, there are types from which you can use to (de)serilalize and translate to the core types. Does al that make any sense?

That all said, this PR is an improvement to current situation and excellent work! The above rather touches upon that we didn't yet decide on a clear strategy on serialization as a whole (and not about specific core types).I think we should merge this PR but keep in mind that there is a bigger question on howto handle serialization (also of different serialization kinds, e.g. binary proto/amino and different kinds of JSON). If we decide to go with the above mentioned approach, some of the custom serialization code (the questioned Visitor), can be deleted.

tendermint/src/serializers.rs
```diff
@@ -96,24 +97,83 @@ pub mod time_duration {
...
             D: Deserializer<'de>,
         {
-            let string = String::deserialize(deserializer)?;
-
-            hex::decode_upper(&string)
-                .or_else(|_| hex::decode(&string))
-                .map_err(deError::custom)
+            deserializer.deserialize_option(super::BytesVisitor {
+                string_type: super::ByteStringType::Hex,
+            })
```
Liamsi, line 176: So this `deserialize_option` here and for the "ByteTypes" introduced the need for a Visitor, right? 

No need to apply this but if replaced by sth along the lines of:

```suggestion
let string = Option::<String>::deserialize(deserializer)?.unwrap_or("".into());
            hex::decode_upper(&string)
                .or_else(|_| hex::decode(&string))
                .map_err(serde::de::Error::custom)
```
here and in the other two places and if I'm not wrong the need for a custom Visitor would be gone?

Liamsi: Confirmed: https://github.com/informalsystems/tendermint-rs/pull/271/files

PR #271, opened 2020-05-14
Alternative to visitor and enum

Liamsi:
Try out if https://github.com/informalsystems/tendermint-rs/pull/269#discussion_r425005565 really works.

* [ ] Wrote tests
* [ ] Updated CHANGES.md


Liamsi:
Thanks Greg 😊I'll mere it into your branch as is. I think we can slightly improve by aliasing the error imports (currently the full path is used in the code, e.g. `serde::de::Error::custom`).

PR #267, opened 2020-05-11
JSON test structure changes

Shivani912:
closes #266 

To make identifying each test case easy, I've split up the test files and changed the directory structure. Now, each JSON file is a single test case.

**Directory Structure**
Under tendermint-rs/tendermint/tests/support/lite:

The four failing tests are found here:
1. ./single_step/sequential/commit/one_third_vals_don't_sign.json
2. ./single_step/sequential/commit/more_than_two_third_vals_sign.json
3. ./single_step/skipping/commit/one_third_vals_dont_sign.json
4. ./single_step/skipping/commit/more_than_two_third_vals_sign.json

**Further changes required:**
Issue #268 
* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGES.md

PR #273, opened 2020-05-15
Full JSON test structure changes

### LINKED ISSUE #268: Further JSON test fixes required
**Context**
PR #267 introduces directory structure changes to the existing JSON tests. 

**Problem**
This requires us to change the way these tests are run in Rust code because the older files no longer exist. What we have currently is configured to read big JSON files that deserialize into a vector of `TestCase` and we loop over them to run our tests. But now, we have split all these huge JSON files into multiple files such that each JSON file is a single `TestCase`. Now that there are so many files, it is messy to read each file manually and run test over it. 

**Proposed Solution**
Configure Rust test to loop over these files instead of manually having to read each one of them.

Liamsi:
For reviewers: the generated JSON files are too large to be reviewed thoroughly. The structure and content was already reviewed previously. now the large JSON files are simply split up in smaller ones (no need to review the JSON files...). 
.rs file diff only: https://github.com/informalsystems/tendermint-rs/pull/273/files?file-filters%5B%5D=.rs

#267 introduced a new test structure (one file per case). This PR here takes the JSON changes and adapts the rust code to these changes.

For more details see:
https://github.com/informalsystems/tendermint-rs/pull/267#issue-416243319

closes #268 
supersedes #267:

Note: one test currently fails (skipped) and also **all** multi-peer tests aren't even run yet (because we didn't implement the logic yet).

* [x] Wrote tests
* [ ] Updated CHANGES.md


tendermint/tests/lite.rs
```diff
@@ -308,7 +297,13 @@ async fn run_bisection_test(case: TestBisection) {
...
             let expected_state = TrustedState::new(untrusted_signed_header, untrusted_next_vals);
             assert_eq!(new_states[new_states.len() - 1], expected_state);
-            assert_eq!(new_states.len() as i32, expected_num_of_bisections);
+            let _expected_num_of_bisections = case.expected_num_of_bisections;
+            // TODO: number of bisections started diverting in JSON tests and Rust impl
+            // assert_eq!(
+            //     new_states.len() as i32,
+            //     expected_num_of_bisections,
+            //     "expected # bisections"
+            // );
```
Liamsi, line 306: This is new behaviour. Not sure why these assertions are failing now. Probably the go version @Shivani912 used to generate the new JSON tests moved on and evolved while the current light client in Rust didn't. 

I think it is OK to comment this out for now. It might be debatable if the number of bisections needs to be tested/asserted anyways.

tendermint/tests/lite.rs
```diff
@@ -121,138 +120,140 @@ impl MockRequester {
...
+                assert!(expects_err);
+            }
+        }
+    }
 }
+#[tokio::test]
+async fn bisection() {
+    // TODO: re-enable multi-peer tests as soon as the light client can handle these:
+    // let dir = "bisection/multi_peer";
+    // run_bisection_tests(dir).await;
```
Liamsi, line 180: I think there is already an open issue about multi peer support. As the light client is currently refactored, it will probably take a little longer until we have multi peer support.

tendermint/tests/lite.rs
```diff
@@ -121,138 +120,140 @@ impl MockRequester {
...
-// https://github.com/Shivani912/tendermint/commit/f8305c9a0e05696340fd8853ed5657a2075df895
-#[tokio::test]
-async fn bisection_happy_path() {
-    let case: TestBisection =
-        serde_json::from_str(&read_json_fixture("many_header_bisection/happy_path")).unwrap();
-    run_bisection_test(case).await;
-}
+fn run_single_step_tests(dir: &str) {
+    // TODO: this test need further investigation:
+    let skipped = ["commit/one_third_vals_don't_sign.json"];
```
Liamsi, line 227: I started with the 4 mentioned tests in that array (from #267). Only that one is failing. So this doesn't really need to be an array now 🤔

tendermint/tests/lite.rs
Shivani912: > This is new behaviour. Not sure why these assertions are failing now (even for the "happy path" ones). Probably the go version @Shivani912 used to generate the new JSON tests moved on and evolved while the current light client in Rust didn't.
> 
That can't be, because I didn't make any changes in the assertions in test cases. I'll take a closer look at this and fix it. But I think we can merge it this way and do the fix in a separate PR.
> I think it is OK to comment this out for now. It might be debatable if the number of bisections needs to be tested/asserted anyways.
>
Yeah, also it's not possible to make this assertion for Go code because we don't have access to this information. So this is implementation specific in a way.

tendermint/tests/lite.rs
Shivani912: The 4 tests mentioned are supposed to fail on parsing because they have Absent CommitSigs without validator address. @greg-szabo Did anything change on serialization side here?

Shivani912: Oh, just realized the serialization code for CommitSig from Greg hasn't been merged with master yet. I'm not sure if we should compare this branch with the one that has Greg's serialization code or merge the changes with master as is?

tendermint/tests/lite.rs
Liamsi: > Yeah, also it's not possible to make this assertion for Go code because we don't have access to this information. So this is implementation specific in a way.

Certainly sounds like, we should remove the assertion here then.

tendermint/tests/lite.rs
Liamsi: I'd merge into master. So if Greg finishes the CommitSig changes, we can merge in master into his branch and see if there is any additional work to do to make now failing tests pass.

tendermint/tests/lite.rs
Liamsi: opened #276 to track this.

PR #275, opened 2020-05-18
a few invariants and a bugfix to block verification

konnov:
This PR contains several Fastsync properties:

 * `Correctness3` and `Correctness3AsInv` that propose a fix to `Correctness2` (satisfied)
 * `SyncFromCorrectInv` that states a naive property that the synchronized blocks come only from the correct processes (violated)
 * `CorrectNeverSuspectedInv` that states that the correct peers are never suspected (violated)
 * `CorrectBlocksInv` that states that the synchronized blocks have correct block ids and enough voting power.

The current version of Fastsync violates `CorrectBlocksInv` in two ways:

 * As `InitNode` does not have the genesis block, a faulty peer may send the node a corrupt genesis block. We fix this by initializing `blockStore` with a correct genesis block at height 1.
 * As `ExecuteBlocks` does not verify a block at the height of 2, the last commit in block 2 may be corrupt. We fix this by introducing an extra test for the height of 2.

### DIFF
--- a/docs/spec/fastsync/fastsync.tla
+++ b/docs/spec/fastsync/fastsync.tla
@@ -75,6 +75,9 @@ HeightsPlus == 1..MAX_HEIGHT+1
 \* a special value for an undefined height
 NilHeight == 0
 
+\* the height of the genesis block
+GenesisHeight == 1
+
 \* the set of all peer ids the node can receive a message from
 AllPeerIds == CORRECT \union FAULTY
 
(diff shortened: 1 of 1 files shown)


ebuchman:
So Tendermint doesn't actually have a genesis block. It just has a genesis state which determines the initial validator set for the first block, block 1. But block 1 does not contain a commit - it's explicitly empty. It might have been nice if the protocol was designed instead to just get the validators to sign the genesis state, but unfortunately it currently does not.  So the first commit is contained in block 2. If this commit isn't valid for block 1, both blocks should be thrown out, like normal.

konnov:
This PR is completely outdated. I will close this one and prepare another one.

Issue #260, opened 2020-05-08
[pending Go changes] CommitSig validator address missing in Absent vote


greg-szabo:
**Go implementation**
In CommitSig, in the BlockIDFlagAbsent type, the validator address field is expected to be missing.

**ADR-025**
The validator address field in Absent signatures should be present. (The validator address field should always be present.)

**Current Rust implemenation**
Rust implemented the ADR.

**Problem**
Tendermint-Rust is incompatible with Tendermint-Go.

**Solution**

* [done #277](https://github.com/informalsystems/tendermint-rs/issues/277) Add a compatibility code to the Rust implementation.
* [done](https://github.com/tendermint/tendermint/issues/4797) Add an issue to the Go repository
* [todo] Remove the compatibility code when the Go code adheres to ADR-025

ebuchman:
Thanks @greg-szabo . If and when this is changed in the Go, it will be breaking, and we can open a new issue for it then, since it otherwise seems we've addressed it for now.

Also as per https://github.com/tendermint/tendermint/issues/4797#issuecomment-637875417, not sure this ever will be addressed, especially since we might look to remove ValidatorAddress from all CommitSigs!

Let's close this then?

Liamsi:
I'm closing this. Feel free to re-open if you feel we should keep it open @greg-szabo.

Issue #259, opened 2020-05-08
CommitSig Timestamp can be zero time


greg-szabo:
**Go implementation**
In CommitSig, in the BlockIDFlagAbsent type, the Timestamp field is a Go "Zero value" (`0001-01-01T00:00:00Z`).

**ADR-025**
The Timestamp field in Absent signatures should be null or omitted.

**Current Rust implemenation**
Rust implemented the ADR.

**Problem**
Tendermint-Rust is incompatible with Tendermint-Go.

**Solution**
- [done #277](https://github.com/informalsystems/tendermint-rs/pr/277) Add a compatibility code to the Rust implementation.
- [todo] Add an issue to the Go repository
- [todo] Remove the compatibility code when the Go code adheres to ADR-025

Relevant discussion:
https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r401027989

greg-szabo:
Next step: open a Go issue and link it here. (I need to reproduce the issue in Go for them.)

Liamsi:
This was resolved in https://github.com/informalsystems/tendermint-rs/pull/277

I think we should rather close this PR. If the changes are made in go (and there will be potenially more substantial changes in the light of https://github.com/tendermint/tendermint/pull/4857). Feel free to re-open this issue if you thinjk we should open an issue in amino or tendermint and link it here.

Issue #266, opened 2020-05-11
Conformance tests compatibility issues


Shivani912:
**Context**
- Rust implementation is incompatible with Go implementation. 
- Details: Issues #260 and #259 

**Problem**
Since the conformance tests (JSON-based tests) are generated using the Go implementation, these incompatibilities make four of the `Commit` tests fail on deserialization. These are:
- one-third vals don't sign (both under single-step sequential and skipping)
- more than two-third vals did sign (both under single-step sequential and skipping)

**Proposed Solution**
After a good discussion with @greg-szabo and @Liamsi on this, it was decided to identify and run these tests separately where we let it fail on deserialization for now with an added warning about the issues around it. 
This way, we wouldn't have to makes changes to our code that are not acceptable i.e. accept an `AbsentCommitSig` without `ValidatorAddress` in this case. Also, once the issue is fixed on the Go implementation side, we'll be able to run these tests normally.

PR #277, opened 2020-05-21
CommitSig refactor

greg-szabo:
This is a second - slightly less complex - stab at refactoring CommitSig.

Resolves the following:
#247 - CommitSig refactor
#259 - CommitSig timestamp can be "0000" time instead of null (Go compatibility issue)
#260 - CommitSig validator address not present in Absent votes (Go compatibility issue)

- The two compatibility issues had to be included or else the tests would have failed. With that said, the compatibility issues should remain open until they are fixed in the Go code. The fixed code is also included here and it's commented out.

Please review and comment. We can change the way we deal with the compatibility code, if this looks unruly.

~~- @melekes had two questions in a previous review of this code: where is the validator_address validation (I think that was not implemented before) and why is there a TODO about full verification and what should we do about it. - I would like to clarify these and open issues around them or resolve them in this PR, if we don't need to do anything about them.~~
Anton's questions were either answered in the PR or a new issue opened about it.

~~Because of the above questions, I'm putting this PR into "draft" mode, although it's ready to be reviewed.~~

* [X] Wrote tests
* [X] Updated CHANGES.md

### DIFF
--- a/tendermint/src/block/commit_sig.rs
+++ b/tendermint/src/block/commit_sig.rs
@@ -12,7 +12,7 @@ use std::convert::TryFrom;
 #[serde(try_from = "RawCommitSig")]
 pub enum CommitSig {
     /// no vote was received from a validator.
-// Todo: https://github.com/informalsystems/tendermint-rs/issues/260 - CommitSig validator address missing in Absent vote
+    // Todo: https://github.com/informalsystems/tendermint-rs/issues/260 - CommitSig validator address missing in Absent vote
     BlockIDFlagAbsent,
     /// voted for the Commit.BlockID.
     BlockIDFlagCommit {
--- a/tendermint/src/block/commit_sig.rs
+++ b/tendermint/src/block/commit_sig.rs
@@ -77,7 +77,7 @@ impl TryFrom<RawCommitSig> for CommitSig {
             }
             BlockIDFlag::BlockIDFlagCommit => {
                 if value.timestamp.is_none() {
-                    Err("timestamp is null for BlockIDFlagCommit CommitSig")
+                    Err("timestamp is missing for BlockIDFlagCommit CommitSig")
                 } else if value.signature.is_none() {
                     Err("signature is null for BlockIDFlagCommit CommitSig")
                 } else {
--- a/tendermint/src/block/commit_sig.rs
+++ b/tendermint/src/block/commit_sig.rs
@@ -45,7 +45,6 @@ pub enum CommitSig {
 impl TryFrom<RawCommitSig> for CommitSig {
     type Error = &'static str;
 
-    // Todo: @melekes asked: where's validator_address validation?
     fn try_from(value: RawCommitSig) -> Result<Self, Self::Error> {
         // Validate CommitSig (strict)
         match value.block_id_flag {
--- a/tendermint/src/block/commit_sig.rs
+++ b/tendermint/src/block/commit_sig.rs
@@ -1,78 +1,117 @@
 //! CommitSig within Commit
 
-use crate::serializers;
+use crate::serializers::BlockIDFlag;
+use crate::serializers::RawCommitSig;
 use crate::{account, Signature, Time};
-use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
-
-/// BlockIDFlag is used to indicate the validator has voted either for nil, a particular BlockID or was absent.
-#[derive(Copy, Clone, Debug, PartialEq)]
-pub enum BlockIDFlag {
-    /// BlockIDFlagAbsent - no vote was received from a validator.
-    BlockIDFlagAbsent = 1,
-    /// BlockIDFlagCommit - voted for the Commit.BlockID.
-    BlockIDFlagCommit = 2,
-    /// BlockIDFlagNil - voted for nil.
-    BlockIDFlagNil = 3,
-}
-
-impl BlockIDFlag {
-    /// Deserialize this type from a byte
-    pub fn from_u8(byte: u8) -> Option<BlockIDFlag> {
-        match byte {
-            1 => Some(BlockIDFlag::BlockIDFlagAbsent),
-            2 => Some(BlockIDFlag::BlockIDFlagCommit),
-            3 => Some(BlockIDFlag::BlockIDFlagNil),
-            _ => None,
-        }
-    }
-
-    /// Serialize this type as a byte
-    pub fn to_u8(self) -> u8 {
-        self as u8
-    }
-
-    /// Serialize this type as a 32-bit unsigned integer
-    pub fn to_u32(self) -> u32 {
-        self as u32
-    }
-}
-
-impl Serialize for BlockIDFlag {
-    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        self.to_u8().serialize(serializer)
-    }
-}
-
-impl<'de> Deserialize<'de> for BlockIDFlag {
-    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        let byte = u8::deserialize(deserializer)?;
-        BlockIDFlag::from_u8(byte)
-            .ok_or_else(|| D::Error::custom(format!("invalid block ID flag: {}", byte)))
-    }
-}
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
 
 /// CommitSig represents a signature of a validator.
 /// It's a part of the Commit and can be used to reconstruct the vote set given the validator set.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-pub struct CommitSig {
-    /// Block ID FLag
-    pub block_id_flag: BlockIDFlag,
-
-    /// Validator address
-    #[serde(deserialize_with = "serializers::parse_non_empty_id")]
-    pub validator_address: Option<account::Id>,
-
-    /// Timestamp
-    pub timestamp: Time,
-
-    /// Signature
-    #[serde(deserialize_with = "serializers::parse_non_empty_signature")]
-    pub signature: Option<Signature>,
+#[serde(try_from = "RawCommitSig")]
+pub enum CommitSig {
+    /// no vote was received from a validator.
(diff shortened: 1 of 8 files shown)


greg-szabo:
Note: I just realized the note in #266 that the compatibility issues don't need to be implemented. But the current tests would fail because the incompatible tests are not separated yet.

greg-szabo:
I'm scratching my head at the test failure, it worked on my machine. I'll look into it in a second.

As for @melekes 's question, I think I realize what he was asking: the validator_address is an `account::Id` which is deserialized from a hexstring into a 20 long byte array. That is restricted during deserialization. Anything else will panic. (Except in the case of CommitSig, where the Go code enables empty strings as no-address values.)

If that was the question and the answer is good enough, I'd like to remove the comment about it in the source code.

tendermint/src/block/commit_sig.rs
```diff
@@ -1,78 +1,119 @@
...
 }
 
-impl CommitSig {
-    /// Checks if a validator's vote is absent
-    pub fn is_absent(&self) -> bool {
-        self.block_id_flag == BlockIDFlag::BlockIDFlagAbsent
+impl TryFrom<RawCommitSig> for CommitSig {
+    type Error = &'static str;
+
+    // Todo: @melekes asked: where's validator_address validation?
```
greg-szabo, line 48: I think I realize what he was asking: the validator_address is an account::Id which is deserialized from a hexstring into a 20 long byte array. That is restricted during deserialization. Anything else will panic. (Except in the case of CommitSig, where the Go code enables empty strings as no-address values.)

If that was the question and the answer is good enough, I'd like to remove the comment about it.

Liamsi:
> But the current tests would fail because the incompatible tests are not separated yet.

Let's jump on a call for this. I think they are separated in the sense that you can skip single JSON files to make an exception and treat them differently (e.g expect a panic, or make sure that they already fail on deserialisation instead of later <- these exceptions aren't written yet, only the possibly to skip and treat them separately)

greg-szabo:
Good news, tests succeed now, no need for a call. I've resolved all outstanding questions one way or another. Feel free to review and comment.

tendermint/src/block/commit_sig.rs
```diff
@@ -1,79 +1,118 @@
...
+            }
+            BlockIDFlag::BlockIDFlagCommit => {
+                if value.timestamp.is_none() {
+                    Err("timestamp is null for BlockIDFlagCommit CommitSig")
+                } else if value.signature.is_none() {
+                    Err("signature is null for BlockIDFlagCommit CommitSig")
+                } else {
+                    Ok(CommitSig::BlockIDFlagCommit {
+                        // <<< Compatibility code for https://github.com/informalsystems/tendermint-rs/issues/260
+                        validator_address: value.validator_address.unwrap(),
```
Liamsi, line 86: Can't this unwrap panic here if `validator_address` is None? You could use `ok_or_else()` in combination with `?` to bubble up an error instead of potentially panicing.

tendermint/src/block/commit_sig.rs
```diff
@@ -1,79 +1,118 @@
...
+            }
+            BlockIDFlag::BlockIDFlagNil => {
+                if value.timestamp.is_none() {
+                    Err("timestamp is null for BlockIDFlagNil CommitSig")
+                } else if value.signature.is_none() {
+                    Err("signature is null for BlockIDFlagNil CommitSig")
+                } else {
+                    Ok(CommitSig::BlockIDFlagNil {
+                        // <<< Compatibility code for https://github.com/informalsystems/tendermint-rs/issues/260
+                        validator_address: value.validator_address.unwrap(),
```
Liamsi, line 105: Same as above: this could panic, better to err, if there was a None in `validator_address`.

tendermint/src/block/commit_sig.rs
```diff
@@ -1,79 +1,118 @@
...
+                        timestamp: value.timestamp.unwrap(),
+                        signature: value.signature.unwrap(),
+                    })
+                }
+            }
+            BlockIDFlag::BlockIDFlagNil => {
+                if value.timestamp.is_none() {
+                    Err("timestamp is null for BlockIDFlagNil CommitSig")
+                } else if value.signature.is_none() {
+                    Err("signature is null for BlockIDFlagNil CommitSig")
```
Liamsi, line 101: s/missing/null/g

tendermint/src/block/commit_sig.rs
```diff
@@ -1,79 +1,118 @@
...
+                if value.signature.is_some() {
+                    return Err("signature is present for BlockIDFlagAbsent CommitSig");
+                }
+                // <<< Compatibility code for https://github.com/informalsystems/tendermint-rs/issues/260
+                Ok(CommitSig::BlockIDFlagAbsent)
+                // === Real code after compatibility issue is resolved
+                /*
+                Ok(CommitSig::BlockIDFlagAbsent {
+                    validator_address: value.validator_address,
+                })
```
Liamsi, line 74: Did we actually clarify if that will be changed on the go-side too? (will the `validator_address ` be added back in the case `BlockIDFlagAbsent`? Or is the ADR simply out of date? Sorry, I can't remember if that was answered somewhere else.

tendermint/src/serializers/raw_commit_sig.rs
```diff
@@ -0,0 +1,66 @@
...
+use serde::de::Error;
+use serde::{Deserialize, Deserializer};
+use serde_repr::{Deserialize_repr, Serialize_repr};
+use std::str::FromStr;
+
+// Implements decision: https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-025-commit.md#decision
+
+/// indicate which BlockID the signature is for
+#[derive(Serialize_repr, Deserialize_repr, PartialEq, Debug)]
+#[repr(u8)]
```
Liamsi, line 12: Neat, I didn't know about this crate.

tendermint/src/block/commit_sig.rs
```diff
@@ -1,79 +1,118 @@
...
+    type Error = &'static str;
+
+    fn try_from(value: RawCommitSig) -> Result<Self, Self::Error> {
+        // Validate CommitSig (strict)
+        match value.block_id_flag {
+            BlockIDFlag::BlockIDFlagAbsent => {
+                if value.timestamp.is_some() {
+                    // <<< Compatibility code for https://github.com/informalsystems/tendermint-rs/issues/259
+                    if value.timestamp.unwrap()
+                        != Time::parse_from_rfc3339("0001-01-01T00:00:00Z").unwrap()
```
Liamsi, line 55: 👍

tendermint/src/block/commit_sig.rs
```diff
@@ -1,79 +1,118 @@
...
+                    validator_address: value.validator_address,
+                })
+                */
+                // >>> end of real code
+            }
+            BlockIDFlag::BlockIDFlagCommit => {
+                if value.timestamp.is_none() {
+                    Err("timestamp is null for BlockIDFlagCommit CommitSig")
+                } else if value.signature.is_none() {
+                    Err("signature is null for BlockIDFlagCommit CommitSig")
```
Liamsi, line 82: s/null/missing

tendermint/src/block/commit_sig.rs
greg-szabo: done

tendermint/src/block/commit_sig.rs
greg-szabo: Yeah, I can do that - although since I'm already checking specifically for timestamp and signature, I could just check for validator_address too. (Giving a more relevant error message.)

If the Go code would be compatible with ADR-025, this wouldn't be a problem, because validator_address wouldn't be an Option<>. (See the "real code" below.)

tendermint/src/block/commit_sig.rs
greg-szabo: It seems to me that the issue was acknowledged on the Go side. It was tagged as a consensus-related issue and there's 73 of that open. Also, based on the note, the fix for this might be postponed or fully cancelled because of discussions on refactoring messaging.

https://github.com/tendermint/tendermint/issues/4797
> it's tricky to get ValidatorAddress from a nil vote. So, if we decide to have ValidatorAddress for absent votes, we'll need to remove the concept of nil votes in Tendermint, which may be a good idea on its own.

tendermint/src/serializers/raw_commit_sig.rs
greg-szabo: Serde and its dark arts... :)

tendermint/src/block/commit_sig.rs
greg-szabo: done

greg-szabo:
So, based on the priorities in the Tendermint Go code, I wouldn't hold my breath that issue #260 (or #259) will be resolved soon. It's not a real problem for us either, the workarounds are ok. But if you want readability, we should just cut our losses ("the real code") and get on with life.

I'll leave a note where the code needs changing that points to the issues. If we close the issues we should fix/remove those comments.

tendermint/src/block/commit_sig.rs
greg-szabo: Panics removed.

tendermint/src/block/commit_sig.rs
greg-szabo: Panics removed.

greg-szabo:
So, all compatibility remarks removed. They were already in issues and the issues were mentioned in that code. The problem is that some of the compatibility changes require changes in multiple files and we're littering our codebase with sub-optimal code because of Go incompatibility things. Right now Issue #260 requires changes in 4 files and a total of 6 places. The code was written for it but now the right code is removed and workarounds were put in place. (I guess whoever fixes #260 can come back to this PR and see what was done.)

If there's a good way to keep track of compatibility decisions in code (things like "why is this an `Option<>`?" or "why is this check here?"), it would be nice to somehow do it. My example was really just testing if this would be ok.

All above remarks are addressed, let's try another review.

Liamsi:
> If there's a good way to keep track of compatibility decisions in code (things like "why is this an Option<>?" or "why is this check here?")

If we followed it consequently, ADRs instead of in the are the best place for longer descriptions IMO. Other than that, regular comments in the code can explain why certain things are as they are.

Issue #278, opened 2020-05-23
Event Subscription response cannot be deserialized correctly


greg-szabo:
**What happens**
When receiving an event subscription response from Tendermint Go, Rust deserializes the incoming data into a `GenericJSONEvent`. This can be observed in the `event_subscription` test.
(Tendermint Go v0.33.4)

**What should happen**
The incoming data should be deserialized into a `JsonRPCTransactionResult` type correctly.

The reason seems to be that the defined Rust structs are not compatible with the incoming Go JSON.
Interesting addition: the Game of Zones compatible Gaia RPC sends the correct JSON and Rust can deserialize it correctly into `JsonRPCTransactionResult`.

One discrepancy found is that the JSON `result_begin_block` field is `{}` in Tendermint, while it's populated with an `events: [...]` field in Gaia. The Rust implementation considers the events field mandatory. There might be other field that don't behave well.

Relevant discussion [here](https://github.com/informalsystems/tendermint-rs/pull/225/files#r429444086). Please note the TODO in the code above the discussion.

ref: #225

greg-szabo:
The below files are example responses from Tendermint Go v0.33.4 and Gaia (commit ID cbc3321 which is the WIP code for GoZ phase 2, also using Tendermint v0.33.4)

[gaia.json.log](https://github.com/informalsystems/tendermint-rs/files/4672113/gaia.json.log)
[tendermint.json.log](https://github.com/informalsystems/tendermint-rs/files/4672114/tendermint.json.log)

Liamsi:
# Context 
Some context on the events; in tendermint there are these types:

Event data from subscriptions:
```go
type ResultEvent struct {
	Query  string              `json:"query"`
	Data   types.TMEventData   `json:"data"`
	Events map[string][]string `json:"events"`
}
```
and the `TMEventData` in the Data field:
```go
// TMEventData implements events.EventData.
type TMEventData interface {
	// empty interface
}

func RegisterEventDatas(cdc *amino.Codec) {
	cdc.RegisterInterface((*TMEventData)(nil), nil)
	cdc.RegisterConcrete(EventDataNewBlock{}, "tendermint/event/NewBlock", nil)
	cdc.RegisterConcrete(EventDataNewBlockHeader{}, "tendermint/event/NewBlockHeader", nil)
	cdc.RegisterConcrete(EventDataTx{}, "tendermint/event/Tx", nil)
[...]

PR #279, opened 2020-05-23
Simplify event subscription

Liamsi:
ref #278 and followup to #225

* [x] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/rpc/event_listener.rs
+++ b/tendermint/src/rpc/event_listener.rs
@@ -79,16 +79,26 @@ impl EventListener {
     }
 
     /// Get the next event from the websocket
-    pub async fn get_event(&mut self) -> Result<ResultEvent, Box<dyn stdError>> {
+    pub async fn get_event(&mut self) -> Result<Option<ResultEvent>, RPCError> {
         let msg = self
             .socket
             .next()
             .await
             .ok_or_else(|| RPCError::websocket_error("web socket closed"))??;
-        let result_event =
-            serde_json::from_str::<WrappedResultEvent>(&msg.to_string())?.into_result()?;
 
-        Ok(result_event)
+        if let Ok(result_event) = serde_json::from_str::<WrappedResultEvent>(&msg.to_string()) {
+            // if we get an rpc error here, we will bubble it up:
+            return Ok(Some(result_event.into_result()?));
+        }
+        dbg!("We did not receive a valid JSONRPC wrapped ResultEvent!");
+        if let Ok(_) = serde_json::from_str::<String>(&msg.to_string()) {
+            // FIXME(ismail): Until this is a proper websocket client
+            // (or the endpoint moved to grpc in tendermint), we accept whatever was read here
+            // dbg! it out and return None below.
+            dbg!("Instead of JSONRPC wrapped ResultEvent, we got:");
+            dbg!(&msg.to_string());
+        }
+        Ok(None)
     }
 }
 
--- a/tendermint/src/rpc/event_listener.rs
+++ b/tendermint/src/rpc/event_listener.rs
@@ -168,23 +168,6 @@ pub struct Attribute {
     value: String,
 }
 
-/// Block Results
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct RPCBlockResult {
-    query: String,
-    data: BlockResultData,
-    events: HashMap<String, Vec<String>>,
-}
-impl response::Response for RPCBlockResult {}
-
-/// Block Results data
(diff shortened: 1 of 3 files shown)


tendermint/src/rpc/event_listener.rs
```diff
@@ -57,102 +57,85 @@ impl EventListener {
...
+        let result_event =
+            serde_json::from_str::<WrappedResultEvent>(&msg.to_string())?.into_result()?;
+
+        Ok(result_event.data)
     }
 }
-/// The Event enum is typed events emmitted by the Websockets
+
+// TODO: (later) this should live somewhere else; these events are also
+// published byte the event bus independent from RPC.
```
xla, line 94:
```suggestion
// published by the event bus independent from RPC.
```

tendermint/src/rpc/event_listener.rs
```diff
@@ -57,102 +57,85 @@ impl EventListener {
...
+// We leave it here for now because unsupported types are still
+// decodeable via fallthrough variants (GenericJSONEvent).
+/// The Event enum is typed events emitted by the Websockets
```
xla, line 97:
```suggestion
/// The Event enum is typed events emitted by the Websockets.
```

tendermint/src/rpc/event_listener.rs
```diff
@@ -57,102 +57,85 @@ impl EventListener {
...
-    value: TxValue,
-}
+/// JSONRPC wrapped ResultEvent
+pub type WrappedResultEvent = Wrapper<ResultEvent>;
 
 /// TX value
 #[derive(Serialize, Deserialize, Debug, Clone)]
-struct TxValue {
+pub struct EventDataTx {
     #[serde(rename = "TxResult")]
```
xla, line 132: This could be a the container attribute `rename_all = "PascalCase"`.

tendermint/src/rpc/event_listener.rs
```diff
@@ -171,65 +154,36 @@ impl response::Response for TxResultResult {}
...
-    value: BlockValue,
-}
-
 ///Block Value
 #[derive(Serialize, Deserialize, Debug, Clone)]
-struct BlockValue {
+pub struct EventDataNewBlock {
     block: Option<Block>,
+
+    // TODO these should be the same as abci::responses::BeginBlock
```
xla, line 174:
```suggestion
    // TODO(liamsi): these should be the same as abci::responses::BeginBlock
```

Maybe worth an issue.

tendermint/src/rpc/event_listener.rs
```diff
@@ -171,65 +154,36 @@ impl response::Response for TxResultResult {}
...
-        &self,
-        action_query: &str,
-    ) -> Result<HashMap<String, Vec<String>>, Box<dyn stdError>> {
-        if let Some(message_action) = self.events.get("message.action") {
-            if message_action.contains(&action_query.to_owned()) {
-                return Ok(self.events.clone());
-            }
-        }
-        Err("Incorrect Event Type".into())
-    }
```
zmanian, line 234: This function is required for actually providing types for the events. It can't be removed. Needs to be added to the new ResultEvent type.

Liamsi: Thanks, I'll add it back now. Did everything else work with gaia too?

Liamsi: Just to confirm my understanding: Above in `get_event` you don't just need the `TMEventData` but the whole `ResultEvent`? This:
```rust
pub struct ResultEvent {
    query: Option<String>,
    data: Option<TMEventData>,
    events: Option<HashMap<String, Vec<String>>>,
}
```
Is that correct?

tendermint/src/rpc/event_listener.rs
Liamsi: I'd rather keep those renamings local to the var. Seems more consistent to the remaining code.

tendermint/src/rpc/event_listener.rs
zmanian: I think it's reasonable for the caller to be interested in both the TMEventData and the Hashmap of Events

Liamsi: Changed to return `ResultEvent` in https://github.com/informalsystems/tendermint-rs/pull/279/commits/0e3e422ae713e439a6a2e2cf04c2a9f59a098dcd

Liamsi:
I'll merged this as is but I still think this still needs some more work. But instead of re-iterating on the implementation only, it would be better to write up an ADR describing the event subscription and the made and to be made design choices briefly. I'll followup on that shortly.

Issue #280, opened 2020-05-25
Light Client: Follow-up


romac:
- [x] Introduce `clockDrift` parameter, as per the spec
- [x] Write documentation
- [x] Tag code as per the traceability proposal
- [x] Add new tags to the English spec
- [x] Add tag versions to the English spec
- [x] See whether we need to re-introduce the `TrustedStoreReader/Writer` wrappers (we don't)
- [ ] Add support for verifying headers lower than the latest trusted state - #361 
- [ ] Add [structured logging](https://docs.rs/tracing)
- [ ] Streamline creation of a light client instance and supervisor
    - [ ] Use the [builder pattern](https://www.ameyalokare.com/rust/2017/11/02/rust-builder-pattern.html) to create and initialize the `LightClient` and `Supervisor`
- [ ] Write units tests for new components - #320
- [ ] Port current light client unit tests - #320 
- [ ] Remove previous lite client code in `tendermint` crate - #342 
- [x] Address comments in #294

---

Other issues we may want to fix as well as part of this follow-up:

- [x] #276
- [x] #137
- [ ] #235
- [x] #138 
- [x] #136 (mostly done in #237)

---

These tasks should be handled in separate issue and PRs:

- Implement CLI for running a standalone light client + RPC server

ebuchman:
I believe all the points have either been addressed or given new issues except for:

- [ ] Add [structured logging](https://docs.rs/tracing)
- [ ] Streamline creation of a light client instance
    - [ ] Use the [builder pattern](https://www.ameyalokare.com/rust/2017/11/02/rust-builder-pattern.html) to create and initialize the `LightClient`

@romac want to open separate issues for these and then we can close this one?

romac:
Done in #421 and #422.

PR #284, opened 2020-05-27
Light Client: Follow-up

romac:
See #280

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/components/scheduler.rs
+++ b/light-client/src/components/scheduler.rs
@@ -76,18 +76,18 @@ pub fn basic_schedule(
 /// - i) If `latest_verified_height == current_height` and `latest_verified_height < target_height`
 /// then `current_height < scheduled_height <= target_height`.
 ///
-/// - ii) If `latest_verified_height < next_height` and `latest_verified_height < target_height`
-/// then `latest_verified_height < scheduled_height < next_height`.
+/// - ii) If `latest_verified_height < current_height` and `latest_verified_height < target_height`
+/// then `latest_verified_height < scheduled_height < current_height`.
 ///
 /// - iii) If `latest_verified_height = target_height` then `scheduled_height == target_height`.
 ///
 /// ## Note
 ///
-/// - Case i. captures the case where the light block at height nextHeight has been verified,
-/// and we can choose a height closer to the targetHeight. As we get the `light_store` as parameter,
-/// the choice of the next height can depend on the `light_store`, e.g., we can pick a height
-/// for which we have already downloaded a light block.
-/// - In Case ii. the header of `next_height` could not be verified, and we need to pick a smaller height.
+/// - Case i. captures the case where the light block at height `current_height` has been verified,
+///   and we can choose a height closer to the `target_height`. As we get the `light_store` as parameter,
+///   the choice of the next height can depend on the `light_store`, e.g., we can pick a height
+///   for which we have already downloaded a light block.
+/// - In Case ii. the header at `current_height` could not be verified, and we need to pick a lesser height.
 /// - In Case iii. is a special case when we have verified the `target_height`.
 ///
 /// ## Implements
--- a/light-client/src/components/scheduler.rs
+++ b/light-client/src/components/scheduler.rs
@@ -39,7 +39,7 @@ where
     }
 }
 
-/// Basic scheduler which picks the appropriate midpoint without
+/// Basic bisecting scheduler which picks the appropriate midpoint without
 /// optimizing for performance using the blocks available in the light store.
 ///
 /// ## Precondition
@@ -49,7 +49,7 @@ where
 /// - The resulting height must be valid according to `valid_schedule`. [LCV-SCHEDULE-POST.1]
 #[pre(light_store.latest(VerifiedStatus::Verified).is_some())]
 #[post(valid_schedule(ret, target_height, current_height, light_store))]
-pub fn basic_schedule(
+pub fn basic_bisecting_schedule(
     light_store: &dyn LightStore,
     current_height: Height,
     target_height: Height,
--- a/light-client/src/components/scheduler.rs
+++ b/light-client/src/components/scheduler.rs
@@ -1,14 +1,25 @@
 use crate::prelude::*;
 use contracts::*;
 
+/// The scheduler decides what block to verify next given the current and target heights.
+///
+/// The scheduler is given access to the light store, in order to optionally
+/// improve performance by picking a next block that has already been fetched.
 #[contract_trait]
 pub trait Scheduler {
+    /// Decides what block to verify next.
+    ///
+    /// ## Precondition
+    /// - The light store contains at least one verified block. [LCV-SCHEDULE-PRE.1]
+    ///
+    /// ## Postcondition
+    /// - The resulting height must be valid according to `valid_schedule`. [LCV-SCHEDULE-POST.1]
     #[pre(light_store.latest(VerifiedStatus::Verified).is_some())]
-    #[post(valid_schedule(ret, target_height, next_height, light_store))]
+    #[post(valid_schedule(ret, target_height, current_height, light_store))]
     fn schedule(
         &self,
         light_store: &dyn LightStore,
-        next_height: Height,
+        current_height: Height,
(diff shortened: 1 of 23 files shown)


brapse:
Amazing documentation with traceability to the english spec 😍

light-client/src/components/verifier.rs
```diff
@@ -78,47 +82,48 @@ impl ProdVerifier {
...
-        crate::predicates::verify_overlap(
-            &*self.predicates,
-            &self.voting_power_calculator,
-            &trusted_state,
-            &light_block,
-            options,
-        )
-        .into()
-    }
+        let verdict = verdict.and_then(|| {
```
OStevan, line 108: If I understand correctly overlap verification should only be done if the two headers are non adjacent and in the case they are adjacent we should just check hashes as is done in `valid_next_validator_set`. While here `verify_overlap` is always checked disregarding adjacency relatinoship.

light-client/src/predicates.rs
```diff
@@ -291,6 +319,8 @@ pub fn verify_overlap(
     Ok(())
 }
 
+/// Assuming the given light block is valid, check whether the voting
+/// power of its validators meet the required threshold.
 pub fn has_sufficient_voting_power(
```
OStevan, line 324: It looks like this function should replace `verify_commit_full` in the current implementation, however it checks the power of commit votes with respect to `trust_threshold`(> 1/3) and the validator set of the header. Which seems like it is not correct as the check we are trying to do is confirm that the consensus has been reached and this should be done with respect to a threshold of 2/3. The mentioned check for >2/3 is done in `has_sufficient_signers_overlap`.

light-client/src/components/verifier.rs
romac: Good catch, that's indeed an issue, which Ethan raised in #294. Glad to see you guys agree :)
Should be fixed in 953591a4296c32f4ebdd7b6158618634889d8ffa if you want to take another look.

light-client/src/predicates.rs
romac: Same as above.

light-client/src/components/scheduler.rs
```diff
@@ -21,51 +32,81 @@ where
...
+    current_height: Height,
     target_height: Height,
 ) -> Height {
     let latest_trusted_height = light_store
         .latest(VerifiedStatus::Verified)
         .map(|lb| lb.height())
         .unwrap();
 
-    if latest_trusted_height == next_height && latest_trusted_height < target_height {
+    if latest_trusted_height == current_height && latest_trusted_height < target_height {
```
brapse, line 62: Will we be able toe eventually do backwards verification like in the go client?

light-client/src/components/scheduler.rs
```diff
@@ -1,14 +1,25 @@
 use crate::prelude::*;
```
brapse, line 1: Maybe in the preamble provide reference to the english spec  such that the invariants can be looked up by the reader.

light-client/src/components/scheduler.rs
```diff
@@ -21,51 +32,81 @@ where
...
+/// ## Precondition
+/// - The light store contains at least one verified block. [LCV-SCHEDULE-PRE.1]
+///
+/// ## Postcondition
+/// - The resulting height must be valid according to `valid_schedule`. [LCV-SCHEDULE-POST.1]
 #[pre(light_store.latest(VerifiedStatus::Verified).is_some())]
-#[post(valid_schedule(ret, target_height, next_height, light_store))]
-pub fn schedule(
+#[post(valid_schedule(ret, target_height, current_height, light_store))]
+pub fn basic_schedule(
```
brapse, line 52: Since this is actually performing bisection does it make sense to call it `bisecting_schedule`?

light-client/src/components/scheduler.rs
```diff
@@ -21,51 +32,81 @@ where
...
 }
 
-fn valid_schedule(
+/// Checks whether the given `scheduled_height` is a valid schedule according to the
+/// following specification.
+///
+/// - i) If `latest_verified_height == current_height` and `latest_verified_height < target_height`
+/// then `current_height < scheduled_height <= target_height`.
+///
+/// - ii) If `latest_verified_height < next_height` and `latest_verified_height < target_height`
```
brapse, line 79: s/next_height/current_height

light-client/src/components/scheduler.rs
```diff
@@ -21,51 +32,81 @@ where
...
+/// then `latest_verified_height < scheduled_height < next_height`.
+///
+/// - iii) If `latest_verified_height = target_height` then `scheduled_height == target_height`.
+///
+/// ## Note
+///
+/// - Case i. captures the case where the light block at height nextHeight has been verified,
```
brapse, line 86: s/nextHeight/current_height

light-client/src/components/scheduler.rs
```diff
@@ -21,51 +32,81 @@ where
...
+/// and we can choose a height closer to the targetHeight. As we get the `light_store` as parameter,
+/// the choice of the next height can depend on the `light_store`, e.g., we can pick a height
+/// for which we have already downloaded a light block.
+/// - In Case ii. the header of `next_height` could not be verified, and we need to pick a smaller height.
```
brapse, line 90: same

light-client/src/components/verifier.rs
```diff
@@ -26,35 +24,29 @@ impl From<Result<(), VerificationError>> for Verdict {
...
 
+/// Production implementation of the verifier.
+///
+/// For testing purposes, this implementation is parametrized by:
+/// - A set of predicates used to validate a light block
+/// - A voting power calculator
+/// - A commit validator
+/// - A header hasher
+///
+/// For regular use, one can summon an implementation with `ProdVerifier::default()`.
```
brapse, line 49: s/summon/construct? 🧙

light-client/src/store/sled/utils.rs
```diff
@@ -117,10 +117,36 @@ where
...
+//         kv.insert(&db, &4, &4).unwrap();
+
+//         let mut iter = kv.iter(&db);
+//         assert_eq!(iter.next_back(), Some(589473798493));
+//         assert_eq!(iter.next_back(), Some(12342425));
+//         assert_eq!(iter.next_back(), Some(4));
+//         assert_eq!(iter.next_back(), Some(1));
+//         assert_eq!(iter.next_back(), None);
+//     }
+// }
```
Liamsi, line 152: Why are these commented out? Maybe better to add a comment or delete.

light-client/src/components/scheduler.rs
romac: Sure, that's indeed a better name. I chose `basic` here to express that this scheduler implementation does not do anything fancy for performance (like picking heights already present in the store, etc.)

romac: So perhaps, `basic_bisecting_schedule` or something like that?

light-client/src/components/verifier.rs
romac: Right, the default instance is indeed not fancy enough to warrant such witchcraft ;)

light-client/src/components/scheduler.rs
romac: @ebuchman suggested getting rid of the prelude entirely for better discoverability so that's perhaps not the right place to put those references. The crate-level documentation in `lib.rs` might be a better fit.

light-client/src/store/sled/utils.rs
romac: Yeah this test is currently disabled because it fails on CI as we don't have access to this temp dir. I need to read up a bit on the GitHub Actions infra to figure out how to specify a proper temp dir, so I will just add a comment in the meantime.

romac:
Thanks for the comments, just addressed them. Feel free to merge once CI goes through (modulo the `test-integration-ignore` job that's currently broken).

Liamsi:
> (modulo the test-integration-ignore job that's currently broken).

You should be able to merge the PR despite the failing integration tests (I think it's OK to ignore these for now).

PR #285, opened 2020-05-28
Light Client: Fork Detection v1

melekes:
TODO:
- [ ] verification
- [x] better errors

### DIFF
--- a/light-client/src/components/fork_detector.rs
+++ b/light-client/src/components/fork_detector.rs
@@ -6,12 +6,12 @@ use crate::prelude::*;
 pub enum ForkDetection {
     // NOTE: We box the fields to reduce the overall size of the enum.
     //       See https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
-    Detected(Box<LightBlock>, Box<LightBlock>),
+    Detected(Vec<Box<LightBlock>>),
     NotDetected,
 }
 
 pub trait ForkDetector {
-    fn detect(&self, light_blocks: Vec<LightBlock>) -> ForkDetection;
+    fn detect(&self, verified_block: &LightBlock, light_blocks: Vec<LightBlock>) -> ForkDetection;
 }
 
 pub struct RealForkDetector {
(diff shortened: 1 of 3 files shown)


melekes:
Looks like bisection algorithm is using a trusted store and not abstract enough for us to run it against a witness (i.e. check that we can bisect from latest trusted state to given header using a witness). This will require some changes.

brapse:
Right. In the latest it [iteration](https://github.com/brapse/light-spike/blob/master/src/light_client.rs) The we haven't solidified the API on the `witnesses` or the `lightstore` so it's great to surface these needs now. Our current thinking is for each `witness` primary or otherwise to have a `verify_to_target` method and it's own trusted store.  We will have to expose an API on the witness's store to trust a header if it matches the verified header on the primary. In this case it might make sense for the latest trusted to be a method on the witness and abstract the storage details of the underlying storage, roughly thinking 🤔

romac:
I have a very much WIP implementation of fork detection roughly based on the [draft spec](https://github.com/informalsystems/tendermint-rs/pull/287) here: https://github.com/informalsystems/tendermint-rs/blob/c1b675510966b47654828d9ad2d374768a2fceed/light-client/src/fork_detector.rs#L33-L75

I currently just expose the `LightClient`'s `LightStore` but as you say, having a specific method on the `LightClient` which returns the latest state would be a lot better.

I also [added the ability](https://github.com/informalsystems/tendermint-rs/blob/c1b675510966b47654828d9ad2d374768a2fceed/light-client/src/light_client.rs#L167-L171) to specify a `State` (which includes the `LightStore`) to perform verification with. I am not so happy with that solution at the moment, since it requires [a hack](https://github.com/informalsystems/tendermint-rs/blob/c1b675510966b47654828d9ad2d374768a2fceed/light-client/src/light_client.rs#L146-L151) to work around the double borrow that arises when we want to work over the light client's current state.

Let's try to sync up tomorrow to discuss the current state of things, as well as the discrepancies between this PR's implementation and the specification?

PS: Great job by the way :)

ebuchman:
In the prior version managing the store was a more external responsibility, so all the internal bisection/verification functions didn't depend on it - ie. the `verify_bisection` took a trusted state and returned a list of new trusted states that the caller could add to the store. This way bisection could easily be done against any peer, without reference to a store. Does it make sense to do something like that? Or maybe to use something like a fresh mem-store for each  bisection run and then sync the mem-store to the underlying db store afterwards as necessary.

romac:
I guess both solutions would work, but I would perhaps favor the one where we pass the store explicitly as it's a bit simpler and more versatile. That's basically what we are doing now in the [supervisor branch](https://github.com/informalsystems/tendermint-rs/tree/romain/supervisor-state), but for now we are still keeping the internal store around, hence the need for the `mem::replace` trick.

Nonetheless, I think it's still worth giving your second solution more thought. For example, we could keep a log of the operations performed on it so that we can reapply them in order on the disk-based store. But this wouldn't solve the issue of supplying the light client with a pre-populated store, like the fork detector requires.

ebuchman:
> But this wouldn't solve the issue of supplying the light client with a pre-populated store, like the fork detector requires.

Why does the fork detector require a prepopulated store?

romac:
> Why does the fork detector require a prepopulated store?

For fork detection, each witness needs a custom light store which only contains the trusted state (marked as verified) and the light block to verify (marked as unverified), you can see [the relevant code here](https://github.com/informalsystems/tendermint-rs/pull/302/files#diff-6a0d7abe1681a600d0f9b2212fa7a1b6R63-R73).

As you can see in the code, since my comment above I changed the light client to not maintain its own state at all, but to rather have it passed as an argument to `verify_to_target`. It's now the responsibility of the supervisor to maintain a mapping between the peers and their light client instance + state.

romac:
I think we can close this PR since it's been integrated within the [supervisor PR](https://github.com/informalsystems/tendermint-rs/pull/302).

The one difference between both the two implementation is that this PR's version compares header hashes, while the supervisor one compares header for equality, as per the fork detection spec. We should discuss which is best somewhere.

melekes:
thanks @romac!

PR #287, opened 2020-05-29
Lightclient detection English Spec draft

josef-widder:
I am still working on a formal treatment of forks to precisely capture the temporal properties, but I guess the spec is good enough to start discussing.

Note that the detector has no "sequential" spec, as it does not make sense to define a detector in a fault-free sequential system.

[Rendered](https://github.com/informalsystems/tendermint-rs/blob/josef/lightclient-detection/docs/spec/lightclient/detection.md)


docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,393 @@
...
+				// header matches. we do nothing.
+	    }
+		else {
+			    // [LCD-REQ-REP]
+			    // header does not match. there is a situation.
+				// we try to verify sh by querying s
+				// we set up an auxiliary lightstore with the highest
+			    // trusted lightblock and the lightblock we want to verify
+				auxLS.Init
+				auxLS.Update(LightStore.LatestTrusted(),StateVerified);
```
melekes, line 318: why do we update the current state to verified here? not sure I understand

josef-widder: Here, I just initialize an auxiliary light store to pass into "VerifyToTarget". The light block LightStore.LatestTrusted() by definition of LatestTrusted has been verified by the verification, so it is OK to do that

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,487 @@
...
+> the following will do:
+
+Let *b* and *c* be two light blocks. We define **supports(b,c,t)** 
+    iff `ValidAndVerified(b, c)` at time *t*
+
+
+> The following formalizes that *b* was properly generated by
+> Tendermint; *b* can be traced back to genesis
+
+#### **[TMBC-SEQ-ROOTED]**
```
OStevan, line 86: Based on the definitions below it looks to me like [TMBC-SEQ-ROOTED] and [TMBC-SKIP-ROOTED] are the other way around.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,487 @@
...
+> the intuition behind **main chain forks**
+> in the document on [forks][tendermintfork]. However, main chain
+> forks were defined more operational "forks that are observed by
+> full nodes as part of normal Tendermint consensus protocol". Please
+> confirm! 
+  
+#### **[TMBC-SIGN-UNIQUE]**
+Let *b* and *c* be  light blocks, we define *sign-unique(b,c) =
+true* iff
+   - *a.Header.Height =  b.Header.Height* and
```
OStevan, line 139: I think this is an error, I don't see how we got to the implication based on `a` and `b`

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,487 @@
...
+> chain but to a light client
+> fork if *t+1 < f < 2t+1*.  
+> F4 and F5 can also lead to bogus blocks
+
+
+
+### Informal Problem statement
+
+> We put tags to informal problem statements as there is no sequential
+> secification.
```
OStevan, line 186:
```suggestion
> specification.
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,487 @@
...
+
+#### **[LCD-IP-Q]**
+
+Whenever the light client verifier performs `VerifyToTarget` with the
+primary and returns with
+`(lightStore, ResultSuccess)`, the
+ detector should query the secondaries by calling `FetchLightBlock` for height
+ *LightStore.LatestVerified().Height* remotely.  
+Then, 
+the detector returns the set of all headers $h'$ downloaded from
```
OStevan, line 209:
```suggestion
the detector returns the set of all headers *h'* downloaded from
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,487 @@
...
+ - *h'* is a (light) fork
+
+
+#### **[LCD-IP-PEERSET]**
+
+Whenever the detector observes misbehavior of a full node from the set
+of Secondaries it should be replaced by a fresh full node.  (A full
+node that has not been primary or secondary before). This includes in
+particular the case where *h'* is different from
+*LightStore.LatestVerified()* but h' is not a fork.
```
OStevan, line 221:
```suggestion
*LightStore.LatestVerified()* but *h'* is not a fork.
```

docs/spec/lightclient/detection.md
josef-widder: sorry, I guess a should be c. I fix that!

docs/spec/lightclient/detection.md
josef-widder: I am not sure to understand your comment. Is it about the labels?

OStevan: Sorry, the explanation is not that good. What I meant is that in the **TMBC-SEQ-ROOTED** **sequ-rooted** such that there exists a block of height *i* between 1 and *h* which supports `supports(a(i), a(i+1), t), for 1 <= i < h`. This definition in my opinion is closer to skipping instead of sequential as the naming would suggest.

josef-widder: Ah! I read "for 1 <= i < h" as "for all i, 1 <= i < h". That's how Knuth writes it ;-) But if it is confusing, I will make this more clear.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,486 @@
...
+> Observe that a light client fork is a special case of a slashable
+> fork. 
+
+
+#### **[TMBC-LC-FORK]**
+Let *a*, *b*, *c*, be light blocks and *t* a time. We define
+*light-client-fork(a,b,c,t)* iff
+   - *sign-skip-match(a,b,c,t) = false* and
+   - *sequ-rooted(b)* and
+   - *b* is "unique", that is, for all *d*,  *sequ-rooted(d)* and
```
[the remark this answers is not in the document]
josef-widder: My understanding is that in many discussions, by a "light client fork" people understand that "there is no fork on the chain", but someone tries to fool the light client. That there is no fork on the main chain is captured by "unique". 
The *slashable fork* does not require this uniqueness, and describes more general scenarios.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,486 @@
...
+> in the document on [forks][tendermintfork]. However, main chain
+> forks were defined more operational "forks that are observed by
+> full nodes as part of normal Tendermint consensus protocol". Please
+> confirm! 
+  
+#### **[TMBC-SIGN-UNIQUE]**
+Let *b* and *c* be  light blocks, we define *sign-unique(b,c) =
+true* iff
+   - *b.Header.Height =  c.Header.Height* and
+   - *rooted(b)* and
```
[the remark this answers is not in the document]
josef-widder: I did not mean to imply this or define what needs to be checked. I agree with you that we just need a common ancestor, but this common ancestor *a* itself need to satisfy *sequ-rooted(a)*, which then by transitivity goes to b and c. 
But you make an important point! As next step, I will have to be more precise, what checks follow from these definitions.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,486 @@
...
+
+#### **[TMBC-SIGN-SKIP-MATCH]**
+Let *a*, *b*, *c*, be light blocks and *t* a time, we define 
+*sign-skip-match(a,b,c,t) = true* iff
+   - *sequ-rooted(a)* and
+   - *sequ-rooted(b)* and
+   - *b.Header.Height = c.Header.Height* and
+   - *skip-root(a,c,t)*
+
+implies *b = c*.
```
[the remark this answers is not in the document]
josef-widder: Exactly. What I want to capture here is that if the predicate "sign-skip-match(a,b,c,t)" is true at time t, the blocks a, b, c do not constitute a fork. If the predicate evaluates to false, one may use a, b, and c to prove the existence of a fork.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,486 @@
...
+> of block b. However, c is of that height, is different, and passes skipping
+> verification  
+> Observe that a light client fork is a special case of a slashable
+> fork. 
+
+
+#### **[TMBC-LC-FORK]**
+Let *a*, *b*, *c*, be light blocks and *t* a time. We define
+*light-client-fork(a,b,c,t)* iff
+   - *sign-skip-match(a,b,c,t) = false* and
```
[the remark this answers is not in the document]
josef-widder: *sign-skip-match(a,b,c,t)* is defined as an implication. If  *sign-skip-match(a,b,c,t)=false* this means that `b != c` (and the precondition of the implication is true), that is, there is a fork.

romac:
@josef-widder Could you please motivate the need to maintain both a list of full nodes and a set of  secondaries of a given size? Is it for performance reasons?

josef-widder:
> @josef-widder Could you please motivate the need to maintain both a list of full nodes and a set of secondaries of a given size? Is it for performance reasons?

I guess this was the assumption from the beginning of the discussions: (1) Most likely the light client will easily be able to obtain a (sizeable) list of full nodes, and for the first version we (2) only want to talk to a small number of secondaries.

At later stages, we might want to think more precisely about how to maintain the list of full nodes (somewhat like PEX) and randomly pick some secondaries (perhaps according to some strategy) in order to provide probabilistic guarantees about being connected to at least one **correct** full node.

Does that help?

romac:
> @josef-widder: Does that help?

It does, makes sense, thanks!

docs/spec/lightclient/verification.md
```diff
@@ -0,0 +1,925 @@
...
+
+const (
+	StateUnverified = iota + 1
+	StateVerified
+	StateFailed
+	StateTrusted
+)
+```
+
+> Only the detector module sets a lightBlock state to `StateTrusted`
```
romac, line 477: It is not clear what is the purpose of this `Trusted` state, since it does not appear anywhere in this spec except in this definition, nor in the verification spec.

romac: Edit: except in this definition and the postcondition of `LatestVerified`

docs/spec/lightclient/detection.md
OStevan: Oh, ok than it makes sense :D

docs/spec/lightclient/verification.md
josef-widder: The idea was that `StateTrusted` is the state after detection was passed. Since we define the data structure I have put it here. But I forgot to set the state at the end of detection. I will add that now! Thanks!!

josef-widder:
> I am still working on a formal treatment of forks to precisely capture the temporal properties, but I guess the spec is good enough to start discussing.

Addresses: https://github.com/tendermint/spec/issues/103

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+leave other uses of the detection mechanism (e.g., when run on a full
+node) to the future.
+
+
+## Context of this document
+
+The light client verification specification [[verification]] is
+designed for the Tendermint failure model (1/3 assumption)
+[TMBC-FM-2THIRDS]. It is safe under this assumption, and live
+if it can reliably (that is, no message loss, no duplication, and
```
milosevic, line 42: Maybe remove clarification from brackets.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+#### **[TMBC-GENESIS]**
+Let *Genesis* be the agreed-upon initial block (file).
+
+#### **[TMBC-FUNC]**
+> **TODO** be more precise. +2/3 of b.NextV = c.Val signed c. For now
+> the following will do:
+
+Let b and c be two light blocks
+with *b.Header.Height + 1 = c.Header.Height*. We define **signs(b,c)**
+    iff `ValidAndVerified(b, c)`
```
milosevic, line 74: Where is ValidAndVerified defined?

josef-widder: In https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification.md. I should have added a pointer...

milosevic: Just link it.

docs/spec/lightclient/detection.md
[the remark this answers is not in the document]
milosevic: Is this formalism for intuition that if you trust a(a is sequ-rooted) and skip-root(a,c,t) is true, then c must be also obtained by the sequential verification?

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+   - *sequ-rooted(a)* and
+   - *sequ-rooted(b)* and
+   - *b.Header.Height = c.Header.Height* and
+   - *skip-root(a,c,t)*
+
+implies *b = c*.
+
+If there exists three light blocks a, b, and c, with 
+*sign-skip-match(a,b,c,t) =
+false* then we have a *slashable fork*.
```
milosevic, line 125: I am wondering if this is too strong. Isn't it sufficient to have skip-root(a,b,t) and skip-root(a,c,t) and b.height = c.height to have a slashable fork? Or we need to have also a block d such that d.height = c.height and sequ-rooted(d)?

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+     *d.Header.Height=b.Header.Height* implies *d = b*
+
+
+> Finally, let's also define bogus blocks that have no support.
+> Observe that bogus is even defined if there is a fork on the chain.
+
+#### **[TMBC-BOGUS]**
+Let *b* be a light block and *t* a time. We define *bogus(b,t)* iff
+  - *sequ-rooted(b) = false* and
+  - for all *a*, *sequ-rooted(a)* implies *skip-root(a,b,t) = false*
```
milosevic, line 171: Should a.height be smaller than b.height here?

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+#### **[LCD-IP-STATE]**
+
+The detector works on a LightStore that contains LightBlocks in one of 
+the state `StateUnverified`, ` StateVerified`, `StateFailed`, or
+`StateTrusted`.
+
+
+#### **[LCD-IP-Q]**
+
+Whenever the light client verifier performs `VerifyToTarget` with the
```
milosevic, line 203: We should clarify what is the state of light blocks required to be used for VerifyToTarget?

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+primary and returns with
+`(lightStore, ResultSuccess)`, the
+ detector should query the secondaries by calling `FetchLightBlock` for height
+ *LightStore.LatestVerified().Height* remotely.  
+Then, 
+the detector returns the set of all headers *h'* downloaded from
+secondaries that satisfy
+ - *h'* is different from *LightStore.LatestVerified()*
+ - *h'* is a (light) fork
```
milosevic, line 212:
```suggestion
 - *h'* is a (light) fork.
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+
+
+#### **[LCD-IP-PEERSET]**
+
+Whenever the detector observes misbehavior of a full node from the set
```
milosevic, line 217: What kind of misbehavior is meant here?

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+the assumptions that correct full nodes reliably talk to the 
+detector.
+
+
+### Assumptions
+
+#### **[LCD-A-CorrFull]**
+
+At all times there is at least one correct full
+node among the primary and the secondary.
```
milosevic, line 247:
```suggestion
node among the primary and the secondaries.
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+the verification conditions [LCD-DIST-*] have preconditions on specific
+cases where primary and/or secondaries are faulty.
+
+#### **[LCD-A-RelComm]**
+
+Communication between the  detector and a correct full node is 
+reliable and bounded in time. Reliable communication means that
+messages are not lost, not duplicated, and eventually delivered. There
+is a (known) end-to-end delay *Delta*, such that if a message is sent
+at time *t* then it is received and processes by time *t + Delta*.
```
milosevic, line 259:
```suggestion
at time *t* then it is received and processed by time *t + Delta*.
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+
+
+## (Distributed) Problem statement
+
+> As the fork detector from the beginning is there to reduce the
+> impact of faulty nodes, and faulty nodes imply that there is a
+> distributed system, there is no sequential specification.
+
+The  detector gets as input a lightstore *lightStore*.
+Let *h-target = lightStore.LatestVerified().Height* and
```
milosevic, line 274: Should we call it h-verified?

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+     *h-trust=lightStore.LatestTrusted().Height* (see
+     [LCV-DATA-LIGHTSTORE]).
+It queries the secondaries for  headers at height *h-target*.
+The  detector returns a set *Forks*, and should satisfy the following
+     temporal formulas
```
milosevic, line 279:
```suggestion
     temporal formulas:
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+The problem is solved by calling  the function `ForkDetector` with
+a lightstore that contains a light block that has
+just been verified by the verifier. 
+
+
+```go
+func ForkDetector(ls LightStore)  {
+    Forks.Init // initialize a container in which we collect forks
+	for i, s range Secondaries {
+		sh := FetchLightBlock(s,LightStore.LatestVerified().Height)
```
milosevic, line 395: Maybe extract `LightStore.LatestVerified()` into a variable.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+		else {
+			    // [LCD-REQ-REP]
+			    // header does not match. there is a situation.
+				// we try to verify sh by querying s
+				// we set up an auxiliary lightstore with the highest
+			    // trusted lightblock and the lightblock we want to verify
+				auxLS.Init
+				auxLS.Update(LightStore.LatestTrusted(), StateVerified);
+				auxLS.Update(sh,StateUnverified);
+				result := VerifyToTarget(s, auxLS, LightStore.LatestVerified().Height)
```
milosevic, line 410:
```suggestion
				result := VerifyToTarget(s, auxLS, sh.Header.Height)
```

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+				if result = (_,ResultSuccess) || (_,EXPIRED) {
```
milosevic, line 411: We should clarify what is the expectation of VerifyToTarget so if it returns TimeoutError it can be assumed faulty. I guess that VerifyToTarget with correct full node should never terminate with TimeoutError.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+				    // we verified header sh which is conflicting to hd
+					// there is a fork on the main blockchain.
+                    // If return code was EXPIRED it might be too late
```
milosevic, line 414: This shouldn't be the case. We should ensure that even if it timeout we can still have time to punish it.

docs/spec/lightclient/detection.md
```diff
@@ -0,0 +1,491 @@
...
+					// If a new node is added to secondaries, this
+					// should not imply an additional loop iteration.
+					// We assume one of the secondaries is correct.
+				}
+			}
+	}
+	if Forks.isEmpty {
+	    LightStore.Update(LightStore.LatestVerified(), StateTrusted)
+	}
+	return (LightStore,Forks,OK)
```
milosevic, line 430: I think that we are missing the main logic, i.e., describing what happens at the caller of ForkDetector. I guess in this case we report evidence and shut down light client?

josef-widder: Yes, we need to clarify this. I will add a TODO in the beginning of the document.

docs/spec/lightclient/detection.md
josef-widder: done. I called it `testedLB`. Not sure it is the best name...

docs/spec/lightclient/detection.md
josef-widder: I am not sure we can enforce this. If the detection starts late, and is based on old light blocks, this might always happen. I have added a TODO to check whether we can do something there.

docs/spec/lightclient/detection.md
josef-widder: will do

docs/spec/lightclient/detection.md
josef-widder: done

docs/spec/lightclient/detection.md
josef-widder: Timeout or if it provides a light block for which it cannot produce a proof (a sequence of other headers). That are the things we can detect locally. I will clarify this

docs/spec/lightclient/detection.md
josef-widder: yes. it is implied by the definitions. but I will add a remark for clarificiation.

docs/spec/lightclient/detection.md
milosevic: That's good point. I was only thinking about going forward case. But if you ask for older height (for example from RPC endpoint) that could be problematic. These timing assumptions are really tricky. I am more convinced that we need some kind of simplistic approach where we don't provide any "service" for headers that are not within some constant of end of expiration period. It basically means enforcing synchrony assumptions at the request time, and being sure that we are within big safe period (order of days) so we never play close to the expiration point. And we do this at the trust period and then diff between trust and unbonding period gives time for evidence processing and fork accountability.

josef-widder:
> After the conflicts with master are resolved I think we are good for this initial version :shipit:

Yes, I agree. I will make sure I address all of Zarko's comments tomorrow morning (even if it is just puttind in TODOs) and then I will take care of conflicts and merge.

docs/spec/lightclient/detection.md
josef-widder: I guess, this should rather go to the part on the main logic. We need to discuss where to put it. I guess we should do a light client spec that uses the functions provided in the verification and detection specs

docs/spec/lightclient/detection.md
josef-widder: Yes! Good catch!!

PR #288, opened 2020-06-01
Make serialization of CommitSig, TrustedState symmetric

### LINKED ISSUE #286: Serialization of CommitSig is not symmetric
Since `CommitSig` is included in `TrustedState`, so it'll need to be persisted, which means serialize and deserialize back.
But [currently](https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/src/block/commit_sig.rs#L12), it's deserialized through `RawCommitSig` but serialized as itself, so it'll fail to deserialize the serialized result.

Proposal:
- serialize through `RawCommitSig` too
```
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-#[serde(try_from = "RawCommitSig")]
+#[serde(try_from = "RawCommitSig", into = "RawCommitSig")]
```
- add serialization roundtrip test of `TrustedState` in `tendermint/tests/lite.rs`.

If it's good to go, I can send the PR.

yihuang:
Fix #286

### DIFF
--- a/tendermint/src/block/commit_sig.rs
+++ b/tendermint/src/block/commit_sig.rs
@@ -9,7 +9,7 @@ use std::convert::TryFrom;
 /// CommitSig represents a signature of a validator.
 /// It's a part of the Commit and can be used to reconstruct the vote set given the validator set.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-#[serde(try_from = "RawCommitSig")]
+#[serde(try_from = "RawCommitSig", into = "RawCommitSig")]
 pub enum CommitSig {
     /// no vote was received from a validator.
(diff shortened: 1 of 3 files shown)


tendermint/src/block/commit_sig.rs
```diff
@@ -9,7 +9,7 @@ use std::convert::TryFrom;
 /// CommitSig represents a signature of a validator.
 /// It's a part of the Commit and can be used to reconstruct the vote set given the validator set.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-#[serde(try_from = "RawCommitSig")]
```
Liamsi, line 12: 👍

PR #295, opened 2020-06-03
light tests: remove assertions about number of bisections

### LINKED ISSUE #276: Light client testing: remove implementation specific assertions about number of bisections
JSON tests: remove assertions for number of bisections as they are an implementation detail
see: https://github.com/informalsystems/tendermint-rs/pull/273#discussion_r426203592

Liamsi:
Quickfix that closes #276 (I had this change uncommited in my workspace and just pushed it)

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/tests/lite.rs
+++ b/tendermint/tests/lite.rs
@@ -63,7 +63,6 @@ struct TestBisection {
     height_to_verify: Height,
     now: Time,
     expected_output: Option<String>,
-    expected_num_of_bisections: i32,
 }
 
 #[derive(Deserialize, Clone, Debug)]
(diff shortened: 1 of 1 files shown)


romac:
Should we remove the field from the JSON as well?

Liamsi:
> Should we remove the field from the JSON as well?

The JSON is generated from the go-code. I wouldn't manually remove it here but let @Shivani912 change the generator code and re-generate the JSON files instead.

romac:
> I've got to learn more about bisections and why different implementations would require such different number of bisections but this change in itself is fine. You're removing unused code.

@josef-widder touched on that in the [verification spec](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification.md):

> `Schedule` decides which height to try to verify next. We keep this underspecified as different implementations (currently in Goland and Rust) may implement different optimizations here. We just provide necessary conditions on how the height may evolve.

As an example, an implementation of `Schedule` could realize that, even if the optimal block to bisect to is of height H, there is already a block of height H+N (for a small N) in the store, and decide to pick this one to avoid fetching it over the network. This could lead to a different number of bisections.

Issue #211, opened 2020-04-10
tendermint/0.33 / #196 followup issue


Liamsi:
Collecting all outstanding discussions for full 0.33 compatibility and review comments from #196 and #199 and related issues here:

- [x] Add changelog entries before release
- [x] address @melekes comments: https://github.com/informalsystems/tendermint-rs/pull/196#pullrequestreview-394541450 (addressed in #296)
---
 - [x] JSON / RPC issues: https://github.com/informalsystems/tendermint-rs/pull/196#issuecomment-611969478 and https://github.com/informalsystems/tendermint-rs/pull/196#issuecomment-611982250
- [x] investigate why above isn't detected by the integration tests

@ancazamfir re-tested this and it 
> Seems ok now

This was likely resolved as part of the serialization refactoring @greg-szabo did.

---
- [x] tests for jsonrpc ID: https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r399790732 (done in #298 )
- [x] err handling weirdness: https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r399782582 (this got resolved / outdated in the several serialzation improvement PRs and using blockIfFlags)
- [x] review of #199: https://github.com/informalsystems/tendermint-rs/pull/199#discussion_r401985251, https://github.com/informalsystems/tendermint-rs/pull/199#discussion_r401987638, https://github.com/informalsystems/tendermint-rs/pull/199#discussion_r406830449 (time was addressed in #277 and the remaining work was done in the commitsig refactoring PRs)
 - [x] delete AsRef completely: https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r400539671 (was done after 0.33 PR was merged)
- [x] move Proof: https://github.com/informalsystems/tendermint-rs/pull/206#pullrequestreview-390017106 (done in: https://github.com/informalsystems/tendermint-rs/pull/297)

Liamsi:
I think https://github.com/informalsystems/tendermint-rs/pull/196#issuecomment-611969478 is resolved. At least I would think this would be caught by the integration tests 🙏

Liamsi:
>  tests for jsonrpc ID: #196 (comment)

This would become obsolete if we switched to using an jsonrpc library instead. Otherwise the tests I had in mind would look similar to: https://github.com/paritytech/jsonrpc/blob/ef92a481d76802da489f0f57dfdf4d459d7f7f76/core/src/types/id.rs#L16

[...]

PR #298, opened 2020-06-03
Add tests for jsonrpc id

Liamsi:
https://github.com/informalsystems/tendermint-rs/pull/196#discussion_r399790732

part of https://github.com/informalsystems/tendermint-rs/issues/211 

* [x] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/rpc/id.rs
+++ b/tendermint/src/rpc/id.rs
@@ -29,3 +29,27 @@ impl Id {
         Id::Str(uuid.to_string())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test::test_serialization_roundtrip;
+
+    #[test]
+    fn round_tripping_id() {
+        let str = r#""42""#;
+        test_serialization_roundtrip::<Id>(str);
+
+        let str2 = r#""936DA01F-9ABD-4D9D-80C7-02AF85C822A8""#;
+        test_serialization_roundtrip::<Id>(str2);
+
+        let num = r#"42"#;
+        test_serialization_roundtrip::<Id>(num);
+
+        let zero = r#"0"#;
+        test_serialization_roundtrip::<Id>(zero);
+
+        let null = r#"null"#;
+        test_serialization_roundtrip::<Id>(null);
+    }
+}
--- a/tendermint/src/rpc/id.rs
+++ b/tendermint/src/rpc/id.rs
@@ -36,7 +36,7 @@ mod tests {
     use crate::test::test_serialization_roundtrip;
 
     #[test]
-    fn round_tripping_id() {
+    fn round_tripping_jsonrpc_id() {
         let str = r#""42""#;
         test_serialization_roundtrip::<Id>(str);
 


tendermint/src/rpc/id.rs
```diff
@@ -29,3 +29,27 @@ impl Id {
...
+        test_serialization_roundtrip::<Id>(str2);
+
+        let num = r#"42"#;
+        test_serialization_roundtrip::<Id>(num);
+
+        let zero = r#"0"#;
+        test_serialization_roundtrip::<Id>(zero);
+
+        let null = r#"null"#;
+        test_serialization_roundtrip::<Id>(null);
```
Liamsi, line 53: @greg-szabo didn't you say that `null` doesn't roundtrip properly? Or did I dream this? It does in fact roundtrip.

greg-szabo: I said that but that's only the case in a JSON->Object->JSON roundtrip. Test_serialization does an Object->JSON->Object roundtrip, which is fine.

tendermint/src/rpc/id.rs
```diff
@@ -29,3 +29,27 @@ impl Id {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test::test_serialization_roundtrip;
+
+    #[test]
+    fn round_tripping_id() {
```
greg-szabo, line 39: Can we name this so it's more on point?

```suggestion
    fn json_id() {
```

Liamsi: I can change it but note that this test will actually shows up as:
`test rpc::id::tests::round_tripping_id ... ok`

which seems clear that we are roundtrippin the rpc id (and all the rpc uses JSON rpc). Maybe that whole module should be renamed to `jsonrpc` though.

tendermint/src/rpc/id.rs
Liamsi: Still confusing because `test_serialization_roundtrip` actually does:
JSON->Object->JSON->Object

tendermint/src/rpc/id.rs
Liamsi: Is `round_tripping_jsonrpc_id` ok for you?

tendermint/src/rpc/id.rs
greg-szabo: Yes, but we don't compare the JSON objects, as far as I understand. Comparing the Rust objects is fine.

`None` is `None`, whether it came from an empty string or a `null` value. The same is not true in JSON.

tendermint/src/rpc/id.rs
greg-szabo: Yeah, I was overeager and didn't check the bigger picture. It's cool.

tendermint/src/rpc/id.rs
Liamsi: Thanks for clarifying!

PR #302, opened 2020-06-05
[light-client] Supervisor

romac:
* [ ] Wrote tests
* [ ] Updated CHANGES.md


light-client/src/supervisor.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+                _ => {
+                    // NoOp?
+                }
+            }
+        }
+    }
+}
+
+pub struct Handler {
```
Liamsi, line 230: I mentioned that in the call but also noting here: we might want to name this differently. I don't have a better suggestion yet 😬

light-client/src/supervisor.rs
```diff
@@ -0,0 +1,302 @@
...
+    sender: channel::Sender<Event>,
+}
+
+// Assume single handler
```
Liamsi, line 234: Does that mean, there will only one handler instance?

light-client/src/supervisor.rs
```diff
@@ -0,0 +1,302 @@
...
+impl Handler {
+    // How do we connect with the runtime?
```
Liamsi, line 236: Sounds like this should be a TODO?

light-client/src/prelude.rs
```diff
@@ -13,3 +13,7 @@ pub use crate::{
     store::{memory::*, sled::*, LightStore, VerifiedStatus},
     types::*,
 };
+
+pub fn todo<A>() -> A {
+    unreachable!()
+}
```
Liamsi, line 19: Is this used anywhere?

light-client/src/fork_detector.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        for secondary in secondaries {
+            let mut state = State::new(MemoryStore::new());
+
+            let secondary_block = secondary
+                .light_client
+                .get_or_fetch_block(light_block.height(), &mut state)
+                .unwrap(); // FIXME: unwrap
+
+            // TODO: Should hash the headers here instead of comparing them for equality?
```
Liamsi, line 58: I think that would be the right thing to do, yes! (theoretically comparing the headers vs comparing their hashes should be equivalent, practically that might differ although I don't think that can be the case here - still using the hashes would be consistent imo).

light-client/src/fork_detector.rs
```diff
@@ -0,0 +1,88 @@
...
+                .update(secondary_block.clone(), VerifiedStatus::Unverified);
+
+            let result = secondary
+                .light_client
+                .verify_to_target(light_block.height(), &mut state);
+
+            match result {
+                Ok(_) => forks.push(Fork::Forked(secondary_block)),
+                Err(e) if e.kind().has_expired() => forks.push(Fork::Forked(secondary_block)),
+                Err(_) => forks.push(Fork::Faulty(secondary_block)),
```
Liamsi, line 78: Wouldn't it be good to know the error here?

light-client/src/prelude.rs
romac: Not anymore, thanks for pointing it out!

light-client/examples/light_client.rs
```diff
@@ -85,13 +85,7 @@ fn sync_cmd(opts: SyncOpts) {
...
     }
 
-    let peers = Peers {
-        primary,
-        witnesses: Vec::new(),
-    };
-
-    let state = State {
-        peers,
+    let mut state = State {
```
brapse, line 88: It seems like we keep leaning on `use ...::prelude::*` which obscures the package structure and makes it pretty difficult to figure out where structs are coming from. From a comprehension perspective it may be better to just import the structs we need from their locations.

light-client/examples/light_client.rs
```diff
@@ -109,20 +103,11 @@ fn sync_cmd(opts: SyncOpts) {
...
-        verifier,
-        fork_detector,
-        io,
-    );
+
+    let mut light_client = LightClient::new(primary, options, clock, scheduler, verifier, io);
 
     loop {
-        match light_client.verify_to_highest() {
+        match light_client.verify_to_highest(&mut state) {
```
brapse, line 110: If the state is `Instance`  specific would it make sense to just make it internal and have the `Instance` expose specific methods for updating the latest trusted state? This might provide better encapsulation. 🤔

light-client/src/supervisor.rs
brapse: I think this is called the Facade pattern. Dunno how comfortable we feel being that pretentious. 🧐

light-client/src/supervisor.rs
brapse: This is a comment from the sketch that should probably be removed 😅

light-client/src/supervisor.rs
brapse: Another artifact from the spike. 😅

light-client/src/supervisor.rs
```diff
@@ -0,0 +1,302 @@
...
+            ForkDetection::NotDetected => None,
+        }
+    }
+
+    pub fn handler(&mut self) -> Handler {
+        Handler::new(self.sender.clone())
+    }
+
+    // Consume the instance here but return a runtime which will allow interaction
+    // Maybe return an output channnel here?
```
Liamsi, line 209:
```suggestion
    // Maybe return an output channel here?
```

light-client/src/supervisor.rs
```diff
@@ -0,0 +1,302 @@
...
+            ForkDetection::Detected(forks) => Some(forks),
+            ForkDetection::NotDetected => None,
+        }
+    }
+
+    pub fn handler(&mut self) -> Handler {
+        Handler::new(self.sender.clone())
+    }
+
+    // Consume the instance here but return a runtime which will allow interaction
```
Liamsi, line 208: IMO it would help immensely, if there was a simple test / example that shows how to actually interact with the instance. e.g. maybe a call to `verify_to_target_async` with a custom callback?

light-client/src/supervisor.rs
Liamsi: Hmm, SthSthFacade wouldn't really help I guess and it sounds like we werehiding a lot of complexity behind that facade (but the design you and @romac came up with looks actually simple, in a good way). 

Maybe as soon as we see both the supervisor and handler in action we can find a better name for the latter 🤔 
From a data pov the handler is really just a wrapper around a sending channel. Similarly the `handler` method "just" enables you to access that channel.  We can brainstorm naming in a followup issue or on a call I guess.

light-client/src/supervisor.rs
brapse: Agreed. I think we are porting over the unit tests now (as part of #280) but created a specific issue for that kind of test (#316). The use case that motivated (but may not be necessary) Was for the relayer which wants to accumulate events from its subscription while it waits for the light client. Here is the pseudo-ish code which we used to orient the discussion. 

```rust
// Snippet from  Relayer event loop
let light_client = ... // some light client handler
let (sender, receiver) = ???;

while Some(event) = recever.recv() {
    match event {
        // Assume this is from subscription
        Event::NewHeader(header) => {
            let callback = sender.clone();
            light_client.validate_async(header, move |result| {
                match result {
                    Ok(header) => {
                        callback.send(Event::ValidatedHeader(header)))
                    },
                    Err(err) => {
                        callback.send(Event::FailedValidation(err));
                    }
                }
            });
        },
        Event::ValidatedHeader(header) => {
           ...
        }
        Event::FailedValidation(err) => {
            // Shut down the relayer
        }
    }
}
```

light-client/src/light_client.rs
```diff
@@ -46,48 +47,53 @@ impl Options {
...
+        let target_block = match self.io.fetch_light_block(self.peer, LATEST_HEIGHT) {
             Ok(last_block) => last_block,
             Err(io_error) => bail!(ErrorKind::Io(io_error)),
         };
 
-        self.verify_to_target(target_block.height())
+        self.verify_to_target(target_block.height(), state)
     }
 
     /// Attemps to update the light client to a block of the primary node at the given height.
```
Liamsi, line 99:
```suggestion
    /// Attempts to update the light client to a block of the primary node at the given height.
```

light-client/src/supervisor.rs
romac: Yeah I'd this is more a handle (as in thread handle) than a handler (as in event handler). So what about `Handle` or `LightClientHandle`?

light-client/examples/light_client.rs
romac: Agreed, I plan on removing it asap.

light-client/src/supervisor.rs
Liamsi: Yes, handle is less confusing 👍

light-client/src/supervisor.rs
romac: See https://github.com/informalsystems/tendermint-rs/pull/302/commits/76857496d8be96344b1348d25b177d545715b4c7.

light-client/src/fork_detector.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        for secondary in secondaries {
+            let mut state = State::new(MemoryStore::new());
+
+            let secondary_block = secondary
+                .light_client
+                .get_or_fetch_block(light_block.height(), &mut state)
+                .unwrap(); // FIXME: unwrap
+
+            let primary_hash = self.header_hasher.hash(&light_block.signed_header.header);
```
melekes, line 62: should be done outside of for loop

light-client/src/fork_detector.rs
```diff
@@ -0,0 +1,97 @@
...
+                .hash(&secondary_block.signed_header.header);
+
+            if primary_hash == secondary_hash {
+                // Hashes match, continue with next secondary, if any.
+                continue;
+            }
+
+            state
+                .light_store
+                .update(trusted_state.clone(), VerifiedStatus::Verified);
```
melekes, line 74: why do we update the current state to verified here? not sure I understand

light-client/src/supervisor.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    #[pre(self.peers.primary().is_some())]
+    fn detect_forks(
+        &mut self,
+        light_block: &LightBlock,
+        trusted_state: &LightBlock,
+    ) -> Result<Option<Vec<Fork>>, Error> {
+        if self.peers.secondaries().is_empty() {
+            return Ok(None);
```
melekes, line 153: this seems very insecure to me.. even though we rely on 1 correct node assumption, my personal opinion is that we need at least 1 witness. we can't assume primary will always be correct.

light-client/examples/light_client.rs
```diff
@@ -109,27 +110,31 @@ fn sync_cmd(opts: SyncOpts) {
...
             Ok(light_block) => {
                 println!("[ info  ] synced to block {}", light_block.height());
             }
             Err(e) => {
                 println!("[ error ] sync failed: {}", e);
             }
-        }
+        });
+
         std::thread::sleep(Duration::from_millis(800));
```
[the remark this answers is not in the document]
melekes: fyi: in Go, we don't have a runtime because we don't know what is a single best runtime. for some, a loop like ^ might be sufficient (e.g. check for a new block every 5s). others may want to subscribe for new headers via WebSocket and verify headers as soon as there are committed by Tendermint.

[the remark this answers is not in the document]
melekes: although, this is an example, so feel free to ignore

light-client/src/fork_detector.rs
romac: The light client needs to have at least one trusted (`Verified`) block to perform verification.
We should perhaps provide a nicer API on `State` for interacting with the store, and make such things more explicit and higher-level.

Issue #290, opened 2020-06-02
Light client conformance tests differences between description and real errors


OStevan:
When running tests for light client I noticed that for `not_enough_commits` (in the current canonical implementation) and `invalid_validator_set` (the new one in `light-client`) the expected error based on the description is not the same as the error we get when the test is run.

`not_enough_commits`: fails with `InvalidNextValidatorSet` for height 5 instead of `InsufficientCommitPower` for height 6

`invalid_validator_set`: fails with `request for height 6` instead of invalid `InvalidNextValidatorSet` for height 11

OStevan:
I'll continue adding differences between errors we would expect based on the description and the ones which are currently registered in this issue.

OStevan:
`wrong_header_height`: fails with `InvalidCommitValue` while the description would suggest a height related error

Shivani912:
I'm looking into this now

Shivani912:
> When running tests for light client I noticed that for `not_enough_commits` (in the current canonical implementation) and `invalid_validator_set` (the new one in `light-client`) the expected error based on the description is not the same as the error we get when the test is run.
> 
> `not_enough_commits`: fails with `InvalidNextValidatorSet` for height 5 instead of `InsufficientCommitPower` for height 6
> 
> `invalid_validator_set`: fails with `request for height 6` instead of invalid `InvalidNextValidatorSet` for height 11

So as @OStevan explained, there were errors in `not_enough_commits` and `invalid_validator_set`. These have been fixed in PR #366 .




> `wrong_header_height`, `wrong_header_timestamp`, `wrong_last_block_id`: fails with `InvalidCommitValue` while the description would suggest a different error.

Well, description in tests explain "where" the error should be instead of "what". Specifying the error type could be tricky here because the above cases are essentially all about detecting differences in header and commit. So for the same error, you could say it's an invalid header. As long as it fails for the expected reason, I wouldn't care much about the error type.

OStevan:
@Shivani912 thanks for following up on this.

> Well, description in tests explain "where" the error should be instead of "what". Specifying the error type could be tricky here because the above cases are essentially all about detecting differences in header and commit. So for the same error, you could say it's an invalid header. As long as it fails for the expected reason, I wouldn't care much about the error type.

[...]

Issue #235, opened 2020-04-23
lite_impl: Commit#voting_power_in func is suboptimal


melekes:
`voting_power_in(&self, validators: &Set)` https://github.com/informalsystems/tendermint-rs/blob/efadd8b4a5032b951e764ad12b2e6aac8bd8e99e/tendermint/src/lite_impl/signed_header.rs#L15 function is suboptimal. 

You don’t have to calculate all the voting power. What you want is to verify that trust threshold (⅔+ in sequential verification or configurable ⅓+ in skipping verification) signed this commit.

In Go Tendermint, Verify function returns as soon as threshold is reached. https://github.com/tendermint/tendermint/blob/7b717c72df1e9d38cdaf110cdbdaa57459272dc1/types/validator_set.go#L664

ebuchman:
This will be especially relevant once #506 is done

marbar3778:
This could be further optimized by utilizing batch verification from dalek or zebra.

PR #306, opened 2020-06-08
Fix `VotingPowerCalculator::voting_power_in`

### LINKED ISSUE #282: `BlockIdFlagNil` votes counted when verifying a commit
Comparing the implementation of Light Client verification with the Go one, there is a difference in the calculation of the voting power for a commit.

Methods `verify_commit_full` and `verify_commit_trusting` in `lite/verifier.rs` use a method [`voting_power_in`](https://github.com/informalsystems/tendermint-rs/blob/afb11bf21fc3899edaf157b4f95e1b94a158215b/tendermint/src/lite_impl/signed_header.rs#L17) which sums up voting power for all validators in the commit which have a non absent vote.

Comparing this to the Go implementation in functions `VerifyCommit` and `VerifyCommitTrusting` in `types/validator_set.go` where voting power is counted by summing up only those which have a [`BlockIdFlagCommit`](https://github.com/tendermint/tendermint/blob/master/types/validator_set.go#L681) vote.

This then results in Rust implementation accepting commits which have no `BlockIdFlagCommit` votes but only `BlockIdFlagNil`.

### LINKED ISSUE #335: Don't add voting power of the same validator multiple times in voting_power_in
There is a TODO already in the code: https://github.com/informalsystems/tendermint-rs/blob/184a1f9925f541473313efcd483c32cb913f8af9/light-client/src/operations/voting_power.rs#L38

romac:
Fix #282 
Fix #335

 - Provide two helper methods for checking validators overlap and if there is enough trust w.r.t. the trusted state
- Only count `Commit` signatures towards the voting power (#282)
- Do not rely on `signed_votes` method to get the votes
- Only count the voting power of a validator once (#335)

### Notes

- [x] Needs to throw a proper error in some places
- [x] Trust threshold is currently hardcoded and should be rather passed in
- [x] The method should likely be renamed or its signature should be changed to reflect the new behavior
- [ ] ~~Most of logic and comments are taken from the Go code, so perhaps some attribution should be added as well~~
- [ ] ~~Tests are failing because the signatures currently fail to verify, likely because of mismatch with their validator~~

---

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -132,7 +132,7 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
             }
 
             // If the vote is neither absent nor nil, tally its power
-            if true || signature.is_commit() {
+            if signature.is_commit() {
                 tallied_voting_power += validator.power();
             } else {
                 // It's OK. We include stray signatures (~votes for nil)
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -55,10 +55,8 @@ pub trait VotingPowerCalculator: Send {
         untrusted_validators: &ValidatorSet,
     ) -> Result<VotingPower, VerificationError> {
         let trust_threshold = TrustThreshold::TWO_THIRDS;
-
         let voting_power =
             self.voting_power_of(untrusted_header, untrusted_validators, trust_threshold)?;
-
         if trust_threshold.is_enough_power(voting_power.tallied, voting_power.total) {
             Ok(voting_power)
         } else {
@@ -93,7 +91,7 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
         let mut tallied_voting_power = 0_u64;
         let mut seen_validators = HashSet::new();
 
-        for (idx, signature) in signatures.into_iter().enumerate() {
+        for (idx, signature) in signatures.iter().enumerate() {
             let vote = vote_from_non_absent_signature(signature, idx as u64, &signed_header.commit);
             let vote = match vote {
                 Some(vote) => vote,
@@ -162,8 +160,8 @@ fn vote_from_non_absent_signature(
             timestamp,
             signature,
         } => (
-            validator_address.clone(),
-            timestamp.clone(),
+            *validator_address,
+            *timestamp,
             signature.clone(),
             Some(commit.block_id.clone()),
         ),
@@ -171,12 +169,7 @@ fn vote_from_non_absent_signature(
             validator_address,
             timestamp,
             signature,
-        } => (
-            validator_address.clone(),
-            timestamp.clone(),
-            signature.clone(),
-            None,
-        ),
+        } => (*validator_address, *timestamp, signature.clone(), None),
     };
 
     Some(Vote {
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -39,7 +39,6 @@ pub trait VotingPowerCalculator: Send {
         untrusted_validators: &ValidatorSet,
         trust_threshold: TrustThreshold,
     ) -> Result<VotingPower, VerificationError> {
-        println!("check_validators_overlap");
         let voting_power =
             self.voting_power_of(untrusted_header, untrusted_validators, trust_threshold)?;
 
@@ -55,7 +54,6 @@ pub trait VotingPowerCalculator: Send {
         untrusted_header: &SignedHeader,
         untrusted_validators: &ValidatorSet,
     ) -> Result<VotingPower, VerificationError> {
-        println!("check_signers_overlap");
         let two_thirds = TrustThreshold::new(2, 3).unwrap();
         let voting_power =
             self.voting_power_of(untrusted_header, untrusted_validators, two_thirds)?;
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -135,7 +135,8 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
                 // to measure validator availability.
             }
 
-            // TODO: Break out when we have enough voting power
+            // TODO: Break out of the loop when we have enough voting power.
+            // See https://github.com/informalsystems/tendermint-rs/issues/235
         }
 
         let voting_power = VotingPower {
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -101,28 +101,16 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
                 None => continue, // Ok, some signatures can be absent
             };
 
+            // Ensure we only count a validator's power once
             if seen_validators.contains(&vote.validator_address) {
-                println!(
-                    "  > already seen vote from this validator: {}",
-                    vote.validator_address
-                );
-
                 continue;
             } else {
                 seen_validators.insert(vote.validator_address);
             }
 
-            // TODO: Check that we didn't see a vote from this validator twice...
             let validator = match validator_set.validator(vote.validator_address) {
                 Some(validator) => validator,
-                None => {
-                    // println!(
-                    //     "  > couldn't find validator with address {}",
-                    //     vote.validator_address,
-                    // );
-
-                    continue;
-                }
+                None => continue, // Cannot find matching validator, so we skip the vote
             };
 
             let signed_vote = SignedVote::new(
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -91,13 +91,16 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
         let mut tallied_voting_power = 0_u64;
         let mut seen_validators = HashSet::new();
 
-        for (idx, signature) in signatures.iter().enumerate() {
-            let vote = vote_from_non_absent_signature(signature, idx as u64, &signed_header.commit);
-            let vote = match vote {
-                Some(vote) => vote,
-                None => continue, // Ok, some signatures can be absent
-            };
+        // Get non-absent votes from the signatures
+        let non_absent_votes = signatures.iter().enumerate().flat_map(|(idx, signature)| {
+            if let Some(vote) = non_absent_vote(signature, idx as u64, &signed_header.commit) {
+                Some((signature, vote))
+            } else {
+                None
+            }
+        });
 
+        for (signature, vote) in non_absent_votes {
             // Ensure we only count a validator's power once
             if seen_validators.contains(&vote.validator_address) {
                 continue;
@@ -149,11 +152,7 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
     }
 }
 
-fn vote_from_non_absent_signature(
-    commit_sig: &CommitSig,
-    validator_index: u64,
-    commit: &Commit,
-) -> Option<Vote> {
+fn non_absent_vote(commit_sig: &CommitSig, validator_index: u64, commit: &Commit) -> Option<Vote> {
     let (validator_address, timestamp, signature, block_id) = match commit_sig {
         CommitSig::BlockIDFlagAbsent { .. } => return None,
         CommitSig::BlockIDFlagCommit {
--- a/light-client/src/operations/voting_power.rs
+++ b/light-client/src/operations/voting_power.rs
@@ -14,13 +14,13 @@ use tendermint::lite::types::ValidatorSet as _;
 use tendermint::vote::{SignedVote, Vote};
 
 #[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct VotingPower {
+pub struct VotingPowerTally {
     pub total: u64,
     pub tallied: u64,
     pub trust_threshold: TrustThreshold,
 }
 
-impl fmt::Display for VotingPower {
+impl fmt::Display for VotingPowerTally {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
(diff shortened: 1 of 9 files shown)


light-client/src/operations/voting_power.rs
```diff
@@ -50,36 +53,104 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
...
+        for (idx, signature) in signatures.into_iter().enumerate() {
+            if signature.is_absent() {
+                continue; // OK, some signatures can be absent.
+            }
+
+            // The vals and commit have a 1-to-1 correspondance (see check above).
+            // This means we don't need the validator address or to do any lookup.
+            let val = validators[idx];
 
-        for vote in &signed_header.signed_votes() {
```
Liamsi, line 59: Thanks for doing this @romac 🙇 Can't we then delete the `signed_votes()` method?

romac: I guess we could, but since this is very much a WIP I didn't remove any code yet.

light-client/tests/light_client.rs
```diff
@@ -152,13 +153,16 @@ fn run_bisection_test(tc: TestBisection<LightBlock>) {
     let trusting_period = tc.trust_options.period;
     let now = tc.now;
 
+    // FIXME: What should this be, and where should it be configured?
```
melekes, line 156: in Go, default is 10sec. once we switch to the proposer based timestamps, it will probably be a consensus parameter

light-client/src/operations/voting_power.rs
```diff
@@ -50,36 +53,104 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
...
     }
 }
+
+fn vote_from_non_absent_signature(
+    commit_sig: &CommitSig,
+    validator_index: u64,
+    commit: &Commit,
+) -> Option<Vote> {
+    let (validator_address, timestamp, signature, block_id) = match commit_sig {
+        CommitSig::BlockIDFlagAbsent { .. } => return None,
```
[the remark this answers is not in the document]
romac: This function is actually the one doing the filtering for non absent signatures, hence the need for the Option. I guess the name isn't great, I am open to suggestions :)

romac:
I pushed a new version which should now be doing the right thing w.r.t. checking trust vs validators signers, but still relies on validator addresses. I guess we can address the latter in a future PR once we are convinced that this version is correct.

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,78 @@
...
     fn total_power_of(&self, validators: &ValidatorSet) -> u64;
-    fn voting_power_in(
+
+    fn check_enough_trust(
+        &self,
+        untrusted_header: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
+        trust_threshold: TrustThreshold,
+    ) -> Result<VotingPower, VerificationError> {
+        println!("check_validators_overlap");
```
brapse, line 42: This should probably be removed.

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,78 @@
...
+            Err(VerificationError::NotEnoughTrust(voting_power))
+        }
+    }
+
+    fn check_signers_overlap(
+        &self,
+        untrusted_header: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
+    ) -> Result<VotingPower, VerificationError> {
+        println!("check_signers_overlap");
```
brapse, line 58: Can probably be removed.

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,78 @@
...
+        let two_thirds = TrustThreshold::new(2, 3).unwrap();
```
brapse, line 59: Should this perhaps be a package level constant?

romac:
It would be great if someone could run the tests of the `light-client` crate and inspect the output to ensure that errors that are raised are indeed the ones we expect from the JSON file. I did so myself 

```
$ cd light-client/
$ cargo test -- --nocapture --test-threads 1
```

I already did so myself, and I am not too sure about this test:

```
Running light client against bisection test-file: ./tests/support/bisection/single_peer/not_enough_commits.json
  - Case: Trusted height=1, fails at height 6 because more than one-third (trust level) vals didn't sign
[light-client/tests/light_client.rs:247] e = Error(
    Context {
        kind: InvalidLightBlock(
            InvalidNextValidatorSet {
                header_next_validators_hash: Hash::Sha256(75B9F27F25F3515EF7CA61DD42C272CCC280C537E6A9ED493710EB361CB90B0D),
                next_validators_hash: Hash::Sha256(26952B5D784A1564D167DF98D2D37376B5E77771928256D25E6FF9AE3AD11564),
            },
        ),
...
```

This test is passing because we expect an error, but I don't think that that is the right one, except if there's perhaps more than one error encoded in the JSON test. In particular, I would have expected  the test to result in either `NotEnoughTrust` or `InsufficientValidatorsOverlap` but we are getting `InvalidNextValidatorSet` instead. 

@Shivani912 Is that something you could look into?

light-client/src/operations/voting_power.rs
```diff
@@ -24,39 +83,111 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
...
-                None => continue,
+        validator_set: &ValidatorSet,
+        trust_threshold: TrustThreshold,
+    ) -> Result<VotingPower, VerificationError> {
+        let signatures = &signed_header.commit.signatures;
+
+        let mut tallied_voting_power = 0_u64;
+        let mut seen_validators = HashSet::new();
+
+        for (idx, signature) in signatures.into_iter().enumerate() {
```
brapse, line 97: Does it make sense to reference the test vector which exercises this particular edge case?

light-client/src/operations/voting_power.rs
```diff
@@ -24,39 +83,111 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
...
+            // If the vote is neither absent nor nil, tally its power
+            if signature.is_commit() {
+                tallied_voting_power += validator.power();
+            } else {
+                // It's OK. We include stray signatures (~votes for nil)
+                // to measure validator availability.
             }
 
-            signed_power += val.power();
+            // TODO: Break out when we have enough voting power
```
brapse, line 140: Reference #235

light-client/src/operations/voting_power.rs
romac: Will do this as part of #342.

Shivani912:
> It would be great if someone could run the tests of the `light-client` crate and inspect the output to ensure that errors that are raised are indeed the ones we expect from the JSON file. I did so myself
> 
> ```
> $ cd light-client/
> $ cargo test -- --nocapture --test-threads 1
> ```
> 
> I already did so myself, and I am not too sure about this test:
> 
> ```
> Running light client against bisection test-file: ./tests/support/bisection/single_peer/not_enough_commits.json
>   - Case: Trusted height=1, fails at height 6 because more than one-third (trust level) vals didn't sign
> [light-client/tests/light_client.rs:247] e = Error(
>     Context {
>         kind: InvalidLightBlock(
>             InvalidNextValidatorSet {
>                 header_next_validators_hash: Hash::Sha256(75B9F27F25F3515EF7CA61DD42C272CCC280C537E6A9ED493710EB361CB90B0D),
>                 next_validators_hash: Hash::Sha256(26952B5D784A1564D167DF98D2D37376B5E77771928256D25E6FF9AE3AD11564),
>             },
>         ),
> ...
> ```
> 
> This test is passing because we expect an error, but I don't think that that is the right one, except if there's perhaps more than one error encoded in the JSON test. In particular, I would have expected the test to result in either `NotEnoughTrust` or `InsufficientValidatorsOverlap` but we are getting `InvalidNextValidatorSet` instead.
> 
> @Shivani912 Is that something you could look into?

@romac There's a bug in this one (issue #290) I'm working on it and will do a PR once I fix it! 

I feel we really need to have checks (in the test itself) on errors even if the tests pass. I tried to do this at the end of Q1/beginning of Q2 but it seemed it was a little too early to be able to do so. Think this should be on priority for next quarter?

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,75 @@
...
+    fn check_enough_trust(
+        &self,
+        untrusted_header: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
+        trust_threshold: TrustThreshold,
+    ) -> Result<VotingPower, VerificationError> {
+        let voting_power =
+            self.voting_power_of(untrusted_header, untrusted_validators, trust_threshold)?;
+
+        if trust_threshold.is_enough_power(voting_power.tallied, voting_power.total) {
```
OStevan, line 45: Mentioned in one of the comments above, I would probably make more sense to just check this in the `voting_power_of` and instead of a value just return a boolean value or something else to tell us if everything is fine or there is some error.

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,75 @@
...
+            self.total, self.tallied, self.trust_threshold
+        )
+    }
+}
 
 pub trait VotingPowerCalculator: Send {
     fn total_power_of(&self, validators: &ValidatorSet) -> u64;
-    fn voting_power_in(
+
+    fn check_enough_trust(
```
OStevan, line 36: The name and the result of this function were for me a bit misleading as the name would suggest a boolean value (or in this case `Ok()` or `Error(...)`) and it is more or less used as such since the result value in the Ok case is just ignored.

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,75 @@
...
+use std::collections::HashSet;
+use std::fmt;
+
+use tendermint::block::CommitSig;
+use tendermint::lite::types::TrustThreshold as _;
 use tendermint::lite::types::ValidatorSet as _;
+use tendermint::vote::{SignedVote, Vote};
+
+#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct VotingPower {
```
OStevan, line 17: For me this is a bit misleading as the struct has more information than what I would expect by reading just the name. Also, I think it would be beneficial to implement the TODO about the breaking out of the loop as than this would go probably go away.

light-client/src/operations/voting_power.rs
```diff
@@ -24,39 +80,107 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
...
+                Some(vote) => vote,
+                None => continue, // Ok, some signatures can be absent
+            };
+
+            // Ensure we only count a validator's power once
+            if seen_validators.contains(&vote.validator_address) {
+                continue;
+            } else {
+                seen_validators.insert(vote.validator_address);
+            }
```
OStevan, line 109: One interesting difference between go code and the implementation here is this `if`. (go) In case of 2/3 threshold this check is done by comparing the size of the signers validator set and the expected validator set, if both are the same length everything is fine and we don't even do these checks. However in the case where we are interested in trust between two non adjacent headers, `then` branch should raise an exception.

light-client/src/operations/voting_power.rs
```diff
@@ -24,39 +83,111 @@ impl VotingPowerCalculator for ProdVotingPowerCalculator {
...
+        validator_set: &ValidatorSet,
+        trust_threshold: TrustThreshold,
+    ) -> Result<VotingPower, VerificationError> {
+        let signatures = &signed_header.commit.signatures;
+
+        let mut tallied_voting_power = 0_u64;
+        let mut seen_validators = HashSet::new();
+
+        for (idx, signature) in signatures.into_iter().enumerate() {
+            let vote = vote_from_non_absent_signature(signature, idx as u64, &signed_header.commit);
```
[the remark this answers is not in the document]
romac: You mean counting absent signatures? I don't think that's possible since a `CommitSig::BlockIDFlagAbsent` does not contain any of the data required to craft a `Vote` (ie. validator_address, timestamp and signature).

light-client/src/operations/voting_power.rs
romac: What edge case are you referring to here?

romac:
> @romac There's a bug in this one (issue #290) I'm working on it and will do a PR once I fix it!

Awesome, thanks :)

> I feel we really need to have checks (in the test itself) on errors even if the tests pass. I tried to do this at the end of Q1/beginning of Q2 but it seemed it was a little too early to be able to do so. Think this should be on priority for next quarter?

Yes, that would definitely increase our confidence in the tests and thus the correctness of implementation :)

light-client/src/operations/voting_power.rs
romac: Agreed that the name is perhaps not the best, but we still need this information to be able to yield a meaningful error in case of lack of trust/signers overlap.

light-client/src/operations/voting_power.rs
romac: @melekes What do you think?

romac: I guess raising an error is probably the proper behavior there. Will already change that, and we can always revert it back if that was wrong.

light-client/src/operations/voting_power.rs
brapse: Oops perhaps I did not comment the correct line and actually there is an informative comment further down `Ensure we only count a validator's power once`

light-client/src/operations/voting_power.rs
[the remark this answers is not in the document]
brapse: True!

light-client/src/operations/voting_power.rs
romac: Unfortunately, we currently do have a test vector which covers this edge case.

cc @Shivani912

light-client/src/operations/voting_power.rs
```diff
@@ -1,19 +1,76 @@
...
+    }
+}
 
 pub trait VotingPowerCalculator: Send {
     fn total_power_of(&self, validators: &ValidatorSet) -> u64;
+
+    fn check_enough_trust(
+        &self,
+        untrusted_header: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
```
ebuchman, line 39: These are trusted validators

romac:
@ebuchman 

> I think we still want to break the voting_power_in into two functions, one for check_signers and one for check_enough_trust

It would be great if you could describe the differences between the two versions somewhere.

> and neither should depend on vote.validator_address

I initially tried to fix that as well but eventually chose to focus on fixing the computation itself, and improve it later.

ebuchman:
Done in https://github.com/informalsystems/tendermint-rs/issues/377

PR #312, opened 2020-06-09
Quickfix: make fields in event subscription accessible publicly

### LINKED ISSUE #308: Need access to block in `EventDataNewBlock`
The IBC relayer subscribes for block notifications and on `EventDataNewBlock` it needs access to the block height only ...for now.
But content is not public:
https://github.com/informalsystems/tendermint-rs/blob/bcef28b55a83dc3917aeedc8acf6429108aead5c/tendermint/src/rpc/event_listener.rs#L193

Is there an access function I don't see?

Liamsi:
Non-elegant quickfix: makes all event subscription fields `pub` s.t. they can be accessed and processed. 

close #308 

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/rpc/event_listener.rs
+++ b/tendermint/src/rpc/event_listener.rs
@@ -101,7 +101,8 @@ impl EventListener {
             dbg!(&msg.to_string());
             return Ok(None);
         }
-
+        dbg!("received neither event nor generic string message:");
+        dbg!(&msg.to_string());
         Err(RPCError::new(
             Code::Other(-1),
             Some("received neither event nor generic string message".to_string()),
--- a/tendermint/src/rpc/event_listener.rs
+++ b/tendermint/src/rpc/event_listener.rs
@@ -151,26 +151,27 @@ pub type WrappedResultEvent = Wrapper<ResultEvent>;
 /// TX value
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct EventDataTx {
+    /// The actual TxResult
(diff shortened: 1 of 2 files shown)


tendermint/src/rpc/event_listener.rs
```diff
@@ -101,7 +101,8 @@ impl EventListener {
             dbg!(&msg.to_string());
             return Ok(None);
         }
-
+        dbg!("received neither event nor generic string message:");
+        dbg!(&msg.to_string());
```
ancazamfir, line 105: I checked this with IBC relayer and it's ""

Liamsi: I think that is a ping message:
https://github.com/tendermint/tendermint/blob/ba3a2dde376f708085726cd778bb361bd8c7c692/rpc/jsonrpc/server/ws_handler.go#L407

 If the server doesn't drop your event subscription, you can ignore it for now.

ancazamfir: > If the server doesn't drop your event subscription, you can ignore it for now.

Yes, I am ignoring but was wondering if this could be handled more gracefully :) Anyway..this is for the other issue. Just saw you added this here and wanted to mention that for the failure I see it is empty string.

Liamsi: > Yes, I am ignoring but was wondering if this could be handled more gracefully :)

Absolutely, the client should actually play ping-pong with the server instead of treating this as an error.

PR #324, opened 2020-06-14
Add integration test pinned to tendermint-go v0.33.5

### LINKED ISSUE #304: RPC integration tests failing on CI
The RPC integration tests are currently failing on CI.

Affected PRs are:
- #302 ([job](https://github.com/informalsystems/tendermint-rs/pull/302/checks?check_run_id=742462716))
- #284 ([job](https://github.com/informalsystems/tendermint-rs/pull/284/checks?check_run_id=742259921))

Here's an excerpt from the build logs for #302:

<details>
<summary>Click to show log</summary>

```
running 11 tests

test rpc::abci_info ... ok
test rpc::abci_query ... ok
test rpc::block ... FAILED
test rpc::block_results ... ok
test rpc::commit ... FAILED
test rpc::blockchain ... FAILED
test rpc::genesis ... ok
test rpc::health ... ok
test rpc::net_info ... ok
test rpc::status_integration ... ok
test rpc::event_subscription ... FAILED

failures:

---- rpc::block stdout ----
thread 'rpc::block' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("invalid type: integer `1`, expected a string at line 8 column 19") }', tendermint/tests/integration.rs:73:26

---- rpc::commit stdout ----
thread 'rpc::commit' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("invalid type: integer `0`, expected a string at line 17 column 23") }', tendermint/tests/integration.rs:111:27

---- rpc::blockchain stdout ----
thread 'rpc::blockchain' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ParseError, message: "Parse error. Invalid JSON", data: Some("invalid type: integer `1`, expected a string at line 11 column 23") }', tendermint/tests/integration.rs:95:31

---- rpc::event_subscription stdout ----
[tendermint/src/rpc/event_listener.rs:95] "We did not receive a valid JSONRPC wrapped ResultEvent!" = "We did not receive a valid JSONRPC wrapped ResultEvent!"
thread 'rpc::event_subscription' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: Other(-1), message: "Error (code: 0)", data: Some("received neither event nor generic string message") }', tendermint/tests/integration.rs:164:34

failures:
    rpc::block
    rpc::blockchain
    rpc::commit
    rpc::event_subscription

test result: FAILED. 7 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out
```
</details>

shonfeder:
Closes #304

With this change, CI runs two integration tests:

1. A `stable` test to protect against regressions, run against a pinned
version of the tendermint/tendermint docker image.
2. A `latest` test to track whether we're maintaining parity with the
latest development version of tendermint-go.

IIUC, we'd ideally like (1) to be a required test, and to get visibility on (2) but not block the PR if it fails. I can configure (2) so that it the job succeeds even if the step fails, but then we lose clear visibility on the failure.

Please see individual commits for more detail.

* [x] Wrote tests
* [x] Updated CHANGES.md

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -61,11 +61,35 @@ jobs:
           command: test
           args: --all-features --no-fail-fast
 
-  test-integration-ignored:
+  # TODO(shonfeder): remove duplication once GitHub addresses one of these
+  #  - https://github.community/t/support-for-yaml-anchors/16128/15
+  #  - https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/13
+  #  - https://github.community/t/using-matrix-variable-in-docker-image-name/17296
+  test-integration-stable:
     runs-on: ubuntu-latest
     services:
       tendermint:
-        image: tendermint/tendermint
+        image: tendermint/tendermint:v0.33.5
+        ports:
+          - 26656:26656
+          - 26657:26657
+          - 26660:26660
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: -p tendermint --test integration --no-fail-fast -- --ignored
+
+  test-integration-latest:
+    runs-on: ubuntu-latest
+    services:
+      tendermint:
+        image: tendermint/tendermint:latest
         ports:
           - 26656:26656
           - 26657:26657
--- a/tendermint/tests/integration.rs
+++ b/tendermint/tests/integration.rs
@@ -35,7 +35,6 @@ mod rpc {
     async fn abci_info() {
         let abci_info = localhost_rpc_client().abci_info().await.unwrap();
 
-        assert_eq!(&abci_info.version, "0.17.0");
         assert_eq!(abci_info.app_version, 1u64);
         // the kvstore app's reply will contain "{\"size\":0}" as data right from the start
         assert_eq!(&abci_info.data, "{\"size\":0}");
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,8 @@ CommitSig:
 - Added CommitSig timestamp zero-check compatibility code [#259](https://github.com/informalsystems/tendermint-rs/issues/259)
 
 Testing:
-- Updated abci_info test to 0.17.0 ([#249](https://github.com/informalsystems/tendermint-rs/issues/249))
+- Add integration test to track tendermint-go v0.33.5 ([#324](https://github.com/informalsystems/tendermint-rs/pull/324))
+- Remove test for hard-coded version in `abci_info` ([#324](https://github.com/informalsystems/tendermint-rs/pull/324))
 
 Serialization:
 - Refactor serializers library to use modules, give a nicer annotation to structs and separated into its own folder. ([#247](https://github.com/informalsystems/tendermint-rs/issues/247))


tendermint/tests/integration.rs
```diff
@@ -35,7 +35,6 @@ mod rpc {
     async fn abci_info() {
         let abci_info = localhost_rpc_client().abci_info().await.unwrap();
 
-        assert_eq!(&abci_info.version, "0.17.0");
```
shonfeder, line 38: See the specific commit message for my reasoning here: b452fcc6d807fe616bcd16ccc99839649d86f212

An important caveat (tho not well placed in the git logs), is that I don't have a very informed position, given that this is my first tiny foray into helping here, so please forgive if my suggestions are wide of the mark.

Some alternatives that might make sense, in addition to the approach I propose in the commit message:

- Specify the expected version via an environment variable during testing (allowing us to support testing against multiple versions, as desired).
-  Change this to a check for a string that matches a semantic version (e.g., `\d+\.\d+\.\d+`)

shonfeder:
This change should make it possible to make the `test-integration-stable` test required (if that's desirable).

.github/workflows/rust.yml
```diff
@@ -61,11 +61,34 @@ jobs:
           command: test
           args: --all-features --no-fail-fast
 
-  test-integration-ignored:
+  # TODO(shonfeder): remove duplication once GitHub addresses
```
shonfeder, line 64: IIUC, we could DRY this out *now* if we want to define our own action for this, but I'm not sure if we want to complicate the CI config that way. I'm up for it tho, if anyone thinks it's preferable.

.github/workflows/rust.yml
```diff
@@ -61,11 +61,34 @@ jobs:
...
+  #  - https://github.community/t/support-for-yaml-anchors/16128/15, or
+  #  - https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/13
+  test-integration-stable:
     runs-on: ubuntu-latest
     services:
       tendermint:
-        image: tendermint/tendermint
+        image: tendermint/tendermint:v0.33.5
```
shonfeder, line 71: It would be nice to set this version somewhere more canonical, but from what I've seen of github actions so far, there's no clear way to parameterize the values of an `image` field here. I'd be interested in suggestions for a more elegant approach here. Failing that, I could at least document this somewhere (and the question is just where that should go).

romac:
> This change should make it possible to make the `test-integration-stable` test required (if that's desirable).

Yeah I think we want to fail the build if the integration tests do not pass on stable.

Great job by the way, thanks a lot!

.github/workflows/rust.yml
Liamsi: I don't have much experience with gh either. I think even having the version more prominently in an env variable, might be ok: https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables

@marbar3778 as you looked a lot more into GA, do you have a suggestion how to parameterize this?

Liamsi:
I'm surprised the integration tests still fail despite fixing the version to 0.33.5 🤔 maybe there were some changes we missed in the meantime.

marbar3778:
> as you looked a lot more into GA, do you have a suggestion how to parameterize this?

I have not. Other than env variables I can't think of anything right now but will think about it for a bit and geet back to you.

> I'm surprised the integration tests still fail despite fixing the version to 0.33.5 🤔 maybe there were some changes we missed in the meantime.

This may have something to do with our new lib for encoding json to replace amino.

Liamsi:
Merging as is. I'll investigate why the tests are failing in a separate pr later.

Liamsi:
> This may have something to do with our new lib for encoding json to replace amino.

This is actually because the tests against `:latest` are still running: 

From the introductory comment:
>  I can configure (2) so that it the job succeeds even if the step fails, but then we lose clear visibility on the failure.

Yes, I think that would be OK for now! Please do (or anyone else with necessary perms)

Issue #219, opened 2020-04-16
Light Client RPC Service


brapse:
_Updated to reflect the conclusion of the discussion in comments and slack_

The light client fetches headers/validator sets from other nodes to perform verification. The light client should also make those headers/validator sets available to the network through an RPC service.

The goal here is to create an RPC service operating in a separate runtime with access to it's own light client handler similar to:
https://github.com/informalsystems/tendermint-rs/blob/1df3ac3aebcd842f54737bb47249e76aca3573e8/light-client/examples/light_client.rs#L155

This issue should exclusively include endpoints for:
* `/status`: returns local status information, like the latest height, latest block and valset hashes, list of of connected full nodes (primary and secondaries), etc.
* `/state`: returns the latest header, commit, and validator set

Follow up issues can address potential further improvements.

ebuchman:
The goal here is essentially a [light-node](https://github.com/informalsystems/tendermint-rs/issues/112): a process that syncs the blockchain using light-client verification and exposes an http API for interacting with the blockchain (ie. queries, transactions). Such an API could be used to build local applications that read/write to the blockchain securely without needing a local full node. 

In Go, we have `tendermint lite`, which runs a light node that exposes the [same RPC API as the full node](https://docs.tendermint.com/master/rpc/), essentially an http proxy that does light client verification. There's also the [gaiacli rest-server](https://hub.cosmos.network/master/resources/service-providers.html#setting-up-the-rest-server), a similar idea but with an application specific REST API (eg. for accounts, staking, governance etc.) rather than the RPC exposed by Tendermint full nodes (which can do raw abci queries, but doesn't know anything about accounts, staking, governance, etc.).

To start, we should expose two http endpoints in the light-node process:

- `/status`: returns local status information, like the latest height, latest block and valset hashes, list of of connected full nodes (primary and secondaries), etc.
- `/state`: returns the latest header, commit, and validator set

Once we have the basics, we can draft a longer term plan for this API, and consider something like the complete `tendermint lite` proxy in Rust, which would effectively wrap our tendermint-rs RPC client with light client verification. Note for full compatibility we'd have to support queries for historical heights, which could require running the light client backwards.

Liamsi:
Cross posting this here:

> You guys asked me last week what methods the rpc would need from the light client. As far as I can see it’s only one method:
> - given a target height h , either return the stored result, (if still valid) or, run the light client to that height ( h might potentially be in the past)

[...]

PR #329, opened 2020-06-16
Bootstrap proxy for light-node API

xla:
Introducing a proxy which serves the Tendermint RPC API with the help of a connected RPC client. Later to be extended to add light client functionality. This is in nature equivalent to the Go proxy[0].

Instead of reaching for hyper this change introduces warp to build an HTTP server. This decision should be evaluated very carefully, while warp is a thinly wrapper around hyper, it brings a lot of very important tooling (including Websockets), it also has its own set of dependencies.

Also currently unclear if the module that implements the RPC API with lite client functionality should live in the RPC source tree.

Part of #219

[0] https://github.com/tendermint/tendermint/tree/master/light/proxy

***

Still under development, open this draft PR for early feedback and course corrections.

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -33,6 +33,7 @@ codecov = { repository = "..."}
 [dependencies]
 anomaly = "0.2"
 async-trait = "0.1"
+async-tungstenite = {version="0.5", features = ["tokio-runtime"]}
 bytes = "0.5"
 chrono = { version = "0.4", features = ["serde"] }
 futures = "0.3"
@@ -42,6 +43,7 @@ hyper = "0.13"
 once_cell = "1.3"
 prost-amino = "0.5"
 prost-amino-derive = "0.5"
+ripemd160 = "0.8"
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1" }
 serde_bytes = "0.11"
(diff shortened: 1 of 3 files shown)


ebuchman:
Thanks @xla !! But see my comment https://github.com/informalsystems/tendermint-rs/issues/219#issuecomment-644751321 - not sure we want to go full proxy yet, probably best to just start with a simple RPC service with two methods and figure out from there how we really want to proceed

xla:
@ebuchman Read all your comments thoroughly and promise to jump the shark here. Boostraping here means pretty much the two endpoints you mentioned.

Liamsi:
I had to google "jump the shark" and ended up watching this: https://www.youtube.com/watch?v=t4ZGKI8vpcg 😬 

This PR, needs to be updated using the new rpc crate: #338 

Now that I learned that tendermint (go) actually offers (98%) jsonrpc compatible endpoints, we could use a jsonrpc server crate to expose the endpoints instead.

xla:
> I had to google "jump the shark" and ended up watching this: https://www.youtube.com/watch?v=t4ZGKI8vpcg grimacing

World-class entertainment.

> This PR, needs to be updated using the new rpc crate: #338

Think I gonna ditch this one and start fresh on top of #338.

> Now that I learned that tendermint (go) actually offers (98%) jsonrpc compatible endpoints, we could use a jsonrpc server crate to expose the endpoints instead.

Agreed!

xla:
Super-seeded by #363

PR #338, opened 2020-06-17
chore: move `net2` dependency out of tendermint crate

### LINKED ISSUE #289: Feature guard the networking stuff
- since we already implemented the networking stuff ourselves, so it would be good for us to cut some dependencies.
- we can even try to make it `no_std`, after separating out networking features.

xla:
This is isolates the code paths which depend on network and http functionality by moving the rpc module into its own crate (tendermint-rpc). The expected outcome is that third-parties which depend on the tendermint crate are not forced to pull in unnecessary dependencies. Furthermore it will isolate any potential crates used to drive an rpc server implementation going forward.

~~Controversially this also removes the dependency on `tokio` as it was mostly used for async test execution, which was achieved by making extensive use of `futures::executor::block_on` - we've come full circle. Eager to hear opinions and alternatives to that.~~
Update: now a dev-dependency only instead of the above.

Follow-up to #7 
Preparation for #219
Closes #289

***

* [ ] ~~Wrote tests~~
* [ ] Updated CHANGES.md


rpc/src/endpoint/block.rs
```diff
@@ -1,11 +1,9 @@
-//! `/block` endpoint JSONRPC wrapper
+//! `/block` endpoint JSONcrate wrapper
```
Liamsi, line 1: This looks like an accidental change? It's still a jsonrpc wrapper.

rpc/src/endpoint/block_results.rs
```diff
@@ -1,8 +1,9 @@
-//! `/block_results` endpoint JSONRPC wrapper
+//! `/block_results` endpoint JSONcrate wrapper
```
Liamsi, line 1: same as above

rpc/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "tendermint-rpc"
```
Liamsi, line 2: 👍

rpc/src/endpoint/abci_query.rs
```diff
@@ -1,13 +1,12 @@
...
-    block,
-    merkle::proof::Proof,
-    rpc, serializers,
-};
 use serde::{Deserialize, Serialize};
 
+use tendermint::abci::{Code, Log, Path};
+use tendermint::block;
+use tendermint::merkle::proof::Proof;
+use tendermint::serializers;
```
Liamsi, line 8: Just out of curiosity: are all these import changes due to a different formatter? Is there a re reason why all the `use crate:: ...`s are replaced in that fashion?

rpc/src/endpoint/blockchain.rs
```diff
@@ -1,9 +1,10 @@
-//! `/block` endpoint JSONRPC wrapper
+//! `/block` endpoint JSONcrate wrapper
```
Liamsi, line 1: same

rpc/src/endpoint/broadcast/tx_async.rs
```diff
@@ -14,17 +12,17 @@ pub struct Request {
 }
 
 impl Request {
-    /// Create a new async transaction broadcast RPC request
+    /// Create a new async transaction broadcast crate request
```
Liamsi, line 15:
```suggestion
    /// Create a new async transaction broadcast RPC request
```

rpc/src/endpoint/broadcast/tx_sync.rs
```diff
@@ -20,11 +18,11 @@ impl Request {
...
 }
 
-impl rpc::Request for Request {
+impl crate::Request for Request {
     type Response = Response;
 
-    fn method(&self) -> rpc::Method {
-        rpc::Method::BroadcastTxSync
+    fn method(&self) -> crate::Method {
+        crate::Method::BroadcastTxSync
```
Liamsi, line 25: Couldn't all these be simply `Method::BroadcastTxSync` etc now that they are in a separate crate?

rpc/src/lib.rs
```diff
@@ -1,11 +1,9 @@
-//! Tendermint RPC: JSONRPC over HTTP support
-//!
-//! Wraps the RPC API described at: <https://tendermint.com/rpc/>
+//! Tendermint RPC definitons and types.
 
-mod client;
-pub mod endpoint;
 pub mod error;
 // TODO(ismail): document fields or re-use the abci types
```
Liamsi, line 4: This comment belongs to the `#[allow(missing_docs)]` below.

tendermint/Cargo.toml
```diff
@@ -55,11 +52,8 @@ subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
...
 toml = { version = "0.5" }
-uuid = { version = "0.8", default-features = false }
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
-async-tungstenite = {version="0.5", features = ["tokio-runtime"]}
-tokio = { version = "0.2", features = ["macros"] }
 ripemd160 = "0.9"
 
 [dev-dependencies]
-serde_json = "1"
+tendermint-rpc = { version = "0.1.0", path = "../rpc" }
```
Liamsi, line 59: Nice, nice.

tendermint/src/lite/verifier.rs
```diff
@@ -528,22 +528,21 @@ mod tests {
...
+    fn assert_bisection_ok(
         req: &MockRequester,
         ts: &TrustedState<MockCommit, MockHeader>,
         untrusted_height: u64,
         expected_num_of_requests: usize,
         expected_final_state: &MockState,
     ) {
         let mut cache: Vec<MockTrustedState> = Vec::new();
-        let ts_new = verify_bisection_inner(
+        let ts_new = futures::executor::block_on(verify_bisection_inner(
```
Liamsi, line 539: is this necessary to get rid of the direct tokio dependency?

tendermint/src/lite/verifier.rs
```diff
@@ -784,20 +782,18 @@ mod tests {
         assert_single_err(ts, vac, err.into());
     }
 
-    #[tokio::test]
-    async fn test_verify_bisection_1_val() {
```
Liamsi, line 788: Ok, answers my question I guess :-D 
https://github.com/informalsystems/tendermint-rs/pull/338/files#r441736094

tendermint/src/serializers.rs
```diff
@@ -50,5 +50,5 @@ pub(crate) use raw_commit_sig::RawCommitSig;
 mod tests;
 
 mod custom;
-pub(crate) use custom::parse_non_empty_block_id;
-pub(crate) use custom::parse_non_empty_hash;
+pub use custom::parse_non_empty_block_id;
+pub use custom::parse_non_empty_hash;
```
Liamsi, line 54: Do we have to publish these beyond the crate? Probably, yes as some might be used in the rpc crate now?

rpc/Cargo.toml
Liamsi, line 2: This probably deserves a short readme (can be done in a followup PR).

rpc/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+async-tungstenite = {version="0.5", features = ["tokio-runtime"]}
+bytes = "0.5"
+futures = "0.3"
+getrandom = "0.1"
+http = "0.2"
+hyper = "0.13"
+serde = { version = "1", features = [ "derive" ] }
+serde_bytes = "0.11"
+serde_json = "1"
+tendermint = { version = "0.13.0", path = "../tendermint" }
```
Liamsi, line 17: I'm always sceptical of circular dependencies (even if it's just a dev-dependency). Are we confident that this does cannot have weird side-effects?

rpc/src/endpoint/abci_query.rs
xla: Personal touch really, optimised for editability and tracebility of dependencies. I usually one liner per top-level module in a crate.

tendermint/src/lite/verifier.rs
Liamsi: Also, I should better read the introductory comments 🙈 😆 

> Controversially this also removes the dependency on tokio as it was mostly used for async test execution, which was achieved by making extensive use of futures::executor::block_on

rpc/Cargo.toml
xla: I would not expect any, but truth be told this was the solution with the least amount of opinionated changes. Would like to get rid of this circle as well, but afraid this goes too far for this PR.

rpc/src/endpoint/broadcast/tx_sync.rs
xla: We could, but there are some reoccurring names here `Request`, `Response`. Wanna keep it clear where all of them come from.

xla:
> But more relevant is that some of the changes seem to have messed with the integration tests:
> `thread 'rpc::abci_info' panicked at 'there is no reactor running, must be called from the context of Tokio runtime',`

On second thought it might be worthwhile to keep tokio in `dev-dependencies`.

tendermint/src/serializers.rs
xla: Indeed, the rpc crate depends on a lot of these serialisers. Not happy with it. Also found the excessive use of `pub(crate)` is unnecessary as the top-level lib determines visibility beyond crate boundaries. Eager to hear how we can address this issue.

Liamsi: If they are *only* used in the rpc lib, they could actually live there. Not sure yet.

rpc/src/endpoint/abci_query.rs
Liamsi: Ok for me. Ideally, these should be consistent in all our crates though.

tendermint/src/serializers.rs
xla: They are also used in `tendermint::block`.

Liamsi: Maybe removing the net dependency outweighs this now publisized types. 

I might have mentioned that somewhere else before but ideally, the serialzation code would simply be abstracted away in the sense that anyone using the tendermint crate can e.g. simply call block::deserialize (or deserialize_json and deserialize_binary) and the that could internally use a dedicated serialization type serialize::block (and impl Froms between both) which uses methods / helpers like the above. Anyone (like the rpc crate) could simply use the core types and wouldn't need to care about serialization specific stuff. Does that make sense? Probably worth another discussion and later another PR.

rpc/Cargo.toml
Liamsi: I can't think of any problems either 🍀 🤞

tendermint/src/serializers.rs
Liamsi: For now: maybe the `serializers` module should include a comment that we don't guarantee any backwards compatibility on the methods of that module and consider them an internal detail (use at you own risk alike).

tendermint/src/serializers.rs
```diff
@@ -2,6 +2,10 @@
 //!
 //! Serializers and deserializers for a transparent developer experience.
 //!
+//! CAUTION: There are no guarantees for backwards compatibility, this module should be considered
+//! an internal implementation detail which can vanish without further warning. Use at your own
+//! risk.
```
Liamsi, line 7: 👍

tendermint/src/serializers.rs
ebuchman: We need to lay out an explicit semantic versioning scheme and declare which modules we're covering and which we're not. Good to start tracking like this though thanks!

Issue #332, opened 2020-06-17
Exit condition in verify_to_target is not correct


milosevic:
See https://github.com/informalsystems/tendermint-rs/blob/184a1f9925f541473313efcd483c32cb913f8af9/light-client/src/light_client.rs#L187
      
What we need is separating logic for the case target_height is smaller than trusted_height from the case target_height is bigger than the trusted_height. In my view, exit condition, in case we don't have verification error, should in both cases be only for equality: target_height == trusted_height.

PR #340, opened 2020-06-18
Draft spec for targetHeight < LatestVerified

josef-widder:
This addresses the case when a smaller height is requested from the lightclient verification (#332)
I have added a section "Supporting the IBC Relayer" that sketches out the spec. A function `Main` just decides which method (forward skipping vs. backwards check hashes) should be used.

I think the description is not very elegant, yet, but I wanted to have something we can discuss quickly.


docs/spec/lightclient/verification.md
```diff
@@ -844,9 +915,130 @@ If on the blockchain the validator set of the block at height
...
+  if targetHeight > lightStore.LatestVerified.height {
+    return VerifyToTarget(primary, lightStore, targetHeight)
+  }
+  else if targetHeight = lightStore.LatestVerified.height {
+    return (lightStore, ResultSuccess)
+  }
+  else {
+    lb, res = lightStore.LatestPrevious(targetHeight);
+    if res = true {
+	  // make auxiliary lightStore auxLS to call VerifyToTarget
```
romac, line 1000: Why do we need an auxiliary light store here, can't we re-use the original one like we currently do when `targetHeight < lightStore.LatestVerified.height`?

josef-widder: I am not sure what you mean with "like we currently do" (did you get the '<' in the right direction?). The problem is that currently in `VerifyToTarget`, the loop condition is
`for lightStore.LatestVerified.height < targetHeight`. If we do this check on the complete Lightstore, we never enter the loop. So the idea was to make an auxiliary light store so that the LatestVerified gets "smaller". This also allows to keep the simple constraints of `Schedule`.

romac: > I am not sure what you mean with "like we currently do" (did you get the '<' in the right direction?).

Sorry, I indeed meant `targetHeight > lightStore.LatestVerified.height`.

> The problem is that currently in VerifyToTarget, the loop condition is
for lightStore.LatestVerified.height < targetHeight. If we do this check on the complete Lightstore, we never enter the loop. So the idea was to make an auxiliary light store so that the LatestVerified gets "smaller". This also allows to keep the simple constraints of Schedule.

I see, makes sense!

docs/spec/lightclient/verification.md
```diff
@@ -844,9 +915,130 @@ If on the blockchain the validator set of the block at height
...
+  latest := lb.Header
+  for i := lb.Header.height - 1; i >= targetHeight; i-- {
+    // here we download height-by-height. We might first download all
+	// headers down to targetHeight and then check them.
+    current := FetchLightBlock(primary,i)
+    if (hash(current) != latest.Header.LastBlockId) {
+      return (lightStore, ResultFailure)
+    }
+	else {
+	  lightStore.Update(current, StateVerified)
```
milosevic, line 976: This is a new type? I guess the idea is to mark headers obtained by backward verification differently? Can they be used as a basis for forward verification?How do we reason about trusted period of those headers?

docs/spec/lightclient/verification.md
```diff
@@ -844,9 +915,130 @@ If on the blockchain the validator set of the block at height
...
+method should be used.
+
+#### **[LCV-FUNC-IBCMAIN.1]**:
+```go
+func Main(primary PeerID, lightStore LightStore,
+                    targetHeight Height) (LightStore, Result) {
+  if targetHeight > lightStore.LatestVerified.height {
+    return VerifyToTarget(primary, lightStore, targetHeight)
+  }
+  else if targetHeight = lightStore.LatestVerified.height {
```
milosevic, line 994: This should probably be just check if the targetHeight is already in the store.

milosevic: I think the main decision tree should be as follows: 1) is `targetHeight` already in the store; if this is the case return the block, else 2) if `targetHeight >  lightStore.LatestVerified.height` do normal forward verification business, else 3) if  `targetHeight > some trusted height minH from the store` then do forward verification starting from `minH`, else 4) do the backward verification from some trusted header mH from the store, but being careful that backward verification must terminate within trusted period of mH. So picking mH as minimum trusted header might not be the best idea. In practice, I would guess that most of the verification requests would be addressed by scenarios 1-3 (at least for relayer cases).

josef-widder: OK.  
A clarification for point 1). `targetHeight` needs to be in store **and** must be verified (I adapted to code accordingly). It might be that we previously downloaded it but could not verify it, and then did some bisection, and eventually jumped over it. Apalache found such traces, and they are fine.  
Regarding point 4, I am not sure. If any header is set to trusted, we treat it as if it comes from the blockchain. By assumption on hashes, all previous blocks that pass the tests, necessarily are also from the chain. The problem of course is whether the application should the use the header of targetHeight, when it was expired. I think this is a question for the specific application. It might be application-specific whether the information in such a block is still usable.

docs/spec/lightclient/verification.md
josef-widder: It might make sense to introduce a new type here. Even if we do not distinguish now, at a later points we might find it useful.  
I guess we should discuss whether they should be treated in a special way. My current understanding is that if we go backwards from a trusted light block, and all hashes match, the trust extends to these light blocks as well. But perhaps I miss something.

docs/spec/lightclient/verification.md
```diff
@@ -844,8 +915,139 @@ If on the blockchain the validator set of the block at height
...
+
+
+```go
+func (ls LightStore) LatestPrevious(height Height) (LightBlock, bool) 
+```
+- Expected postcondition
+  - returns a light block *lb* that satisfies:
+      - *lb* is in lightStore
+      - *lb* is verified and not expired
+	  - *lb.Header.Height < height*
```
brapse, line 932: @romac is the post condition that the header fetched has not experired satisfied in the current implementation?
Or maybe it's handled somewhere else? 🤔 
https://github.com/informalsystems/tendermint-rs/blob/118b4aee254d417446888f4cd53d5c54ac4c871e/light-client/src/store/memory.rs#L65-L71

docs/spec/lightclient/verification.md
```diff
@@ -844,8 +915,139 @@ If on the blockchain the validator set of the block at height
...
+    latest := current
+  }
+  return (lightStore, ResultSuccess)
+ }
+```
 
+The following function just decided based on the required height which
+method should be used.
 
+#### **[LCV-FUNC-IBCMAIN.1]**:
```
brapse, line 988: Nit but can we ensure the spacing and line format is consistent here? Should adhere to `go fmt` for readability.

josef-widder: Yeah. was a bit ugly. I tried to fix that by hand now. `gofmt` complained as it might be too much pseudo code. But it looks better now. please approve!

docs/spec/lightclient/verification.md
romac: It's checked here: https://github.com/informalsystems/tendermint-rs/blob/118b4aee254d417446888f4cd53d5c54ac4c871e/light-client/src/light_client.rs#L174-L180

docs/spec/lightclient/verification.md
konnov: `gofmt` apparently has a different opinion on documenting things :-)

Issue #339, opened 2020-06-17
[rpc] crate follow-ups


xla:
As leftovers from #338 we need to address to main points

* [x] crate documentation https://github.com/informalsystems/tendermint-rs/pull/338#discussion_r441742695
* [x] a satisfying solution to the now public serialisers: https://github.com/informalsystems/tendermint-rs/pull/338#discussion_r441768296
* [x] potentially make only endpoints available: https://github.com/informalsystems/tendermint-rs/issues/289#issuecomment-645719342

yihuang:
It seems keeping the endpoint data types inside tendermint crate, can help keeping the serializers private.

Liamsi:
> It seems keeping the endpoint data types inside tendermint crate, can also help keeping the serializers private.

@yihuang from your point of view and keeping the serializer helper methods aside, are their other good reasons to keep the rpc endpoints (or the whole module) inside the tendermint crate?

yihuang:
> > It seems keeping the endpoint data types inside tendermint crate, can also help keeping the serializers private.
> 
> @yihuang from your point of view and keeping the serializer helper methods aside, are their other good reasons to keep the rpc endpoints (or the whole module) inside the tendermint crate?

I don't have other reasons for that, I think feature guarding is good too ;D

ebuchman:
Given all the recent serialization work, can we close this @thanethomson @greg-szabo ?

Issue #337, opened 2020-06-17
[rpc] Overhaul rpc module


Liamsi:
## Motivation

As we are introducing an rpc service (aka light-node #219), now is a good time to look at different options regarding the rpc client (and soon server) lib.

The current approach seems to be to write all types and logic from scratch. 

This has the major benefit that we don't introduce any dependencies, and, that we can manually guarantee compatibility with the go implementation (which might not always adhere to standards, e.g. see: https://github.com/tendermint/tendermint/issues/2949).

The major drawback here is that we need to implement details that are not directly tied to tendermint core types or business logic. e.g., we need to make the jsonrpc wrapper type is properly tested (#298), or that we care about control messages in the web-socket implementation (https://github.com/informalsystems/tendermint-rs/issues/311). 

## Actionable Items

Reconsider using a library for most of the rpc server (and partly client) logic. 

For that a brief writeup (adr) on pros and cons on libraries leaving a documentation trace why a particular choice was made would be ideal.
Currently paritiy's crates look most promising here:
- https://github.com/paritytech/jsonrpc/tree/master/core
- https://github.com/paritytech/jsonrpc/tree/master/core-client
- more: https://github.com/paritytech/jsonrpc#sub-projects
But they might soon be replaced by an asny/awaitified variant: https://github.com/paritytech/jsonrpsee

## Drawbacks

- might introduce yet another dependency (always comes with hidden costs, e.g. see: https://github.com/informalsystems/tendermint-rs/issues/289 and https://github.com/crypto-com/chain/issues/1675#issuecomment-636871417)
- ecosystem is still moving fast (especially everything around runtimes / async/await etc)

### related

 - tendermint go rpc issue: https://github.com/tendermint/tendermint/issues/3367
 - adr regarding rpc in tendermint go: https://github.com/tendermint/tendermint/pull/4857

 - subscript / event listener issue: https://github.com/informalsystems/tendermint-rs/issues/313
 - async/await and rpc: https://github.com/informalsystems/tendermint-rs/issues/318
 - light node: #219

brapse:
Just to relay the conclusion from out of band conversations, I think this makes sense. Going with `jsonrpc` to provide consistent documentation is a win. Let's continue to be mindful of dependencies and the complexities that come with them. 🙏

Liamsi:
referencing this: https://github.com/informalsystems/tendermint-rs/issues/289#issuecomment-645867633

Makes most sense to me.

[...]

PR #343, opened 2020-06-18
rpc: guard client with feature flag

xla:
Guard client code paths To avoid the transient dependency on net related crates for users who are primarily interested in the core types.

Follow-up #339
Ref #337

* [ ] ~~Updated all code comments where relevant~~
* [ ] ~~Wrote tests~~
* [ ] Updated CHANGES.md


xla:
@yihuang Does this approach work for your use-case?

PR #349, opened 2020-06-19
Add list of spare and faulty nodes to PeerList

romac:
Neither list are currently used but have been added to match the spec more closely.

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/peer_list.rs
+++ b/light-client/src/peer_list.rs
@@ -5,18 +5,42 @@ use crate::{
     types::PeerId,
 };
 
-use contracts::pre;
-use std::collections::HashMap;
+use contracts::{post, pre};
+use std::collections::{HashMap, HashSet};
 
 /// A mapping from PeerIds to Light Client instances.
 /// Keeps track of which peer is deemed the primary peer.
 #[derive(Debug)]
 pub struct PeerList {
-    peers: HashMap<PeerId, Instance>,
+    instances: HashMap<PeerId, Instance>,
     primary: PeerId,
+    witnesses: HashSet<PeerId>,
+    full_nodes: HashSet<PeerId>,
+    faulty_nodes: HashSet<PeerId>,
 }
 
 impl PeerList {
+    /// Invariant maintained by a `PeerList`
+    ///
+    /// ## Implements
+    /// - [LCD-INV-NODES]
+    pub fn invariant(peer_list: &PeerList) -> bool {
+        peer_list.full_nodes.is_disjoint(&peer_list.witnesses)
+            && peer_list.full_nodes.is_disjoint(&peer_list.faulty_nodes)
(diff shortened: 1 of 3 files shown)


light-client/src/peer_list.rs
```diff
@@ -5,111 +5,169 @@ use crate::{
...
+            && peer_list.witnesses.is_disjoint(&peer_list.faulty_nodes)
+    }
+
+    /// Transition invariant maintained by a `PeerList`
+    ///
+    /// ## Implements
+    /// - [LCD-INV-NODES]
+    pub fn transition_invariant(_prev: &PeerList, _next: &PeerList) -> bool {
+        true
+        // TODO
```
Liamsi, line 39: Can you add a slightly more descriptive TODO?

light-client/src/peer_list.rs
```diff
@@ -5,111 +5,169 @@ use crate::{
...
+use std::collections::{HashMap, HashSet};
 
 /// A mapping from PeerIds to Light Client instances.
 /// Keeps track of which peer is deemed the primary peer.
 #[derive(Debug)]
 pub struct PeerList {
-    peers: HashMap<PeerId, Instance>,
+    instances: HashMap<PeerId, Instance>,
     primary: PeerId,
+    witnesses: HashSet<PeerId>,
```
Liamsi, line 17: witnesses are what is currently called Secondaries in the spec. Will the naming be aligned or is there a particular reason for the difference?

romac: I picked `witnesses` to be consistent with the naming in the multi-peer JSON tests, but you are right that we should align them. Let's already merge this and I'll open an issue to track this and to discuss which name we want to keep.

light-client/src/peer_list.rs
```diff
@@ -5,111 +5,169 @@ use crate::{
...
     pub fn primary_mut(&mut self) -> Option<&mut Instance> {
-        self.peers.get_mut(&self.primary)
+        self.instances.get_mut(&self.primary)
     }
 
     /// Get a list of references to all the witnesses,
     /// (ie. all peers which are not the primary).
     pub fn witnesses(&self) -> Vec<&Instance> {
-        self.peers
+        self.instances
```
OStevan, line 72: This method now seems to be wrong as it will return all `Instance` which are not `primary` (as per the docs above but not what we want when doing fork detection), therefore we might be returning faulty nodes, and we are going to contact all instances in this peer list. Effectively communicating with all full nodes instead of a subset defined in `witnesses`. This could probably be fixed in #397 @romac

romac: Fixed in https://github.com/informalsystems/tendermint-rs/pull/397/commits/e371271820e11568244b13ada2616e90b63133a9.

Issue #342, opened 2020-06-18
Deduplicate Light Client code


romac:
A lot of code in the `light-client` code was ported over from the `tendermint` crate without deleting the original code. As such, we have a lot of duplicated code between the two crates.

xla:
Ref #7

romac:
### Prerequisites:

- #341 
- #359

xla:
Superseeded by #437

PR #358, opened 2020-06-23
Refactor CommitValidator and port over code from the `lite_impl` module

### LINKED ISSUE #341: Extract `tendermint::lite_impl::SignedHeader::validate` into `CommitValidator`
[This method](https://github.com/informalsystems/tendermint-rs/blob/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2/tendermint/src/lite_impl/signed_header.rs#L48-L91) should be moved into the `light-client`'s [`CommitValidator`](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/operations/commit_validator.rs).

romac:
Close: #341
Ref: #342 

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/operations/commit_validator.rs
+++ b/light-client/src/operations/commit_validator.rs
@@ -4,15 +4,21 @@ use crate::{
     types::{SignedHeader, ValidatorSet},
 };
 
-use anomaly::BoxError;
-use tendermint::lite::types::Commit as _;
+use tendermint::block::CommitSig;
+use tendermint::lite::types::ValidatorSet as _;
 
 pub trait CommitValidator: Send {
     fn validate(
         &self,
         signed_header: &SignedHeader,
         validators: &ValidatorSet,
-    ) -> Result<(), BoxError>;
+    ) -> Result<(), VerificationError>;
+
+    fn validate_full(
+        &self,
+        signed_header: &SignedHeader,
+        validator_set: &ValidatorSet,
+    ) -> Result<(), VerificationError>;
 }
 
 #[derive(Copy, Clone)]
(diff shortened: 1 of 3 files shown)


light-client/src/operations/commit_validator.rs
```diff
@@ -39,7 +46,39 @@ impl CommitValidator for ProdCommitValidator {
...
+        Ok(())
+    }
+
+    // This check is only necessary if we do full verification (2/3)
+    //
+    // See https://github.com/informalsystems/tendermint-rs/issues/281
+    //
+    // It returns `ImplementationSpecific` error if it detects a signer
+    // that is not present in the validator set
+    fn validate_full(
```
liamsi, line 58: it makes sense to decouple these and have a separate `validate_full` method instead 👍

light-client/src/predicates.rs
```diff
@@ -73,12 +73,10 @@ pub trait VerificationPredicates: Send {
...
         validators: &ValidatorSet,
-        validator: &dyn CommitValidator,
+        commit_validator: &dyn CommitValidator,
     ) -> Result<(), VerificationError> {
-        // FIXME: Do not discard underlying error
-        validator
-            .validate(signed_header, validators)
-            .map_err(|e| VerificationError::InvalidCommit(e.to_string()))?;
+        commit_validator.validate(signed_header, validators)?;
+        commit_validator.validate_full(signed_header, validators)?;
```
liamsi, line 79: Is there actually a place where `commit_validator.validate` is used without `validate_full`?

romac: No there's not. We currently perform both checks, and I don't see the need/place for doing just `validate` since we do 2/3 full verification of every block by that point. But perhaps that's not right, and we need to only to `validate_full` somewhere here: https://github.com/informalsystems/tendermint-rs/pull/358/files/b1aee8d1f4b42d56a5e43dda9fc52572455ff4ea#diff-1911c69336354ecbef5a4524721976faR236-R263

ebuchman: if we do https://github.com/informalsystems/tendermint-rs/issues/377, we can prob get rid of validate_full, especially since we're trying not to rely on the underlying validator_address. though I suppose so long as it is there it should be checked.

PR #363, opened 2020-06-23
light-node: RPC server

### LINKED ISSUE #219: Light Client RPC Service

xla:
Adds an RPC server to expose internal state of a light node. Besides adding the general infrastructure it provides the first endpoint under `/state` as described in #219.

First part of #219
Depends on #394 
Depends on #403


light-node/Cargo.toml
```diff
@@ -9,6 +9,10 @@ publish = false
 abscissa_tokio = "0.5"
 async-trait = "0.1"
 gumdrop = "0.7"
+jsonrpc-core = "14.2"
+jsonrpc-core-client = "14.2"
+jsonrpc-http-server = "14.2"
+jsonrpc-derive = "14.2"
```
liamsi, line 15: do we plan to derive rpc functionality from exiting structs / methods (I'm all for it but that means less control over the implementation I guess).

xla: The upside of that approach is that you get a client and a server. Not fuzzed either way, just thought it would be helpful for utilitiies that operate against the running API.

light-client/src/supervisor.rs
```diff
@@ -17,21 +17,26 @@ use tendermint::evidence::{ConflictingHeadersEvidence, Evidence};
...
-/// Type alias for readability
+/// Verification operation [`Result`] type alias for readability.
 pub type VerificationResult = Result<LightBlock, Error>;
 
-/// Events which are exchanged between the `Supervisor` and its `Handle`s.
+/// Input events which are sent by the [`Handle`] to its [`Supervisor`]..
 #[derive(Debug)]
-pub enum Event {
-    // Inputs
+pub enum Input {
```
liamsi, line 25: Might be a good call to split this up. 
Throwing in the suggestion to rename this to `InputEvents` / `OutputEvents`?

light-client/src/supervisor.rs
```diff
@@ -17,21 +17,26 @@ use tendermint::evidence::{ConflictingHeadersEvidence, Evidence};
...
     /// Terminate the supervisor process
     Terminate(Callback<()>),
     /// Verify to the highest height, call the provided callback with result
     VerifyToHighest(Callback<VerificationResult>),
     /// Verify to the given height, call the provided callback with result
     VerifyToTarget(Height, Callback<VerificationResult>),
 
-    // Outputs
+    /// Get the latest block from the primary.
+    GetLatest(Callback<Result<LightBlock, Error>>),
```
liamsi, line 34: `LatestLightBlock`?

light-client/src/supervisor.rs
```diff
@@ -17,21 +17,26 @@ use tendermint::evidence::{ConflictingHeadersEvidence, Evidence};
 use contracts::pre;
 use crossbeam_channel as channel;
 
-/// Type alias for readability
+/// Verification operation [`Result`] type alias for readability.
 pub type VerificationResult = Result<LightBlock, Error>;
 
-/// Events which are exchanged between the `Supervisor` and its `Handle`s.
+/// Input events which are sent by the [`Handle`] to its [`Supervisor`]..
```
liamsi, line 23: 👍

xla:
Supervisor changes have been split out into #394.

xla:
Depends on #401

light-node/src/rpc.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    impl Handle for MockHandle {
+        fn latest_trusted(&self) -> Result<Option<LightBlock>, Error> {
+            let block: LightBlock = serde_json::from_str(LIGHTBLOCK_JSON).unwrap();
+
+            Ok(Some(block))
+        }
+    }
+
+    const LIGHTBLOCK_JSON: &str = r#"
```
liamsi, line 122: Is that needed twice? Here and above?

xla: The example is temporary to help with the integration work, wouldn't it to leave as is mid-term.

light-node/src/error.rs
```diff
@@ -1,80 +1,38 @@
...
-}
-
-impl ErrorKind {
-    /// Create an error context from this error
-    pub fn context(self, source: impl Into<BoxError>) -> Context<ErrorKind> {
-        Context::new(self, Some(source.into()))
-    }
-}
+    #[error("i/o error: {0}")]
+    Io(String),
```
liamsi, line 22: Just out of curiosity: why did you decide to wrap the underlying error as a string here (as opposed to using the `#[from]` directive)?

light-node/examples/rpc.rs
```diff
@@ -0,0 +1,146 @@
+//! Basic example of running the RPC server. This is a temporary show-case and should be removed
+//! once integrated in the light node proper. To test the `/state` endpoint run:
```
liamsi, line 2: 👍

light-node/src/error.rs
```diff
@@ -1,80 +1,38 @@
...
+        Self::Io(format!("{}", err))
     }
 }
 
-impl From<io::Error> for Error {
-    fn from(err: io::Error) -> Self {
-        ErrorKind::Io.context(err).into()
+impl Into<jsonrpc_core::types::Error> for Kind {
+    fn into(self) -> jsonrpc_core::types::Error {
+        todo!()
```
liamsi, line 36: Did you leave this a `todo!()` on purpose?

light-node/src/rpc.rs
```diff
@@ -0,0 +1,240 @@
...
+
+use crate::error;
+
+pub use sealed::{Client, Rpc, Server};
+
+/// Run the given [`Server`] on the given address and blocks until closed.
+///
+/// n.b. The underlying server has semantics to close on drop. Also does it not offer any way to
+/// get the underlying Future to await, so we are left with this rather rudimental way to control
+/// the lifecyel. Should we be interested in a more controlled way to close the server we can
```
liamsi, line 14:
```suggestion
/// the lifecycle. Should we be interested in a more controlled way to close the server we can
```

light-node/src/error.rs
xla: Because `io::Error` doesn't implement `Clone`. Which we mind not need anymore on our `Kind` as I refrained from returning our custom errors in the rpc server, as I wasn't able to get mapping to the jsonrpc error to work.

light-node/src/error.rs
xla: See https://github.com/informalsystems/tendermint-rs/pull/363/files#r450373699 - I think the whole conversion can go. We don't use it at the moment.

light-node/Cargo.toml
```diff
@@ -7,11 +7,18 @@ publish = false
 
 [dependencies]
 abscissa_tokio = "0.5"
+anomaly = { version = "0.2", features = [ "serializer" ] }
```
liamsi, line 10: @tarcieri @tony-iqlusion Mind chiming in here? It's been a while since I actively used all 3 dependencies: `anomaly`, `thiserror`, and `abscissa`. Does it make sense to mix those here? Doesn't make abscissa using anomaly in the same crate obsolete? I'm a bit lost here.

tarcieri: Error handling in Abscissa is in the middle of an overhaul to switch to [`eyre`](https://github.com/yaahc/eyre) (or if the improvements in `eyre` end up being upstreamable, potentially [`anyhow`](https://github.com/dtolnay/anyhow)) /cc @yaahc

liamsi: Good to know thanks!

yaahc: It looks like there's a decent chance that the changes will get upstreamed back into `anyhow`, not sure how long it will be though.

light-node/src/rpc.rs
```diff
@@ -0,0 +1,242 @@
...
+use tendermint_light_client::supervisor::Handle;
+
+use crate::error;
+
+pub use sealed::{Client, Rpc, Server};
+
+/// Run the given [`Server`] on the given address and blocks until closed.
+///
+/// n.b. The underlying server has semantics to close on drop. Also does it not offer any way to
+/// get the underlying Future to await, so we are left with this rather rudimental way to control
```
brapse, line 13: s/rudimental/rudimentary

brapse:
```suggestion
/// get the underlying Future to await, so we are left with this rather rudimentary way to control
```

light-node/src/error.rs
romac: One way to work around the lack of `Clone` on `io::Error` is to use the `anomaly::Context` wrapper (as shown in the [docs](https://docs.rs/anomaly/0.2.0/anomaly/index.html#usage)), which would look something like that:

```diff
diff --git a/light-node/src/error.rs b/light-node/src/error.rs
index 094799b..7162c03 100644
--- a/light-node/src/error.rs
+++ b/light-node/src/error.rs
@@ -1,8 +1,9 @@
 //! Error types
 
-use std::io;
 use std::net;
 
+use anomaly::{BoxError, Context};
+
 /// Error type
 pub type Error = anomaly::Error<Kind>;
 
@@ -18,12 +19,15 @@ pub enum Kind {
     Config,
 
     /// Input/output error
-    #[error("i/o error: {0}")]
-    Io(String),
+    #[error("i/o error")]
+    Io,
 }
 
-impl From<io::Error> for Kind {
-    fn from(err: io::Error) -> Self {
-        Self::Io(format!("{}", err))
+impl Kind {
+    /// Add additional context (i.e. include a source error and capture a backtrace).
+    /// You can convert the resulting `Context` into an `Error` by calling `.into()`.
+    pub fn context(self, source: impl Into<BoxError>) -> Context<Self> {
+        Context::new(self, Some(source.into()))
     }
 }
+
diff --git a/light-node/src/rpc.rs b/light-node/src/rpc.rs
index 544dcb4..f6b190f 100644
--- a/light-node/src/rpc.rs
+++ b/light-node/src/rpc.rs
@@ -25,7 +25,7 @@ where
             AccessControlAllowOrigin::Any,
         ]))
         .start_http(&addr.parse().map_err(error::Kind::from)?)
-        .map_err(error::Kind::from)?;
+        .map_err(|e| error::Kind::Io.context(e))?;
 
     srv.wait();
```

romac: PR at #413

xla: Yeah was looking at the `Context` as we also use it in the light-client crate. You do loose the `From` semantics with it, correct? As in you have to know which variant to construct when mapping.

romac:
Great work @xla! 🥇

light-node/src/error.rs
romac: > You do loose the From semantics with it, correct? As in you have to know which variant to construct when mapping.

I would say that you just have to be more explicit, rather than let the compiler pick the appropriate `From` instance based on the (static) type of the underlying error, which in this case is a `io::Error`. So you don't really lose anything aside from the `From` instance, which in this case is not appropriate for `io::Error` because of the lack of `Clone` bound. If `io::Error` was `Clone`, then you would use `#[from]` like you did for `net::AddrParseError`.

PR #365, opened 2020-06-23
Fix exit condition in `verify_to_target`

romac:
Close: #332

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/light_client.rs
+++ b/light-client/src/light_client.rs
@@ -171,6 +171,13 @@ impl LightClient {
                 .highest(VerifiedStatus::Verified)
                 .ok_or_else(|| ErrorKind::NoInitialTrustedState)?;
 
+            if target_height < trusted_state.height() {
+                bail!(ErrorKind::TargetLowerThanTrustedState {
+                    target_height,
+                    trusted_height: trusted_state.height()
+                });
+            }
+
             // Check invariant [LCV-INV-TP.1]
             if !is_within_trust_period(&trusted_state, options.trusting_period, options.now) {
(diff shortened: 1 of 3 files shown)


light-client/src/light_client.rs
```diff
@@ -182,9 +182,8 @@ impl LightClient {
             // Trace the current height as a dependency of the block at the target height
             state.trace_block(target_height, current_height);
 
-            // If the trusted state is now at the height greater or equal to the target height,
-            // we now trust this target height, and are thus done :) [LCV-DIST-LIFE.1]
-            if target_height <= trusted_state.height() {
+            // If the trusted state is now at a height equal to the target height, we are done. [LCV-DIST-LIFE.1]
+            if target_height == trusted_state.height() {
```
liamsi, line 186: The case  `target_height` < `trusted_height` still needs to be spec'ed out I assume?

romac: It's already spec'ed out in a draft PR (#340) but implementing it is not a priority since the IBC relayer will in the end not need it.

romac: We might need some tests to make sure we error out properly when that case occurs, though.

milosevic: Do we still need to add somewhere logic for `target_height < trusted_height` case? We can just error for the moment in that case.

romac: Yes, I am on it :)

romac: Done

romac:
Should we already merge this and address the case where the target height is below the trusted state but we have a previous trusted state within the trusting period that's below the target height?

Issue #320, opened 2020-06-11
[light-client] Testing Regimen meta-issue


brapse:
In anticipation of the release of Tendermint-rs Light Client we outline here want we want to be tested. This issue is meant to collect a bunch of sub issues and provide an overview of how correctness will be codified in unit and integration tests. If you are interested in contributing to the test, simply create a break out issue of one of the items and update this issue with a reference ☺️ 

## Verification
The verifier is composed of multiple predicates, each of which should be tested for both the positive and negative case using concrete types. `operations` should be mocked out.

## Fork Detection: `light-client/src/fork_detector.rs`
The Fork detector should be tested in isolation using concrete types to detect the following cases
* Header matches
* Headers differ but verification fails
* Headers differ but verification succeeds

## Bisection: `light-client/tests/light_client.rs`
(#290)
* All conformance Tests should pass including `commit/one_third_vals_don’t_sign.json"`

## Operations: `light-client/src/operaitons/*`
The operators provide an abstraction around crypto functions and allow traits to help mock them out during testing. Eventually these should be moved to the main crate but they should be moved there with accompanying tests. Each operator should be tested using concrete types.

* Voting power calculation
* Commit validation
* Header Hashing

## Supervisor verify_to_target integration test: `light-client/src/supervisor.rs`
(#316)
The supervisor provides verification using multiple peers. It integrates the io, verification, bisection scheduling, fork detection and evidence reporting. All of which can be mocked out to exercise complex scenarios including
* Perform Bisection into a fork with failed verification (true fork)
    * Should report evidence
* Switching from primary to secondary
* Exhausting peer list

## Node Integration
(#304)
*  Launch a node with the CLI and sync against the Tendermint proxy App

It’s possible and even likely that this doesn’t cover everything we want to test. Please feel free to discuss in the comments and edit this description as need be.

josef-widder:
It would be good to have tests when the primary (or secondaries) are not responding. In this case we should timeout and FetchLightBlock should report timeout errors. In case of verification we just terminate with failure then. In the detection we remove the secondary from the peer list

romac:
Just added support for timeout to the `Io` component in the supervisor branch: https://github.com/informalsystems/tendermint-rs/pull/302/commits/8fca5c184edbf78c72f9ad5b412e04fe57babe93

romac:
Implemented removal of witness which timed out in 56bc022.

brapse:
> It would be good to have tests when the primary

[...]

PR #366, opened 2020-06-23
Conformance test fixes

### LINKED ISSUE #290: Light client conformance tests differences between description and real errors

Shivani912:
closes #290 
partially addresses: #320 (bisection)

**UPDATE**: moved `single_peer/invalid_validator_set` to `multi_peer/malicious_validator_set` because it seemed more appropriate. 

**QUESTION**: Why do we don't run the `multi_peer/...` tests yet? 

* [ ] Wrote tests
* [ ] Updated CHANGES.md


light-client/tests/light_client.rs
```diff
@@ -254,23 +254,12 @@ fn run_bisection_test(tc: TestBisection<LightBlock>) {
...
     for file_path in paths {
         let dir_entry = file_path.unwrap();
         let fp_str = format!("{}", dir_entry.path().display());
 
-        if skipped
-            .iter()
-            .any(|failing_case| fp_str.ends_with(failing_case))
-        {
-            println!("Skipping JSON test: {}", fp_str);
-            return;
```
liamsi, line 271: ouch, this should have been `continue` instead of `return`

light-client/tests/light_client.rs
```diff
@@ -319,9 +308,9 @@ fn bisection() {
 #[test]
 fn single_step_sequential() {
     let dirs = [
-        "single_step/sequential/commit",
+        // "single_step/sequential/commit",
         "single_step/sequential/header",
-        "single_step/sequential/validator_set",
+        // "single_step/sequential/validator_set",
```
liamsi, line 313: Why are these dirs now commented out?

Shivani912: Oops, my bad!

liamsi:
> QUESTION: Why do we don't run the multi_peer/... tests yet?

I think when you had them ready, we added them to this repo without the rust code being ready? 
https://github.com/informalsystems/tendermint-rs/blob/27298f0e7a5089096893f097d910f3e9271bf850/tendermint/tests/lite.rs#L193-L196

light-client/tests/light_client.rs
liamsi: All good, thanks!

light-client/tests/light_client.rs
liamsi: I think I introduced this and it was copied over and while it was fixed in the `lite` module, I didn't noticed the same here.

https://github.com/informalsystems/tendermint-rs/blob/27298f0e7a5089096893f097d910f3e9271bf850/tendermint/tests/lite.rs#L251-L256

Shivani912:
> > QUESTION: Why do we don't run the multi_peer/... tests yet?
> 
> I think when you had them ready, we added them to this repo without the rust code being ready?
> 
> https://github.com/informalsystems/tendermint-rs/blob/27298f0e7a5089096893f097d910f3e9271bf850/tendermint/tests/lite.rs#L193-L196

Possible, `multi_peer` tests fail at the moment in rust code. Seems to me it's not fetching headers from a witness. Will have to take a closer look though!

Shivani912:
Thanks @liamsi ! Will do that :)

Issue #369, opened 2020-06-23
Missing multi peer conformance tests


Shivani912:
Initially, `bisection/multi_peer` tests were not being used because the light client wasn't capable of handling multiple peers at that time, but we definitely need it now.

**Need help here**:
We already have those JSON test files in code and only need a way to run them. When I run it using the existing test code (`tests/light_client.rs`), it fails because it doesn't really seem to fetch headers from witness/secondary provider and probably needs the `supervisor` here. I'm not too familiar with the current light client implementation and would appreciate some help here!

PR #371, opened 2020-06-24
Multi-peer conformance tests

romac:
Closes: #369

* [ ] Wrote tests
* [ ] Updated CHANGES.md


Shivani912:
The test currently panics at `'Pre-condition of mark_witness_as_faulty violated: peer_id != self . primary && self . witnesses . contains (& peer_id)'` ([here](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/peer_list.rs#L85) in code) 

This is probably because it marks the primary provider faulty here(?) 
Maybe @romac can help here!

**Run test:** 
`cargo test run_multipeer_tests -- --nocapture`

**Gives output:**
Screen Shot 2020-06-30 at 6 13 19 PM

light-client/src/tests.rs
```diff
@@ -75,6 +82,72 @@ impl Trusted {
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct MockEvidenceReporter;
+
+#[contract_trait]
+impl EvidenceReporter for MockEvidenceReporter {
+    fn report(&self, _e: Evidence, _peer: PeerId) -> Result<Hash, IoError> {
+        Ok(Hash::new([0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]))
```
romac, line 141: Rust has a [short-hand](https://doc.rust-lang.org/stable/rust-by-example/primitives/array.html) for initializing arrays and slices with the same value :)

```suggestion
        Ok(Hash::new([0; 32]))
```

romac:
> This is probably because it marks the primary provider faulty here(?)

Yes, it's a flaw in the current PeerList API, which I've attempted to fix in #397. I think you can disregard this error until we merge #397 and adapt this PR to the new, improved API.

romac:
On second thought, perhaps that's an actual bug in the supervisor. Will take a closer look.

romac:
So yeah it's a bug. We should check whether the faulty peer is the primary or not and call the appropriate method if that's the case (`swap_primary`), but the PeerList API currently does not let us perform this check. In #397, we now expose this information, but could envision having a single method on `PeerList` which would perform the check and dispatch to the right method for us.

romac:
Scratch that, the issue is actually the faulty witness is not part of the witnesses list. Need to investigate further to see why that's the case.

romac:
After merging #397 in this branch, the error goes away (likely due to a bug that was fixed PeerList where it would not perform primary/witness swap properly).

Tests then fail with `ErrorKind::NoTrustedState(Trusted)`, because the light store was initialized with a `Verified` block instead of a `Trusted` one.

After fixing this, the first two tests pass but the third one fails with:

```
Running Test Case: Case: Trusted height=1, found conflicting valid commit at height=11 from only one of the two witnesses, should not expect error
with Primary Peer: node::Id(BADFADAD0BEFEEDC0C0ADEADBEEFC0FFEEFACADE)
Witness: ADBEEFC0FFEEFACADEBADFADADE0BEFEEDC0C0AD
Witness: BADFADBEEFC0FFEEFACADEAD0BEFEEDC0C0ADEAD
[light-client/tests/supervisor.rs:145] e = Error(
    Context {
        kind: NoWitnessLeft,
        source: None,
        backtrace: Some(

/* snip */
```

To be investigated further.

light-client/src/tests.rs
```diff
@@ -129,16 +208,46 @@ impl From<Provider<AnonLightBlock>> for Provider<LightBlock> {
...
+    peer_ids[count]
+}
+
 impl From<TestBisection<AnonLightBlock>> for TestBisection<LightBlock> {
     fn from(tb: TestBisection<AnonLightBlock>) -> Self {
+        let mut witnesses: Vec<WitnessProvider<LightBlock>> = tb.witnesses.into_iter().map(Into::into).collect();
+
+        for (count, provider) in witnesses.iter_mut().enumerate() {
+            for lb in provider.value.lite_blocks.iter_mut() {
+                lb.provider = random_peer_id(count);
```
brapse, line 237: Is this random or is it deterministic? 🤔

romac:
@Shivani912 Just merged `master` in now that #397 has been merged, and fixed the trusted state issue. We are now left with the `NoWitnessLeft` error.

light-client/src/tests.rs
romac: This is actually deterministic: https://github.com/informalsystems/tendermint-rs/pull/371/files#diff-11a3f1d34a7a9515595b37dba6035f46R222-R232

light-client/src/tests.rs
Shivani912: 🤦‍♀️

light-client/src/tests.rs
Shivani912: Initially, it was to be used for something else and then I didn't think about renaming it. Thanks for the catch @brapse !

Shivani912:
Thanks @romac for your help here! I'll take care of the failing test

Shivani912:
This is interesting... So the test case here is that there is a primary peer and 2 witnesses. One of the witnesses supports the data sent by the primary one and the other witness doesn't. Looking at the backtrace of the test, supervisor actually tries to replace the primary peer marking it "faulty". Shouldn't it mark the witness who sent conflicting data faulty?

romac:
Merged `master` in after merging #407. This got rid of the error above, but the test now fails because when we detect an actual fork, [we raise a `ForkDetected` error](https://github.com/informalsystems/tendermint-rs/blob/547379daf73839883037b65f97d3ceb373f499b2/light-client/src/supervisor.rs#L202), which the test is not expecting.

@Shivani912 @josef-widder @OStevan @melekes: What do you think should be the proper behavior when we detect a fork? Here is what we currently do:

- If verification with the primary fails for any reason, we replace the faulty primary and perform verification again
- Otherwise, if the fork detector suspects
   - a) a faulty witness, we replace the faulty witness and perform verification again
   - b) that a witness timed out, we replace the faulty witness and perform verification again
   - c) an actual fork, we report the fork, raise an error and abort

Case a) and b) seem correct to me, but perhaps we should replace the witness and keep trying to verify in case c), rather than abort?

Shivani912:
> 
> @Shivani912 @josef-widder @OStevan @melekes: What do you think should be the proper behavior when we detect a fork? Here is what we currently do:
> 
> * If verification with the primary fails for any reason, we replace the faulty primary and perform verification again
> * Otherwise, if the fork detector suspects
>   
>   * a) a faulty witness, we replace the faulty witness and perform verification again
>   * b) that a witness timed out, we replace the faulty witness and perform verification again
>   * c) an actual fork, we report the fork, raise an error and abort
> 
> Case a) and b) seem correct to me, but perhaps we should replace the witness and keep trying to verify in case c), rather than abort?

After our discussion on the call, it seems appropriate that if we detect an actual fork, option (c) is the right expectation here and test should be modified to expect so. This reveals an inconsistency between Go (v 0.33) and Rust code as the Go light client does not raise an error on finding a fork and simply ignores it, marks it faulty and moves to verify headers from other witnesses. Also, this have been improved in their latest (master) code and reports on finding a fork. (So not raising an issue for this anymore @milosevic )

Shivani912:
>  After fixing this, the first two tests pass but the third one fails with:
>  ```
>  Running Test Case: Case: Trusted height=1, found conflicting valid commit at height=11 from only one of the two witnesses, should not expect error
>  with Primary Peer: node::Id(BADFADAD0BEFEEDC0C0ADEADBEEFC0FFEEFACADE)
>  Witness: ADBEEFC0FFEEFACADEBADFADADE0BEFEEDC0C0AD
>  Witness: BADFADBEEFC0FFEEFACADEAD0BEFEEDC0C0ADEAD
>  [light-client/tests/supervisor.rs:145] e = Error(
>      Context {
>          kind: NoWitnessLeft,
>          source: None,
>          backtrace: Some(
>  
>  /* snip */
 > ```


Also, this test fails with `NoWitnessLeft` error which should instead be something like `ForkDetected` or `FoundConflictingHeaders`. This fix is outside the scope of this PR, but just noting it here.

melekes:
> Case a) and b) seem correct to me, but perhaps we should replace the witness and keep trying to verify in case c), rather than abort?

We're considering it as a future improvement. See https://github.com/tendermint/tendermint/issues/4694. However, it won't be so simple if it's a full fork (2/3+ forked the chain), in which case it's probably safer to stop everything.

brapse:
Yes it seems like we want to do `C` and have this be an error instead of a panic.

We really should have clarity on the proper termination behaviour of light clients. I made an issue in the spec repo here https://github.com/tendermint/spec/issues/114.

romac:
> Yes it seems like we want to do `C` and have this be an error instead of a panic.

Perhaps I wasn't clear, but the bullet points above describe the current behavior, not alternatives to be chosen from, so that's already what we are doing.

We indeed raise an error, and it's then up to the consumer of the API to either handle it or let it bubble up and eventually panic in the main function.

romac:
@brapse Should we merge this already until we get clarification in tendermint/spec#114?

light-client/tests/supervisor.rs
```diff
@@ -128,20 +114,25 @@ fn run_multipeer_test(tc: TestBisection<LightBlock>) {
     let target_height = tc.height_to_verify.try_into().unwrap();
 
     match handle.verify_to_target(target_height) {
-        Ok(_new_states) => {
+        Ok(new_state) => {
+            /// Check that the expected state and new_state match
```
romac, line 118:
```suggestion
            // Check that the expected state and new_state match
```

light-client/tests/supervisor.rs
```diff
@@ -128,20 +114,25 @@ fn run_multipeer_test(tc: TestBisection<LightBlock>) {
...
+            let untrusted_light_block = io
+                .fetch_light_block(primary, AtHeight::At(target_height))
+                .expect("header at untrusted height not found");
+
+            let expected_state = untrusted_light_block;
+            assert_eq!(new_state.height(), expected_state.height());
+            assert_eq!(new_state, expected_state);
+
+            /// Check the verdict
```
romac, line 127:
```suggestion
            // Check the verdict
```

Issue #362, opened 2020-06-23
Introduce `Trusted` verification state in light store


romac:
See the draft English spec for fork detection in #287.

PR #375, opened 2020-06-25
Introduce a `Trusted` status for light blocks which passed fork detection

romac:
Close: #362
cf. #287

* [ ] Wrote tests
* [ ] Updated CHANGES.md


romac:
This PR unfortunately [introduces a hack](https://github.com/informalsystems/tendermint-rs/pull/375/files#diff-7798509be617f6e992ae4bccb801ebb3R160-R176) to work around the limitations of the borrow checker, which might warrant a more thorough refactor of the supervisor, perhaps by using events internally. Will open an issue to discuss a potential design.

light-client/src/contracts.rs
```diff
@@ -40,6 +38,6 @@ pub fn light_store_contains_block_within_trusting_period(
 //     target_height: Height,
 // ) -> bool {
 //     light_store
-//         .all(VerifiedStatus::Verified)
+//         .all(Status::Verified)
 //         .all(|lb| lb.height() < target_height)
 // }
```
liamsi, line 43: Why is this commented out method in here?

light-client/src/supervisor.rs
```diff
@@ -154,6 +157,23 @@ impl Supervisor {
     /// Verify either to the latest block (if `height == None`) or to a given block (if `height == Some(height)`).
     #[pre(self.peers.primary().is_some())]
     fn verify(&mut self, height: Option<Height>) -> VerificationResult {
+        self.verify_inner(height).map(|trusted_lightblock| {
+            // We can now trust this block
+            self.peers
+                .primary_mut()
+                .unwrap() // SAFETY:`verify_inner` wouldn't have returned `Ok` if there were no more primary
```
liamsi, line 164: So this can't panic because `verify_inner` would have failed before. Is this (unit)testable?

light-client/src/supervisor.rs
```diff
@@ -154,6 +157,23 @@ impl Supervisor {
...
+                .trust_block(&trusted_lightblock);
+
+            trusted_lightblock
+        })
+    }
+
+    /// Helper for `verify` to work around limitations of the borrow checker.
```
liamsi, line 171: Isn't it possible to hide this by making it a pure (private) function?

light-client/src/contracts.rs
romac: It's here because it is needed by the (commented-out) precondition of `LightClient::verify_to_target`. It's disabled because it used to cause problems before we had a `Clock` component. I just didn't get around to re-enabling the precondition since we have that. Thanks for reminding me to look into that, tracked in #382.

light-client/src/supervisor.rs
romac: I agree that it's not great, will think about a better and safer way of doing this!

light-client/src/supervisor.rs
romac: I guess we could declare it as a function within `verify` itself. Or did you mean as a plain function outside of the supervisor impl?

romac:
@liamsi Found a way to get rid of `verify_helper`: https://github.com/informalsystems/tendermint-rs/pull/375/commits/de5c5876754caf8ec67fdcbbf598ade7e807e122

light-client/src/supervisor.rs
liamsi: either way (nor sure what works best)

light-client/src/supervisor.rs
```diff
@@ -204,6 +187,16 @@ impl Supervisor {
                             }
                         }
                         ForkDetection::NotDetected => {
+                            // We need to re-ask for the primary here as the compiler
+                            // is not smart enough to realize that we do not mutate
+                            // the `primary` field of `PeerList` between the initial
+                            // borrow of the primary and here (can't blame it, it's
+                            // not that obvious).
```
liamsi, line 194: 👍

light-client/src/light_client.rs
```diff
@@ -168,8 +164,8 @@ impl LightClient {
             // Get the highest trusted state
             let trusted_state = state
                 .light_store
-                .highest(VerifiedStatus::Verified)
-                .ok_or_else(|| ErrorKind::NoInitialTrustedState)?;
+                .highest(Status::Verified)
```
OStevan, line 167: @romac  please confirm that the following explanation makes sense and I can open an issue. Give that the light store does not have a notion that `Trusted` blocks are also `Verified` we will probably always run into a situation where we immediately fail to do `verify_to_target` as the only block we have would be a block which is `Trusted`. Also, it will happen that after one run of `Supervisor` verification the highest block we'll get from here is not the one which was previously the target as that block is in the `Trusted` state. It might also make sense to update single peer tests to start of with `Trusted` light blocks as that would immediately show that we can not do correct verification if the only block we have is trusted.

romac: Yeah you are right, this should actually fetch the latest verified or trusted block from the store. Will fix!

romac: Fixed in https://github.com/informalsystems/tendermint-rs/pull/407

PR #386, opened 2020-06-27
lightclient spec readme

ebuchman:
Add a README to the docs/spec/lightclient to give an overview of the work.

I'm not sure what to do with all the MC files but we should clean that up (presumably we'll collect more of these for other specs here and we need to namespace them in a dir or something) and provide some more details on what they are and how to run things - @konnov maybe you can help here?


josef-widder:
Thanks for taking the initiative on the readme file! I was on my todo list as well. Regarding the model checking results, I discussed with @konnov on Friday, and the easiest would be to copy the table that summarizes the model checking result from the paper we recently submitted.  
Regarding the TLA+ and the model checking files we have to discuss. It would be good to have them close to the English spec. However, if we think about CI in the future, and the troubles we had with github the last days, perhaps we want them in a different directory...

konnov:
I agree that having the `MC*.tla` files is annoying. But I would consider them as functional tests, as they are in fact inputs to the model checker. From an organization perspective, it makes sense to move them in a separate directory, but then TLA+ tools require all files to be in the same directory, so we would have to keep two copies of `Lightclient_A_1.tla`.

josef-widder:
perhaps for now we just add a tla subdirectory (docs/specs/lightclient/tla) and move all tla related stuff there?

ebuchman:
Can we have one dir per spec, so like a `light-client/verification` and ` light-client/fork-accountability`, or are they going to need to reference each other possibly? If they're all in the same dir we need to prefix the MC files with the spec they're relevant for

PR #394, opened 2020-06-29
light-client: expose latest_trusted on Supervisor Handle

xla:
Part of #219 and preparation for #363. Wanted to split out the new additions to the light-client `Supervisor` and its `Handle` to make it easier to review and move forward. The declared goal of this change-set is to introduce new API to full-fill the planned endpoints of #219.

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## pending
+
+Light Client:
+
+- Expose latest_trusted from Supervisor Handle ([#394])
+
 ## [0.14.1] (2020-06-23)
 
 - Update `prost-amino`/`prost-amino-derive` to v0.6 ([#367])
--- a/light-client/src/supervisor.rs
+++ b/light-client/src/supervisor.rs
@@ -1,3 +1,8 @@
+use contracts::pre;
+use crossbeam_channel as channel;
+
+use tendermint::evidence::{ConflictingHeadersEvidence, Evidence};
+
 use crate::{
     bail,
     callback::Callback,
@@ -10,17 +15,13 @@ use crate::{
     types::{Height, LightBlock, PeerId, Status},
 };
 
-use tendermint::evidence::{ConflictingHeadersEvidence, Evidence};
-
-use contracts::pre;
-use crossbeam_channel as channel;
-
 /// Type alias for readability
 pub type VerificationResult = Result<LightBlock, Error>;
 
-/// Events which are exchanged between the `Supervisor` and its `Handle`s.
+/// Input events sent by the [`Handle`]s to the [`Supervisor`]. They carry a [`Callback`] which is
+/// used to communicate back the responses of the requests.
 #[derive(Debug)]
-pub enum Event {
+enum HandleInput {
     // Inputs
     /// Terminate the supervisor process
     Terminate(Callback<()>),
@@ -28,14 +29,6 @@ pub enum Event {
     VerifyToHighest(Callback<VerificationResult>),
     /// Verify to the given height, call the provided callback with result
     VerifyToTarget(Height, Callback<VerificationResult>),
-
-    // Outputs
-    /// The supervisor has terminated
-    Terminated,
-    /// The verification has succeded
-    VerificationSuccess(Box<LightBlock>),
-    /// The verification has failed
-    VerificationFailure(Error),
 }
 
 /// An light client `Instance` packages a `LightClient` together with its `State`.
@@ -108,9 +101,9 @@ pub struct Supervisor {
     /// Reporter of fork evidence
     evidence_reporter: Box<dyn EvidenceReporter>,
     /// Channel through which to reply to `Handle`s
-    sender: channel::Sender<Event>,
+    sender: channel::Sender<HandleInput>,
     /// Channel through which to receive events from the `Handle`s
-    receiver: channel::Receiver<Event>,
+    receiver: channel::Receiver<HandleInput>,
 }
 
 impl std::fmt::Debug for Supervisor {
@@ -131,7 +124,7 @@ impl Supervisor {
         fork_detector: impl ForkDetector + 'static,
         evidence_reporter: impl EvidenceReporter + 'static,
     ) -> Self {
-        let (sender, receiver) = channel::unbounded::<Event>();
+        let (sender, receiver) = channel::unbounded::<HandleInput>();
 
         Self {
             peers,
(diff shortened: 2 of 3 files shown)


light-client/src/supervisor.rs
```diff
@@ -313,51 +314,58 @@ impl Supervisor {
...
 impl Handle {
     /// Crate a new handle that sends events to the supervisor via
     /// the given channel. For internal use only.
-    pub fn new(sender: channel::Sender<Event>) -> Self {
+    fn new(sender: channel::Sender<HandleInput>) -> Self {
         Self { sender }
     }
 
+    /// Get latest trusted block from the [`Supervisor`].
+    pub fn latest_trusted(&mut self) -> Result<Option<LightBlock>, Error> {
```
liamsi, line 328: In what case does this return Ok(`None`)?

xla: When the underlying primary returns `None` for its `latest_trusted`. This is really a direct pass-through, and judging by the method on the Instance the latest trusted being none should be expected and is non-errornous behaviour.

light-client/src/supervisor.rs
```diff
@@ -10,32 +15,21 @@ use crate::{
...
     /// Verify to the given height, call the provided callback with result
     VerifyToTarget(Height, Callback<VerificationResult>),
-
-    // Outputs
-    /// The supervisor has terminated
-    Terminated,
-    /// The verification has succeded
-    VerificationSuccess(Box<LightBlock>),
-    /// The verification has failed
-    VerificationFailure(Error),
```
liamsi, line 38: Interesting. I wasn't fully aware that these are needed for internal communication only. Looks like a good simplification. Especially removing the `pub` from `Events` (now `HandleInput`) makes sense then 👍  I would definitely like to hear what @romac / @brapse think about this, too.

light-client/src/supervisor.rs
liamsi: Makes sense. Need to check this out locally. I'm still wondering, when a primary doesn't yet have any trusted light block in store:
https://sourcegraph.com/github.com/informalsystems/tendermint-rs@b3a5b1e70cd2d7fb06361950d10951d51b3dce06/-/blob/light-client/src/supervisor.rs#L53-55

xla: If the primary always returns a trusted block, we should encode that in our types. While primary/witness might only be a operational concern as @milosevic pointed out during the call yesterday, there might a possibility to otherwise encode this expectation. For example with an enum distinguishing between node types/states. What would be the right way to find answers to these questions?

light-client/src/supervisor.rs
brapse: Yes they are internal only so a simplification here might makes sense.

xla: To give a bit more context, up until now this enum was a mix of events. Some of them used to send callbacked events from the `Handle` up to its `Supervisor`, while others where only used as intermediates inside of `Handle` methods. This led to the use of wildcard match arms in places where only a subset of events where relevant, with undefined or panic behaviour. Additionally it would mask the case where a new variant was added but not accounted for in vital places, e.g. the `Supervisor` run loop. By isolating the enum to the single purpose of `Handle -> Supervisor` comms, all these downsides are removed and we get some help from the compiler when extending the surface between the two of them.

light-client/src/supervisor.rs
liamsi: This might be orthogonal to this PR but I think, we'd need to enumerate in which cases the primary does not yet have a trusted light block and then figure out if we want to treat these cases as errors for some use-cases instead (maybe further up, e.g. in the rpc).

xla: Follow-up in #395

light-client/src/supervisor.rs
```diff
@@ -10,32 +15,21 @@ use crate::{
     types::{Height, LightBlock, PeerId, Status},
 };
 
-use tendermint::evidence::{ConflictingHeadersEvidence, Evidence};
-
-use contracts::pre;
-use crossbeam_channel as channel;
-
 /// Type alias for readability
 pub type VerificationResult = Result<LightBlock, Error>;
```
romac, line 19: Since we are using a fully expanded type in `HandleInput::LatestTrusted`, I think we can remove this type alias.

light-client/src/supervisor.rs
```diff
@@ -10,32 +15,21 @@ use crate::{
...
+/// Input events sent by the [`Handle`]s to the [`Supervisor`]. They carry a [`Callback`] which is
+/// used to communicate back the responses of the requests.
 #[derive(Debug)]
-pub enum Event {
-    // Inputs
+enum HandleInput {
     /// Terminate the supervisor process
     Terminate(Callback<()>),
     /// Verify to the highest height, call the provided callback with result
     VerifyToHighest(Callback<VerificationResult>),
```
romac, line 28:
```suggestion
    VerifyToHighest(Callback<Result<LightBlock, Error>>),
```

light-client/src/supervisor.rs
```diff
@@ -10,32 +15,21 @@ use crate::{
...
     /// Verify to the given height, call the provided callback with result
     VerifyToTarget(Height, Callback<VerificationResult>),
```
romac, line 30:
```suggestion
    VerifyToTarget(Height, Callback<Result<LightBlock, Error>>),
```

light-client/src/supervisor.rs
romac: Good catch! I wasn't so happy either with having both input and outputs/internal events in the same enum, but didn't see/look for the opportunity to simplify all this. 👍

light-client/src/supervisor.rs
romac: Left a comment in #395. Otherwise, I fully agree that we should try encode such invariants as possible in the types themselves.

light-client/src/supervisor.rs
xla: More thoughts on the matter in https://github.com/informalsystems/tendermint-rs/pull/185#pullrequestreview-439830876

liamsi:
What would it take to get this out of draft into ready for review? I think the changes look good and make sense 👍

xla:
> What would it take to get this out of draft into ready for review? I think the changes look good and make sense +1

I still would like to extend the API to full-fill all needs of the `/status` endpoint as described in #219. Also happy to split that up if we want smaller chunks for review. Latter might even be better, What ya all think?

liamsi:
> Also happy to split that up if we want smaller chunks for review. Latter might even be better, What ya all think?

I'd slightly prefer to merge this and get the remaining changes in a separate PR but either way is good to me.

light-client/src/supervisor.rs
```diff
@@ -142,6 +133,17 @@ impl Supervisor {
         }
     }
 
+    /// Create a new handle to this supervisor.
+    pub fn handle(&mut self) -> Handle {
+        Handle::new(self.sender.clone())
+    }
+
+    #[pre(self.peers.primary().is_some())]
```
romac, line 141: I think we should remove this precondition (and other similar ones on `verify_to_highest/target`) since we are actually raising a proper error if there is no primary, but this can be done in a follow-up (eg. by me).

romac: Tracking this in #400

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+## pending
+
+Light Client:
+
+- Expose latest_trusted from Supervisor Handle ([#394])
```
liamsi, line 5: Did we decide to have the light client changelog here or in the separate crate? 
@brapse 

Probably the changes related to the light client need a more general description and intro in the changelog but this is good for now 👌

xla: Just looked for precedence and there were entries for `Light Client` further down.

PR #397, opened 2020-06-30
Rework PeerList, improve its API, and fix a bug in `swap_primary`

romac:
- Removes the newly assigned `primary` from the `witnesses`.
- Ensures the primary is not part of any of the peer lists (`witnesses`, `full_nodes`, `faulty_nodes`).
- Fixes a bug in `swap_primary` (now `replace_faulty_primary`) where we would iterate the first element of `witnesses` over and over.
- Rework and generalize the PeerList for easier testing and safer API

    This commit changes the `PeerList` type to be parametrized by the type of values associated with a `PeerId`. This enables easier testing of the peer list without having to construct light client`Instance`s.

    This commit additionally introduces a safer API by taking into account the invariant associated with a `PeerList`, which lets us return unconditionally the value associated with the primary peer.

    In turn, this simplifies the verification loop in the supervisor, which is now expressed more simply as a recursive function whose correctness is more obvious than the imperative one, at least to me.

---

* [x] Wrote tests
* [ ] Updated CHANGES.md


romac:
Thanks @OStevan for prompting this little refactor and reviewing it.

OStevan:
@romac thanks for following up on this.

romac:
> Module level tests would help asserting the right behaviour.

Yep, working on those :) Thanks for the review!

romac:
Adding tests prompted me to refactor the PeerList definition and API in https://github.com/informalsystems/tendermint-rs/pull/397/commits/cbf9753de7fb2535eda0dbaf278fe342d752fdb0.

light-client/src/peer_list.rs
```diff
@@ -1,156 +1,206 @@
...
 #[derive(Debug)]
-pub struct PeerList {
-    instances: HashMap<PeerId, Instance>,
+pub struct PeerList<T> {
+    values: HashMap<PeerId, T>,
     primary: PeerId,
-    witnesses: HashSet<PeerId>,
-    full_nodes: HashSet<PeerId>,
-    faulty_nodes: HashSet<PeerId>,
+    witnesses: BTreeSet<PeerId>,
```
xla, line 19: Maybe I missed some context in a commit, but what's the motivation to move to a `BTreeSet`? Asking purely out of curioisity.

light-client/src/peer_list.rs
```diff
@@ -1,156 +1,206 @@
...
+    full_nodes: BTreeSet<PeerId>,
+    faulty_nodes: BTreeSet<PeerId>,
 }
 
-impl PeerList {
+impl<T> PeerList<T> {
```
xla, line 24: Is there any constraints we want to have on `T` that would help make the construction of the `PeerList` even more safe?

light-client/src/peer_list.rs
```diff
@@ -1,156 +1,206 @@
...
+/// A generic container mapping `PeerId`s to some type `T`,
+/// which keeps track of the primary peer, witnesses, full nodes,
+/// and faulty nodes. Provides lifecycle methods to swap the primary,
+/// mark witnesses as faulty, and maintains an `invariant` for
+/// correctness.
 #[derive(Debug)]
-pub struct PeerList {
-    instances: HashMap<PeerId, Instance>,
+pub struct PeerList<T> {
+    values: HashMap<PeerId, T>,
```
xla, line 17: Why the quite generic rename here?

light-client/src/peer_list.rs
```diff
@@ -1,156 +1,206 @@
...
-    instances: HashMap<PeerId, Instance>,
+pub struct PeerListBuilder<T> {
+    values: HashMap<PeerId, T>,
     primary: Option<PeerId>,
-    witnesses: HashSet<PeerId>,
-    full_nodes: HashSet<PeerId>,
-    faulty_nodes: HashSet<PeerId>,
+    witnesses: BTreeSet<PeerId>,
+    full_nodes: BTreeSet<PeerId>,
+    faulty_nodes: BTreeSet<PeerId>,
```
xla, line 158: It seems wasteful to see the exact same fields as in `PeerList`, why does the builder not hold on to a fresh `PeerList` and returning it on `build` instead of constructing it.

light-client/src/peer_list.rs
```diff
@@ -161,13 +211,110 @@ impl PeerListBuilder {
...
+        fn to_vec(&self) -> Vec<T>;
+    }
+
+    impl BTreeSetExt<PeerId> for BTreeSet<PeerId> {
+        fn to_vec(&self) -> Vec<PeerId> {
+            self.iter().copied().collect()
+        }
+    }
+
+    fn a() -> PeerId {
```
xla, line 239: You could name these according to their use case further down.

light-client/src/peer_list.rs
romac: This is to provide a stable ordering on the peer ids, to make the whole process (more) deterministic.

Good question by the way, I failed to mention that in the commit.

light-client/src/peer_list.rs
romac: Since the peer list can now hold to any datatype rather than just light client instances, I thought `values` was more appropriate than `instances`, but I am open to suggestions.

light-client/src/peer_list.rs
romac: I don't see any constraints which would make it safer, but perhaps I am not fully understanding/seeing what you mean by safer in this context?

light-client/src/peer_list.rs
romac: I am more or less following the internal representation used by the [`derive_builder`](https://docs.rs/derive_builder/0.9.0/derive_builder/) crate, and didn't give it much more thought. But I agree that it's a bit wasteful and that what you suggest would work as well. I don't feel strongly either way.

light-client/src/peer_list.rs
xla: I think what I'm getting at is partially what I said in https://github.com/informalsystems/tendermint-rs/pull/397#pullrequestreview-441029225 - given this is the `PeerList` of the light client crate, it might not be all that beneficial to generalise entirely over `T` and rather constraint it to a trait (capabilities) that indicate that whatever is held as a peer value can be used as an `Instance`, or whatever other abstraction has the right surface for other components to rely/integrate on. It is in the same direction as making the `Handle` a trait, I suspect these changes to make it easier to build up the object graph in tests and safer to construct. It's mostly food for thought and can be iterated on over time.

light-client/src/peer_list.rs
romac: @xla Thanks for the review! :) See my answers inline.

> This is a great refactoring and I agree it makes the core of the verification indeed easier to follow and comprehend. My main concern is the very constrained T on the PeerList. Could it not give stronger guarantees if Instance would be a trait and PeerList expects that as its type parameter?

I don't really see the value of having `Instance` defined as a trait and constraining `T` over that, since the behavior of `PeerList` does not depend in any way on the choice of `T`. My general approach to building data structures is to keep them as generic as possible, and only constrain on types if absolutely required by the definition of the struct, or on a per-method basis. Admittedly, this is a mindset I inherited from working in Haskell, but sees shared by the Rust community as well: https://github.com/rust-lang/rust-clippy/issues/1689.

Of course, if there is actual (type) safety to be gained by having a bound on `T` (that I am at the moment not seeing, but feel free to expand on that), I'd be happy to revisit that decision in that case.

romac:
@xla Thanks for the review! :) See my answers inline.

> This is a great refactoring and I agree it makes the core of the verification indeed easier to follow and comprehend. My main concern is the very constrained T on the PeerList. Could it not give stronger guarantees if Instance would be a trait and PeerList expects that as its type parameter?

I don't really see the value of having Instance defined as a trait and constraining T over that, since the behavior of PeerList does not depend in any way on the choice of T. My general approach to building data structures is to keep them as generic as possible, and only constrain on types if absolutely required by the definition of the struct, or on a per-method basis. Admittedly, this is a mindset I inherited from working in Haskell, but it seems to be mostly shared by the Rust community as well: rust-lang/rust-clippy#1689.

Of course, if there is actual (type) safety to be gained by having a bound on T (that I am at the moment not seeing, but feel free to expand on that), I'd be happy to revisit that decision in that case.

romac:
> As we not building general container datastructures but rather specific ones, i.e. the peer list of what the light client understands as a peer.

I agree, and that's why the peer list was initially specialized to `Instance`, but in this case having it being generic over the value associated with peers helped a lot to keep the tests concise as well, without eg. having to introduce an `Instance` trait.

I still wonder what you had in mind w.r.t. to safety, because I don't see any constraints on `T` that would improve on the current implementation, ie. `PeerList<Instance>` is functionally equivalent to `PeerList<T>` for any T. Would you mind expanding on what you had in mind, because I might very well be missing something?

light-client/src/peer_list.rs
```diff
@@ -164,13 +211,110 @@ impl PeerListBuilder {
...
+    }
+
+    #[test]
+    #[should_panic(expected = "Pre-condition of replace_faulty_witness violated")]
+    fn replace_faulty_witness_fails_if_not_witness() {
+        let mut peer_list = dummy_peer_list();
+        let _ = peer_list.replace_faulty_witness(d());
+        unreachable!();
+    }
+}
```
liamsi, line 320: Thanks for adding in the tests 👍

liamsi:
> I still wonder what you had in mind w.r.t. to safety,

My understanding is that this isn't necessarily about safety per se, only that a completely generic `T` might be too general where an actual constraining trait could make sense. On the other hand, no particular behaviour (or capabilities) seem to be required by `T`, so it can be that generic.

xla:
> My understanding is that this isn't necessarily about safety per se, only that a completely generic `T` might be too general where an actual constraining trait could make sense. On the other hand, no particular behaviour (or capabilities) seem to be required by `T`, so it can be that generic.

Spot on, sorry @romac - what I was going at is not so much safety but communication of intent.

Issue #398, opened 2020-06-30
light-client: ergonomic limitations of Callback abstraction


brapse:
Captured from the concluding conversation in #185 

The Callback abstraction used in the supervisor has some concerns:
https://github.com/informalsystems/tendermint-rs/blob/b004a3ad7d99c1f6b1b4aef8b4001133b7c0fe22/light-client/src/supervisor.rs#L343-L351

The Callback abstraction has some drawbacks, i.e. try semantics, scope and borrow complexities. As the general pattern seems to be that a bounded channel is constructed to transport the response it could be reworked to be the actual value send with the event to the Supervisor. Which in turn send back the response on the typed channel to the Handle. As an added side-effect this could remove the need for the Callback abstraction all together and fully embraces CSP style concurrency.

This issue would start with an ADR capturing this context and allow feedback of the suggested API changes.

PR #401, opened 2020-06-30
light-client: turn Handle into a trait

xla:
As we start to depend on the surface of the `Handle` we benefit from it being a trait that can be implemented on a per need basis. This will result in less overhead constructing object graphs in places where we want to assert behaviour of other types in remote places, i.e. the light-node RPC server. Overall we hope for an increased ease in writing tests on module level.

Ref #219
Ref #398

* [ ] ~~Wrote tests~~
* [x] Updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Light Client:
 
 - Expose latest_trusted from Supervisor Handle ([#394])
+- Turn `Handle` into a trait for ease of integration ([#401])
+
+[#394]: https://github.com/informalsystems/tendermint-rs/pull/394
+[#401]: https://github.com/informalsystems/tendermint-rs/pull/401
 
 ## [0.14.1] (2020-06-23)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 Light Client:
 
 - Expose latest_trusted from Supervisor Handle ([#394])
-- Turn `Handle` into a trait for ease of integration ([#401])
+- Turn `Handle` into a trait for ease of integration and testability ([#401])
 
 [#394]: https://github.com/informalsystems/tendermint-rs/pull/394
 [#401]: https://github.com/informalsystems/tendermint-rs/pull/401
--- a/light-client/src/supervisor.rs
+++ b/light-client/src/supervisor.rs
@@ -15,6 +15,39 @@ use crate::{
     types::{Height, LightBlock, PeerId, Status},
 };
 
+pub trait Handle {
+    /// Get latest trusted block from the [`Supervisor`].
+    fn latest_trusted(&mut self) -> Result<Option<LightBlock>, Error>;
+
+    /// Verify to the highest block.
+    fn verify_to_highest(&mut self) -> Result<LightBlock, Error>;
+
+    /// Verify to the block at the given height.
+    fn verify_to_target(&mut self, height: Height) -> Result<LightBlock, Error>;
+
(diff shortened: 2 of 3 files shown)


light-client/src/supervisor.rs
```diff
@@ -15,6 +15,39 @@ use crate::{
...
+    /// verification result.
+    fn verify_to_target_async(
+        &mut self,
+        height: Height,
+        callback: impl FnOnce(Result<LightBlock, Error>) -> () + Send + 'static,
+    );
+
+    /// Terminate the underlying [`Supervisor`].
+    fn terminate(&mut self);
+}
```
liamsi, line 49: Looks good to me 👍 My understanding of why making this a trait is cool is because it makes testability easier. Is there a (wip) example of such a test that would be annoying to write without this?

CHANGES.md
```diff
@@ -3,6 +3,10 @@
 Light Client:
 
 - Expose latest_trusted from Supervisor Handle ([#394])
+- Turn `Handle` into a trait for ease of integration ([#401])
```
liamsi, line 6: maybe? 
```suggestion
- Turn `Handle` into a trait for ease of integration and testability ([#401])
```

light-client/src/supervisor.rs
xla: The commit https://github.com/informalsystems/tendermint-rs/pull/363/commits/a452244ab2ab4802bb6dd4a85a3c87aec5e5f453 shows the start, where the test code implements a mockHandle. Generally it means a reduction of the objects that need constructing. If something was dependent on the `Handle` before, one had to provide a `PeerList`, `ForkDetector`, `EvidenceProvider` to construct the `Supervisor`. On all the state inside had to be correct in order to produce output as expected, which means the correct state in peer(s) stores, etc.

light-client/src/supervisor.rs
```diff
@@ -338,50 +388,23 @@ impl Handle {
...
         let event = HandleInput::VerifyToHighest(Callback::new(callback));
         self.sender.send(event).unwrap();
     }
 
-    /// Async version of `verify_to_target`.
-    ///
-    /// The given `callback` will be called asynchronously with the
-    /// verification result.
-    pub fn verify_to_target_async(
+    fn verify_to_target_async(
```
brapse, line 407: Should we continue to include the inline docs here?

light-client/src/supervisor.rs
```diff
@@ -390,8 +413,7 @@ impl Handle {
         self.sender.send(event).unwrap();
     }
 
-    /// Terminate the underlying supervisor.
-    pub fn terminate(&mut self) {
+    fn terminate(&mut self) {
```
brapse, line 416: Here as well

light-client/src/supervisor.rs
xla: The trait methods are documented for that reason. For doc generation purposes developers will always see the docs from the impled trait. Keeping this in two places, just lends itself to divergence and unnecessary overhead.

light-client/src/supervisor.rs
xla: See https://github.com/informalsystems/tendermint-rs/pull/401#discussion_r448160787

romac:
Looks good :)

light-client/src/supervisor.rs
liamsi: Thanks!

PR #402, opened 2020-06-30
[ADR] Light Client Supervisor Ergonomics

xla:
Follow-up to #398

* [ ] ~~Updated all code comments where relevant~~
* [ ] ~~Wrote tests~~
* [ ] ~~Updated CHANGES.md~~

### DIFF
--- a/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
+++ b/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
@@ -8,7 +8,7 @@
 
 The initial approach to use callbacks in order to capture async responses
 from requests issued by the `Handle` to its `Supervisor` introduces a couple of
-has some drawbacks, i.e. try semantics, scope and borrow complexities. As the
+drawbacks, i.e. try semantics, scope and borrow complexities. As the
 general pattern seems to be that a bounded channel is constructed to transport
 the response it could be reworked to be the actual value send with the event to
 the Supervisor. Which in turn send back the response on the typed channel to the
@@ -25,7 +25,7 @@ Remove the callback abstraction in favour of passing channels in the events
 exchanged directly as the pattern is present already in the current concrete
 implementation of the `supervisor::Handle` anyway.
 
-Remove async versions of the public methods until use-case drive their
+Remove async versions of the public methods until a use-case drives their
 implementation.
 
 ## Status
--- a/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
+++ b/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
@@ -16,8 +16,8 @@ Handle. As an added side-effect this could remove the need for the Callback
 abstraction all together and fully embraces CSP style concurrency.
 
 Furthermore we have async versions of most `Handle` methods, but no real
-use-case driving it. If we find such needs, we should look into an interface
-embracing `async/await`.
+use-case driving it. At this point we can't anticipate the limitations of the
+current design with regards to concurrent access.
 
 ## Decision
 
@@ -26,7 +26,9 @@ exchanged directly as the pattern is present already in the current concrete
 implementation of the `supervisor::Handle` anyway.
 
 Remove async versions of the public methods until a use-case drives their
-implementation.
+implementation. We will leave concurrency to be informed by emerging use-cases
+and field reports. With the first real integration likely happen in the RPC
+server and ibc relayer.
 
 ## Status
 
--- a/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
+++ b/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
@@ -48,6 +48,5 @@ Proposed
 
 ## References
 
-* [Feedback on preceeding
-  ADR](https://github.com/informalsystems/tendermint-rs/pull/185#pullrequestreview-439830876)
+* [Feedback on preceeding ADR](https://github.com/informalsystems/tendermint-rs/pull/185#pullrequestreview-439830876)
 * [Tracking issue](https://github.com/informalsystems/tendermint-rs/issues/398)
--- a/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
+++ b/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
@@ -1,0 +1,53 @@
+# ADR 007: Light Client Supervisor Ergonomics
+
+## Changelog
+
+* 2020-06-30: Initial draft
+
+## Context
+
+The initial approach to use callbacks in order to capture async responses
+from requests issued by the `Handle` to its `Supervisor` introduces a couple of
+has some drawbacks, i.e. try semantics, scope and borrow complexities. As the
+general pattern seems to be that a bounded channel is constructed to transport
+the response it could be reworked to be the actual value send with the event to
+the Supervisor. Which in turn send back the response on the typed channel to the
+Handle. As an added side-effect this could remove the need for the Callback
+abstraction all together and fully embraces CSP style concurrency.
+
+Furthermore we have async versions of most `Handle` methods, but no real
+use-case driving it. If we find such needs, we should look into an interface
+embracing `async/await`.
+
+## Decision
+
+Remove the callback abstraction in favour of passing channels in the events
+exchanged directly as the pattern is present already in the current concrete
+implementation of the `supervisor::Handle` anyway.
+
+Remove async versions of the public methods until use-case drive their
+implementation.
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+* Smaller public surface of the Handle abstraction
+* Removal of Callback type
+* Easier to reason code
+
+### Negative
+
+### Neutral
+
+* No explicit async version of Supervisor methods
+
+## References
+
+* [Feedback on preceeding
+  ADR](https://github.com/informalsystems/tendermint-rs/pull/185#pullrequestreview-439830876)
+* [Tracking issue](https://github.com/informalsystems/tendermint-rs/issues/398)


docs/architecture/adr-007-light-client-supervisor-ergonomics.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Changelog
+
+* 2020-06-30: Initial draft
+
+## Context
+
+The initial approach to use callbacks in order to capture async responses
+from requests issued by the `Handle` to its `Supervisor` introduces a couple of
+has some drawbacks, i.e. try semantics, scope and borrow complexities. As the
```
liamsi, line 11:
```suggestion
drawbacks, i.e. try semantics, scope and borrow complexities. As the
```

docs/architecture/adr-007-light-client-supervisor-ergonomics.md
```diff
@@ -0,0 +1,53 @@
...
+use-case driving it. If we find such needs, we should look into an interface
+embracing `async/await`.
+
+## Decision
+
+Remove the callback abstraction in favour of passing channels in the events
+exchanged directly as the pattern is present already in the current concrete
+implementation of the `supervisor::Handle` anyway.
+
+Remove async versions of the public methods until use-case drive their
```
liamsi, line 28:
```suggestion
Remove async versions of the public methods until a use-case drives their
```

docs/architecture/adr-007-light-client-supervisor-ergonomics.md
```diff
@@ -0,0 +1,53 @@
...
+## Context
+
+The initial approach to use callbacks in order to capture async responses
+from requests issued by the `Handle` to its `Supervisor` introduces a couple of
+has some drawbacks, i.e. try semantics, scope and borrow complexities. As the
+general pattern seems to be that a bounded channel is constructed to transport
+the response it could be reworked to be the actual value send with the event to
+the Supervisor. Which in turn send back the response on the typed channel to the
+Handle. As an added side-effect this could remove the need for the Callback
+abstraction all together and fully embraces CSP style concurrency.
```
liamsi, line 16: I can't think of any drawback removing the Callback abstractions. @brapse @romac We aren't missing any larger context here, right? (e.g. the callbacks weren't supposed to later carry some traceability function, like a logger etc? and either way, this can equivalently be modelled using the channel / csp abstraction).

docs/architecture/adr-007-light-client-supervisor-ergonomics.md
```diff
@@ -0,0 +1,53 @@
...
+The initial approach to use callbacks in order to capture async responses
+from requests issued by the `Handle` to its `Supervisor` introduces a couple of
+drawbacks, i.e. try semantics, scope and borrow complexities. As the
+general pattern seems to be that a bounded channel is constructed to transport
+the response it could be reworked to be the actual value send with the event to
+the Supervisor. Which in turn send back the response on the typed channel to the
+Handle. As an added side-effect this could remove the need for the Callback
+abstraction all together and fully embraces CSP style concurrency.
+
+Furthermore we have async versions of most `Handle` methods, but no real
```
brapse, line 18: As mentioned in #403, The purpose of the async interface would be to facilitate the relayer dispatching verification from an event loop to allow events from queries/subscripts to accumulate concurrently. The result of the callback would then be written to the relayer event loop when verification results are ready.

Maybe this could also be done using async/await (as @romac  has suggested) however my understanding is that await calls within the relayer would block the relayer loop and defeat the concurrency that we are looking for. This discussion is probably outside the scope of this ADR but are there similarly simple (channels) methods of achieving the concurrency we want (accumulate events while waiting for verification) without implementing a scheduler?

xla: Interesting, I'd be curious to get more context on the claim of the blocking of the relayer loop.

Generally this sounds like the relayer will have very specific requirements for its concurrency, which in turn means driving the design of the interface of the `Handle` with it will likely result in brittle coupling. What should give us ease of mind is that that by design the `Handle` can be safely shared between threads, so whatever the relayer needs w.r.t. scheduling can be done on top of the exisiting synchronous methods. This will keep the surface small and guarantees flexibility on the caller side. We can codify and enforce that with marker traits (`Send` + `Sync`) and the already in use [static_assertions](https://docs.rs/static_assertions/1.1.0/static_assertions/).

xla:
@brapse It's the right place to discuss these concerns, see https://github.com/informalsystems/tendermint-rs/pull/402#discussion_r448166015 for the follow-up.

docs/architecture/adr-007-light-client-supervisor-ergonomics.md
brapse: I think in my head, sequences of `await` calls are serialized even if there is no dependence:
```rust
loop {
  foo.compute().await;
  baz.compute().await;
}
```

Is this not the case? 

What we want is loop over a queue of unordered responses:

(excuse the pseudo-ish code here)
```rust
let sender, receiver = ...
while let Some(event) = receiver.recv() {
  foo.compute(sender.clone());
  baz.compute(sender.clone());
}
```

There could be a better way here, but this is the requirement we anticipate for the relayer.

brapse: The above is related to the current concurrency architecture proposed in https://github.com/informalsystems/ibc-rs/issues/121

image

xla: Thanks for the detailed explanations and extended context for the needs of downstream consumers. As I suspected the concurrency model of the relayer is quite specific and in essence would necessitate somethig like a [`select`](https://docs.rs/crossbeam-channel/0.4.2/crossbeam_channel/macro.select.html#examples) to provide the response to whatever computation has been completed. And again one can build that easily in the context of the ibc/relayer codebase with the synchronous API methods of the `Handle`. Esp. as you want to control the construction and type of the channels used to select over.

> I think in my head, sequences of `await` calls are serialized

That is correct, such is the nature of `await`.

brapse: >  And again one can build that easily in the context of the ibc/relayer codebase with the synchronous API methods of the Handle

I believe the concurrency model we are looking for is one in which calls to the `chain light client` (and all other arrows) would specifically not block the `event handler`.

xla: I understnad the original design intend. My point is that the `_async` variants only partially alleviate the problem and introduce coupling with downstream consumers for not much gain. The non-blocking nature in the event loop of the relayer can be easily achieved in the context of the ibc repository on top of the synchronous methods, while leaving fine-grained control over the concurrency model to the implementation of the relayer. As an exmaple:

``` rust
let (first_sender, first_receiver) = channel::bounded::<_>(1);
let (second_sender, second_receiver) = channel::bounded::<_>(1);

thread::spawn(move || {
    let res = blocking_computation();
    first_sender.send(res);
});

thread::spawn(move || {
    let res = another_blocking_computation();
    second_sender.send(res);
});

loop {
    select!{
        recv(first_receiver) -> res => todo!(),
        recv(first_receiver) -> res => todo!(),
    }
}
```

brapse: Makes sense. However this doesn't appear to embrace `async/await` or provide any clarity around the suggestion in the paragraph. Perhaps you could add a more illustrative example to the ADR or remove that particular comment if it's not critical to the major thrust of the argument.

liamsi: Apparently, you can also `select` over async functions / futures: https://rust-lang.github.io/async-book/06_multiple_futures/03_select.html

docs/architecture/adr-007-light-client-supervisor-ergonomics.md
romac: > We aren't missing any larger context here, right? (e.g. the callbacks weren't supposed to later carry some traceability function, like a logger etc?

No, the callbacks were just meant as a thin abstraction over the underlying channels, that's it.

> and either way, this can equivalently be modelled using the channel / csp abstraction).

Yep, I concur in https://github.com/informalsystems/tendermint-rs/pull/403#discussion_r448474189

brapse:
> Sounds good to me, though I wonder if we shouldn't go a bit further and actually return the underlying receiver rather than block on it within each Handle's method?

I think this is a really cool idea. I'm not sure how much of this we need to solve here I just want to highlight @xla's insight within the context of @romac suggestion in hopes that we can integrate them into this ADR in short order.

It is the user of the handle's job to manage concurrency concerns as they have the best context to get it right for their use case. The handle can expose a simple synchronous interface and it's the job of the user (and owner) of that handle to integrate it into its own runtime. I personally like channels and @romac suggestion but what if the user is using built in mpsc instead of crossbeam? There may be advantages of having minimal assumptions  encoded in the interface that we expose via the handle.

xla:
> It is the user of the handle's job to manage concurrency concerns as they have the best context to get it right for their use case. The handle can expose a simple synchronous interface and it's the job of the user (and owner) of that handle to integrate it into its own runtime. I personally like channels and @romac suggestion but what if the user is using built in mpsc instead of crossbeam? There may be advantages of having minimal assumptions encoded in the interface that we expose via the handle.

That's a better summary of my thoughts than I managed to put together. In essence when exposing channels (CSP style concurrency is one of the best tools we have at the moment) we not only leak the underlying concurrency model of the implementation - of the `Supervisor` and its `Hande` - but also the channel implementation in use (crossbeam). As @brapse said what if the caller is using a different channel implementation, or none at all. Any synchronous API will allow for the user to build their concurrency on top of it. It gets a bit more invovled with regards to async runtimes, but let's leave that aside.

My suggestion going forward is to remove the mention of `async/await` and make it explicit in the ADR that we don't know yet what the limitations of the current deisgn are w.r.t. users needs for concurrent access and leave that to be informed by emerging use-cases and field reports. Where the first real world test will be the integration in the relayer. How does that sound @brapse @romac ?

PR #403, opened 2020-06-30
light-client: improve supervisor ergonomics

### LINKED ISSUE #398: light-client: ergonomic limitations of Callback abstraction

xla:
* replace callbacks with channels
* remove `Callback` abstraction
* relax mutability on `Handle`
* provide **noop** default implementations for `Handle` trait methods
* handle crossbeam errors honestly
* update ADR status
* add changes entry

Closes #398
Implements [ADR-007](https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-007-light-client-supervisor-ergonomics.md)
Depends on #401

***

* [ ] ~~Wrote tests~~
* [x] Updated CHANGES.md


light-client/src/supervisor.rs
```diff
@@ -25,25 +22,6 @@ pub trait Handle {
     /// Verify to the block at the given height.
     fn verify_to_target(&mut self, height: Height) -> Result<LightBlock, Error>;
 
-    /// Async version of `verify_to_highest`.
-    ///
-    /// The given `callback` will be called asynchronously with the
-    /// verification result.
-    fn verify_to_highest_async(
```
brapse, line 32: Just for context here. The intention of developing an async alternative here was to provide asynchronous verification to the relayer such that it could accumulate events from subscriptions/queries while waiting for verification. We can easily add this back when the requirement materialises in the relayer implementation.

romac: Perhaps we could make the API exposed by the Handle a bit more generic by returning the underlying receiver instead of the result of blocking on it. This way, consumers of the API can decide whether they:

a) want the result right away and call `.recv()/.recv_timeout()` on it themselves
b) store the channel away for use at a later time
c) pipe its result into another unbounded channel that will be iterated over at a later time

romac: This would allow for both sync and async needs, and give the consumers of the API a bit more control over when they want to access the result of the computation.

romac: Moved this discussion in https://github.com/informalsystems/tendermint-rs/pull/402#pullrequestreview-441022995, where I should have posted first thing anyway.

light-client/src/errors.rs
```diff
@@ -57,6 +60,9 @@ pub enum ErrorKind {
 
     #[error("invalid light block: {0}")]
     InvalidLightBlock(#[source] VerificationError),
+
+    #[error("internal channel disconnected")]
+    ChannelDisconnected,
```
liamsi, line 65: Wouldn't it be nice to "decorate" the error with the underlying crossbeam error message?

light-client/src/supervisor.rs
```diff
@@ -2,63 +2,49 @@ use crossbeam_channel as channel;
...
+use crate::light_client::LightClient;
+use crate::peer_list::PeerList;
+use crate::state::State;
+use crate::types::{Height, LightBlock, PeerId, Status};
 
 pub trait Handle {
     /// Get latest trusted block from the [`Supervisor`].
-    fn latest_trusted(&mut self) -> Result<Option<LightBlock>, Error>;
+    fn latest_trusted(&self) -> Result<Option<LightBlock>, Error> {
+        todo!()
```
liamsi, line 17: Will this contain the default impl? The `todo!()` doesn't really communicate what is supposed to happen here.

Same with the other 3 functions below.

CHANGES.md
```diff
@@ -1,12 +1,14 @@
-## pending
+# pending
 
 Light Client:
 
 - Expose latest_trusted from Supervisor Handle ([#394])
 - Turn `Handle` into a trait for ease of integration and testability ([#401])
+- Improve `Supervisor` ergonomics ([#403])
```
liamsi, line 7: Not sure what others think about this but if there is a realted ADR, it be cool to link it in the changelog, too.

light-client/examples/light_client.rs
```diff
@@ -160,19 +160,20 @@ fn sync_cmd(opts: SyncOpts) {
...
+        let maybe_block = handle.verify_to_highest();
+        match maybe_block {
             Ok(light_block) => {
-                println!("[ info  ] synced to block {}", light_block.height());
+                println!("[info] synced to block {}", light_block.height());
             }
-            Err(e) => {
-                println!("[ error ] sync failed: {}", e);
+            Err(err) => {
+                println!("[error] sync failed: {}", err);
```
liamsi, line 174: Out of curiosity: what triggered removing the whitespaces here?

light-client/src/errors.rs
```diff
@@ -107,3 +113,15 @@ impl ErrorExt for ErrorKind {
         }
     }
 }
+
+impl<T: Debug + Send + Sync + 'static> From<crossbeam::SendError<T>> for ErrorKind {
+    fn from(_err: crossbeam::SendError<T>) -> Self {
+        Self::ChannelDisconnected
```
liamsi, line 119: Why not additionally wrap the underlying error msg? (see above)

light-client/src/supervisor.rs
```diff
@@ -314,26 +301,26 @@ impl Supervisor {
...
+                HandleInput::VerifyToTarget(height, sender) => {
                     let outcome = self.verify_to_target(height);
-                    callback.call(outcome);
+                    sender.send(outcome).map_err(ErrorKind::from)?;
                 }
-                HandleInput::VerifyToHighest(callback) => {
+                HandleInput::VerifyToHighest(sender) => {
                     let outcome = self.verify_to_highest();
-                    callback.call(outcome);
+                    sender.send(outcome).map_err(ErrorKind::from)?;
```
liamsi, line 323: If you find this too explicit, you can also write this and other places as:
```suggestion
                    sender.send(outcome).into()?;
```

light-client/examples/light_client.rs
xla: Looked like a mistake, what would be the reason for it?

light-client/src/supervisor.rs
xla: The intention is that impls can overwrite the methods they depend on, but the default implementation should never be used without being overwritten. This way mocks and partial implementations are leaner, but can't be miss-used.

light-client/src/errors.rs
xla: The reason we jump through these hoops is that the crossbeam errors don't derive `Deserialize` and `Serialize` which we need for our errors. Furthermore they only carry the message that failed to send, Would question how much value that has.

light-client/src/errors.rs
xla: https://github.com/informalsystems/tendermint-rs/pull/403#discussion_r449255706

light-client/examples/light_client.rs
liamsi: As long as this consistent throughout the codebase, I don't really care. @romac would know if there was any particular reason. I don't know and can't think of any reason.

light-client/src/supervisor.rs
liamsi: Actually, nvm, doesn't compile

light-client/src/supervisor.rs
```diff
@@ -314,26 +301,26 @@ impl Supervisor {
     /// Run the supervisor event loop in the same thread.
     ///
     /// This method should typically be called within a new thread with `std::thread::spawn`.
-    pub fn run(mut self) {
+    pub fn run(mut self) -> Result<(), Error> {
         loop {
-            let event = self.receiver.recv().unwrap();
+            let event = self.receiver.recv().map_err(ErrorKind::from)?;
```
brapse, line 306: nice improvement 😍

light-client/src/errors.rs
liamsi: Oh, I thought you can wrap errors in the error kinds which do not derive (de)serialize and still have serialization via anomaly but I might be wrong (or too tired). Doesn't sound like it's worth to further investigate here.

PR #404, opened 2020-07-01
Minor comment updates and var renames

ebuchman:
Update a few comments.

Rename `light_block -> verified_block` and `trusted_state -> trusted_block` to clarify the distinction during fork detection.

Also questions about:

- (re-) marking the trusted block as verified 
- reporting evidence to all peers

### DIFF
--- a/light-client/src/fork_detector.rs
+++ b/light-client/src/fork_detector.rs
@@ -34,7 +34,7 @@ pub enum Fork {
 
 /// Interface for a fork detector
 pub trait ForkDetector: Send {
-    /// Detect forks using the given the verified block, trusted block,
+    /// Detect forks using the given verified block, trusted block,
     /// and list of witnesses to verify the given light block against.
     fn detect_forks(
         &self,
--- a/light-client/src/fork_detector.rs
+++ b/light-client/src/fork_detector.rs
@@ -34,12 +34,12 @@ pub enum Fork {
 
 /// Interface for a fork detector
 pub trait ForkDetector: Send {
-    /// Detect forks using the given light block, trusted state,
+    /// Detect forks using the given the verified block, trusted block,
     /// and list of witnesses to verify the given light block against.
     fn detect_forks(
         &self,
         verified_block: &LightBlock,
-        trusted_state: &LightBlock,
+        trusted_block: &LightBlock,
         witnesses: Vec<&Instance>,
     ) -> Result<ForkDetection, Error>;
 }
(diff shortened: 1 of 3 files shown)


light-client/src/fork_detector.rs
```diff
@@ -34,12 +34,12 @@ pub enum Fork {
 
 /// Interface for a fork detector
 pub trait ForkDetector: Send {
-    /// Detect forks using the given light block, trusted state,
+    /// Detect forks using the given the verified block, trusted block,
```
liamsi, line 37: Why the extra "the"?
```suggestion
    /// Detect forks using the given verified block, trusted block,
```

light-client/src/fork_detector.rs
```diff
@@ -101,22 +103,24 @@ impl ForkDetector for ProdForkDetector {
                 continue;
             }
 
-            state.light_store.update(&trusted_state, Status::Verified);
+            // XXX: Shouldn't this have already happened? Why do we need
+            // to re-mark the trusted block as verified?
```
liamsi, line 107: cc @romac

brapse: Yes, I believe that to be the case:
https://github.com/informalsystems/tendermint-rs/blob/d498fe2d5518a8887f501006a6b477f24621b5cc/light-client/src/light_client.rs#L203-L206

Fork detection on this particular light block will only occur on verdict successful so it should be in the store marked as verified.

This has come up before by I just drop it again here as it seems relevant. Sharing mutable access to the light store between components feels a bit cumbersome. It might be better in future iterations have a single owner (probably the supervisor) to ensure consistent mutation.

melekes: lol. I've asked the same question a few weeks ago. https://github.com/informalsystems/tendermint-rs/pull/287#discussion_r438590326

romac: The ‘state’ here is a fresh in-memory state, not the primar’s or any of the witnesses’ own states, so we need to insert the trusted block to have a trusted state to work from for verification. I guess the mistake I made here is to call ‘update’ instead of ‘insert’, which in this case are equivalent but does not communicate the proper intent. Will fix!

romac: Fixed in https://github.com/informalsystems/tendermint-rs/pull/404/commits/b45851877f695da5e162713b5ecc50d2d5feef55

romac:
Just merged `master` in

PR #407, opened 2020-07-02
Fetch latest trusted or verified block from the store as latest trusted state

romac:
See https://github.com/informalsystems/tendermint-rs/pull/375/files#r448951616 for context.

This PR also renames `Store::highest` to `Store::latest` for consistency with the spec.

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/store.rs
+++ b/light-client/src/store.rs
@@ -31,13 +31,13 @@ pub trait LightStore: std::fmt::Debug + Send {
     /// Remove the light block with the given height and status, if any.
     fn remove(&mut self, height: Height, status: Status);
 
-    /// Get the latest light block with the given status.
+    /// Get the light block of greatest height with the given status.
     fn latest(&self, status: Status) -> Option<LightBlock>;
 
     /// Get an iterator of all light blocks with the given status.
     fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock>>;
 
-    /// Get the latest trusted or verified block from the store.
+    /// Get the light block of greatest height with the trusted or verified status.
     fn latest_trusted_or_verified(&self) -> Option<LightBlock> {
         let latest_trusted = self.latest(Status::Trusted);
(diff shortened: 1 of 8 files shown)


romac:
I am not a big fan of this new `latest_trusted_or_verified` method, but it keeps the light client code short and clean. Ideally we'd like a way of combining `Status` values using boolean algebra to express AND/OR queries, but that seems a bit overkill at the moment.

OStevan:
@romac thanks for the quick answer and follow up.

romac:
I just updated the conformance test to use the `Trusted` status for the initial trusted block, which should prevent regressions (eg. we only ask for a "verified" block where we should ask for "verified or trusted").

light-client/src/store.rs
```diff
@@ -20,15 +20,38 @@ pub mod sled;
...
     /// Overrides any other block with the same height and status.
     fn insert(&mut self, light_block: LightBlock, status: Status);
+
     /// Remove the light block with the given height and status, if any.
     fn remove(&mut self, height: Height, status: Status);
-    /// Get the highest light block with the given status.
-    fn highest(&self, status: Status) -> Option<LightBlock>;
+
+    /// Get the latest light block with the given status.
+    fn latest(&self, status: Status) -> Option<LightBlock>;
```
liamsi, line 35: Just to remove any remaining ambiguity: previously this was called "highest" and it was pretty clear that this refers to the height. Latest is a bit unclear: latest added? Latest as in newest regarding block time, or, height? I think the doc comment could be explicit (the name is pr. fine though).

romac: Good point! As in the spec, it means latest as in newest regarding height, will clarify in doc comment.

romac: See https://github.com/informalsystems/tendermint-rs/pull/407/commits/dc63808b6f7b89ee37b7d6b837623757ea9ebda9

PR #410, opened 2020-07-06
close #409: sort validators by address on deserialization

### LINKED ISSUE #409: Semantics of validator::Set
I have a question wrt. to the semantics/implementation of [validator::Set](https://github.com/informalsystems/tendermint-rs/blob/547379daf73839883037b65f97d3ceb373f499b2/tendermint/src/validator.rs#L17), and in particular [validator::Set::new()](https://github.com/informalsystems/tendermint-rs/blob/547379daf73839883037b65f97d3ceb373f499b2/tendermint/src/validator.rs#L25). Actually, for me this feels more like a bug, but I need some clarification. The reason I am asking is because while writing my generator for Tendermint types I've stumbled into a behavior that was unexpected, and I've spent quite some time debugging the issue...

The question: **is it a well-established convention, that validators in a validator set should be sorted by their address?** This is how the implementation of `new()` works. My problem was that the validators in a JSON file were listed in a different order, therefore the hash computed for them, when `validator::Set` is deserialized using [parse_vals()](https://github.com/informalsystems/tendermint-rs/blob/547379daf73839883037b65f97d3ceb373f499b2/tendermint/src/validator.rs#L48) function, is different, than the hash computed from the same set, but constructed using [validator::Set::new()](https://github.com/informalsystems/tendermint-rs/blob/547379daf73839883037b65f97d3ceb373f499b2/tendermint/src/validator.rs#L25). Both ways should produce the same result, right? So, either of the two variants should be true:
1. it is a well established convention, and then all serialized validators should be in the right order. Then it should be explicitly documented somewhere.
2. it is not a convention, and then it's a bug in the deserialization function, which should actually also sort what it receives. 

Also, [validator::Set::new()](https://github.com/informalsystems/tendermint-rs/blob/547379daf73839883037b65f97d3ceb373f499b2/tendermint/src/validator.rs#L25) is modifying its argument, which somehow doesn't feel right for me -- a constructor should leave its arguments untouched, imho...

Will be grateful for clarifications.

andrey-kuprianov:
As discussed in #409, in order to maintain the type invariant of `validator::Set`, the validators are now sorted correctly also on deserializing the structure.

*NB! when migrating from v. 0.33 to 0.34, the sorting order in `sort_validators` function needs to be changed.*

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/tendermint/src/validator.rs
+++ b/tendermint/src/validator.rs
@@ -55,7 +55,8 @@ fn parse_vals<'de, D>(d: D) -> Result<Vec<Info>, D::Error>
 where
     D: Deserializer<'de>,
 {
-    let mut vals: Vec<Info> = Deserialize::deserialize(d).map(|x: Option<_>| x.unwrap_or_default())?;
+    let mut vals: Vec<Info>
+        = Deserialize::deserialize(d).map(|x: Option<_>| x.unwrap_or_default())?;
     Set::sort_validators(&mut vals);
     Ok(vals)
 }
--- a/tendermint/src/validator.rs
+++ b/tendermint/src/validator.rs
@@ -35,7 +35,7 @@ impl Set {
     /// Sort the validators according to the current Tendermint requirements
     /// (v. 0.33 -> by validator address, ascending)
     fn sort_validators(vals: &mut Vec<Info>) {
-        vals.sort_by(|v1, v2| v1.address.partial_cmp(&v2.address).unwrap());
+        vals.sort_by_key(|v| v.address);
     }
 }
 
--- a/tendermint/src/validator.rs
+++ b/tendermint/src/validator.rs
@@ -55,8 +55,8 @@ fn parse_vals<'de, D>(d: D) -> Result<Vec<Info>, D::Error>
 where
     D: Deserializer<'de>,
 {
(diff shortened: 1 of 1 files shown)


tendermint/src/validator.rs
```diff
@@ -23,14 +23,20 @@ impl Set {
...
 
     /// Get Info of the underlying validators.
     pub fn validators(&self) -> &Vec<Info> {
         &self.validators
     }
+
+    /// Sort the validators according to the current Tendermint requirements
+    /// (v. 0.33 -> by validator address, ascending)
+    fn sort_validators(vals: &mut Vec<Info>) {
+        vals.sort_by(|v1, v2| v1.address.partial_cmp(&v2.address).unwrap());
```
romac, line 38:
```suggestion
        vals.sort_by_key(|v| v.address);
```

More idiomatic since `account::Id` implements `Ord` and improve type-safety by avoiding the `unwrap` call (which always succeeds because of the `Ord` bound, but could silently start failing if the bound was somehow removed).

andrey-kuprianov: nice:) thanks for pointing out!

PR #412, opened 2020-07-07
light-client constraint Handle to Clone + Send + Sync

yihuang:
Since the `SupervisorHandle` is a channel sender, it indeed support `Send + Sync + Clone`, but it can't be used that way if we only return `impl Handle`.

### DIFF
--- a/light-client/src/supervisor.rs
+++ b/light-client/src/supervisor.rs
@@ -153,7 +153,7 @@ impl Supervisor {
     }
 
     /// Create a new handle to this supervisor.
-    pub fn handle(&mut self) -> impl Handle {
+    pub fn handle(&mut self) -> SupervisorHandle {
         SupervisorHandle::new(self.sender.clone())
     }
 
@@ -329,7 +329,8 @@ impl Supervisor {
 
 /// A [`Handle`] to the [`Supervisor`] which allows to communicate with
 /// the supervisor across thread boundaries via message passing.
-struct SupervisorHandle {
+#[derive(Clone)]
+pub struct SupervisorHandle {
     sender: channel::Sender<HandleInput>,
 }
 


yihuang:
> Good find, could you provide some context how this was an issue when using the `Handle. Unfortunately this change will leak a private type. In it's current design the caller shouldn't care about the actual implementation returned, hence the return `impl`. It should get the same results by constraining `Handle`:
> 
> ```rust
> pub trait Handle: Clone + Senc + Sync {
> ```
> 
> and then derive `Clone` on `SupervisorHandle`.

In the chain client, it'll have a single shared light client supervisor, and different parts of apps running in different threads might want to call it through the handle.
Make `Clone + Senc + Sync` the dependency of `Handle` also works in my scenario.
It seems `SupervisorHandle` is the only implementation of that trait right now? what's the difference between returning a trait and returning a concrete type with private members.

xla:
> In the chain client, it might have a single shared light client supervisor, and different parts of apps running in different threads might want to call it through the handle, so it's good to support these traits. Make `Clone + Senc + Sync` the dependency of `Handle` also works in my scenario I think.

Great!

> It seems `SupervisorHandle` is the only implementation of that trait right now? what's the difference between returning a trait and returning a concrete type with private members?

Sorry, I was a bit quick with my response. What I meant was that in the current implementation (on master) `SupervisorHandle` is private, so that a user of the `Supervisor` doesn't depend on a specific implementation, rather on the functional surface defined by the `Handle` trait. In the ideal this leads to less coupling between components and more robust way going forward evolving the API of `Handle`. You are right there is only one implementation but as soon as #363 lands you can see how we use the trait in order for a dependent component to be tested in isolation, without the need to construct an entire `Supervisor` with its dependencies.

yihuang:
> > It seems `SupervisorHandle` is the only implementation of that trait right now? what's the difference between returning a trait and returning a concrete type with private members?
> 
> Sorry, I was a bit quick with my response. What I meant was that in the current implementation (on master) `SupervisorHandle` is private, so that a user of the `Supervisor` doesn't depend on a specific implementation, rather on the functional surface defined by the `Handle` trait. In the ideal this leads to less coupling between components and more robust way going forward evolving the API of `Handle`. You are right there is only one implementation but as soon as #363 lands you can see how we use the trait in order for a dependent component to be tested in isolation, without the need to construct an entire `Supervisor` with its dependencies.

I think it's better to make the decision at the caller side rather than callee, like:
```
impl Supervisor {
  fn handler(&self) -> SupervisorHandler;
}

struct User<H: Handle + Send + Sync> {
  handler: H,
}

let real_one = User {
  handler: supervisor.handler()
};

let mock_one = User {
  handler: some_mock_handler()
};
```
So we can still keep the `Handle` trait as simple as now, and user have the freedom to decide the combinations of traits?

xla:
> ```
> 
> ```
> 
> 
> So we can still keep the `Handle` trait as simple as now, and user have the freedom to decide the combinations of traits?
> 
> If user are running in single threaded environment, he might not care about Send/Sync.

Thinking more about it, you should be able to have a separate `Handle` per thread of execution in the chain client. And I agree the caller should constraint, but that is orthogonal to returning a concrete type.

light-client/src/supervisor.rs
```diff
@@ -329,7 +329,8 @@ impl Supervisor {
 
 /// A [`Handle`] to the [`Supervisor`] which allows to communicate with
 /// the supervisor across thread boundaries via message passing.
-struct SupervisorHandle {
+#[derive(Clone)]
+pub struct SupervisorHandle {
```
liamsi, line 333: I'm also wary to make this `pub`.

liamsi:
@yihuang do you mind sharing a link to the branch with your code so we can better understand what you are doing? (It doesn't matter if it is still work in progress etc)

yihuang:
@xla @liamsi
I changed the light client example to demonstrate it. https://github.com/yihuang/tendermint-rs/blob/test-light-client/light-client/examples/light_client.rs#L168
the gist of it is we want to share a supervisor process among multiple worker threads by clone and move the handle.

xla:
@yihuang Thanks for sharing, it's intended to be used a bit different. Instead of trying to clone the handle, you create as many handles as you need before you set up the supervisors event loop. Like this:

```rust
let handle = supervisor.handle();
    let handle1 = supervisor.handle();
    let handle2 = supervisor.handle();
    let handle3 = supervisor.handle();

    std::thread::spawn(|| supervisor.run());
```

This way you won't need any changes to the current implementation. Does that help?

yihuang:
> @yihuang Thanks for sharing, it's intended to be used a bit different. Instead of trying to clone the handle, you create as many handles as you need before you set up the supervisors event loop. Like this:
> 
> ```rust
> let handle = supervisor.handle();
>     let handle1 = supervisor.handle();
>     let handle2 = supervisor.handle();
>     let handle3 = supervisor.handle();
> 
>     std::thread::spawn(|| supervisor.run());
> ```
> 
> This way you won't need any changes to the current implementation. Does that help?

But other tasks might need to be spawn as needed, for example reacting to user request.

brapse:
> But other tasks might need to be spawn as needed, for example reacting to user request.

We anticipated tasks being long lived and re-usable between requests. What would be a compelling example of needing a different light client per user request?

yihuang:
> > But other tasks might need to be spawn as needed, for example reacting to user request.
> 
> Interesting, we anticipated tasks being long lived and re-usable between requests. What would be a compelling example of needing a different light client per user request?

I mean the handle is copied between requests, the light client supervisor is long lived the re-used. I guess below code capture our use case better:
```rust
pub struct App<L: Handle + Send + Sync> {
    pub handle: L,
}

impl<L: Handle + Send + Sync> App<L> {
    pub fn sync(&self) {
       self.handle.verify_to_highest();
    }
}

pub trait TestTrait: Send + Sync {
    fn test(&self);
}

impl<L: Handle> TestTrait for App<L> {
    fn test(&self) {}
}
```

> > But other tasks might need to be spawn as needed, for example reacting to user request.
> 
> Interesting, we anticipated tasks being long lived and re-usable between requests. What would be a compelling example of needing a different light client per user request?

We do share a single light client supervisor between requests, it's just we need to clone the `Handle`.
By the way, I find that a simple wrapper can solve my issue and working on current implementation:
```rust
pub struct LightClientHandle<L> {
    inner: Arc<L>,
}

impl<L> Clone for LightClientHandle<L> {
    fn clone(&self) -> Self {
        Self {
            inner: self.inner.clone(),
        }
    }
}

impl<L: Handle> Handle for LightClientHandle<L> {
    fn latest_trusted(
        &self,
    ) -> result::Result<Option<LightBlock>, tendermint_light_client::errors::Error> {
        self.inner.latest_trusted()
    }

    /// Verify to the highest block.
    fn verify_to_highest(
        &self,
    ) -> result::Result<LightBlock, tendermint_light_client::errors::Error> {
        self.inner.verify_to_highest()
    }

    /// Verify to the block at the given height.
    fn verify_to_target(
        &self,
        height: u64,
    ) -> result::Result<LightBlock, tendermint_light_client::errors::Error> {
        self.inner.verify_to_target(height)
    }

    /// Terminate the underlying [`Supervisor`].
    fn terminate(&self) -> result::Result<(), tendermint_light_client::errors::Error> {
        self.inner.terminate()
    }
}

pub fn spawn_light_client_supervisor(...) -> LightClientHandle<impl Handle + 'static> {}
```

This `LightClientHandle` will meet the requirement of `Handle + Send + Sync + Clone`.

brapse:
Maybe I completely misunderstand the requirement here, but if you need a handler per request context, would it be possible to initiate a middleware layer with a handle which clones it on each request by parameterizing the request handler with that specific instance? 🤔

yihuang:
> Maybe I completely misunderstand the requirement here, but if you need a handler per request context, would it be possible to initiate a middleware layer with a handle which clones it on each request by parameterizing the request handler with that specific instance? 🤔

I'm afraid I don't get what you mean :(

yihuang:
I guess I should close this for now since I have my solution for my original problem.

PR #419, opened 2020-07-08
light-client: treat `Trusted` status as a special case of `Verified` as per the spec

### LINKED ISSUE #418: Light client might downgrade already `Trusted` blocks to `Verified`
In the light client verification loop, we only look in the store for a block at the current height that is either `Verified` or `Unverified` (ie. we fetch it from a node), which may miss a block that has previously gone through fork detection and has been marked as `Trusted`. We then fetch it again from the peer, go through verification again, and then override its status to `Verified`, thus downgrading it from `Trusted` to `Verified`.

This is an issue because:

a) it's wasteful to re-fetch and re-verify a block from a node when we already trust it.
b) we may end up downgrading the status of a trusted block to verified, whether or not we fix a)

So I guess the question here is: what do we do if, during bisection, we encounter a block to verify that has i) already been verified, or ii) is already trusted?

Do we skip verification altogether and just move on to (schedule) the next block to verify?
Do we reverify the block anyway, and ensure we don't downgrade it's status if verification succeeds?
Something else?

---

- https://github.com/informalsystems/tendermint-rs/blob/142796b0cc37319f91af55cf2703b96a0cb80230/light-client/src/light_client.rs#L157
- https://github.com/informalsystems/tendermint-rs/blob/142796b0cc37319f91af55cf2703b96a0cb80230/light-client/src/light_client.rs#L246

### LINKED ISSUE #427: Light client needlessly re-fetch already `Trusted` blocks from its peer
Moved over from #418

In the light client verification loop, we only look in the store for a block at the current height that is either `Verified` or `Unverified` (ie. we fetch it from a node), which may miss a block that has previously gone through fork detection and has been marked as `Trusted`. We thus have to fetch it again from the peer, which is wasteful.

romac:
Close #427 

This PR also addresses a concern raised in #418. While has been brought to light that there is no actual case where a trusted block would be downgraded to verified, I believe it is a bit cleaner to assume it could happen and handle it properly.

---

* [x] Wrote tests
* [x] Updated CHANGES.md


light-client/src/light_client.rs
```diff
@@ -195,31 +194,34 @@ impl LightClient {
...
-            let current_block = self.get_or_fetch_block(current_height, state)?;
+            // Fetch the block at the current height from the light store if already present,
+            // or from the primary peer otherwise.
+            let (current_block, status) = self.get_or_fetch_block(current_height, state)?;
 
-            // Validate and verify the current block
-            let verdict = self
-                .verifier
-                .verify(&current_block, &trusted_state, &options);
+            if status != Status::Trusted {
```
OStevan, line 201: If I understood correctly given that we should always have just one verifier per peer I'm not sure if we are ever going to be in a situation where an intermediate header moves to the `Trusted` state concurrently. And if say that there might be another concurrent `verifier/supervisor` updating the state of the blocks, this if does not prevent the status of this specific light block to be downgraded. I think that this check might make more sense if it was combined with a `bail` as the current concurrency model does not really allow us to get into this state.

romac: It's not about concurrent access, more about accidentally stumbling upon a block we already trust (has status `Trusted`) during verification/bisection and then downgrading it to `Verified`.

See the updated issue at #418 for a hopefully clearer description of the problem.

light-client/src/store.rs
```diff
@@ -37,21 +38,35 @@ pub trait LightStore: std::fmt::Debug + Send {
     /// Get an iterator of all light blocks with the given status.
     fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock>>;
 
+    /// Get a block at a given hieht whatever its verification status
```
liamsi, line 41:
```suggestion
    /// Get a block at a given height whatever its verification status
```

light-client/src/types.rs
```diff
@@ -49,30 +49,33 @@ pub type SignedHeader = TMSignedHeader;
...
+        use Status::*;
+        static ALL: &[Status] = &[Unverified, Verified, Trusted, Failed];
         ALL
     }
+
+    pub fn most_trusted(a: Self, b: Self) -> Self {
+        std::cmp::max(a, b)
+    }
+
+    pub fn least_trusted(a: Self, b: Self) -> Self {
```
liamsi, line 76: Do we foresee an application for this method?

light-client/src/store.rs
```diff
@@ -37,21 +38,35 @@ pub trait LightStore: std::fmt::Debug + Send {
...
+    fn get_any(&self, height: Height) -> Option<(LightBlock, Status)> {
```
liamsi, line 42: it's not really "any" because that would contain `Failed` too?

romac: No, will remove it!

romac:
@josef-widder This PR also addresses a concern raised in #418. While has been brought to light that there is no actual case where a trusted block would be downgraded to verified, I believe it is a bit cleaner to assume it could happen and handle it properly. We can revisit this when we add support for verifying blocks below the latest trusted state. What do you think?

light-client/src/store.rs
romac: Oops, wrong thread.

Good point, will find a better name.

light-client/src/types.rs
romac: No, will remove it!

light-client/src/light_client.rs
```diff
@@ -230,35 +232,32 @@ impl LightClient {
...
-            .map(|current_block| {
-                state
-                    .light_store
-                    .insert(current_block.clone(), Status::Unverified);
+        let block = self
+            .io
+            .fetch_light_block(self.peer, AtHeight::At(height))
+            .map_err(ErrorKind::Io)?;
+
+        state.light_store.insert(block.clone(), Status::Unverified);
```
xla, line 259: It's strange to see mutations being done in a method that is called `get_or_fetch_block`. I know it has been here before but it indicates something being of about where we manage this concern. As now there is an implicit behavioural assumption that the fetched block will be in the light store going forward.

light-client/src/light_client.rs
```diff
@@ -230,35 +232,32 @@ impl LightClient {
...
 
-                current_block
-            })
-            .map_err(|e| ErrorKind::Io(e).into())
+        Ok((block, Status::Unverified))
```
xla, line 261: Encoding this information as tuples is surprising, esp. as we construct permutations of it on the spot. Can't we more strongly encode this in an enum, which for some variants carries the `LightBlock`?

light-client/src/utils.rs
```diff
@@ -0,0 +1,20 @@
+/// Stable version of `std::cmp::max_by_key`.
```
xla, line 1: See comment about this module above.

light-client/src/types.rs
```diff
@@ -116,3 +115,40 @@ impl LightBlock {
         self.signed_header.header.height.into()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
xla, line 120: :black_heart:

light-client/src/types.rs
```diff
@@ -49,30 +49,29 @@ pub type SignedHeader = TMSignedHeader;
...
     /// Return a slice of all the possible values for this enum.
-    pub fn iter() -> &'static [Status] {
-        static ALL: &[Status] = &[
-            Status::Unverified,
-            Status::Verified,
-            Status::Trusted,
-            Status::Failed,
-        ];
-
+    pub fn iter() -> &'static [Self] {
```
xla, line 66: Where are we in need of iterating over all statuses?

light-client/src/light_client.rs
xla: This is in line with our discussions about reasoning about the state of the light client and how we should thrive for stronger abstractions which make it hard to construct impossible states.

light-client/src/lib.rs
```diff
@@ -25,6 +25,7 @@ pub mod state;
 pub mod store;
 pub mod supervisor;
 pub mod types;
+mod utils;
```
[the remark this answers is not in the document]
romac: Agreed. Perhaps something like `std_ext` since those are basically either stabilized versions of std functions or additions to the std lib?

light-client/src/light_client.rs
romac: Agreed it's not ideal. Allhough we don't really rely on this assumption, it's more of a performance optimization (ie. caching of fetched blocks). Perhaps the method should be renamed, or even refactored into a helper which caches blocks in the store and the original `get_or_fetch_block`. This would make the light store mutation more explicit at call site. Will think about it.

light-client/src/light_client.rs
romac: > Encoding this information as tuples is surprising, esp. as we construct permutations of it on the spot.

To me a tuple is basically an anonymous struct, and I figured since the status of a block acts like a key, it would be okay to package both the block and its status as an enum, much like part of the API of Rust HashMap etc. where they return the value and its key as a tuple, rather than making up a struct whose only purpose is to be deconstructed at call site.

> Can't we more strongly encode this in an enum, which for some variants carries the LightBlock?

I don't quite see how an enum would fit here, could you expand on that?

light-client/src/types.rs
romac: https://github.com/informalsystems/tendermint-rs/blob/9631ead98bc55b0189462450aba7fbc9e2754833/light-client/src/store/sled.rs#L55

romac:
Thanks @xla for the valuable comments!

> In general the changes don't really reflect the premise, as Trusted and Verified are still distinct states. More inline.

Agreed, going further would require a better API for the Store, something I have somewhat expressed in https://github.com/informalsystems/tendermint-rs/issues/428.

light-client/src/lib.rs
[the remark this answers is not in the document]
xla: That sounds like a good idea!

light-client/src/light_client.rs
xla: Ideally behaviour like this (sounds like write-through caching) happens on a layer wrapping this component. Often referred to as decorators.This way each layer can be looked at in isolation and concerns are separated.

xla: I can give examples and/or go into more detail if it is of interest.

light-client/src/light_client.rs
xla: I think this goes in line with https://github.com/informalsystems/tendermint-rs/pull/419#issuecomment-656591438 and the general discussion around representing states in the crate.

Specifically here I'm wondering are we ever interested in a `LightBlock` without the status information? How would the rest of the crate look like if this would be our `LightBlock`:

```rust
enum LightBlock {
    Failed {
        provider: PeerId,
    },
    Unverified {
        provider: PeerId,
    },
    Verified {
        signed_header: SignedHeader,
        validators: ValidatorSet,
        next_validators: ValidatorSet,
        provider: PeerId,
    },
    Trusted {
        signed_header: SignedHeader,
        validators: ValidatorSet,
        next_validators: ValidatorSet,
        provider: PeerId,
    },
}
```

Generally what I think we see here is that we let the actual domain of block storage creep into our higher level abstractions, which is informed too much by the key/value store semantics and it's physicality.

light-client/src/light_client.rs
romac: > Ideally behaviour like this (sounds like write-through caching) happens on a layer wrapping this component. Often referred to as decorators.This way each layer can be looked at in isolation and concerns are separated.

Agreed, that's actually what I had in mind when I wrote "refactored into a helper which caches blocks in the store" but the name *decorator* didn't spring to my mind at the time :)

light-client/src/light_client.rs
romac: > Generally what I think we see here is that we let the actual domain of block storage creep into our higher level abstractions, which is informed too much by the key/value store semantics and it's physicality.

Agree 100%. We should think hard about this as part of https://github.com/informalsystems/tendermint-rs/issues/428.

xla: Absolutely, let's move the discussion there!

romac: I am not quite convinced that the enum above is the right way to express this, but I agree that we should look into this more closely.

light-client/src/types.rs
xla: Thanks for the pointer.

light-client/src/light_client.rs
xla: Can we capture this somewhere so it doesn't get lost?

romac: Captured in #434.

Issue #293, opened 2020-06-03
Light Node Initialization


ebuchman:
The light node already support config file and flags through Abscissa.

Configuration file for a light node should include:

- chain id
- list of full node urls (prod versions should require them to be https)
- trust period duration (eg. 1week)
- trust threshold (numerator and denominator, eg. 1/3)

When starting the light node for the first time, we need to provide subjective initialization in the form of a block height and hash, eg. `--height 962118 --hash 28B97BE9F6DE51AC69F70E0B7BFD7E5C9CD1A595B7DC31AFF27C50D4948020CD`. See the [`tendermint lite` command](https://github.com/tendermint/tendermint/blob/master/cmd/tendermint/commands/lite.go#L26).

If we restart the process, we shouldn't need to provide this info, as we should have a trusted state in our db by then. 

Of course if the trusted state has expired, the node should fail, and we should be required to re-initialize with a new hash and height.

Another possible situation for initialization is starting from a new blockchain, in which case we can potentially pass in a `--genesis` flag with a path to a tendermint genesis file to initialize the trusted state from. But we also need to add support for initializing from genesis (see #201).

We might want to separate initialization and normal operation into two distinct commands.

ebuchman:
This isn't just for the Light Node - any use of the Light Client will require some way to subjectively initialize it, so it should expose some initialization method that takes a height and hash and be able to start from there.

Note for off chain uses (eg. light node, relayer), we specify a height and hash and the LC fetches the header and validator set, checks everything matches, and trusts it and starts from there. For on chain uses (eg. IBC on chain), the header and validator set would be provided directly and just trusted (ie. as part of client creation on chain)

romac:
The light client (supervisor) will also need to be [provided](https://github.com/informalsystems/tendermint-rs/pull/302/files#diff-7798509be617f6e992ae4bccb801ebb3R112) with a [mapping](https://github.com/informalsystems/tendermint-rs/pull/302/files#diff-7798509be617f6e992ae4bccb801ebb3R55-R58) between peer ids and net addresses.

liamsi:
> We might want to separate initialization and normal operation into two distinct commands.

My current thinking is ideally there would be two commands:
 - `light-node initialize -h <trusted height> <trusted hash> [-f]` <- inits the light node subjectively (optionally force overwrite existing initialized store)
-  `light-node start [-c]` starts main loop (light client handle for `verify_to_highest` and the rpc server)

[...]

Issue #428, opened 2020-07-09
Improve LightStore API to remove need for ad-hoc methods


romac:
The current `LightStore` API is a bit underpowered, and forces us to implement ad-hoc method like `get_non_failed` or `latest_trusted_or_verified` to keep the downstream code clean.

It would be nice to have a slightly more powerful, perhaps query-based, API that would remove the need for such ad-hoc methods.

romac:
Something else to keep in mind and think hard about when designing the new API/abstraction:

> @xla: Generally what I think we see here is that we let the actual domain of block storage creep into our higher level abstractions, which is informed too much by the key/value store semantics and it's physicality. — https://github.com/informalsystems/tendermint-rs/pull/419#discussion_r452765452

xla:
Maybe a broader look at the LightStore would include https://github.com/informalsystems/tendermint-rs/pull/430#discussion_r452745330

romac:
We should also keep https://github.com/informalsystems/tendermint-rs/pull/449#issuecomment-659527984 in mind.

Issue #422, opened 2020-07-08
Streamline creation of a light client instance


romac:
Perhaps via the [builder pattern](https://www.ameyalokare.com/rust/2017/11/02/rust-builder-pattern.html).

ebuchman:
Closing for #484

PR #430, opened 2020-07-09
light-node: cli

liamsi:
Light node commands as described in https://github.com/informalsystems/tendermint-rs/issues/293#issuecomment-655466205

* [ ] Wrote tests
* [ ] Updated CHANGES.md


light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,92 @@
...
+        let mut peer_map = HashMap::new();
+        peer_map.insert(lc.peer_id, lc.address.clone());
+
+        let io = ProdIo::new(peer_map, Some(app_cfg.rpc_config.request_timeout));
+
+        initialize_subjectively(self.height, subjective_header_hash, &lc, &io);
+    }
+}
+
+fn initialize_subjectively(
```
xla, line 52: This function does too many things I believe, the database setup seems in the right place but the actual logic smells of library code.

xla: As it is something we potentially wanna test.

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,92 @@
...
+        std::process::exit(1);
+    }
+    // TODO(ismail): actually verify more predicates of light block before storing!?
+    let got_header_hash = trusted_state.signed_header.header.hash();
+    if got_header_hash != subjective_header_hash {
+        println!(
+            "[error] received LightBlock's header hash: {} does not match the subjective hash: {}",
+            got_header_hash, subjective_header_hash
+        );
+        std::process::exit(1);
```
xla, line 89: :crying_cat_face: Is there no way to return errors from these functions?

light-node/src/commands/start.rs
```diff
@@ -4,94 +4,106 @@
...
-use tendermint::Hash;
-
-use tendermint_rpc as rpc;
-
-use crate::application::APPLICATION;
-use crate::config::LightNodeConfig;
-use crate::prelude::*;
-use crate::requester::RPCRequester;
-use crate::store::{MemStore, State};
+use crate::application::{app_config, APPLICATION};
```
xla, line 8: Imports.

light-node/src/config.rs
```diff
@@ -4,55 +4,100 @@
...
+                request_timeout: Duration::from_secs(60),
+            },
+            light_clients: vec![LightClientConfig::default()],
+        }
+    }
 }
 
-impl Default for SubjectiveInit {
-    fn default() -> Self {
+impl From<LightNodeConfig> for light_client::Options {
```
xla, line 94: :black_heart:

light-node/src/commands.rs
```diff
@@ -1,17 +1,19 @@
...
 //!
 //! See the `impl Configurable` below for how to specify the path to the
 //! application's configuration file.
 
+mod initialize;
 mod start;
 mod version;
 
 use self::{start::StartCmd, version::VersionCmd};
+use crate::commands::initialize::InitCmd;
```
xla, line 16: See general imports comment.

light-node/src/commands/start.rs
```diff
@@ -4,94 +4,106 @@
...
     fn run(&self) {
         if let Err(err) = abscissa_tokio::run(&APPLICATION, async {
-            let config = app_config();
+            StartCmd::assert_init_was_run();
 
-            let client = rpc::Client::new(config.rpc_address.parse().unwrap());
-            let req = RPCRequester::new(client);
-            let mut store = MemStore::new();
+            // TODO(ismail): we need to verify the addr <-> peerId mappings somewhere!
+            let mut peer_map = HashMap::new();
```
xla, line 58: There seems to be overlap with some of the setup needed in the init command. We should extract the common utility.

light-node/src/commands/start.rs
```diff
@@ -108,53 +120,59 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
+        // Todo figure out if other options would be reasonable to overwrite via CLI arguments.
+        if let Some(addr) = self.listen_addr {
+            config.rpc_config.listen_addr = addr;
         }
-
         Ok(config)
     }
 }
+impl StartCmd {
+    fn make_instance(
```
xla, line 131: Mentioned it in https://github.com/informalsystems/tendermint-rs/pull/431#discussion_r452653735 - there are enough occurrences to make this a library method,

light-node/src/commands/start.rs
```diff
@@ -108,53 +120,59 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
+        let db = sled::open(primary_db_path).unwrap_or_else(|e| {
+            println!("[error] could not open database: {}", e);
+            std::process::exit(1);
+        });
 
-    store.add(trusted_state.to_owned())?;
+        let primary_store = SledStore::new(db);
 
-    Ok(())
+        if primary_store.latest(Status::Verified).is_none() {
```
xla, line 173: Why don't we assert that the latest is `Status::Trusted`?

light-node/src/commands/start.rs
```diff
@@ -4,94 +4,106 @@
...
-                )
-                .await
-                .unwrap();
-
-                println!("Succeeded bisecting!");
-
-                // notifications ?
-
-                // sleep for a few secs ?
+                std::thread::sleep(Duration::from_millis(800));
```
xla, line 106: Why do we have this random sleep here?

light-node/src/commands/start.rs
```diff
@@ -4,94 +4,106 @@
...
+    #[options(short = "c", long = "config", help = "path to light_node.toml")]
+    pub config: Option<PathBuf>,
 }
 
 impl Runnable for StartCmd {
     /// Start the application.
     fn run(&self) {
         if let Err(err) = abscissa_tokio::run(&APPLICATION, async {
-            let config = app_config();
+            StartCmd::assert_init_was_run();
```
xla, line 55: Generally I think the run method is a bit bloated, it would be nice to break out the nitty-gritty of the setup and just see a flow like:

* constructs supervisor
* constructs rpc
* spawns supervisor run loop
* spawns rpc
* runs verify continously

light-node/src/config.rs
```diff
@@ -4,55 +4,100 @@
 //! application's configuration file and/or command-line options
 //! for specifying it.
 
+use abscissa_core::path::PathBuf;
```
xla, line 7: See import comment.

light-node/light_node.toml.example
```diff
@@ -0,0 +1,28 @@
...
+nanos = 0
+
+[clock_drift]
+secs = 5
+nanos = 0
+
+[rpc_config]
+listen_addr = "127.0.0.1:8888"
+
+    [rpc_config.request_timeout]
```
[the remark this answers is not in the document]
liamsi: Yes, it's valid toml and I thought it better captures that it belongs under `rpc_config`, in rust the structure is:
```rust
struct RpcConfig {
    /// The address the RPC server will serve.
    pub listen_addr: SocketAddr,
    /// The duration after which any RPC request to tendermint node will time out.
    pub request_timeout: Duration,
}
```

light-node/README.md
```diff
@@ -1,14 +1,5 @@
 # LightNode
 
-Tendermint light client node.
+Tendermint light client node can be used as
```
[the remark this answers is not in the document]
liamsi: I didn't (really) start with updating the  readme yet. Will address the other points of the review and focus on the text here afterwards. Didn't realize that I committed that half-modified sentence 😆  
Thanks for the thorough review btw.

light-node/Cargo.toml
```diff
@@ -16,6 +16,7 @@ jsonrpc-http-server = "14.2"
 jsonrpc-derive = "14.2"
 serde = { version = "1", features = ["serde_derive"] }
 serde_json = "1.0"
+sled = "0.31.0"
```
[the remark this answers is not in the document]
liamsi: To use the ` tendermint_light_client::store::sled::SledStore` you need to pass in a `SledDb`. This could be abstracted away in that library s.t. users don't need the extra sled dependency. Wanted to keep changes as contained in the light-node crate as much as possible. Maybe a good candidate for a followup PR?

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,92 @@
+//! `intialize` subcommand
+
+use crate::application::app_config;
```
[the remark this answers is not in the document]
liamsi: I like that. Will change. Can flagging this be somehow enforced by clippy btw?

light-node/src/commands/initialize.rs
liamsi: I agree, a version of this should potentially live in the light-client crate (with tests)! 
Do you want to help out here? Otherwise I think it makes sense to capture this in an issue (actually #293 is related) before this gets merged and address it in a followup PR to not block this one.

light-node/src/commands/initialize.rs
liamsi: There is of course. If this was a library function it definitely would have to.  In this context, that caller of this private helper would `exit` on an error anyways.

light-node/src/commands/start.rs
liamsi: Absolutely. I was just eager to test against a running tendermint node as quickly as possible. Will refactor this before setting it to ready for review.

light-node/Cargo.toml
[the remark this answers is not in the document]
xla: I think it's fine, if we decide to make `sled` a public dependency of our API then the caller ought to construct the db. A potential change would be to hide away sled entirely, which I think is not the worst idea, and rework it from `SledDb` to `OnDisk` or similar.

light-node/light_node.toml.example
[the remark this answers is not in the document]
xla: Don't think it helps much by ways of hierarchy, the dot notation in the key is indicator enough. That said I don't feel strongly either way.

light-node/src/commands/initialize.rs
xla: Which is already an improvement as from my understanding `std::process::exit` unwinds the stack and doesn't allow for `Drop`s to kick in, a behaviour we should be very conscious about.

light-node/src/commands/initialize.rs
xla: Capturing these things in a follow-up sounds great.

light-node/src/commands/start.rs
xla: Dope!

light-node/src/commands/start.rs
liamsi: Hmm, that needs more thoughts: when catching up to the current height, sleeping doesn't make sense here. But as soon as the light client caught up, it does not really make sense to request the latest light block several times before a new block is created 🤔 

On the command line this will look like:
```
[info] synced to block 830
[info] synced to block 831
[info] synced to block 831
[info] synced to block 831
[info] synced to block 831
[info] synced to block 831
[info] synced to block 831
[info] synced to block 831
[info] synced to block 831
```

light-node/src/commands/start.rs
liamsi: Yes, makes sense. I thought as part of the integration tests work with @Shivani912 there now was such a method but it seems that a slightly modified version is copied where needed currently. Can we address this separately?

light-node/src/commands/start.rs
liamsi: Good question. I thought a lightblock is only considered trusted only if it is supported by witnesses. On init there is only one instance. Need to run this by @milosevic and check the spec.

light-node/src/commands/start.rs
xla: What's the behaviour we want? That it syncs on an interval? Or that it loops as quickly until a certain condition is met?

light-node/src/commands/start.rs
xla: Already captured in #422 thanks @romac for flagging.

light-node/src/commands/start.rs
liamsi: My understanding is: if lagging behind, catch up without delay as quickly as possible. If caught up, try sync every blocktime/2 or sth like that?

light-node/src/commands/initialize.rs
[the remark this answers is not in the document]
romac: Let's track this in https://github.com/informalsystems/tendermint-rs/issues/433.

light-node/src/commands/start.rs
xla: Would be nice to surface that behaviour explicitly. And in the synced state use a timer, for example from tokio, are we using pure threads in the light-node?

light-node/src/commands/initialize.rs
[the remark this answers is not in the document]
xla: Yes!

light-node/Cargo.toml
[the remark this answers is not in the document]
xla: Linked in #428

light-node/src/commands/start.rs
milosevic: The initial lightblock can be considered as a special case as it is considered trusted by the user. The potential problem I see is that there is an invariant that a trusted block is confirmed with all witnesses. What if initial light block turns out not to be equal to what witnesses consider lightblock for that height? Smells like potential source of bugs.

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    let mut light_store = SledStore::new(db);
+
+    if light_store.latest(Status::Verified).is_some() {
+        println!("[ warning ] overwriting trusted state in database");
+    }
+
+    let trusted_state = io
+        .fetch_light_block(l_conf.peer_id, AtHeight::At(height))
```
milosevic, line 70: Why we are doing IO outside light client (core)? Also all the checks that follow. I was expecting cli code to parse config file, do basic checks on the parameters obtained and then pass them to the light client init code. I am not sure even about db initialisation as we need to cover restarting scenario, and it feels more natural to let core library deals with this.

liamsi: What do you mean by light client (core)? Neither the [`Handle`](https://github.com/informalsystems/tendermint-rs/blob/b1252c639ece21aef5c6ac46f6e5b86187880480/light-client/src/supervisor.rs#L14-L34) nor the [`LightClient`](https://github.com/informalsystems/tendermint-rs/blob/b1252c639ece21aef5c6ac46f6e5b86187880480/light-client/src/light_client.rs#L51) expose a method for intialization. Maybe I'm missing sth here? cc @romac @brapse

I agree that this should live in the light-client crate instead. There is a related discussion around what should live in the core library vs here:  
- https://github.com/informalsystems/tendermint-rs/pull/430#discussion_r452662130
- https://github.com/informalsystems/tendermint-rs/pull/430#discussion_r452656444

light-node/src/commands/start.rs
brapse: Or I think we can use a [timer](https://docs.rs/crossbeam-channel/0.4.2/crossbeam_channel/fn.tick.html) if it's lighter.

xla: Thanks, good minimal solution right there.

light-node/src/commands/start.rs
ebuchman: We should just check the initial light block with all witnesses, no?

liamsi: So, we should actually check the initial light block with every peer / witness? Seems a bit weird bc if the header hash matches what is the attack we are trying to prevent by checking it with all witnesses? What would be the consequence. The human component was fooled (subjectively)?

light-node/src/commands/start.rs
liamsi: Does that look better now? https://github.com/informalsystems/tendermint-rs/pull/430/commits/08ce5cc1f6fe63b8f176a8a4447782cc27ac5574

light-node/tests/acceptance.rs
```diff
@@ -45,46 +43,21 @@ fn start_no_args() {
...
-    let mut runner = RUNNER.clone();
-    let mut cmd = runner
-        .config(&config)
-        .args(&["start", "other:26657"])
-        .capture_stdout()
-        .run();
-
-    cmd.stdout().expect_line("Requesting from other:26657.");
-    cmd.wait().unwrap().expect_success();
+    todo!()
```
liamsi, line 60: Note that we should also capture running these tests against a fullnode in integration tests. Should also capture this in a followup issue.

light-node/src/commands/start.rs
liamsi: `crossbeam_channel::tick` looks like a good fit but I'm wondering if we should stick to `std::thread::sleep` as it doesn't require yet another dependency. LMK what you think.

light-node/src/commands/start.rs
```diff
@@ -108,53 +88,103 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
-) -> Result<(), Error> {
-    if store.get(height).is_ok() {
-        // we already have this !
-        return Ok(());
+impl StartCmd {
+    fn assert_init_was_run() {
+        // TODO: handle errors properly:
+        let primary_db_path = app_config().light_clients.first().unwrap().db_path.clone();
+        let db = sled::open(primary_db_path).unwrap_or_else(|e| {
+            println!("[error] could not open database: {}", e);
```
xla, line 103: Don't we have proper logging facilities through abcissa?

light-node/src/commands/start.rs
```diff
@@ -4,94 +4,74 @@
...
-                )
-                .await
-                .unwrap();
-
-                println!("Succeeded bisecting!");
-
-                // notifications ?
-
-                // sleep for a few secs ?
+                // TODO: use ticks and make this configurable:
```
xla, line 73:
```suggestion
                // TODO(liamsi): use ticks and make this configurable:
```

light-node/src/commands/start.rs
xla: The only solution is fine for now, we can look into `crossbeam` once we have more components to coordinate.

light-node/src/commands/start.rs
```diff
@@ -108,53 +88,103 @@ impl config::Override<LightNodeConfig> for StartCmd {
         &self,
         mut config: LightNodeConfig,
     ) -> Result<LightNodeConfig, FrameworkError> {
-        if !self.rpc_addr.is_empty() {
-            config.rpc_address = self.rpc_addr.to_owned();
+        // Todo figure out if other options would be reasonable to overwrite via CLI arguments.
```
xla, line 91:
```suggestion
        // TODO(liamsi): figure out if other options would be reasonable to overwrite via CLI arguments.
```

light-node/tests/acceptance.rs
xla: Do we have one for this?

light-node/src/config.rs
```diff
@@ -4,55 +4,101 @@
...
+
+/// Default light client config settings.
+impl Default for LightClientConfig {
     fn default() -> Self {
         Self {
-            rpc_address: "localhost:26657".to_owned(),
-            trusting_period: Duration::new(6000, 0),
-            subjective_init: SubjectiveInit::default(),
+            address: "tcp://127.0.0.1:26657".parse().unwrap(),
+            peer_id: "BADFADAD0BEFEEDC0C0ADEADBEEFC0FFEEFACADE".parse().unwrap(),
```
xla, line 67: In what context will this default work?

light-node/src/config.rs
```diff
@@ -4,55 +4,101 @@
...
     pub trusting_period: Duration,
-    /// Subjective initialization.
-    pub subjective_init: SubjectiveInit,
+    /// Correction parameter dealing with only approximately synchronized clocks.
+    pub clock_drift: Duration,
+
+    /// RPC related config parameters.
+    pub rpc_config: RpcConfig,
+
+    // TODO "now" should probably always be passed in as `Time::now()`
```
xla, line 31:
```suggestion
    // TODO(liamsi): "now" should probably always be passed in as `Time::now()`
```

light-node/src/commands/start.rs
```diff
@@ -108,53 +88,103 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
+    }
 
-    store.add(trusted_state.to_owned())?;
+    fn start_rpc_server<H>(h: H)
+    where
+        H: Handle + Send + Sync + 'static,
+    {
+        let server = Server::new(h);
+        let laddr = app_config().rpc_config.listen_addr;
+        // TODO: figure out howto handle the potenial error on run
```
xla, line 152:
```suggestion
        // TODO(liamsi): figure out howto handle the potenial error on run
```

light-node/src/commands/start.rs
```diff
@@ -108,53 +88,103 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
+        });
+
+        let primary_store = SledStore::new(db);
+
+        if primary_store.latest(Status::Verified).is_none() {
+            println!("[error] no trusted state in store for primary, please initialize with the `initialize` subcommand first");
+            std::process::exit(1);
+        }
     }
+    // TODO: this should do proper error handling, be gerneralized
```
xla, line 114:
```suggestion
    // TODO(liamsi): this should do proper error handling, be gerneralized
```

light-node/src/commands/start.rs
```diff
@@ -108,53 +88,103 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
-    vals_hash: Hash,
-    store: &mut MemStore,
-    req: &RPCRequester,
-) -> Result<(), Error> {
-    if store.get(height).is_ok() {
-        // we already have this !
-        return Ok(());
+impl StartCmd {
+    fn assert_init_was_run() {
+        // TODO: handle errors properly:
```
xla, line 100:
```suggestion
        // TODO(liamsi): handle errors properly:
```

light-node/src/commands/start.rs
liamsi: Good point. I think it does.

liamsi: image

light-node/src/config.rs
liamsi: Tracked in #433

light-node/src/config.rs
liamsi: We currently do not verify that the peerId matches the address we are talking to ... So it will work if it can talk to 127.0.0.1:26657.

liamsi:
> We should make sure that for all conversations that are left for the future we have follow-up issues.

I was planning to go through each comment and capture everything (that's not yet addressed) in a followup issue after this gets merged. Ok?

> Also I find it strange that we have a dedicated light-node bin in the light-node crate, can't we just use the main main?

That's the way absicca generates the create structure on init. e.g. running `abscissa new test-app` yields:
```
tree                                                                                                                                                                                
.
`-- test-app
    |-- Cargo.lock
    |-- Cargo.toml
    |-- README.md
    |-- src
    |   |-- application.rs
    |   |-- bin
    |   |   `-- test_app
    |   |       `-- main.rs
    |   |-- commands
    |   |   |-- start.rs
    |   |   `-- version.rs
    |   |-- commands.rs
    |   |-- config.rs
    |   |-- error.rs
    |   |-- lib.rs
    |   `-- prelude.rs
    `-- tests
        `-- acceptance.rs
```

light-node/light_node.toml.example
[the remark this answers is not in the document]
liamsi: >  I don't feel strongly either way. 
but I would still insist 😆

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,110 @@
...
+use tendermint_light_client::predicates::{ProdPredicates, VerificationPredicates};
+use tendermint_light_client::store::sled::SledStore;
+use tendermint_light_client::store::LightStore;
+use tendermint_light_client::types::Status;
+
+/// `initialize` subcommand
+#[derive(Command, Debug, Default, Options)]
+pub struct InitCmd {
+    #[options(
+        free,
```
romac, line 29: What's the rationale for having the `height` and `hash` as free options rather than names ones? Is that because those are required options, and we'd rather leave named options for optional parameters?

While it's unlikely/almost impossible one would switch the two options around since the hash could never be parsed as a u64, I would personally rather have them named explicitly, although I don't feel strongly either way.

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,110 @@
...
+        help = "subjective height of the initial trusted state to initialize the node with"
+    )]
+    pub height: u64,
+
+    #[options(
+        free,
```
romac, line 35: Same as above.

light-node/src/commands/initialize.rs
romac: FYI: The `panic!` macro will properly unwind the stack if compiled with `panic=unwind` (the default), so it could perhaps be used here instead of `process::exit(1)`. One thing to note is that `panic!` will set the status code to `101` and this is AFAIK not configurable.

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,110 @@
...
+            std::process::exit(1);
+        });
+
+    let predicates = ProdPredicates;
+    let hasher = ProdHasher;
+    if let Err(err) = predicates.validator_sets_match(&trusted_state, &hasher) {
+        status_err!("invalid light block: {}", err);
+        std::process::exit(1);
+    }
+    // TODO(ismail): actually verify more predicates of light block before storing!?
```
romac, line 97: Good question! We should clarify how much of the light block we need to validate when doing subjective initialization. Is this documented somewhere? What does our Go counterpart do here?

light-node/src/commands/start.rs
```diff
@@ -108,53 +94,102 @@ impl config::Override<LightNodeConfig> for StartCmd {
...
+            peer_map.clone(),
+            Some(app_config().rpc_config.request_timeout),
+        );
+        let conf = app_config().deref().clone();
+        let options: light_client::Options = conf.into();
+
+        let mut peer_list: PeerListBuilder<Instance> = PeerList::builder();
+        for (i, light_conf) in app_config().light_clients.iter().enumerate() {
+            let instance = self.make_instance(light_conf, io.clone(), options);
+            if i == 0 {
```
romac, line 180: Should we perhaps make the initial primary/witness distinction more explicit by introducing a `primary` option and a `witnesses` array in the config?

light-node/src/commands/initialize.rs
liamsi: > Is that because those are required options, and we'd rather leave named options for optional parameters?

Yes exactly. Seems like in absicca or rather gumdrop this seems the way to say these are mandatory. I think we can remove the `free` make them named and manually check if they were provided. Let's move this to a followup issue.

light-node/src/commands/initialize.rs
liamsi: Not that I'm aware of. I think we always thought of it to be obvious how subjective initialization works. But it seems we didn't think through how it is supposed to work in the context of a fork-detector.

light-node/src/commands/initialize.rs
liamsi: But doesn't the panic macro litter the output with the unwinded stack? I'll open a followup issue and we should add a proper way to exit here instead.

light-node/src/commands/start.rs
liamsi: Probably a good idea. If we make this more explicit, the question though is what happens if the (e.g.) the primary doesn't reply? Does the light-node try the other nodes and makes the first one primary? Or do we just return and let the user figure it out? I'll leave this one open too and capture it in a followup issues, too.

romac: Yeah that's a good question, both in general and from a UX perspective. Let's discuss it in a follow-up!

light-node/src/commands/initialize.rs
```diff
@@ -0,0 +1,110 @@
...
+    if got_header_hash != subjective_header_hash {
+        status_err!(
+            "received LightBlock's header hash: {} does not match the subjective hash: {}",
+            got_header_hash,
+            subjective_header_hash
+        );
+        std::process::exit(1);
+    }
+    // TODO(liamsi): it is unclear if this should be Trusted or only Verified
+    //  - update the spec first and then use library method instead of this:
```
liamsi, line 108: Capturing this in a followup issue too.

liamsi:
Captured all feedback and open issues in https://github.com/informalsystems/tendermint-rs/issues/442

Issue #373, opened 2020-06-25
light-client: test against a real full node in CI


ebuchman:
We should run the light node against a real full node in CI per https://github.com/informalsystems/tendermint-rs/issues/120#issuecomment-644173634

We already run the RPC integration tests against a local full node, but we're not running the light node against it. 

The first and simplest thing to do is to run the light-node against that full node. Since it will just be a 1 validator blockchain that doesn't change, this will not test much of the light-client functionality, but is more about ensuring the light-node init/config/process is working correctly and can sync with a real full node.

Future iterations could introduce a lot more complexity:
- talking to multiple full nodes (where one of the full nodes may be faulty or may serve a fork)
- enabling validator set changes on the chain (this would require other full nodes to actually be set up to be the new validators, even if the light node doesn't talk directly to them!)
- talking to remote full nodes from a live mainnet/testnet Gaia blockchain 

Since there's a lot to do re scoping the more complex set ups, let's start with the simplest for now.

liamsi:
ref: https://github.com/tendermint/tendermint/pull/5087
Context: when new docker images for tendermint-go are published, it should simplify testing against CI slightly.

ebuchman:
Hey is this done with #431?

brapse:
Yeah this is done but made a follow up for the binary initiazation process specifically in #445.

PR #431, opened 2020-07-09
light-client: integration test

romac:
cf. #373

Adds an integration test for the light client.

The test spins up a light client against a single Tendermint node, where both the primary and a single witness point to that node.
It tries to verify to the highest block 20 times, at an interval of 800ms to give the chain time to advance.
The test uses a custom evidence reporter which panics if the supervisor tries to report a fork, as no fork should happen with this setup.

This PR also fixes a bug where the postcondition of `LightClient::verify_to_target` was only looking for verified but not trusted blocks in the light store. This bug occurs when the highest block is still the initial trusted block, which happens in this case because we start syncing to the latest height right after fetching and initializing the trusted block.

---

* [x] Wrote tests
* [x] Updated CHANGES.md

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,7 +65,7 @@ jobs:
   #  - https://github.community/t/support-for-yaml-anchors/16128/15
   #  - https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/13
   #  - https://github.community/t/using-matrix-variable-in-docker-image-name/17296
-  test-integration-stable:
+  test-integration-tendermint-stable:
     runs-on: ubuntu-latest
     services:
       tendermint:
@@ -84,6 +84,26 @@ jobs:
         with:
           command: test
           args: -p tendermint --test integration --no-fail-fast -- --ignored
+
+  # TODO(shonfeder): remove duplication once GitHub addresses one of these
+  #  - https://github.community/t/support-for-yaml-anchors/16128/15
+  #  - https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/13
+  #  - https://github.community/t/using-matrix-variable-in-docker-image-name/17296
+  test-integration-light-client-stable:
+    runs-on: ubuntu-latest
+    services:
+      tendermint:
+        image: tendermint/tendermint:v0.33.5
+        ports:
+          - 26656:26656
+          - 26657:26657
+          - 26660:26660
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
       - uses: actions-rs/cargo@v1
         with:
           command: test
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -84,6 +84,10 @@ jobs:
         with:
           command: test
           args: -p tendermint --test integration --no-fail-fast -- --ignored
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: -p tendermint-light-client --test integration --no-fail-fast -- --ignored
 
   test-integration-latest:
     runs-on: ubuntu-latest
--- a/light-client/src/contracts.rs
+++ b/light-client/src/contracts.rs
@@ -12,6 +12,7 @@ pub fn trusted_store_contains_block_at_target_height(
     target_height: Height,
 ) -> bool {
     light_store.get(target_height, Status::Verified).is_some()
+        || light_store.get(target_height, Status::Trusted).is_some()
 }
 
 pub fn is_within_trust_period(
--- a/light-client/src/contracts.rs
+++ b/light-client/src/contracts.rs
@@ -30,15 +30,6 @@ pub fn light_store_contains_block_within_trusting_period(
     now: Time,
 ) -> bool {
     light_store
-        .all(Status::Verified)
+        .all(Status::Trusted)
         .any(|lb| is_within_trust_period(&lb, trusting_period, now))
 }
-
-// pub fn target_height_greater_than_all_blocks_in_trusted_store(
-//     light_store: &dyn LightStore,
-//     target_height: Height,
-// ) -> bool {
-//     light_store
-//         .all(Status::Verified)
-//         .all(|lb| lb.height() < target_height)
-// }
--- a/light-client/src/light_client.rs
+++ b/light-client/src/light_client.rs
@@ -161,6 +161,7 @@ impl LightClient {
         // Override the `now` fields in the given verification options with the current time,
         // as per the given `clock`.
         let options = self.options.with_now(self.clock.now());
+        dbg!(&self.options, &options, self.clock.now());
 
         let mut current_height = target_height;
 
--- a/light-client/src/light_client.rs
+++ b/light-client/src/light_client.rs
@@ -161,7 +161,6 @@ impl LightClient {
         // Override the `now` fields in the given verification options with the current time,
         // as per the given `clock`.
         let options = self.options.with_now(self.clock.now());
-        dbg!(&self.options, &options, self.clock.now());
 
         let mut current_height = target_height;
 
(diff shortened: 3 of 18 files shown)


light-client/tests/integration.rs
```diff
@@ -0,0 +1,121 @@
...
+#[ignore]
+fn sync() {
+    let primary: PeerId = "BADFADAD0BEFEEDC0C0ADEADBEEFC0FFEEFACADE".parse().unwrap();
+    let witness: PeerId = "CEFEEDBADFADAD0C0CEEFACADE0ADEADBEEFC0FF".parse().unwrap();
+
+    let node_address: tendermint::net::Address = "tcp://127.0.0.1:26657".parse().unwrap();
+
+    let mut peer_map = HashMap::new();
+    peer_map.insert(primary, node_address.clone());
+    peer_map.insert(witness, node_address);
```
liamsi, line 76: Here the same RPC end-point is used twice. Do you think it makes sense add a comment that in a realistic scenario these would be different and that you'd check if the peerId matches?

light-client/src/contracts.rs
```diff
@@ -12,6 +12,7 @@ pub fn trusted_store_contains_block_at_target_height(
     target_height: Height,
 ) -> bool {
     light_store.get(target_height, Status::Verified).is_some()
+        || light_store.get(target_height, Status::Trusted).is_some()
```
liamsi, line 15: I'm wondering if there are more places like this. reminded me a little of #419

romac:
@liamsi Looking into it!

light-client/src/contracts.rs
romac: Yeah I thought I had caught them all in #419 but apparently not. Will go over the codebase again to make sure.

romac:
The failure came from a ~2min clock drift between the Docker container and the GitHub CI worker. Increasing the clock drift parameter to 5min (to be safe) has fixed it. @liamsi Does that seem like an acceptable fix to you?

light-client/tests/integration.rs
romac: Addressed in 41283822e5a131ff1866fc174785b12b07d8af30.

light-client/src/light_client.rs
```diff
@@ -26,23 +26,16 @@ pub struct Options {
...
     /// unbonding period)
     pub trusting_period: Duration,
+
     /// Correction parameter dealing with only approximately synchronized clocks.
     /// The local clock should always be ahead of timestamps from the blockchain; this
     /// is the maximum amount that the local clock may drift behind a timestamp from the
     /// blockchain.
     pub clock_drift: Duration,
-    /// The current time
-    pub now: Time,
```
liamsi, line 38: I'm actually using these "Options" in #430 too. I was wondering how this made sense as an option. Much clearer when it isn't 👍

liamsi:
>  @liamsi Does that seem like an acceptable fix to you?

Does sound reasonable. Do you understand why the clock drift is so large?

romac:
> Do you understand why the clock drift is so large?

I unfortunately don't, would require more investigation. Perhaps someone else knows better? @greg-szabo?

liamsi:
BTW, I also get a lot of `[error] sync failed: no witness left
` in https://github.com/informalsystems/tendermint-rs/pull/430
Is that always clock drift issue?

.github/workflows/rust.yml
```diff
@@ -84,6 +84,10 @@ jobs:
         with:
           command: test
           args: -p tendermint --test integration --no-fail-fast -- --ignored
+      - uses: actions-rs/cargo@v1
```
xla, line 87: I believe it would be better to isolate the jobs per crate. i.e. `test-integration-light-client-stable` under jobs.

light-client/src/predicates.rs
```diff
@@ -202,6 +202,7 @@ pub trait VerificationPredicates: Send {
 /// check that their (next) validator sets hashes match.
 /// - Otherwise, ensure that the untrusted block has a greater height than
 /// the trusted block.
+#[allow(clippy::too_many_arguments)]
```
xla, line 205: This is not a good sign, we should look into how to refactor this. Maybe we capture a follow-up issue.

light-client/tests/integration.rs
```diff
@@ -0,0 +1,125 @@
...
+    },
+    evidence::{Evidence, EvidenceReporter},
+    fork_detector::ProdForkDetector,
+    light_client::{self, LightClient},
+    peer_list::PeerList,
+    state::State,
+    store::{memory::MemoryStore, LightStore},
+    supervisor::{Handle, Instance, Supervisor},
+    types::{PeerId, Status, TrustThreshold},
+};
```
xla, line 25:
```suggestion
use tendermint_light_client::components::clock::SystemClock;
use tendermint_light_client::components::io::{AtHeight, Io, IoError, ProdIo};
use tendermint_light_client::components::scheduler;
use tendermint_light_client::components::verifier::ProdVerifier;
use tendermint_light_client::evidence::{Evidence, EvidenceReporter};
use tendermint_light_client::fork_detector::ProdForkDetector;
use tendermint_light_client::light_client::{self, LightClient};
use tendermint_light_client::peer_list::PeerList;
use tendermint_light_client::state::State;
use tendermint_light_client::store::{memory::MemoryStore, LightStore};
use tendermint_light_client::supervisor::{Handle, Instance, Supervisor};
use tendermint_light_client::types::{PeerId, Status, TrustThreshold};
```

This style of imports is way easier to edit.

light-client/tests/integration.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use tendermint::abci::transaction::Hash as TransactionHash;
+
+use std::collections::HashMap;
+use std::time::Duration;
+
+fn make_instance(peer_id: PeerId, options: light_client::Options, io: ProdIo) -> Instance {
```
xla, line 32: I believe we have a similar method in the example, if this is common enough utility maybe it's time to lift it into the library.

romac: Yes, we plan to eventually deduplicate this type of code (#422), and clean up the tests utilities as well.

xla: Dope!

light-client/tests/integration.rs
romac: Yeah we are not very consistent with the style of imports we use. I started switching to the single import style since that's apparently what the Rust community in general is moving towards, but I don't feel strongly either way. Let's track this in #433.

light-client/src/predicates.rs
romac: Agreed, it's not ideal indeed. And we could perhaps improve things further by introducing typed wrappers for the untrusted/trusted blocks to prevent mixing them up. So yeah, let's follow-up on that!

.github/workflows/rust.yml
romac: Alright!

romac:
> BTW, I also get a lot of `[error] sync failed: no witness left ` in #430
> Is that always clock drift issue?

@liamsi Hard to say like that. Might need to [add some logging statements](https://github.com/informalsystems/tendermint-rs/pull/431/commits/5b8186edc622d6b0aafc830f1012833b142614c6) to figure out what's the underlying error. Perhaps we could/should even track those somehow.

light-client/tests/integration.rs
romac: Since this import style shows up in most places in the `light-client` crate already, I am going to leave it like this until we settle on an import style in #433.

light-client/src/predicates.rs
xla: Great, think this discussion is happening in other places and shouldn't be a blocker in any form for this PR!

light-client/tests/integration.rs
xla: A follow-up on this sounds perfect. Thanks for capturing it.

.github/workflows/rust.yml
romac: Done

xla: Dope!

romac:
@xla @liamsi Can one of please you approve for merging? :)

PR #432, opened 2020-07-10
abci: fix rpc response deserialization

### LINKED ISSUE #423: rpc response decode failed with recent released version of tendermint
Does it only works with tendermint git version?
I used the most recent released version(0.33.6), I got many null decoding issues, also the `gasUsed -> gas_used/gasWanted -> gas_wanted` naming differences.

An example block_results response:
```json
{
  "begin_block_events":null,
  "consensus_param_updates":null,
  "end_block_events":[
    {
      "attributes":[
        {
          "key":"ZXRoYmxvb20=",
          "value":"AAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=="
        }
      ],
      "type":"block_filter"
    }
  ],
  "height":"4",
  "txs_results":[
    {
      "code":0,
      "codespace":"",
      "data":null,
      "events":[
        {
          "attributes":[
            {
              "key":"ZmVl",
              "value":"MC4wMDAwMDMwOA=="
            },
            {
              "key":"dHhpZA==",
              "value":"Y2UxYzc3NDIyM2Q2ZWNjMDQ4MWIxZTVkMjc2MmE4Njc0ZDhkODc2MjI2MjRlNTdlYmZiMzk4NTEzZmJhNDVmNg=="
            }
          ],
          "type":"valid_txs"
        },
        {
          "attributes":[
            {
              "key":"c3Rha2luZ19hZGRyZXNz",
              "value":"MHgzMzUwMmVkMzlkMGM0ZTIwNDRmYjM3ZmRjZDUxNjE0OTNmNTkwMGMz"
            },
            {
              "key":"c3Rha2luZ19vcHR5cGU=",
              "value":"d2l0aGRyYXc="
            },
            {
              "key":"c3Rha2luZ19kaWZm",
              "value":"W3sia2V5IjoiVW5ib25kZWQiLCJ2YWx1ZSI6Ii01MDAwMDAwMDAwMDAwMDAwMDAifV0="
            }
          ],
          "type":"staking_change"
        }
      ],
      "gasUsed":"0",
      "gasWanted":"0",
      "info":"",
      "log":""
    }
  ],
  "validator_updates":null
}
```

Compare to the one in unit test: https://github.com/informalsystems/tendermint-rs/blob/master/rpc/tests/support/block_results.json

yihuang:
Fix rpc response deserialization issues. fix #423 

By the way, the `null_as_default` deserializer might be good to use on some other fields.

### DIFF
--- a/tendermint/src/abci/responses.rs
+++ b/tendermint/src/abci/responses.rs
@@ -1,7 +1,7 @@
 //! ABCI response types used by the `/block_results` RPC endpoint.
 
 use super::{code::Code, data::Data, gas::Gas, info::Info, log::Log, tag::Tag};
-use crate::{consensus, validator};
+use crate::{consensus, serializers, validator};
 use serde::{Deserialize, Deserializer, Serialize};
 use std::fmt::{self, Display};
 
@@ -50,6 +50,7 @@ pub struct DeliverTx {
     pub code: Code,
 
     /// ABCI application data
+    #[serde(deserialize_with = "serializers::null_as_default")]
     pub data: Data,
 
     /// ABCI log data (nondeterministic)
@@ -59,9 +60,11 @@ pub struct DeliverTx {
     pub info: Info,
 
     /// Amount of gas wanted
+    #[serde(rename = "gasWanted")]
     pub gas_wanted: Gas,
 
     /// Amount of gas used
+    #[serde(rename = "gasUsed")]
     pub gas_used: Gas,
 
     /// Events
--- a/rpc/tests/support/block_results.json
+++ b/rpc/tests/support/block_results.json
@@ -8,9 +8,9 @@
         "code": "0",
         "log": "[{\"msg_index\":\"0\",\"success\":true,\"log\":\"\"}]",
         "info": "",
-        "gas_wanted": "200000",
-        "gas_used": "105662",
-        "data": "",
+        "gasWanted": "200000",
+        "gasUsed": "105662",
+        "data": null,
         "events": [
           {
             "type": "someevent1",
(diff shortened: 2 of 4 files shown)


tendermint/src/abci/responses.rs
```diff
@@ -59,9 +60,11 @@ pub struct DeliverTx {
     pub info: Info,
 
     /// Amount of gas wanted
+    #[serde(rename = "gasWanted")]
```
xla, line 63: You can avoid all of these field attribute by putting a `#[serde(rename_all = "camelCase")]` on the struct. See the [section](https://serde.rs/container-attrs.html#rename_all) in the container attribute documentation.

yihuang:
> Thanks for the concise change-set. Could you explain why the null as default behaviour is needed/desired and if it is treated like that somewhere else in tendermint like that.

- The `"data":null,` is what I observed from the response return from tendermint 0.33.6, I pasted that response here: https://github.com/informalsystems/tendermint-rs/issues/423#issue-653716758
- I guess we don't have a proper spec on protobuf json format convertion, but from my understanding, protobuf don't distinguish between default value and non-exist ones(non-exist value can be ignored in wire message).

tendermint/src/abci/responses.rs
yihuang: I think the camelCase here is a special case, because other places tendermint actually use `_` format.

yihuang:
> It be great, if this PR also comes with an (updated?) test-case. Note that currently, there is a test-failure:

The test case is fixed now.

PR #438, opened 2020-07-10
Fix header.hash for height 1

### LINKED ISSUE #436: light-node: can't initialise with header at height 1
An attempt to init with the header at height `1` fails.

Potentially reproducible with:
* run a tendermint full node
* `curl -X GET "http://localhost:26657/block?height=1” -H  "accept: application/json" `
* deserialise in Rust and investigate why `header.hash() != block_id.hash`

Potential causes:
* latest 0.33 changed something
* a potential bug when one of the `last_` fields is not present in

https://github.com/informalsystems/tendermint-rs/blob/142796b0cc37319f91af55cf2703b96a0cb80230/tendermint/src/lite_impl/header.rs#L28-L58

liamsi:
closes #436 

* [x] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/rpc/tests/integration.rs
+++ b/rpc/tests/integration.rs
@@ -225,6 +225,16 @@ mod endpoints {
     }
 
     #[test]
+    fn commit_height_1() {
+        let response =
+            endpoint::commit::Response::from_string(&read_json_fixture("commit_1")).unwrap();
+        let header = response.signed_header.header;
+        let commit = response.signed_header.commit;
+        let block_id = commit.block_id;
+        assert_eq!(header.hash(), block_id.hash);
+    }
+
+    #[test]
     fn genesis() {
         let response =
             endpoint::genesis::Response::from_string(&read_json_fixture("genesis")).unwrap();
--- a/tendermint/src/lite_impl/header.rs
+++ b/tendermint/src/lite_impl/header.rs
@@ -67,7 +67,6 @@ fn encode_bytes(bytes: &[u8]) -> Vec<u8> {
     } else {
         vec![]
     }
-
 }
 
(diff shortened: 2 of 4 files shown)


rpc/tests/integration.rs
```diff
@@ -224,6 +224,16 @@ mod endpoints {
         assert_eq!(header.hash(), block_id.hash);
     }
 
+    #[test]
+    fn commit_height_1() {
```
xla, line 228: Nice to have a case, when I tried it out I went straight into the `lite_impl` module. As we likely gonna fix the `fn hash` implementation, shouldn't we test deeper in the stack?

liamsi: Probably both?

tendermint/src/lite_impl/header.rs
```diff
@@ -48,28 +48,80 @@ impl lite::Header for block::Header {
...
-        fields_bytes.push(bytes_enc(self.proposer_address.as_bytes()));
+        fields_bytes.push(encode_bytes(self.proposer_address.as_bytes()));
 
         Hash::Sha256(simple_hash_from_byte_vectors(fields_bytes))
     }
 }
 
-fn bytes_enc(bytes: &[u8]) -> Vec<u8> {
-    let mut chain_id_enc = vec![];
-    prost_amino::encode_length_delimiter(bytes.len(), &mut chain_id_enc).unwrap();
```
liamsi, line 62: The problem was here: we length-encoded even zero length bytes. In this particular case the `app_hash`.

liamsi:
Thanks @romac! I Forgot that this was copied over. The deduplication should probably go in the other direction (into the light-client crate) but it makes sense to have only one tested and working implementation in place 👍

PR #441, opened 2020-07-13
make dependency secp256k1 optional, fix #391

### LINKED ISSUE #391: Optional dependencies
Not sure if this will be resolved with https://github.com/informalsystems/tendermint-rs/issues/387
but could some tendermint-rs dependencies be made optional (with a feature-guard) -- e.g. if one's TM genesis.json is configured with:
```
    "validator": {
      "pub_key_types": [
        "ed25519"
      ]
    }
```

It's unnecessary to have `signatory` with `ecdsa`-feature, `signatory-secp256k1` and `ripemd160`?

yihuang:
fix #391

### DIFF
--- a/tendermint/src/public_key.rs
+++ b/tendermint/src/public_key.rs
@@ -3,7 +3,9 @@
 use crate::error::{Error, Kind};
 use anomaly::fail;
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
-use signatory::{ecdsa::curve::secp256k1, ed25519};
+#[cfg(feature = "secp256k1")]
+use signatory::ecdsa::curve::secp256k1;
+use signatory::ed25519;
 use std::{
     fmt::{self, Display},
     ops::Deref,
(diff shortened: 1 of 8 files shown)


tendermint/src/public_key.rs
```diff
@@ -3,7 +3,9 @@
 use crate::error::{Error, Kind};
 use anomaly::fail;
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
-use signatory::{ecdsa::curve::secp256k1, ed25519};
+#[cfg(feature = "secp256k1")]
```
liamsi, line 6: If we moved everything secp-related into a separate file/module, we can likely just annotate it on the `mod` level instead? Not blocking this PR but probably worth separating this out?

tendermint/src/public_key.rs
```diff
@@ -49,6 +53,7 @@ impl PublicKey {
 
     /// Get Ed25519 public key
     pub fn ed25519(self) -> Option<ed25519::PublicKey> {
+        #[allow(unreachable_patterns)]
```
romac, line 56: I understand that those `_ => unreachable!()` patterns were here previously to this PR, but just wanted to mention that those seem inherently unsafe to me (as in: may lead to panics). This PR improves the state of things by making them actually unreachable if the `secp265k1` feature is not enabled, but we should perhaps take a closer look at them in a follow-up issue.

Issue #420, opened 2020-07-08
Ensure a trusted light block stays trusted permanently, even if its peer fails later


romac:
> > @milosevic I had a bit of different issue in mind, which more about how we deal with state of different peers. During the `Supervisor` verification the only instance which updates it's `LightBlockStore` is the primary, while during fork detection the witnesses are using freshly created in memory stores which after that get discarded. Therefore, even though we have stored a `LightBlock` which we consider trusted if at any point in time we find out the primary is faulty we swap it and use the `State` associated with the new primary instance and as soon as we do that we basically go back in time to the state before any verification. For me this behaviours is strange as the history of previously verified `LightBlocks` is lost and we might need to restart the light client just because the primary was faulty.
> 
> @josef-widder  I think once a LightBlock is trusted, it should stay trusted forever, even if the primary fails later. Otherwise, it is not clear what is the meaning of "trusted".

— https://github.com/informalsystems/tendermint-rs/issues/395#issuecomment-653508642

xla:
This should include the API to not return an `Option`.

milosevic:
I agree that trusted light block is permanently trusted, even after primary fails (we don't consider forks in this case). We should rely on this invariant when electing new primary; we can continue operating from the latest trusted state, i.e., we don't need to throw away complete state and reinitialise the light client.

PR #446, opened 2020-07-14
Sharing the state between peers.

OStevan:
__Issue and how it solves the problem?__
#420 , this PR proposes a solution in which the `Supervisor` has a state which would be updated with trusted `LightBlocks` after each successful pass of fork detection. Subsequent verification requests first read the highest trusted LightBlock in the supervisor state and use that `LightBlock` as the starting point for verification.

__Concerns related to the solution?__
- Core verification on the primary and witnesses is always using `InMemoryVersion` of `LightStore`. We could either simplify the instance and remove associated `LightStores` or try to always use the one associated with the instance
- Should we actually always use the one associated with an `Instance` so that for example we can examine the store in case of failure?

__Other related discussions?__
The solution from #479 deletes the light blocks higher that the previously trusted one in the shared state. The approach here just discards the light store used by the primary.

* [ ] Wrote tests
* [ ] Updated CHANGES.md


light-node/src/config.rs
```diff
@@ -89,6 +92,7 @@ impl Default for LightNodeConfig {
             // TODO(ismail): need at least 2 peers for a proper init
             // otherwise the light node will complain on `start` with `no witness left`
             light_clients: vec![LightClientConfig::default()],
+            shared_state_config: "./lightstore".parse().unwrap(),
```
liamsi, line 95: Not sure I have the right context but doesn't this mean you can remove the separate stores in each light client?
See above:
```rust
/// LightClientConfig contains all options of a light client instance.
#[serde(deny_unknown_fields)]
pub struct LightClientConfig {
    /// Address of the Tendermint fullnode to connect to and
    /// fetch LightBlock data from.
    pub address: tendermint::net::Address,
    /// PeerID of the same Tendermint fullnode.
    pub peer_id: PeerId,
    /// The data base folder for this instance's store.
    pub db_path: PathBuf,
}
```

OStevan: This is something I wanted to discuss. In practice we can use in-memory state for each peer instance enabling us to remove the separate stores for each light client. I'm however unsure if we might still want to keep these stores to be able to inspect what each light client verified by itself(do we even want such a feature?). @liamsi

liamsi: >  I'm however unsure if we might still want to keep these stores to be able to inspect what each light client verified by itself(do we even want such a feature?).

Not sure tbh, but there might be some cases where this is required for the fork-detector!

light-node/src/commands/initialize.rs
```diff
@@ -51,7 +50,22 @@ impl Runnable for InitCmd {
 
         let io = ProdIo::new(peer_map, Some(app_cfg.rpc_config.request_timeout));
 
-        initialize_subjectively(self.height, subjective_header_hash, &lc, &io);
+        let lock = APPLICATION.read();
+
+        let factory: &dyn LightStoreFactory = lock
+            .state()
+            .components
+            .get_downcast_ref::<ProdLightStoreFactory>()
```
romac, line 58: I don't believe we need the flexibility afforded by the `LightStoreFactory` trait/component, a plain method which converts a `LightStoreConfig` to a `impl LightStore` would likely do the job as well.

This line is also a bit weird because we are always casting to a concrete type, which defeats the purpose of the trait.

romac:
Looks good to me 👍 Thanks for opening this draft PR for discussion!

I just have [one comment](https://github.com/informalsystems/tendermint-rs/pull/446#discussion_r458782915) on the need for a `LightStoreFactory`.

That said, I think the general idea of [sharing the state between all peers](https://github.com/informalsystems/tendermint-rs/issues/420) should be carefully considered and reviewed by @josef-widder and @milosevic.

light-node/src/commands/initialize.rs
OStevan: Changed it to a function.

OStevan:
ping @milosevic @josef-widder

josef-widder:
In #479, we stated the invariant [[LCD-INV-TRUSTED-AGREED.1]](https://github.com/informalsystems/tendermint-rs/blob/df8069a53389fd60b359305e3dc6d71858e407c1/docs/spec/lightclient/detection/detection.md#lcd-inv-trusted-agreed1). It says that the primary and the secondary always agree on the lightblock for height lightStore.LatestTrusted().Height. This allows us to not store any blocks downloaded from secondaries after fork detection returned: More precisely, fork detection is started from this latest trusted block at all secondaries, and if there is no fork, all downloaded blocks can be dropped. If there is a fork, the downloaded blocks are used to generate "proof of fork".

So I am not sure we need to share the state between peers, but perhaps I didn't get the question...

(@milosevic is on vacations this week)

brapse:
This is a great first step but should integrate the work in #479 and be preceded by an ADR outlining the approach. Might make sense to do this as a separate PR. Will leave the branch for reference.

PR #447, opened 2020-07-14
Add logo

liamsi:
Nothing big. Just add the logo which will show up in docs.rs (for an example see the upper left corner on https://docs.rs/signatory/0.20.0/signatory/)


tendermint/src/lib.rs
```diff
@@ -13,7 +13,10 @@
     unused_qualifications
 )]
 #![forbid(unsafe_code)]
-#![doc(html_root_url = "https://docs.rs/tendermint/0.14.1")]
+#![doc(
+    html_root_url = "https://docs.rs/tendermint/0.14.1",
+    html_logo_url = "https://raw.githubusercontent.com/informalsystems/tendermint-rs/master/img/logo-tendermint-rs.png"
```
liamsi, line 18: This URL will become valid when this is merged (see binary png file above).

PR #448, opened 2020-07-14
Cache rust deps during CI runs

shonfeder:
Just trying this on a hunch that it might give us a bit of a CI speedup.

Cribbed from https://github.com/actions/cache/blob/master/examples.md#rust---cargo

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,7 +28,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ runner.os }}-cargo-fmt-${{ hashFiles('**/Cargo.toml') }}
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
@@ -49,7 +49,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ runner.os }}-cargo-clippy-${{ hashFiles('**/Cargo.toml') }}
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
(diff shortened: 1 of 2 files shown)


shonfeder:
Not having the expected effect, and this is work out of band, just on a lark.

.github/workflows/rust.yml
```diff
@@ -34,6 +42,14 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - name: Rust cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-clippy-${{ hashFiles('**/Cargo.toml') }}
```
liamsi, line 52: It probably won't help with the speedup but wouldn't this need to be the top-level `Cargo.lock` file? Also some projects seems to use a different cache for the cargo registry and for the index (`.cargo/git`). Not sure if the latter helps either.

shonfeder: 1. This will match any cargo manifest in the project (iiuc). Should it be *only* the to level lock file?
2. We don't seem to have any lock files committed to this repo (including at the top level) which is why I changed this to just be the manifests: https://github.com/informalsystems/tendermint-rs/blob/2761a9d8cd6eae79eca9d7368a9bc6d1ed4452c3/.gitignore#L5
  - 2.1. Do we want to change this?
  - 2.1. Pinning the cache to the cargo manifest is too specific of a constraint to be ideal (since we don't want to dump the cash just because of some comment change in the manifest or something), but seemed ok for a trial run.
3. using different caches makes sense!

Overall, do you think this would be helpful? If so, given that I couldnt fix it in an hour or so of work, I'm happy to open an issue so we can budget work for it when appropriate.

liamsi: >  Should it be only the to level lock file?

I think there will only be one Cargo.lock per workspace if I'm not mistaken (in a train with a shaky inet connection rn but that's at least my observation locally). 

> 2.1. Do we want to change this? 

Yes, for the binary packages, we [should](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) actually include these 🤔 Also we could try this to make the most out of caches  by adding a lock file (and potentially to increase reproducibility?), not sure tbh.

> Overall, do you think this would be helpful? If so, given that I couldnt fix it in an hour or so of work, I'm happy to open an issue so we can budget work for it when appropriate.

I think reducing CI (and also local build) times could be useful. But caches aren't the only thing we could improve on: reducing or updating dependencies, remove not absolutely necessary jobs (like the integration tests against go-tendermint's master branch). Probably not a high priority though. And we already have so many open issues 🙈

shonfeder: > And we already have so many open issues

lol -- fortunately backlogs have an inexhaustible space.

PR #449, opened 2020-07-15
light-client: add missing RPC endpoint and extend handler/supervisor

### LINKED ISSUE #219: Light Client RPC Service

### LINKED ISSUE #450: light-client: light store does not seem to update
Reproduce by inspecting `trust_block` on an `Instance`:
https://github.com/informalsystems/tendermint-rs/blob/b90604790bd40982dd0371d37cde61d2ae7d6073/light-client/src/supervisor.rs#L72-L74

to:
```rust
pub fn trust_block(&mut self, lb: &LightBlock) {
  self.state.light_store.update(lb, Status::Trusted);
  dbg!(lb.height(), self.latest_trusted().unwrap().height());
}
```
currently prints two different heights:
```
synced to block: 145
[light-client/src/supervisor.rs:81] lb.height() = 147
[light-client/src/supervisor.rs:81] self.latest_trusted().unwrap().height() = 2
synced to block: 147
```
Same if we would look into the latest verified instead of trusted:
```rust
 dbg!(lb.height(), self.state.light_store.latest(Status::Verified).unwrap().height());
```

Found in https://github.com/informalsystems/tendermint-rs/pull/449/files#r455709387

liamsi:
https://github.com/informalsystems/tendermint-rs/issues/219#issuecomment-658041389
and contains a preliminary fix for #450 (thanks @OStevan)
closes #450 
closes #219 

* [x] Wrote tests
* [x] Updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@
 - Correctly handle blocks marked `Trusted` in accordance with the specification ([#407])
 - Treat `Trusted` status as a special case of `Verified` as per the spec ([#419])
 - Add integration test ([#431])
-- Reworked light-node to `Supervisor` / `Handle` ([#430])
+- Rework light-node CLI to use `Supervisor` / `Handle` ([#430])
 - Add `latest_status` to the supervisor `Handle` ([#449])
 - Add JSONRPC endpoints to query the light-node ([#363], [#449])  
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,10 +17,14 @@
 - Correctly handle blocks marked `Trusted` in accordance with the specification ([#407])
 - Treat `Trusted` status as a special case of `Verified` as per the spec ([#419])
 - Add integration test ([#431])
+- Reworked light-node to `Supervisor` / `Handle` ([#430])
+- Add `latest_status` to the supervisor `Handle` ([#449])
+- Add JSONRPC endpoints to query the light-node ([#363], [#449])  
 
 [#302]: https://github.com/informalsystems/tendermint-rs/pull/302
 [#336]: https://github.com/informalsystems/tendermint-rs/pull/336
 [#360]: https://github.com/informalsystems/tendermint-rs/pull/360
+[#363]: https://github.com/informalsystems/tendermint-rs/pull/363
 [#364]: https://github.com/informalsystems/tendermint-rs/pull/364
 [#365]: https://github.com/informalsystems/tendermint-rs/pull/365
 [#371]: https://github.com/informalsystems/tendermint-rs/pull/371
@@ -32,7 +36,9 @@
 [#403]: https://github.com/informalsystems/tendermint-rs/pull/403
 [#407]: https://github.com/informalsystems/tendermint-rs/pull/407
 [#419]: https://github.com/informalsystems/tendermint-rs/pull/419
+[#430]: https://github.com/informalsystems/tendermint-rs/pull/430
 [#431]: https://github.com/informalsystems/tendermint-rs/pull/431
+[#449]: https://github.com/informalsystems/tendermint-rs/pull/449
 
 [ADR-007]: https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-007-light-client-supervisor-ergonomics.md
 
--- a/light-client/src/supervisor.rs
+++ b/light-client/src/supervisor.rs
@@ -18,7 +18,7 @@ pub trait Handle {
         todo!()
     }
 
-    fn latest_status(&self) -> Result<Option<LatestStatus>, Error> {
+    fn latest_status(&self) -> Result<LatestStatus, Error> {
         todo!()
     }
 
@@ -51,7 +51,7 @@ enum HandleInput {
     /// Get the latest trusted block.
     LatestTrusted(channel::Sender<Option<LightBlock>>),
     /// Get the current status of the LightClient
-    GetStatus(channel::Sender<Option<LatestStatus>>),
+    GetStatus(channel::Sender<LatestStatus>),
 }
 
 /// A light client `Instance` packages a `LightClient` together with its `State`.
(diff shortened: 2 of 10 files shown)


liamsi:
TODOs:

 - [ ] add tests
 - [ ] update light-node Readme
 - [ ] changelog?

light-node/src/rpc.rs
```diff
@@ -38,13 +38,18 @@ mod sealed {
...
 
     #[rpc]
     pub trait Rpc {
         /// Returns the latest trusted block.
         #[rpc(name = "state")]
         fn state(&self) -> FutureResult<Option<LightBlock>, Error>;
+
+        /// Returns the latest status.
+        #[rpc(name = "status")]
+        fn status(&self) -> FutureResult<Option<LatestStatus>, Error>;
```
liamsi, line 52: Result for this endpoint looks like this:
```json
{
  "jsonrpc": "2.0",
  "result": {
    "block_hash": "1BA78A40F4D785E6F30E01E1B706546EAB0B88B6005BE8E175198E74C3E53F40",
    "connected_nodes": [
      "BADFADAD0BEFEEDC0C0ADEADBEEFC0FFEEFACADE",
      "CEFEEDBADFADAD0C0CEEFACADE0ADEADBEEFC0FF"
    ],
    "height": 2,
    "valset_hash": "74F2AC2B6622504D08DD2509E28CE731985CFE4D133C9DB0CB85763EDCA95AA3"
  },
  "id": 1
}
```

Unfortunately, it does not seem to update and always remains the same height?! But at least the height is always on par with the other RPC endpoint :-D  @romac is currently investigating.

liamsi: Just noting that there seems to be a bug with the light store. it doesn't seem to update:
```rust
pub fn trust_block(&mut self, lb: &LightBlock) {
        self.state.light_store.update(lb, Status::Trusted);
        dbg!(lb.height(), self.latest_trusted().unwrap().height());
}
```

OStevan: Maybe changing the `latest_trusted` to something like @liamsi @romac 
``` 
/// in SledStore
fn latest(&self, status: Status) -> Option<LightBlock> {
    self.db(status)
        .iter(&self.db)
        .max_by(|first, second| first.height().cmp(&second.height()))
}
/// in MemoryStore
fn latest(&self, status: Status) -> Option<LightBlock> {
    self.store
        .iter()
        .filter(|(_, e)| e.status == status)
        .max_by_key(|(&height, _)| height)
        .map(|(_, e)| e.light_block.clone())
}
```
This is not good performance for sure but might be worth trying to see if the nature previous usage of iterators was wrong. Since I don;t think the sled one forces ordering on inserts.

liamsi: Thanks!

liamsi: At least that does something:
```
 cargo run -- start                                                                                                                      15:32:09
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `[...]/tendermint-rs/target/debug/light-node start`
started RPC server: 127.0.0.1:8888
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
error: sync failed: no witness left
^C^J                                                     
```

liamsi: Actually, that fixes the bug. I had to clean up the previous dangling light store. @OStevan you're a hero  👍  😄

CHANGES.md
```diff
@@ -17,10 +17,14 @@
 - Correctly handle blocks marked `Trusted` in accordance with the specification ([#407])
 - Treat `Trusted` status as a special case of `Verified` as per the spec ([#419])
 - Add integration test ([#431])
+- Reworked light-node to `Supervisor` / `Handle` ([#430])
```
liamsi, line 20:
```suggestion
- Rework light-node CLI to use `Supervisor` / `Handle` ([#430])
```

liamsi:
> The issue I see is that we make the entire LatestStatus optional.

I think the reason is exactly the same as why the endpoint you created is optional too but yeah, you are right in this context it is important to return any information if it is available. Do you think that all these fields should basically be optional then?

light-client/src/supervisor.rs
```diff
@@ -64,7 +64,7 @@ pub struct Instance {
 }
 
 impl Instance {
-    /// Constructs a new instance from the given light client and its state.
+    /// Constructs a new instance from the givlight-client/src/supervisor.rs:399:42en light client and its state.
```
liamsi, line 67:
```suggestion
    /// Constructs a new instance from the given light client and its state.
```

light-client/src/supervisor.rs
```diff
@@ -57,7 +64,7 @@ pub struct Instance {
...
```
liamsi, line 67:
```suggestion
    /// Constructs a new instance from the given light client and its state.
```

romac:
Looks great! :+1:

Just a quick note on the new `LightStore::latest` method: `self.db(status).iter(&self.db).max_by` does a full linear scan of the database, so this might perform very badly as the light store grows. Not a blocker for this PR, but just something to keep in mind for when we revamp the LightStore. Will add a reference to this comment to #428.

Issue #303, opened 2020-06-05
Adopt Property Based Testing


ebuchman:
We want to produce an example of using property based testing in our code.

This issue will be complete when a single example is completed. This can be as simple as a spike.

Tony mentioned [`proptest`](https://altsysrq.github.io/proptest-book/intro.html) in - https://github.com/informalsystems/tendermint-rs/issues/261#issuecomment-634352985 and https://github.com/informalsystems/tendermint-rs/pull/248#issuecomment-624645263

tony-iqlusion:
Some alternatives to consider:

- [quickcheck](https://github.com/BurntSushi/quickcheck) - original Rust port of QuickCheck
- [cargo-crux](https://github.com/GaloisInc/mir-verifier) - symbolic execution of Rust MIR code inside the [yices](https://yices.csl.sri.com/) SMT solver, with the ability to write tests in Rust. Requires an awful lot of RAM for anything remotely complicated.

All that said, I use `proptest`.

ebuchman:
@thanethomson want to look into this? We've had a lot of serialization issues like #451 from insufficient testing ...

thanethomson:
@ebuchman sure thing! Added it to the backlog 👍

ebuchman:
Interesting emerging relevant work from Google: https://alastairreid.github.io/why-not-both/

konnov:
Nice that Google joins the race. Microsoft Research has been generating unit tests with model checking for ages: https://www.microsoft.com/en-us/research/publication/pex-white-box-test-generation-for-net/

tarcieri:
The Project Oak approach is quite a bit different from "generating unit tests with model checking".

It provides a common API for property-based testing and symbolic execution systems, which is nice because those systems have a common developer-facing API (test oracle versus assertions) but different tradeoffs (exhaustiveness vs system resource requirements).

shonfeder:
I talked with @thanethomson and @ebuchman about picking up this issue. Thane and I are going to sync next week, and I've reached out to Romain to get some pointers from him as well.

thanethomson:
I'd say that we now have a few examples of PBT in the codebase and it certainly adds some additional assurances to the correctness of the code it's testing, so subsequent issues regarding PBT can focus on more specific areas to be tested and this one can be closed.

PR #451, opened 2020-07-17
fix deserialize power field in validator_update

yihuang:
When removing validator, the power field is set to zero and ignored in json format.

### DIFF
--- a/tendermint/src/validator.rs
+++ b/tendermint/src/validator.rs
@@ -192,6 +192,7 @@ pub struct Update {
     pub pub_key: PublicKey,
 
     /// New voting power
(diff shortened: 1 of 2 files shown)


tendermint/src/validator.rs
```diff
@@ -192,6 +192,7 @@ pub struct Update {
     pub pub_key: PublicKey,
 
     /// New voting power
+    #[serde(default)]
```
liamsi, line 195: Ouch, thanks! I'm wondering if there shouldn't be two test cases (one with power and one without) instead of one. This is all so brittle :-/ We do not test all edge cases well. ref #303

PR #453, opened 2020-07-17
light-client: minor cleanup in light-node crate

liamsi:
- delete unused store module 
- remove `allow(missing_docs)` for clippy 
- add general comment about light-node


light-node/src/lib.rs
```diff
@@ -23,4 +21,3 @@ pub mod error;
 pub mod prelude;
 pub mod requester;
 pub mod rpc;
-pub mod store;
```
liamsi, line 26: I forgot to delete this ...

Issue #313, opened 2020-06-09
Event subscription ADR and refactor event_listener module


liamsi:
https://github.com/informalsystems/tendermint-rs/blob/71d7e238a0fb4b5d7e9902875f09a0cf10ac4746/tendermint/src/rpc/event_listener.rs#L70-L71

https://github.com/informalsystems/tendermint-rs/blob/71d7e238a0fb4b5d7e9902875f09a0cf10ac4746/tendermint/src/rpc/event_listener.rs#L112-L115

https://github.com/informalsystems/tendermint-rs/blob/71d7e238a0fb4b5d7e9902875f09a0cf10ac4746/tendermint/src/rpc/event_listener.rs#L196-L197

Subscribe should return a stream of events tied to the particular query. Pretty much what the go-code returns on subscribe but "rust style". E.g. see: https://docs.tendermint.com/master/rpc/#/Websocket/subscribe

Furthermore, the current code only supports two events: 
https://github.com/informalsystems/tendermint-rs/blob/cc6a4cf5f241d14069972396927dd374d12af8cb/tendermint/src/rpc/event_listener.rs#L33-L34
While tendermint actually has more event types:
https://github.com/tendermint/tendermint/blob/6961c7e5d1f2a187bc5f77ac0056ce5ec5bf7ec6/types/events.go#L21-L24

also https://github.com/informalsystems/tendermint-rs/blob/cc6a4cf5f241d14069972396927dd374d12af8cb/tendermint/src/rpc.rs#L8
ref: #311

thanethomson:
What level of granularity of event routing to subscribers is needed here? Do we need fine-grained routing to subscribers the way the [Go code handles it](https://github.com/tendermint/tendermint/blob/2ac5a559b4c88385e173b673d1f6250ff7c55263/libs/pubsub/pubsub.go#L65), or will matching purely based on event type (`tm.event`) suffice?

liamsi:
Good question. I think ideally we would replicate the (fine-grained) go behaviour here. It just makes the most sense to me: a dev coming from tendermint go will immediately be familiar with the rust version and it's likely that you want to build (complex) queries and only receive the events for those for instance. 

On the other hand the current simpler approach was successfully used in GoZ and as far as I understand the main use case will be specific events in ibc. So if mimicking the go behaviour overcomplicates this we could go with a simpler variant as long as all use cases in ibc can be handled with it.

xla:
[...]

PR #458, opened 2020-07-17
rpc: handle ping messages in WebSocket listener

### LINKED ISSUE #311: event_listener Web socket error
IBC relayer event monitor, subscribed to Block and Tx events on gaia node, sees this error every ~27sec. It is issued from https://github.com/informalsystems/tendermint-rs/blob/bcef28b55a83dc3917aeedc8acf6429108aead5c/tendermint/src/rpc/event_listener.rs#L105
```
[/Users/ancaz/rust/tendermint-rs/tendermint/src/rpc/event_listener.rs:96] "We did not receive a valid JSONRPC wrapped ResultEvent!" = "We did not receive a valid JSONRPC wrapped ResultEvent!"
Jun 09 12:48:20.007 DEBUG relayer::event_monitor: Web socket error: Error (code: 0): received neither event nor generic string message (code: -1)`
```

thanethomson:
Closes #311.

Adds responding to WebSockets ping messages during the attempt to fetch the next event from the event stream.

I debated whether or not to wrap the fetching of the next message in a finite or an infinite loop, but wrapping it in a finite loop would just push that same responsibility further up the stack. Using this approach allows us to do away with the optionality of a `ResultEvent` and simplifies the method signature.

Also, I added in a `close` method for the listener to allow one to gracefully close the connection (avoids those ugly `websocket: close 1006 (abnormal closure): unexpected EOF` messages in the Tendermint logs).

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/rpc/src/client/event_listener.rs
+++ b/rpc/src/client/event_listener.rs
@@ -119,11 +119,11 @@ impl EventListener {
                     self.socket
                         .send(Message::Pong(Vec::new()))
                         .await
-                        .or_else(|e| {
-                            Err(RPCError::websocket_error(format!(
+                        .map_err(|e| {
+                            RPCError::websocket_error(format!(
                                 "failed to send pong response: {}",
                                 e
-                            )))
+                            ))
                         })?
                 }
                 Message::Pong(_) => (),
@@ -147,11 +147,8 @@ impl EventListener {
                 reason: Cow::from("client closed connection"),
             }))
             .await
-            .or_else(|e| {
-                Err(RPCError::websocket_error(format!(
-                    "failed to close web socket connection: {}",
-                    e
-                )))
+            .map_err(|e| {
+                RPCError::websocket_error(format!("failed to close web socket connection: {}", e))
             })?;
         // try to gracefully close the connection
         match self.socket.next().await {
@@ -161,7 +158,7 @@ impl EventListener {
                 Err(e) => match e {
                     // this is what we want
                     tungsteniteError::ConnectionClosed | tungsteniteError::AlreadyClosed => Ok(()),
-                    _ => return Err(RPCError::websocket_error(e.to_string())),
+                    _ => Err(RPCError::websocket_error(e.to_string())),
                 },
             },
             None => Ok(()),
--- a/rpc/src/client/event_listener.rs
+++ b/rpc/src/client/event_listener.rs
@@ -155,7 +155,6 @@ impl EventListener {
                 // we didn't get the connection closed message we wanted, so force connection closure
                 Ok(_) => Ok(()),
                 Err(e) => match e {
-                    // this is what we want
                     tungsteniteError::ConnectionClosed | tungsteniteError::AlreadyClosed => Ok(()),
                     _ => Err(RPCError::websocket_error(e.to_string())),
                 },
--- a/rpc/src/client/event_listener.rs
+++ b/rpc/src/client/event_listener.rs
@@ -150,7 +150,6 @@ impl EventListener {
             .map_err(|e| {
                 RPCError::websocket_error(format!("failed to close web socket connection: {}", e))
             })?;
-        // try to gracefully close the connection
         match self.socket.next().await {
             Some(r) => match r {
                 // we didn't get the connection closed message we wanted, so force connection closure
--- a/rpc/src/client/event_listener.rs
+++ b/rpc/src/client/event_listener.rs
@@ -3,9 +3,14 @@
 // TODO(ismail): document fields or re-use the abci types
 #![allow(missing_docs)]
 
-use async_tungstenite::{tokio::connect_async, tokio::TokioAdapter, tungstenite::Message};
+use async_tungstenite::{
+    tokio::connect_async, tokio::TokioAdapter, tungstenite::protocol::frame::coding::CloseCode,
+    tungstenite::protocol::CloseFrame, tungstenite::Error as tungsteniteError,
+    tungstenite::Message,
+};
 use futures::prelude::*;
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::error::Error as stdError;
 use tokio::net::TcpStream;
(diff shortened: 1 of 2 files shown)


rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,84 @@ impl EventListener {
...
+                        .or_else(|e| {
+                            Err(RPCError::websocket_error(format!(
+                                "failed to send pong response: {}",
+                                e
+                            )))
+                        })?
+                }
+                Message::Pong(_) => (),
+                Message::Close(_) => return Err(RPCError::websocket_error("web socket closed")),
+                Message::Binary(_) => {
```
thanethomson, line 131: Are we expecting binary messages from the WebSocket endpoint?

liamsi: Not that I'm aware of. Everything should be JSON (text) or these Ping msgs.

rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,81 @@ impl EventListener {
...
+                                    "failed to decode incoming message as an event: {}",
+                                    e
+                                )),
+                            ))
+                        }
+                    }
+                }
+                Message::Ping(_) => {
+                    self.socket
+                        .send(Message::Pong(Vec::new()))
```
liamsi, line 120: Did you try that against a running tendermint node? Not sure any of these will sent in the integration test below 🤔

rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,81 @@ impl EventListener {
...
-    pub async fn get_event(&mut self) -> Result<Option<ResultEvent>, RPCError> {
-        let msg = self
-            .socket
-            .next()
-            .await
-            .ok_or_else(|| RPCError::websocket_error("web socket closed"))??;
+    /// Get the next event from the websocket. Automatically handles websocket
+    /// protocol details, like responding to ping messages, so it can either
+    /// produce events or errors.
+    pub async fn get_event(&mut self) -> Result<ResultEvent, RPCError> {
```
liamsi, line 94: I think it would be much nicer if the `subscribe` would actually return events (a stream of events) for that particular subscription. Then, the even-listener would also run a "infinite" loop where it would capture control messages but it would route the only the relevant events to the particular subscriptions. 
But that could probably be done in a separate PR.

rpc/src/client/event_listener.rs
thanethomson: Seems like we shouldn't be expecting any binary messages: https://github.com/tendermint/tendermint/blob/master/rpc/jsonrpc/server/ws_handler.go#L395

rpc/src/client/event_listener.rs
thanethomson: I did run it against a Tendermint node, and the pings did come through (I used `dbg!` statements to show how we send pong responses, but subsequently removed the `dbg!` statements to clean things up a little).

We definitely need more extensive testing here, but I'm still debating exactly how to go about doing that. Will definitely be looking into more extensive testing in a future PR.

rpc/src/client/event_listener.rs
thanethomson: Yes, that sounds like a much more intuitive developer interface. I'll make a note of that to work on that in a future PR 👍

thanethomson: Just came to me now: this sounds like what you've recommended on #313, right?

rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,81 @@ impl EventListener {
...
+                .await
+                .ok_or_else(|| RPCError::websocket_error("web socket closed"))??;
 
-        if let Ok(result_event) = serde_json::from_str::<WrappedResultEvent>(&msg.to_string()) {
-            // if we get an rpc error here, we will bubble it up:
-            return Ok(Some(result_event.into_result()?));
+            match msg {
+                Message::Text(msg_text) => {
+                    match serde_json::from_str::<WrappedResultEvent>(msg_text.as_str()) {
+                        // if we get an rpc error here, we will bubble it up:
```
xla, line 105: This comment doesn't add much to understand the flow of the program.

rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,81 @@ impl EventListener {
...
+            .socket
+            .close(Some(CloseFrame {
+                code: CloseCode::Normal,
+                reason: Cow::from("client closed connection"),
+            }))
+            .await
+            .map_err(|e| {
+                RPCError::websocket_error(format!("failed to close web socket connection: {}", e))
+            })?;
+        // try to gracefully close the connection
```
xla, line 153:
```suggestion
```

Would remove this one as well. The code is quite self-evident. :>

rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,81 @@ impl EventListener {
...
+        match self.socket.next().await {
+            Some(r) => match r {
+                // we didn't get the connection closed message we wanted, so force connection closure
+                Ok(_) => Ok(()),
+                Err(e) => match e {
+                    // this is what we want
```
xla, line 159:
```suggestion
```

rpc/src/client/event_listener.rs
xla: The issue with this design is, that it hinges on the caller to continuously drive the websocket event machinery. Likely expected to be run in a loop. Which can lead to subtle bugs, when we imagine the caller not calling `get_event` in time and the required `Pong` message to keep the connection alive is never fired. For a user to understand why the connection was closed will be hard.

IMHO we need to rework this so the event loop is running continuously and acts more like a stream. We can still build a synchronous API on top of it, if that is desired.

rpc/src/client/event_listener.rs
```diff
@@ -83,33 +88,81 @@ impl EventListener {
...
+            let msg = self
+                .socket
+                .next()
+                .await
+                .ok_or_else(|| RPCError::websocket_error("web socket closed"))??;
 
-        if let Ok(result_event) = serde_json::from_str::<WrappedResultEvent>(&msg.to_string()) {
-            // if we get an rpc error here, we will bubble it up:
-            return Ok(Some(result_event.into_result()?));
+            match msg {
```
xla, line 102: This match block could benefit from the different arms be in their own methods. Not a huge thing tho.

thanethomson:
Totally agree on the need to restructure the design here 👍 

Since nothing else in this codebase is using the `EventListener` at the moment, merging this PR probably wouldn't really improve things in a meaningful way right now. It's probably a good idea to rather close this PR and focus on #313, where the first step would be to add an ADR for a subscriptions architecture, and then implement it using some of the WebSockets mechanics added in this PR to address #311.

I'll start working on an ADR today!

Issue #416, opened 2020-07-08
Type Generelization


brapse:
_This issue is a Refactor of #387 with the intention of being more actionable_

## Problem
The core types defined in the tendermint crate and used in the light-client crate are broad, complex and subject to change between versions of tendermint. This makes them difficult to generate during tests and cumbersome to integrate against. We want looser coupling between functions and types allowing our core logic to be robust against minor changes in the serialized form of data structures.

## Goals
* Reduce the size of the dependency graph
* Have a clear mocking & testing strategy for core data types
* Have a clear migration path for tendermint core data structure changes

##  Ideal
* All of our data types are generic enough to be invariant to tendermint core data structure changes 

## Approach
This task is complex and should be broken down into several steps. It is likely that we will have to compromise between generality and developer economics. The first step is to produce an ADR outlining a potential strategy to navigating such a tradeoff.

ebuchman:
I think we pretty much achieved all this with the traits in the original version, which are documented in [ADR-02](https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-003-light-client-core-verification.md#decision). Maybe would be helpful to first document the ergonomic challenges with that version, and if/how they might be overcome within the refactored structure.

PR #460, opened 2020-07-17
light-client: make generic over LightBlock type

romac:
See #416 for background information.

This PR is a first rough draft meant to trigger discussion around the exact dependency graph and API we want to expose when we abstract over the concrete `LightBlock` type.

* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/types.rs
+++ b/light-client/src/types.rs
@@ -1,17 +1,12 @@
 //! Defines or just re-exports the main datatypes used by the light client.
 
+use std::fmt::Debug;
+
 use derive_more::Display;
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use tendermint::{
-    account::Id as TMAccountId,
-    block::{
-        header::Header as TMHeader, signed_header::SignedHeader as TMSignedHeader,
-        Commit as TMCommit,
-    },
-    lite::TrustThresholdFraction,
-    validator::Info as TMValidatorInfo,
-    validator::Set as TMValidatorSet,
+    account::Id as TMAccountId, evidence::Evidence as TMEvidence, lite::TrustThresholdFraction,
 };
 
 pub use tendermint::{hash::Hash, lite::Height, time::Time};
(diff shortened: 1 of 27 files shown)


light-client/src/types.rs
```diff
@@ -74,46 +66,115 @@ impl Status {
...
+    fn provider(&self) -> PeerId;
+}
+
+impl Commit for TMCommit {
+    fn block_hash(&self) -> Hash {
+        self.block_id.hash
+    }
+}
+
+impl LightBlock for TMLightBlock {
```
brapse, line 98: Definitely looks like a good first step. I think what we are looking for is to simplify the object graph to the point where it's easy to test and be general over changes to the underlying data structure. I think what might be good here is actually look not just at the source of the dependency graph (TMLightBlock) but the dependencies. There are so many uses of the LightBlock that are related to internal validation, for instance insuring the validator set hashes actually match the validator set. It these cases it seems like we could simplify the dependency graph and uses by having a separate type, ValidLightBlock which provides these guarantees at construction. 

I ask raise this here as it isn't entirely clear to me how much of the underlying data structures we need to expose here. Do we need the entire `signed_header` or just properties of it? I know this goes against my argument that motivated the `operations` I'm just wondering if it's not worthwhile to revisit.

romac:
Yeah those concerns are valid, and exactly the type of discussion this PR is meant to raise. Once we settle on the right API, we should probably codify it in an ADR before altering the code.

light-client/src/types.rs
romac: >  I think what might be good here is to look not just at the source of the dependency graph (TMLightBlock) but the uses. 

Yeah that's actually what I did, I only added methods to the `LightBlock` trait when they were required at use site. But that doesn't mean that this interface is as minimal as it could be, as the code at use site was written with the concrete `TMLightBlock` in mind.

> There are so many uses the LightBlock that are related to internal validation, for instance ensuring the validator set hashes actually match the validator set. In these cases it may be simpler have a separate type, ValidLightBlock which provides these guarantees at construction.

Interesting idea, I definitely want to explore encoding more invariants in the types themselves.

> I raise this here as it isn't entirely clear to me how much of the underlying data structures we need to expose. Do we need the entire signed_header or just properties of it?

Good question! For `SignedHeader`, I don't believe it's needed anymore, as it was only used to implement header hashing, voting power calculation and commit validation via the `Commit` trait, which has been split up in separate traits.

romac: Actually, `SignedHeader` is currently also used in the [`ConflictingHeadersEvidence` struct](https://github.com/informalsystems/tendermint-rs/blob/2112202e79a77ded10114733138675cceb60cc60/tendermint/src/evidence.rs#L39-L44), but could easily be removed by introducing a new private struct to the evidence module with the same definition.

xla:
Should we drop this in favour of an ADR? It's been stale for a bit and we should exercise our discipline when it comes to design first. What you think @brapse @romac?

xla:
> Do we want to leverage an intermediate state like this or wait to further simplify the object graph?

The branch diverged quite a bit I wonder how much overhead it would be to get up to speed with mainline. Maybe a better strategy is to jump into the ADR, extract the code from the PR and use it as part of the proposed design. @romac If you pressed on time I can take this over, including speccing.

romac:
> @romac If you pressed on time I can take this over, including speccing.

Yes, I am going to be busy this week with the IBC handler ADR, so if you want to take a stab at it, that'd be great! :)

brapse:
This is a great first step but will better serve as inspiration. Expect the next iteration as a new branch from @xla

Issue #393, opened 2020-06-29
Generator of Tendermint types for unit, integration, and model-based testing


andrey-kuprianov:
As it is documented already in many places, we need to develop a __simple__ generator for complex tendermint-rs types.

Such a generator can be used for multiple purposes:
* generate complex types for unit tests ([#381](https://github.com/informalsystems/tendermint-rs/issues/381), [comment on #330](https://github.com/informalsystems/tendermint-rs/issues/330#issuecomment-645868104), [#383](https://github.com/informalsystems/tendermint-rs/pull/383))
* generate input for integration tests ([#331](https://github.com/informalsystems/tendermint-rs/issues/331), [290](https://github.com/informalsystems/tendermint-rs/issues/290))
* in model-based testing (MBT), transform an abstract counterexample from a model checker (e.g. [Apalache](https://github.com/informalsystems/apalache)) into a concrete integration test 
* may be for isolation purposes, like to isolate LIghtClient types from complete Tendermint types ([#387](https://github.com/informalsystems/tendermint-rs/issues/387))

The generator should have the following characteristics for the above applications:
* the generated object graphs should have correct relationships between objects (e.g., a header and a commit for it)  
* have a programmer-friendly Rust API allowing to generate types with minimal input. The API should allow both to use defaults for many parameters, as well as to set them explicitely.
* for MBT, it should provide a CLI, allowing to access it, obeying the above principles from other languages / frameworks. 

Here I describe the proposed solution, which is currently being implemented in [tendermint-testgen](https://github.com/informalsystems/tendermint-rs/blob/andrey/mbt-utils/tendermint-testgen/) 

## Rust API

Each tendermint type that need to be generated receives a companion type, which lists all possible options for generating this type.
Here is an example `Validator` struct, a companion for `tendermint::validator::Info`:

```rust
pub struct Validator {
  id: Option<String>,
  voting_power: Option<u64>,
  proposer_priority: Option<i64>,
}
```

Such validator companion can be generated then using the builder pattern as follows:

```rust
let val = Validator::new("a").voting_power(10).proposer_priority(100);
```

Only the parameters of the `new()` constructor are necessary, everything else is optional and can be omitted. In the case of a validator, it can be generated from a simple string identifier
Similarly, for a header the only required parameter is a set of validators (e.g. `time` then defaults to `now()`), and for a commit only requires a header.

Thus, the following code could then generate a header and commit companions from the two validators _a_ and _b_:

```rust
[...]

Issue #330, opened 2020-06-16
[light-client] Verification Testing


brapse:
part of #320

The verifier is composed of multiple predicates, each of which should be tested for both the positive and negative case using concrete types. operations should be mocked out.

* validator_sets_match
* next_validators_match
* header_matches_commit
* valid_commit
* is_within_trust_period
* is_monotonic_bft_time
* is_monotonic_height
* has_sufficient_voting_power (related #235)
* has_sufficient_validators_overlap
* has_sufficient_signers_overlap
* valid_next_validator_set

Setup a simple test harness which facilitates to construction of concrete types to limit the dependency graph and make room for more general constructions.

brapse:
So this as it turns out is trickier than it seems. In order to effectively exercise the predicates mentioned above we need to actually construct concrete types which are very complex. One approach would be to provide factories for these concrete types which extend the fixtures provides in the json tests but allow overriding specific fields. This way the tests can focus on specific field interactions without getting lost in details.

PR #468, opened 2020-07-21
Generator of Tendermint types for unit, integration, and model-based testing

andrey-kuprianov:
Here we introduce both the code-level helpers for generating tendermint object hierarchies in tests, as well as the command-line generator of those hierarchies.

For the high-level explanation please see #393.

This is still a WIP, but as it is already usable it probably makes sense to merge now, and then extend in separate PRs/issues.

Will be grateful for your reviews!

* [X] Wrote tests
* [ ] Updated CHANGES.md


tendermint-testgen/src/helpers.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub fn choose_or<T>(input: Option<T>, default: T) -> T {
+    if let Some(x) = input {
+        x
+    } else {
+        default
+    }
+}
+
+pub fn choose_from<T: Clone>(cli: &Option<T>, input: &Option<T>) -> Option<T> {
```
shonfeder, line 58: iiuc, this just [Option::or](https://doc.rust-lang.org/std/option/enum.Option.html#method.or) with a clone at the end? So I wonder if you might just use `foo.or(bar).map(|x| x.clone)`, or at least implement this in terms of that? Personall, id be in favor of the first option, since it's nice for readers to be able to think in terms of standard library functions when possible, rather than having to lookup idiosyncratic, shallow wrappers around them.

If the latter, then this function should perhaps be renamed to `option_or_clone` or something, so it's clear how this is related to `Option::or`, with the addition of a clone.

tendermint-testgen/src/helpers.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Tries to parse STDIN as the given type; otherwise returns the input wrapped in SimpleError
+pub fn parse_stdin_as<T: DeserializeOwned>() -> Result<T, SimpleError> {
+    let mut buffer = String::new();
+    match io::stdin().read_to_string(&mut buffer) {
+        Err(_) => Err(SimpleError::new("")),
+        Ok(_) => parse_as::<T>(&buffer),
+    }
+}
+
+pub fn choose_or<T>(input: Option<T>, default: T) -> T {
```
shonfeder, line 50: I think this is identical to https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or, and that should be used instead.

tendermint-testgen/src/helpers.rs
```diff
@@ -0,0 +1,83 @@
+//! Helper functions
+
+use serde::de::DeserializeOwned;
+use simple_error::*;
+use std::io::{self, Read};
+use tendermint::{vote, signature::Signature, amino_types};
+use signatory_dalek::Ed25519Verifier;
+use signatory::signature::Verifier;
+
+#[macro_export]
```
shonfeder, line 10: Some documentation for tihs would be helpful.

tendermint-testgen/src/commit.rs
```diff
@@ -0,0 +1,159 @@
...
+            .enumerate()
+            .map(val_to_vote)
+            .collect();
+        self.votes = Some(votes);
+        self
+    }
+
+    // get a mutable reference to the vote of the given validator
+    // this function will panic if the votes or the validator vote is not present
+    pub fn vote_of(&mut self, val: &str) -> &mut Vote {
```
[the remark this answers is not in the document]
brapse: perhaps `vote_of_validator` is a better naming.

brapse:
Went through this at a high level so excuse the lack of detail but in general I would say this is pretty great.

Besides the thoughtful refinements mentioned by @shonfeder I would look forward to this landing sooner rather than later. As a validation and natural next step I would look to validate the capabilities and abstractions by making progress on #330. Perhaps adding a generator for light blocks might help there.

tendermint-testgen/bin/tendermint-testgen.rs
```diff
@@ -0,0 +1,129 @@
...
+
+fn encode_with_stdin<Opts: Generator<T> + Options, T: serde::Serialize>(cli: &Opts) -> Result<String, SimpleError>
+{
+    let stdin = read_stdin()?;
+    let default = Opts::from_str(&stdin)?;
+    let producer = cli.merge_with_default(&default);
+    producer.encode()
+}
+
+fn run_command<Opts: Generator<T> + Options, T: serde::Serialize>(cli: Opts, read_stdin: bool) {
```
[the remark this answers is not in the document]
andrey-kuprianov: well, I also considered using `Abscissa` first, but then I saw that `light-client/examples/light-client.rs` uses `Gumdrop`, and I liked it as more easy to use, at least for me, at my present state of knowledge. It doesn't introduce almost any additional overhead, and please also consider that it allows me to employ the same structs both for Rust API, and for CLI command options, which I think is very neat.

Let's leave it as is for now, and we'll have a chance to reconsider later.

tendermint-testgen/bin/tendermint-testgen.rs
```diff
@@ -0,0 +1,129 @@
...
+                .map(|s| s.split_whitespace().next().unwrap())
+            {
+                eprintln!("\n{} parameters:", cmd);
+                print_params(CliOptions::command_usage(cmd).unwrap())
+            }
+            std::process::exit(1);
+        }
+        Some(Command::Validator(cli)) => run_command(cli, opts.stdin),
+        Some(Command::Header(cli)) => run_command(cli, opts.stdin),
+        Some(Command::Vote(cli)) => eprintln!("{:?}", cli),
```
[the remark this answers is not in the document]
andrey-kuprianov: thanks for spotting this! it is actually implemented, only this last step was missing...

tendermint-testgen/src/commit.rs
```diff
@@ -0,0 +1,159 @@
...
+        Commit {
+            header: Some(header.clone()),
+            round: None,
+            votes: None
+        }
+    }
+    set_option!(votes, Vec<Vote>);
+    set_option!(round, u64);
+
+    // generate commit votes from all validators in the header
```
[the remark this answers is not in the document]
andrey-kuprianov: Nice suggestion! Added these and other doc comments.

tendermint-testgen/src/commit.rs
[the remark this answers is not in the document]
andrey-kuprianov: thanks for the name suggestion, applied

tendermint-testgen/src/commit.rs
```diff
@@ -0,0 +1,159 @@
...
+    // this function will panic if the votes or the validator vote is not present
+    pub fn vote_of(&mut self, val: &str) -> &mut Vote {
+        self.votes.as_mut().unwrap().iter_mut().find(
+            |v| *v.validator.as_ref().unwrap() == Validator::new(val)
+            ).unwrap()
+    }
+
+    // get a mutable reference to the vote at the given index
+    // this function will panic if the votes or the vote at index is not present
+    pub fn vote_at(&mut self, index: usize) -> &mut Vote {
```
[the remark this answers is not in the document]
andrey-kuprianov: thanks for the name suggestion, applied also

tendermint-testgen/src/helpers.rs
andrey-kuprianov: very nice! I was simply unaware of that function; replaced my homegrown one with this everywhere...

tendermint-testgen/src/helpers.rs
andrey-kuprianov: thanks! same as above: was not aware of this. replaced everywhere

tendermint-testgen/src/helpers.rs
andrey-kuprianov: done

tendermint-testgen/src/header.rs
```diff
@@ -84,15 +84,15 @@ impl Generator<block::Header> for Header {
...
             Ok(id) => id,
             Err(_) => bail!("failed to construct header's chain_id")
         };
         let header = block::Header {
             version: block::header::Version { block: 0, app: 0 },
             chain_id,
-            height: block::Height(choose_or(self.height, 1)),
-            time: choose_or(self.time, Time::now()),
+            height: block::Height(self.height.unwrap_or(1)),
+            time: self.time.unwrap_or(Time::now()),
```
romac, line 100:
```suggestion
            time: self.time.unwrap_or_else(|| Time::now()),
```

tendermint-testgen/src/header.rs
```diff
@@ -84,15 +84,15 @@ impl Generator<block::Header> for Header {
             None => valset.clone(),
         };
         let  chain_id = match chain::Id::from_str(
-            choose_or(self.chain_id.clone(), "test-chain".to_string()).as_str()) {
+            self.chain_id.clone().unwrap_or("test-chain".to_string()).as_str()) {
```
romac:
```suggestion
            self.chain_id.clone().unwrap_or_else(|| "test-chain".to_string()).as_str()) {
```

tendermint-testgen/bin/tendermint-testgen.rs
```diff
@@ -0,0 +1,130 @@
...
+
+fn encode_with_stdin<Opts: Generator<T> + Options, T: serde::Serialize>(cli: &Opts) -> Result<String, SimpleError>
+{
+    let stdin = read_stdin()?;
+    let default = Opts::from_str(&stdin)?;
+    let producer = cli.clone().merge_with_default(default);
+    producer.encode()
+}
+
+fn run_command<Opts: Generator<T> + Options, T: serde::Serialize>(cli: Opts, read_stdin: bool) {
```
romac, line 79: Makes the function definition a bit more readable imho.

```suggestion
fn run_command<Opts, T>(cli: Opts, read_stdin: bool)
where
    Opts: Generator<T> + Options,
    T: serde::Serialize,
{
```

tendermint-testgen/src/vote.rs
```diff
@@ -0,0 +1,140 @@
...
+            index: None,
+            header: Some(header.clone()),
+            precommit: None,
+            height: None,
+            time: None,
+            round: None
+        }
+    }
+    set_option!(index, u64);
+    set_option!(header, &Header, Some(header.clone()));
```
romac, line 46: Usually preferable to leave it up to the callee to clone the data passed to a function. This can avoid unnecessary clones in cases where the callee already has ownership of the data passed to the function.

```suggestion
    set_option!(header, Header);
```

tendermint-testgen/src/vote.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[options(help = "commit round (default: from commit)")]
+    pub round: Option<u64>,
+}
+
+impl Vote {
+    pub fn new(validator: &Validator, header: &Header) -> Self {
+        Vote {
+            validator: Some(validator.clone()),
+            index: None,
+            header: Some(header.clone()),
```
romac, line 38: Usually preferable to leave it up to the callee to clone the data passed to a function. This can avoid unnecessary clones in cases where the callee already has ownership of the data passed to the function.

```suggestion
    pub fn new(validator: Validator, header: Header) -> Self {
        Vote {
            validator: Some(validator),
            index: None,
            header: Some(header),
```

tendermint-testgen/src/commit.rs
```diff
@@ -0,0 +1,159 @@
...
+    parse(try_from_str = "parse_as::<Vec<Vote>>"))]
+    pub votes: Option<Vec<Vote>>,
+    #[options(help = "commit round (default: 1)")]
+    pub round: Option<u64>
+}
+
+impl Commit {
+    pub fn new(header: &Header) -> Self {
+        Commit {
+            header: Some(header.clone()),
```
romac, line 24: Usually preferable to leave it up to the callee to clone the data passed to a function. This can avoid unnecessary clones in cases where the callee already has ownership of the data passed to the function.

```suggestion
    pub fn new(header: Header) -> Self {
        Commit {
            header: Some(header),
```

romac:
> Also, this needs a run of cargo fmt.

Not anymore, didn't see you pushed a commit while I was reviewing :)

tendermint-testgen/bin/tendermint-testgen.rs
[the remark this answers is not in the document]
romac: Note: `abscissa` uses `gumdrop` as well, but comes with more bells and whistle which I agree might not be needed for this specific project. As you say, we can still consider moving to `abscissa` a later point if the CLI grows or undergoes substantial changes.

tendermint-testgen/src/commit.rs
andrey-kuprianov: thanks:) yes; ownership and borrowing is one of the rusty things I still struggle with...

romac:
Just want to emphasize how good the CLI interface is by the way:

```
    tendermint-testgen validator --id a --voting-power 3
     echo -n '{"id": "a", "voting_power": 3}' | tendermint-testgen --stdin validator
     echo -n a | tendermint-testgen --stdin validator --voting-power 3
     echo -n '{"id": "a"}' | tendermint-testgen --stdin validator --voting-power 3
     echo -n '{"id": "a", "voting_power": 100}' | tendermint-testgen --stdin validator --voting-power 3
```

😍

tendermint-testgen/src/helpers.rs
```diff
@@ -0,0 +1,64 @@
...
+    let mut buffer = String::new();
+    try_with!(io::stdin().read_to_string(&mut buffer), "");
+    Ok(buffer)
+}
+
+/// Tries to parse STDIN as the given type; otherwise returns the input wrapped in SimpleError
+pub fn parse_stdin_as<T: DeserializeOwned>() -> Result<T, SimpleError> {
+    let mut buffer = String::new();
+    match io::stdin().read_to_string(&mut buffer) {
+        Err(_) => Err(SimpleError::new("")),
```
romac, line 45: It would be nice to keep the underlying error around, even if just as a string, to ease debugging a bit perhaps.

tendermint-testgen/src/helpers.rs
```diff
@@ -0,0 +1,64 @@
...
+pub fn parse_as<T: DeserializeOwned>(input: &str) -> Result<T, SimpleError> {
+    match serde_json::from_str(input) {
+        Ok(res) => Ok(res),
+        Err(_) => Err(SimpleError::new(input)),
+    }
+}
+
+pub fn read_stdin() -> Result<String, SimpleError> {
+    let mut buffer = String::new();
+    try_with!(io::stdin().read_to_string(&mut buffer), "");
```
romac, line 37: It would be nice to keep the underlying error around, even if just as a string, to ease debugging a bit perhaps.

tendermint-testgen/src/helpers.rs
romac: Also `read_stdin` and `parse_stdin_as` could perhaps be rewritten in the same style, ie. with `try_with!` or with a `match`.

tendermint-testgen/src/header.rs
```diff
@@ -0,0 +1,171 @@
...
+            version: block::header::Version { block: 0, app: 0 },
+            chain_id,
+            height: block::Height(self.height.unwrap_or(1)),
+            time: self.time.unwrap_or(Time::now()),
+            last_block_id: None,
+            last_commit_hash: None,
+            data_hash: None,
+            validators_hash: valset.hash(),
+            next_validators_hash: next_valset.hash(),
+            consensus_hash: valset.hash(), // TODO: currently not clear how to produce a valid hash
```
romac, line 106: @andrey-kuprianov Could you expand on this a bit?

tendermint-testgen/src/vote.rs
```diff
@@ -0,0 +1,165 @@
...
+        help = "validator of this vote (required; can be passed via STDIN)",
+        parse(try_from_str = "parse_as::<Validator>")
+    )]
+    pub validator: Option<Validator>,
+    #[options(help = "validator index (default: from commit header)")]
+    pub index: Option<u64>,
+    #[options(help = "header to sign (default: commit header)")]
+    pub header: Option<Header>,
+    #[options(help = "vote type; 'precommit' if set, otherwise 'prevote' (default)")]
+    pub precommit: Option<()>,
```
romac, line 24: Since `gumdrop` supports boolean options, could you explain the choice of `Option<()>` instead?

tendermint-testgen/src/header.rs
andrey-kuprianov: well, it's that from my current limited understanding of the code base I don't see an obvious way on how to compute the consensus hash... If you know the way, please share it with me

tendermint-testgen/src/helpers.rs
andrey-kuprianov: yes, sure, you are right

tendermint-testgen/src/helpers.rs
andrey-kuprianov: `parse_stdin_as` was not used anymore actually; so I removed it; and `read_stdin` I've rewritten using `match`

tendermint-testgen/src/vote.rs
andrey-kuprianov: The alternatives are:
* using plain `bool`: doesn't fit well with other fields being options, so contradicts the overall architecture.
* using `Option<bool>` doesn't work either, because then `gumdrop` expects the input of the form `--precommit true` or `--precommit false`.

Using `Option<()>` on the other hand, allows in the CLI just to write `--precommit`.

tendermint-testgen/bin/tendermint-testgen.rs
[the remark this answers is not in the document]
shonfeder: I was assuming Abscissa would handle stuff like reporting supported arguments and error return codes. But agree there's no reason you have to do it here and now -- or at all :)

andrey-kuprianov:
> Just want to emphasize how good the CLI interface is by the way:
> 

@romac  cool, thanks; glad you like it :) What I particularly like about `gumdrop` is that it's so transparent that I can reuse the same structs for the Rust API and for the CLI...

tendermint-testgen/src/helpers.rs
shonfeder: Thanks!

tendermint-testgen/src/header.rs
andrey-kuprianov: yes, right:) done

andrey-kuprianov:
@shonfeder @romac @brapse I think I've addressed all the issues raised in your reviews; thank you, that was very helpful, and allowed to improve the code a lot:)

tendermint-testgen/src/header.rs
liamsi: The consensus hash is the hash of the consensus params: https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/src/consensus/params.rs

More precisely only Block.MaxBytes and Block.MaxGas are included in the message that gets hashed:
https://github.com/tendermint/tendermint/blob/940e68292cd2e710525c35df08d7e007a7c04f9d/types/params.go#L147-L166

(tmhash is just sha256)

tendermint-testgen/src/commit.rs
```diff
@@ -0,0 +1,190 @@
...
+            round: None,
+            votes: None,
+        }
+    }
+    set_option!(votes, Vec<Vote>);
+    set_option!(round, u64);
+
+    /// Generate commit votes from all validators in the header.
+    /// This function will panic if the header is not present
+    pub fn generate_default_votes(mut self) -> Self {
```
Shivani912, line 34: Shouldn't this be a part of `Commit::new` ? Or is there a reason we're not adding values for votes and round there?

andrey-kuprianov: I don't think so. There are two scenarios we want to cover, probably:
* all (or most) votes are present, and either there are no or very little deviation in the votes from the auto-generated default votes.  In that cause the test writer will go over the route `Commit::new()` -> `generate_default_votes()` -> `modify the votes`
* the desired votes structure differs substantially from the auto-generated one. in that case the user would go over `Commit::new()` -> `add votes manually`.

In general, as a design principle of this library, we want to leave as many parameters optional as possible, but to allow the user to set them explicitely when needed. In the cause of `round`, there is a reasonable default value, `1`, which should probably suffice in most cases, that's why we leave it optional.

Shivani912: Aha, you're right. Thanks for explaining @andrey-kuprianov !

tendermint-testgen/src/header.rs
andrey-kuprianov: @liamsi thanks for the clarification! Still, seeing how it's implemented in Go, and with my current limited knowledge of Rust and how hashing is done in Tendermint, I don't feel comfortable enough to reproduce it, and be confident that what I write is correct. Also, in the current version of the light client this is not checked at all (should it, by the way?), as far as I can say.

If someone else volunteers to write this hashing part for consensus params, I am both hands for it! :) Otherwise, I would suggest to leave it for the future update.

tendermint-testgen/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "tendermint-testgen"
+version = "0.1.0"
+authors = ["Andrey Kuprianov <andrey@informal.systems>"]
+edition = "2018"
+
+[dependencies]
+tendermint = { path = "../tendermint" }
```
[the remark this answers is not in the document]
andrey-kuprianov: yes, you are right :+1:

tendermint-testgen/src/header.rs
ebuchman: This is the hash of the consensus params which light clients currently don't pay attention to but which can change (though very infrequently) in the life of a blockchain. Best to open a follow up issue for things like this

andrey-kuprianov: That's true; created a follow up issue #480.

Cargo.toml
```diff
@@ -5,4 +5,5 @@ members = [
     "light-node",
     "rpc",
     "tendermint",
+    "tendermint-testgen"
```
romac, line 8: How about naming the folder just `testgen` for consistency with the other crates in the workspace? We would still keep `tendermint-testgen` as the crate name in `testgen/Cargo.toml`.

tendermint-testgen/src/commit.rs
andrey-kuprianov: @Shivani912 actually, you were right:) See 44c3f92.

It is a bit difficult to achieve the right balance here, because `round` and `votes` parameters of a `Commit` are not independent. Would love to hear your opinion on the new variant.

Cargo.toml
andrey-kuprianov: yes, nice idea, applied:)

andrey-kuprianov:
@brapse @romac I think I've addressed all the issues raised. For some reason Github says there are still pending changes suggested by @romac, but I don't see anything left, and don't know how to resolve this... 

If everything is OK from your side, could you please merge the PR?

andrey-kuprianov:
Btw, the failing `cargo fmt` has nothing to do with this PR...

PR #474, opened 2020-07-22
fix predicate bug

### LINKED ISSUE #473: Bug in predicate - is_within_trust_period
**Context:** `is_within_trust_period` predicate is intended to check if the latest `trusted` header is still within trusting period while we verify a new, untrusted header. 

We also need to perform another check for header time that it's "header from future", that means, the `untrusted` header that we're verifying must not have a time that is beyond our current time (= now + clock_drift)

**Problem:** Code below for the `is_within_trust_period` predicate includes the "header from future" check.

This is a bug because the header we have there is the `trusted` one instead of the `untrusted` one and so the condition will always evaluate to `true`.

https://github.com/informalsystems/tendermint-rs/blob/548d76d8b7025f1a3ad1ee9adf32a5e6d24e2197/light-client/src/predicates.rs#L99-L121

**Fix:** Will be clearer if we have a separate predicate for "header from future" check.

Shivani912:
closes #473 
* [ ] Wrote tests
* [ ] Updated CHANGES.md

### DIFF
--- a/light-client/src/predicates.rs
+++ b/light-client/src/predicates.rs
@@ -127,7 +127,6 @@ pub trait VerificationPredicates: Send {
         clock_drift: Duration,
         now: Time,
     ) -> Result<(), VerificationError> {
-
         ensure!(
             header.time < now + clock_drift,
             VerificationError::HeaderFromTheFuture {
@@ -251,11 +250,7 @@ pub fn verify(
     )?;
 
     // Ensure the header isn't from a future time
-    vp.is_header_from_past(
-        &untrusted.signed_header.header,
-        options.clock_drift,
-        now,
-    )?;
+    vp.is_header_from_past(&untrusted.signed_header.header, options.clock_drift, now)?;
 
     // Ensure the header validator hashes match the given validators
     vp.validator_sets_match(&untrusted, &*hasher)?;
--- a/light-client/src/predicates.rs
+++ b/light-client/src/predicates.rs
@@ -98,20 +98,11 @@ pub trait VerificationPredicates: Send {
     /// Check that the trusted header is within the trusting period, adjusting for clock drift.
     fn is_within_trust_period(
         &self,
-        header: &Header,
+        trusted_header: &Header,
         trusting_period: Duration,
-        clock_drift: Duration,
         now: Time,
     ) -> Result<(), VerificationError> {
-        ensure!(
-            header.time < now + clock_drift,
-            VerificationError::HeaderFromTheFuture {
-                header_time: header.time,
-                now
-            }
-        );
-
-        let expires_at = header.time + trusting_period;
+        let expires_at = trusted_header.time + trusting_period;
         ensure!(
             expires_at > now,
             VerificationError::NotWithinTrustPeriod { expires_at, now }
@@ -123,14 +114,14 @@ pub trait VerificationPredicates: Send {
     /// Check that the untrusted header is from past.
     fn is_header_from_past(
         &self,
-        header: &Header,
+        untrusted_header: &Header,
         clock_drift: Duration,
         now: Time,
     ) -> Result<(), VerificationError> {
         ensure!(
-            header.time < now + clock_drift,
+            untrusted_header.time < now + clock_drift,
             VerificationError::HeaderFromTheFuture {
-                header_time: header.time,
+                header_time: untrusted_header.time,
                 now
             }
         );
@@ -245,7 +236,6 @@ pub fn verify(
     vp.is_within_trust_period(
         &trusted.signed_header.header,
         options.trusting_period,
-        options.clock_drift,
         now,
     )?;
 
--- a/light-client/src/predicates.rs
+++ b/light-client/src/predicates.rs
@@ -95,7 +95,7 @@ pub trait VerificationPredicates: Send {
         Ok(())
     }
 
-    /// Check that the given header is within the trusting period, adjusting for clock drift.
+    /// Check that the trusted header is within the trusting period, adjusting for clock drift.
     fn is_within_trust_period(
         &self,
         header: &Header,
@@ -115,6 +115,25 @@ pub trait VerificationPredicates: Send {
         ensure!(
             expires_at > now,
             VerificationError::NotWithinTrustPeriod { expires_at, now }
+        );
+
+        Ok(())
+    }
+
+    /// Check that the untrusted header is from past.
+    fn is_header_from_past(
+        &self,
+        header: &Header,
+        clock_drift: Duration,
+        now: Time,
+    ) -> Result<(), VerificationError> {
+
+        ensure!(
+            header.time < now + clock_drift,
+            VerificationError::HeaderFromTheFuture {
+                header_time: header.time,
+                now
+            }
         );
 
         Ok(())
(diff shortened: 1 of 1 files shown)


Shivani912:
The error is from tokio library. Do we still merge?

light-client/src/predicates.rs
```diff
@@ -95,7 +95,7 @@ pub trait VerificationPredicates: Send {
         Ok(())
     }
 
-    /// Check that the given header is within the trusting period, adjusting for clock drift.
+    /// Check that the trusted header is within the trusting period, adjusting for clock drift.
     fn is_within_trust_period(
         &self,
         header: &Header,
```
liamsi, line 101: Nit: maybe we can make it more explicit that we are expecting a trusted header by renaming the param to: `trusted_header`?

light-client/src/predicates.rs
```diff
@@ -120,6 +120,24 @@ pub trait VerificationPredicates: Send {
         Ok(())
     }
 
+    /// Check that the untrusted header is from past.
+    fn is_header_from_past(
+        &self,
+        header: &Header,
```
liamsi, line 126: Nit: similar to above this could be `untrusted_header`?

light-client/src/predicates.rs
Shivani912: I think that's a great idea

light-client/src/predicates.rs
```diff
@@ -120,6 +120,24 @@ pub trait VerificationPredicates: Send {
...
+        clock_drift: Duration,
+        now: Time,
+    ) -> Result<(), VerificationError> {
+        ensure!(
+            header.time < now + clock_drift,
+            VerificationError::HeaderFromTheFuture {
+                header_time: header.time,
+                now
+            }
+        );
```
liamsi, line 136: Isn't that already ensured in `is_within_trust_period`?

light-client/src/predicates.rs
```diff
@@ -231,6 +249,9 @@ pub fn verify(
         now,
     )?;
 
+    // Ensure the header isn't from a future time
+    vp.is_header_from_past(&untrusted.signed_header.header, options.clock_drift, now)?;
```
liamsi, line 253: Great find! 👍

light-client/src/predicates.rs
Shivani912: I think we should remove it from `is_within_trust_period` because the header is already trusted. What do you think?

Shivani912:
> Left some minor (optional) renamings that could clarify wether the functions are expecting a trusted/vs untrusted header (maybe we can use the type system for that on the long-run instead). Can be addressed here or in followups.
> 
> Thanks @Shivani912 🙏 Great work!

Thanks for reviewing @liamsi ! I think doing that makes a lot of sense. @romac would be best person to comment on this as he has a better view of it than I do.

light-client/src/predicates.rs
```diff
@@ -95,28 +95,37 @@ pub trait VerificationPredicates: Send {
...
     fn is_within_trust_period(
         &self,
-        header: &Header,
+        trusted_header: &Header,
         trusting_period: Duration,
+        now: Time,
+    ) -> Result<(), VerificationError> {
+        let expires_at = trusted_header.time + trusting_period;
+        ensure!(
+            expires_at > now,
```
romac, line 107: I wonder if need to account for clock drift here as well?

romac:
Great stuff @Shivani912, I just have one question about clock drift inlined above.

light-client/src/predicates.rs
liamsi: I think it make sense to remove this check. With the new check you introduced (`is_header_from_past`), it should be guaranteed that we do not trust and store headers from the future. Hence, the check on a already trusted header seems redundant.

light-client/src/predicates.rs
liamsi: As far as I remember the spec applies the clockdrift only to untrusted headers. If we change that here we should feed that back into the spec too:
https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification/verification.md#lcv-func-valid1

Found this gem from the spec 🤣 
> If trusted.Header.Time > now - trustingPeriod the blabla

Shivani912: A benefit to not adding the `clock_drift` is that we have some extra time to account for the time taken by the next verification steps. Also this could mean that if the `header` has already expired like a second ago but because we add the `clock_drift` there, it could satisfy the condition and consider it valid.

ebuchman: I think we shouldn't need to worry about clock drift here since trusting_period is orders of magnitude greater than clock_drift but I'm not sure how best to surface that reasoning in the code (maybe just a comment). Also cc @josef-widder

light-client/src/predicates.rs
ebuchman: This is a great find. It's also a reminder that there's an underlying assumption in the spec which should perhaps be more explicitly surfaced that the light client's local clock is closely synchronized (within clock_drift) to that of the chain's BFT time (ie. currently the median timestamp from the last commit).

We might want to better motivate this assumption and its relationship with the trusting period and expired headers. Currently, if our trusted header hasn't expired and `trusting_period >> clock_drift`, this `is_header_from_past` check ensures the new header is also within the trusting period, even though we don't run `is_within_trusting_period(untrusted)` directly. But. `is_header_from_past(untrusted)` is a much stronger condition than `is_within_trusting_period(untrusted)` and it may have subtle consequences in IBC, so we may want to reconsider it (!).

In IBC, a client's "local" clock is the BFT time of that chain (ie. the median timestamp of the last commit from the validator set ...), so this check would imply that the BFT time ("on chain clock") of any two chains in IBC is within clock_drift of each other, which might be too strong (?!).

Note that BFT time in Tendermint is currently subject to unaccountable control by +1/3 validators - they can arbitrarily fast forward time.  This is somewhat counteracted by a punishment mechanism that uses both a hybrid of unbonding period (eg. ~3 weeks) AND minimum number of blocks (eg. ~30,000 blocks) to determine evidence validity, so even if validators fast forward the clock beyond the unbonding period, they can still be punished for some number of blocks. 

And of course there are plans to change how BFT time is generated (proposer-based instead of median of timestamps in commits) and that would prevent +1/3 from controlling it, but would assume +2/3 have synchronized clocks on a given chain, and would still assume synchronize clocks between chains. 

If validators fast forward the clock, they can make it difficult for light clients to keep up and act on current information (since it will look like the chain is always ahead). This may lead to certain kinds of attacks on IBC channels where validators on one chain can delay the validity of client updates on a chain its connected to (maybe there's an equivalent of Miner Extractable Value attacks on IBC here ...) . Not sure how big a problem this is, and/or how much it can be addressed by just checking `is_within_trusting_period(untrusted)` instead of `is_header_from_past(untrusted)` cc @cwgoes @josef-widder ...

ebuchman: Ok I gave it its own issue: https://github.com/informalsystems/tendermint-rs/issues/478

ebuchman:
Great catch @Shivani912 - how'd you find this? 

I think this is fine to merge but it does open a can of worms (which already existed but we may not have been paying attention to): #478

Shivani912:
Thanks for looking into this @ebuchman ! 

> Great catch @Shivani912 - how'd you find this?
> 
I wrote a test specifically for the case where verification must fail because of this: `header_time > now + clock_drift`. And upon investigating why the test failed, it wasn't too difficult to find this!

Shivani912:
Thanks everyone for your input here! Merging this now and clock synchronization issue can be followed at #478

PR #475, opened 2020-07-23
Adds more verification tests

### LINKED ISSUE #330: [light-client] Verification Testing

Shivani912:
Closes #330 - I think we should close this one now as all of the predicates are being tested by conformance tests. We can open a separate issue for the lack of unit tests for these.

* [ ] Wrote tests
* [ ] Updated CHANGES.md


romac:
I guess this depends on #474?

Shivani912:
> I guess this depends on #474?

Yes :)



> Looks great, thanks a lot @Shivani912! Let's merge this after #474

Thanks for an early review @romac ! Will do...

romac:
@Shivani912 Any idea why the `single_step_skipping` test is failing now?

Shivani912:
> @Shivani912 Any idea why the `single_step_skipping` test is failing now?

Aha, the older tests (`tendermint/tests/lite.rs`) is failing because we haven't fixed the predicate issue there. I had in mind that it's already deprecated. Are we still maintaining that code? @romac

romac:
Right, this is an old test, my bad! No we don't maintain that code anymore so I guess it's fine to just comment it out until we remove that code altogether.

Shivani912:
Thank you @romac for actively reviewing this PR, appreciate it! I have addressed all your comments and I think it's ready to merge now. Will be great if you could approve it to be able to merge. Thanks again !

romac:
Great stuff, thanks @Shivani912!

ebuchman:
Aren't all these tests being added in the old `lite` pkg? Do those files get run from the new `light-client`? It looks like they're duplicated in both the old and the new test dirs?

Also, is there some way to see the mapping between the tests we have and the predicates? Like how should someone verify that there's a positive and negative test for each predicate?

PR #476, opened 2020-07-23
rpc: ADR for client event subscription mechanism

thanethomson:
In partial fulfillment of #313 (implementation will close it).

I've gone through several mental iterations of what the various touchpoints for the event pub/sub interface could look like and I think I've pared it down to the simplest form I can come up with.

More scrutiny would be greatly appreciated! 😁 

[Rendered](https://github.com/informalsystems/tendermint-rs/blob/adr/subscription/docs/architecture/adr-008-event-subscription.md)

---

* [ ] Wrote tests
* [ ] Updated CHANGES.md


docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+## Context
+
+The [Tendermint Light Client](../../light-client/) is one of many important
+applications that will make use of the [RPC client](../../rpc/) to query
+Tendermint full nodes for information relating to the blockchain.
+
+Tendermint servers (e.g. full nodes) provide an [event
+subscription][tm-event-subs] RPC endpoint to allow clients to receive
+notifications of specific events as they happen. We need to expose this
+subscription mechanism to the RPC client.
```
liamsi, line 15: Maybe here or in the title mention that this is about the websocket client particular?

romac:
I too very much like the interface 👍 

I think it could be useful to sketch out the `Subscription` type to see how easily we can mock out the `Client` interface to construct a `Subscription` which emits a static list of events.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+## Context
+
+The [Tendermint Light Client](../../light-client/) is one of many important
+applications that will make use of the [RPC client](../../rpc/) to query
+Tendermint full nodes for information relating to the blockchain.
+
+Tendermint servers (e.g. full nodes) provide an [event
+subscription][tm-event-subs] RPC endpoint to allow clients to receive
+notifications of specific events as they happen (currently via a WebSockets
+connection). We need to expose this subscription mechanism to the RPC client.
```
xla, line 15:
```suggestion
connection). We need to expose this subscription mechanism through the RPC client.
```

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+         incoming events only relevant to that subscription (i.e. it should
+         produce an **event stream**).
+      4. It must be possible to, from outside of the RPC client, receive events
+         from multiple subscriptions' event streams concurrently without
+         interfering with/blocking each other.
+      5. It must be possible to, from outside of the RPC client, handle
+         subscriptions' event streams without blocking other RPC operations.
+   2. Offer the ability to **unsubscribe** from specific queries (i.e. to
+      terminate specific subscriptions).
+2. A pub/sub architecture (with an appropriate concurrency model) that
```
xla, line 36: This is a bit vague, could we expand on this?

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+      Ok(event) => { /* handle event */ },
+      Err(e) => { /* terminate subscription and report error */ },
+   }
+}
+```
+
+Since the underlying events may not be produced at the same cadence as that of
+the `Subscription` owner's processing, it is proposed here to make use of an
+appropriate type of buffered channel to realize this. The size of this buffer
+must be configurable through the RPC Client.
```
xla, line 74: This needs further detailing as to what the desired issue we like to address with buffered channels. Furthermore the behaviour needs specifying as to how the implementation should behave when it encounters a filled up/blocked channel.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+
+### RPC Pub/Sub
+
+The RPC Client must asynchronously spawn a component `PubSub` whose
```
xla, line 78: It ought to be the responsibility of the user/caller to spawn this "run loop". Otherwise we take away control about lifetime. Instead of having the client silently spawn things in the background it should be made explicit through its interface.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+3. Route `Event`s to `Subscription`s whose `Query` matches the `Event`.
+4. Manage transport-level concerns (i.e. WebSockets connection: keep-alive
+   messages, graceful connection closure, etc.).
+
+Items (3) and (4) could be encapsulated into their own sub-components to allow
+for better separation of concerns and therefore easier testing. For example, (3)
+could be encapsulated into a `PubSubRouter` component and (4) into a component
+that only handles transport-level concerns.
+
+This is an *internal* component whose interface only matters to the RPC Client.
```
xla, line 92: We should break this down, as this seems to be the core of the new implementation it would help to iron out the details of where concerns are separated. A design for this similar to the high-level overview further up would accompany this greatly.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+    /// Terminates the given subscription and consumes it.
+    pub async fn unsubscribe(&mut self, subs: Subscription) -> Result<(), Error> {
+       // ...
+    }
+}
+```
+
+### Query
+
+Due to the potential complexity of queries, the proposed `Query` entity should
```
xla, line 119: This premise is not advisable, rather we should explore how a querying could look like in the context of the Rust language and runtime. This is an opportunity to design an interface with strong and safe guarantees.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+functionally take its lead from the [Go-based `Query` interface][tm-go-query].
+This type of interface allows for simple short-term implementation (e.g. just
+querying by way of event type), with eventual full implementation of the [query
+PEG][query-peg].
+
+`Query` should implement `FromStr`, which should parse queries and emit the
+appropriate parsing errors:
+
+```rust
+let query = Query::from_str("tm.event='NewBlock'")?;
```
xla, line 129: As `FromStr` gives us parse we can do:

```suggestion
let query = "tm.event='NewBlock'".parse::<Query>()?;
```

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+### Query
+
+Due to the potential complexity of queries, the proposed `Query` entity should
+functionally take its lead from the [Go-based `Query` interface][tm-go-query].
+This type of interface allows for simple short-term implementation (e.g. just
+querying by way of event type), with eventual full implementation of the [query
+PEG][query-peg].
+
+`Query` should implement `FromStr`, which should parse queries and emit the
+appropriate parsing errors:
```
xla, line 126: Do we want free string input in the first iteration?

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+}
+```
+
+### Query
+
+Due to the potential complexity of queries, the proposed `Query` entity should
+functionally take its lead from the [Go-based `Query` interface][tm-go-query].
+This type of interface allows for simple short-term implementation (e.g. just
+querying by way of event type), with eventual full implementation of the [query
+PEG][query-peg].
```
xla, line 123: If this PEG "source-of-truth" exists, why aren't we going to aim for support of it from the get-go?

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,178 @@
...
+* Provides relatively intuitive developer ergonomics (`Subscription` iteration
+  to produce `Event`s).
+* Can be easily tested if the WebSockets transport layer is clearly encapsulated
+  from the perspective of the RPC Pub/Sub component.
+
+### Negative
+
+* Requires an additional concurrent, potentially long-running `async` task to be
+  concerned about.
+* Requires careful selection of `Subscription` buffer size to avoid lost events
```
xla, line 157: This could be a non-issue if we communicate clearly the guarantees of the client and/or actually have ways of tuning these.

docs/architecture/adr-008-event-subscription.md
thanethomson: Definitely. I was using last week as a spike to de-risk this, because I wasn't 100% sure how to practically implement this in Rust. I think I have a good idea along these lines, and will hopefully submit a draft PR in the next day or so with an example 👍

docs/architecture/adr-008-event-subscription.md
thanethomson: My ideal here would be to have compile-time query validation - is that what you're aiming for too?

docs/architecture/adr-008-event-subscription.md
thanethomson: I thought we could do that in a separate ADR? I'll be expanding on this ADR a little more shortly, so I thought I wouldn't muddy the waters with too much focus on the querying.

docs/architecture/adr-008-event-subscription.md
xla: I'd include the design in this PR, as it is vital for the proposal.

docs/architecture/adr-008-event-subscription.md
thanethomson: Good question. Do we actually want free string input at all? 😁 Or do we want to restrict the client to being able to exclusively construct queries we can handle and validate? I'd prefer the latter actually.

docs/architecture/adr-008-event-subscription.md
xla: That be the best possible outcome, an API which is correct by construction.

docs/architecture/adr-008-event-subscription.md
thanethomson: Yeah sorry I wasn't clear, I meant to say I'll add the design to this PR ASAP 👍

docs/architecture/adr-008-event-subscription.md
thanethomson: Excellent! I'll do that then 👍

docs/architecture/adr-008-event-subscription.md
xla: Super fair, let's capture a follow-up.

docs/architecture/adr-008-event-subscription.md
xla: I also prefer the later as well, as  we can build free string input on top of that if need arises.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,519 @@
...
+         [here][query-peg]).
+      2. Callers must be able to create multiple distinct subscriptions.
+      3. A subscription must offer an interface to allow for iteration over
+         incoming events only relevant to that subscription (i.e. it should
+         produce an **event stream**).
+      4. It must be possible to, from outside of the RPC client, receive events
+         from multiple subscriptions' event streams concurrently without
+         interfering with/blocking each other.
+      5. It must be possible to, from outside of the RPC client, handle
+         subscriptions' event streams without blocking other RPC operations.
```
ebuchman, line 34: What about the ability to receive all subscriptions on a single stream? eg. if I want to subscribe to a bunch of distinct event types and receive them in order?

And would it make sense / simplify things for the base subscription system to be like that, ie. a single stream with all subscriptions, and have the multi streams (I guess the Router) be an optional layer on top?

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,519 @@
...
+   2. Offer the ability to **unsubscribe** from specific queries (i.e. to
+      terminate specific subscriptions).
+2. An appropriate concurrency model for drivers of the transport layer that
+   allows the transport layer to operate independently of consumers of
+   subscriptions. This is so that consumers don't block transport layer
+   activities and vice-versa.
+
```
ebuchman, line 41: Would be helpful if we also had context on what currently exists in the code base and what the issues are.

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,519 @@
...
+**Note**: All `async` traits are facilitated by the use of [async-trait].
+
+```rust
+pub type Result<R> = std::result::Result<R, Error>;
+
+#[async_trait]
+pub trait Client {
+    /// `/abci_info`: get information about the ABCI application.
+    async fn abci_info(&self) -> Result<abci_info::AbciInfo>;
```
ebuchman, line 142: Not sure how valuable it is to list more than a few of these here since the tendermint RPC endpoints are already well defined and the methods all follow a common pattern. So just need an indication that there's a method for every RPC endpoint, some representative examples, any unique aspectss, and then the perform method

docs/architecture/adr-008-event-subscription.md
```diff
@@ -0,0 +1,519 @@
...
+
+### `SubscriptionRouter`
+
+All possible [`SubscriptionClient`](#subscriptionclient) implementations would
+need some form of subscription management and event/result routing. A
+`SubscriptionRouter` is proposed whose interface facilitates:
+
+1. Immediate subscribe/unsubscribe request fulfilment.
+2. Two-stage subscribe/unsubscribe request management (where a
+   subscription/unsubscribe request can first be created in a "pending" state,
```
ebuchman, line 297: What's the motivation for this?

docs/architecture/adr-008-event-subscription.md
thanethomson: My understanding of the initial reasoning around having `Subscription` implement the [`Stream`](https://docs.rs/futures/0.3.5/futures/stream/trait.Stream.html) trait is that it opens up possibilities around the use of more general `futures` functionality like [`select_all`](https://docs.rs/futures/0.3.5/futures/stream/fn.select_all.html), which would allow your code to combine multiple subscription streams into a single one and iterate through them as one stream.

docs/architecture/adr-008-event-subscription.md
thanethomson: Good point - I'll condense this 👍

docs/architecture/adr-008-event-subscription.md
thanethomson: Since the client multiplexes multiple subscriptions on a single WebSocket connection, there are no request/reply mechanics. You send messages and you continuously receive them in any order (i.e. subscription request responses could be mixed in with incoming events related to a different subscription). For example:

1. The client sends a subscription request via the WebSocket connection. This creates a subscription in a `Pending` state, but it's not yet active (since the RPC can still respond with an error). We don't publish events to `Pending` subscriptions.
2. The client may, while waiting for a response, receive events for a different subscription (made earlier than (1)).
3. When a response comes back whose ID matches the request from (1), if it's successful, the subscription is marked as `Active`, and if it failed we cancel and remove it.

Similar dynamics happen when unsubscribing, so unsubscribing is also currently a two-stage process.

docs/architecture/adr-008-event-subscription.md
ebuchman: Interesting, and that would preserve order from the underlying socket? Does that make more sense than an approach where all events on the websocket itself are a single Stream and clients have option of partitioning that up into possibly multiple streams per subscription?

thanethomson: The docs for `select_all` claim that elements from the underlying streams are returned while preserving the order in which they are produced by the underlying streams. I can add some tests into #516 that verify this, because I haven't actually tested it yet.

In theory it should work like this. The WebSocket driver I've implemented in #516 pushes events out to subscriptions (into their internal `tokio::mpsc` buffers) immediately as they're received before trying to read another event from the WebSocket connection, so this approach should preserve ordering when using `select_all`.

docs/architecture/adr-008-event-subscription.md
thanethomson: The current proposal (and implementation) now makes exclusive use of **unbounded** channels. This doesn't do away with the need for discussion around handling blocked/full channels, but it buys us a bit of time.

docs/architecture/adr-008-event-subscription.md
thanethomson: I've implemented a handle/driver architecture for most of the parts of the system that need to operate in two async contexts. The "handle" part allows for total control of the "driver" part.

Unfortunately, carelessly dropping the handle part right now with the current architecture still leaves the driver dangling (which will hopefully be more easily addressed when [async destructors](https://internals.rust-lang.org/t/asynchronous-destructors/11127) are available in Rust).

docs/architecture/adr-008-event-subscription.md
thanethomson: No longer relevant (yet), since I've taken out the bounded buffers from the architecture.

docs/architecture/adr-008-event-subscription.md
ebuchman: So when/how can we ever not use the two-stage subscription? ie when can we use the immediate one?

PR #477, opened 2020-07-24
Use primary error as context of `NoWitnessLeft` error

yihuang:
So user can see the real error for verification failure.

### DIFF
--- a/light-client/src/supervisor.rs
+++ b/light-client/src/supervisor.rs
@@ -248,11 +248,9 @@ impl Supervisor {
                 }
             }
             // Verification failed
-            Err(_err) => {
-                // TODO: Log/record error
-
+            Err(err) => {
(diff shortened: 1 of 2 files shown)


light-client/src/supervisor.rs
```diff
@@ -248,11 +248,9 @@ impl Supervisor {
                 }
             }
             // Verification failed
-            Err(_err) => {
-                // TODO: Log/record error
-
+            Err(err) => {
                 // Swap primary, and continue with new primary, if there is any witness left.
-                self.peers.replace_faulty_primary()?;
+                self.peers.replace_faulty_primary(Some(err))?;
```
liamsi, line 253: When putting together the light-node, I also stumbled upon this: I was seeing "no witness left" and it wasn't really informing what went wrong. 
@romac does anything speak against this approach? The todo is indicating you already had sth in mind for this?

Issue #461, opened 2020-07-19
spec: proof of fork


ebuchman:
Draft of the problem with more of an IBC lens is [here](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/detection/req-ibc-detection.md).

Primary next step is to flesh out the "proof of fork" data structures and protocol for light clients to submit to full nodes and to on-chain handlers. 

See the [next steps](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/detection/req-ibc-detection.md#next-steps) in the doc and further discussion in the [first PR reviews](https://github.com/informalsystems/tendermint-rs/pull/424)

ebuchman:
As far as relationship between this follow up and the Go code:

1) "proof of fork" for on-chain clients - may require breaking changes to ICS07
2) "proof of fork" for full nodes - non-breaking changes to Tendermint Core
3) "evidence of misbehaviour" - breaking changes to Tendermint Core

Thus we should prioritize (1) and (3) as these may be breaking on IBC and Tendermint respectively, while (2) would be less disruptive changes to the RPC. That said, of course figuring out (2) may impact (3), so they need to be somewhat thought of together.

Since these changes are to a generally underused component of the system (evidence handling), these changes should not block other integrations against Stargate releases, but they may require testnet restarts.

ebuchman:
@milosevic @josef-widder should we close this?

josef-widder:
Yes. The on-chain accountability protocol is the next thing we are doing, but there are other issues for that.

PR #479, opened 2020-07-27
spec: Fork Detector + Evidence (proof of fork) for discussion.

josef-widder:
This document ([rendered](https://github.com/informalsystems/tendermint-rs/blob/df8069a53389fd60b359305e3dc6d71858e407c1/docs/spec/lightclient/detection/detection.md)) can be a basis for follow-up discussions contains the following sections:

- the outcome of recent discussion
- a sketch of the light client supervisor to provide the context in
  which fork detection happens
- a discussion about lightstore semantics
- a draft of the light node fork detection including "proof of fork"
  definition, that is, the data structure to submit evidence to full
  nodes.

Addresses point 2. in #461.  
Defines "proof of fork" as required by https://github.com/tendermint/tendermint/issues/5149


cmwaters:
Hey josef, I just skimmed through the document yesterday. I think the biggest question I have outstanding is in regards to the the PoF with the lightblock traces. I feel there's no much mentioned of the reasoning behind how the PoF actually proves that there is a fork (this might be discussed in more depth in a document that I haven't yet seen) or more specifically what I feel is the second half of the problem which is what the full node does with the PoF, how it verifies it and extracts out the respective evidence that it will need to commit on the chain (assuming that is the ultimate goal of fork accountability). I thought I'd just drop my thoughts in this PR now so that we can discuss this for our call later today.

josef-widder:
Thanks, @cmwaters for looking at the document! Indeed, this document does not capture the questions you raised. The plan is that that together with @milosevic we will address these questions in a different specification. We already created the following issue for that  https://github.com/tendermint/tendermint/issues/5149.

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+>     the fork happened before or after 5
+>   - The light node cannot just send b10, as the primary would also 
+>     need  b5 for verification
+>   - In order to minimize the evidence, the light node may try to
+>     figure out where the branch happens, e.g., by asking the primary
+>     for height 5 (it might be that more queries are required, also
+>     to the secondary. However, assuming that in this scenario the
+>     primary is faulty it may not respond.
+	  
+	  As the main goal is to catch misbehavior of the primary,
```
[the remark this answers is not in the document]
josef-widder: We want to catch all of them. I guess this remark came out of discussion with @milosevic that also underly the comments in #494: we may use whatever is verified, and roll back if we discover a problem. In this view, the attack on the light client we have to worry about is done by the primary

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
-- We should clarify what is the expectation of VerifyToTarget so if it
-  returns TimeoutError it can be assumed faulty. I guess that
-  VerifyToTarget with correct full node should never terminate with
-  TimeoutError.
+- For submission of proof of fork, we may do some optimizations, for
+  instance, we might just submit  a trace of lightblocks that verifies a block
+  different from the one the full node knows (we do not send the trace
+  the primary gave us back to the primary)
+
+- The light client attack is via the primary. Thus we try to
```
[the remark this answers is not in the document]
josef-widder: As said above, it may only lead to a safety issue, if it comes from the primary, and we gave the verified header to the use. If the verified header that comes from the primary is OK, the secondary may just force the lightclient to shut down and report evidence, but it will not make the light client have a bad verified block.

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+   - as a result we do not have finality
+   - once the light node reconnects with a correct full node, it
+     should detect the light client attack and submit evidence.
+
+Under these assumptions, `StateTrusted` loses its meaning. As a
+result, it should be removed from the API. We suggest that we replace
+it with a flag "trusted" that can be used 
+- internally for efficiency reasons (to maintain
+  [LCD-INV-TRUSTED-AGREED.1] until a fork is detected)
+- by light client based on the "one correct full node" assumption
```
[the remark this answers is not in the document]
josef-widder: I like the idea a lot. In some sense, it makes the requirements about the trusting and unbonding periods more concrete. The unbonding period must contain the following times: 
- download and verify a block
- cross check a block (= the time parameter you are suggesting)
- compute and report proof of fork
- process proof of fork, and punish misbehavior.

Right now, all this is somewhat hidden in the assumption that the trusting period << unbonding period.

docs/spec/lightclient/detection/detection.md
```diff
@@ -363,8 +560,92 @@ The detector shall maintain the following invariants:
...
+> The following data structure [LCV-DATA-POF.1]
+> defines a **proof of fork**. Following
+> [TMBC-SIGN-FORK.1], we require two blocks *b* and *c* for the same
+> height that can both be verified from a common root block *a* (using
+> the skipping or the sequential method).
+
+> [LCV-DATA-POF.1] mirrors the definition [TMBC-SIGN-FORK.1]:
+> *TrustedBlock* corresponds to *a*, and *PrimaryTrace* and *SecondaryTrace*
+> are traces to two blocks *b* and *c*. The traces establish that both
+> *skip-root(a,b,t)* and *skip-root(a,c,t)* are satisfied.
```
cmwaters, line 598: Do we need to enforce that the heights used in verification via skipping must be the same for both `skip-root(a,b,t)` and `skip-root(a,c,t)`

josef-widder: No. Just the blocks *b* and *c* need to be the same height. The other blocks on the traces need not be at the same height. Due to the dynamic changes in the validator set, in general skipping via the same heights might not work on different branches. 

Actually, if there would be earlier two conflicting blocks with the same height on two different traces, we could use those earlier blocks as proof of fork. Therefore, attackers will most likely try to avoid earlier conflicting block.

docs/spec/lightclient/detection/detection.md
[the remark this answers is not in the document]
milosevic: The key property detection should ensure is the following: if a bad light block is "consumed" (either verified in the light node case or installed on chain in IBC case), it will be detected within bounded time. In case of the light node, this is ensured by checking verified block against witnesses, while in IBC case it done by a correct relayer checking every installed header. This ensures that exposed misbehaviour is always slashable which incentivise correct behaviour. In the scenario with correct primary and fork coming from the secondary, we will as Josef pointed out, also submit fork and shut down, but this scenario is less severe as "consumed" block is from the correct branch.

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+
+**TODO:** decide where (into which specification) to put the
+following:
+
+
+We describe the context on which the fork detector is called by giving
+a sequential version of the supervisor function.
+Roughly, it alternates two phases namely:
+   - Light Client Verification. As a result, a header of the required
+     height has been downloaded from and verified with the primary
```
milosevic, line 84:
```suggestion
     height has been downloaded from and verified with the primary.
```

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+   - Light Client Fork Detections. As a result the header has been
+     cross-checked with the secondaries. In case the is a fork we
```
milosevic, line 86:
```suggestion
     cross-checked with the secondaries. In case there is a fork we
```

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+```go
+func Sequential-Supervisor () (Error) {
+    loop {
+	    // get the next height
+        nextheight := input();
+		
+		// Verify
+        result := NoResult;
+        while result != ResultSuccess {
+            lightStore,result := VerifyToTarget(primary, lightStore, nextheight);
```
milosevic, line 105:
```suggestion
            lightStore,result := VerifyToTarget(primary, lightStore, nextHeight);
```

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+            if result == ResultFailure {				
+				// pick new primary (promote a secondary to primary)
+				/// and delete all lightblocks above
+	            // LastTrusted (they have not been cross-checked)
+	            Replace_Primary();
+			}
```
milosevic, line 111: Align it with if.

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
-- clarify EXPIRED case. Can we always punish? Can we give sufficient
-  conditions. 
+
+#### **[LC-FUNC-SUPERVISOR.1]:**
+
+```go
+func Sequential-Supervisor () (Error) {
+    loop {
+	    // get the next height
+        nextheight := input();
```
milosevic, line 100:
```suggestion
        nextHeight := input();
```

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+	            Replace_Primary();
+			}
+        }
+		
+		// Cross-check
+        PoFs := Forkdetector(lightStore, PoFs);
+        if PoFs.Empty {
+		    // no fork detected with secondaries, we trust the new
+			// lightblock
+            LightStore.Update(testedLB, StateTrusted);
```
milosevic, line 119: Where is testedLB defined?

docs/spec/lightclient/detection/detection.md
```diff
@@ -1,19 +1,199 @@
...
+```
+**TODO:** finish conditions
+- Implementation remark
+- Expected precondition
+    - *lightStore* initialized with trusted header
+	- *PoFs* empty
+- Expected postcondition
+    - runs forever, or
+	- is terminated by user and satisfies LightStore invariant, or **TODO**
+	- has submitted proof of fork upon detecting a fork
```
milosevic, line 139: We might want to clarify timing expectation in case of fork on the primary: if a forked block is verified at time t, the corresponding fork is retrieved the latest at time t + Fork_Detection_Timeout.

docs/spec/lightclient/detection/detection.md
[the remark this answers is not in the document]
milosevic: I think that this time parameter should be part of the model and API. In that sense it does not make StateTrusted obsolete: we just need to change its semantic. The right semantic could be the following: light block is in StateTrusted if it was StateVerified and Fork_Detection_Timeout has passed without finding a valid fork. Note that one possible implementation of this guarantee is what we currently have with always assuming correct node in the peer list as it provides bounded detection time. Other possible implementation would involve dynamic peer set, and assumptions on periodic presence of the correct nodes in the peer set. For efficiency reasons we could imagine also probabilistic variants where not every light block will be checked but only the random one.

docs/spec/lightclient/detection/detection.md
```diff
@@ -136,10 +316,18 @@ Let *a*, *b*, *c*, be light blocks and *t* a time, we define
 
 implies *b = c*.
 
+
+#### **[TMBC-SIGN-FORK.1]**
+
 If there exists three light blocks a, b, and c, with 
 *sign-skip-match(a,b,c,t) =
-false* then we have a *slashable fork*.
+false* then we have a *slashable fork*.
```
milosevic, line 324: I think we need to be more precise here: sign-skip-match(a,b,c,t) = false might also hold when sequ-rooted(a) = false. So we should probably give the full condition and not negation of the previous one.

docs/spec/lightclient/detection/detection.md
```diff
@@ -149,7 +337,7 @@ false* then we have a *slashable fork*.
 > full nodes as part of normal Tendermint consensus protocol". Please
 > confirm! 
   
-#### **[TMBC-SIGN-UNIQUE]**
+#### **[TMBC-SIGN-UNIQUE.1]**
```
milosevic, line 340: The same comment here: we need to give full condition and not saying sign-unique(b,c) is false as it can be false when heights differ.

docs/spec/lightclient/detection/detection.md
```diff
@@ -169,7 +357,7 @@ false* then we have a *fork on the chain*.
 > fork. 
 
 
-#### **[TMBC-LC-FORK]**
+#### **[TMBC-LC-FORK.1]**
 Let *a*, *b*, *c*, be light blocks and *t* a time. We define
 *light-client-fork(a,b,c,t)* iff
    - *sign-skip-match(a,b,c,t) = false* and
```
milosevic, line 363: I think the condition should be slightly different: sign-skip-match(a,b,c,t) = true, sequ-rooted(b), b is unique and b != c.

docs/spec/lightclient/detection/detection.md
josef-widder: I think we are precise: Sign-skip-match(a,b,c,t) is defined as implication, A -> B with

A = 
   - *sequ-rooted(a)* and
   - *b.Header.Height = c.Header.Height* and
   - *skip-root(a,b,t)* and
   - *skip-root(a,c,t)*

B = "b = c"

So if sequ-rooted(a) = false this means A is false and so the whole implication must be true, and thus sign-skip-match(a,b,c,t) = true

josef-widder:
I have added a file ([rendered](https://github.com/informalsystems/tendermint-rs/blob/7188c3f83d68af0dd30a657ff72ad73e7096c04a/docs/spec/lightclient/detection/draft-functions.md)) with sketches of functions for fork detection in the IBC context. They contain suggestions to updates to the on-chain IBC component, and for the relayer.

docs/spec/lightclient/detection/detection.md
```diff
@@ -363,8 +560,92 @@ The detector shall maintain the following invariants:
...
+*lightStore.LatestTrusted().Height* 
+from each of the current primary and the secondary, that all reported
+the identical lightblock for that height.
+
+> In the above, I guess "the identical" might be replaced with "a
+> matching" to cover commits that might be different.
+
+> The above requires us that before we pick a new secondary, we have to
+> query the secondary for the header of height 
+> *lightStore.LatestTrusted().Height*.
```
milosevic, line 579: Note that this doesn't ensure that trusted store of the secondary is consistent with the primary, i.e., only the latest trusted light block must be consistent. Could this potentially be problematic if secondary (with invalid store except latest trusted light block) is later promoted to a primary?

docs/spec/lightclient/detection/detection.md
```diff
@@ -363,8 +560,92 @@ The detector shall maintain the following invariants:
...
+<!-- }  -->
+<!-- ``` -->
+
+#### **[LCV-DATA-POFSTORE.1]**:
+
+Proofs of Forks are stored in a structure which stores all  proofs
+generated during detection.
+
+```go
+type PoFStore struct {
```
milosevic, line 629: We probably don't need this; using array of LightNodeProofOfFork should be sufficient.

docs/spec/lightclient/detection/detection.md
```diff
@@ -363,8 +560,92 @@ The detector shall maintain the following invariants:
...
+the [verification specification][verification], the 
+LightStore exposes the following function
+
+#### **[LCD-FUNC-SUBTRACE.1]:**
+```go
+func (ls LightStore) Subtrace(from int, to int) LightStore
+```
+- Expected postcondition
+   - returns a lightstore that contains all lightblocks *b* from *ls*
+     that satisfy: *from < b.Header.Height <= to*
```
milosevic, line 645: Do you need returned light blocks to satisfy that every next block is verified from the previous one?

docs/spec/lightclient/detection/detection.md
```diff
@@ -375,14 +656,110 @@ func FetchLightBlock(peer PeerID, height Height) LightBlock
...
+
+
+#### **[LCD-FUNC-REPLACE-PRIMARY.1]:**
+```go
+Replace_Primary()
+```
+**TODO:** formalize conditions
+- Implementation remark
+    - the primary is replaced by a secondary, and lightblocks above
+      trusted blocks are removed
```
milosevic, line 718: Maybe those blocks just need to go through fork detection procedure instead of removing them.

docs/spec/lightclient/detection/detection.md
```diff
@@ -435,41 +828,47 @@ func ForkDetector(ls LightStore)  {
 			auxLS.Init
 			auxLS.Update(LightStore.LatestTrusted(), StateVerified);
 			auxLS.Update(sh,StateUnverified);
-			result := VerifyToTarget(s, auxLS, sh.Header.Height)
-			if result = (_,ResultSuccess) || (_,EXPIRED) {
+			LS,result := VerifyToTarget(secondary, auxLS, sh.Header.Height)
+			if (result = ResultSuccess || result = EXPIRED) {
```
milosevic, line 832: In case result = EXPIRED that means that secondary is not responsive. But as this does not allow us to verify if conflicting header is a fork, we shouldn't report it. Note that this is not a problem as the important scenario is faulty-primary, correct secondary, in which case we will not have timeout.

josef-widder: Actually, `EXPIRED` referred to some older version of verification.md. The Error code does not exist anymore, and we should remove it and just check for `ResultSuccess`

docs/spec/lightclient/detection/detection.md
josef-widder: We could do that. It will just make the control flow a bit more complicated.

docs/spec/lightclient/detection/detection.md
josef-widder: Yes, actually subtrace should be replaced by something more clever, as you suggest. The problem with consequent blocks not supporting each other, occurred to me only later, and I didn't want to change anything before we agree on (that there is a problem) and the solution.

docs/spec/lightclient/detection/detection.md
josef-widder: This is subtle. First, this current supervisor is simplified as it considers only going forward (`VerifyToTarget` only goes forward). In this case, it is not a problem. 

If we want the supervisor to be more general it need to use [Main](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification/verification.md#lcv-func-ibcmain1) instead of `VerifyToTarget`. Then we would need to query the (new) primary for an older trusted block as reference. This older trusted block then also needs to be passed into Forkdetection (which also assumes going forward). This is less critical as we discuss to anyway only pass in one block in the future. @OStevan has prepared a Rust code snippet and will create a PR on data structures and what information to pass around.

docs/spec/lightclient/detection/detection.md
josef-widder: Let's add a discussion with this explanation into the text.

konnov:
Dr. Widder held a gun against my head and asked me to approve it %)

PR #482, opened 2020-07-28
Cleanup and standardize the READMEs

ebuchman:
A few changes here:

- `CHANGES.md` is now `CHANGELOG.md` to be more standard
- Minor updates to CONTRIBUING.md and link it from the main readme
- Add and standardize READMEs in each crate
- Crate-level READMEs have minimal badges, just crate specific stuff, and point to the root README for badges, license, etc. 
- Root README has repo-wide badges, license, etc., but points to the crate READMEs for documentation

I wasn't sure what the best way to handle the root-level vs. crate-level readmes and in general some of the subtleties of multiple crates in a repo but this is what seems cleanest in my head for now.


ebuchman:
fmt is suddenly failing? This doesn't touch any Rust code? Did I inadvertently update the version of Rust we're using somehow? Btw is that fixed in the CI somewhere? We need to ensure the version of rust in the README (the badge and the Requirements section) is aligned with that in the CI - see #483

liamsi:
> fmt is suddenly failing? 
I don't think we pin the rust toolchain in CI besides saying: use stable for fmt (see: https://github.com/informalsystems/tendermint-rs/blob/db540503e874b0fcc327748e6102ef6298b11fb1/.github/workflows/rust.yml#L20-L26) 

And rust stable got updated recently:
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1450-2020-07-16

Issue #388, opened 2020-06-27
High Level Goals for the next Third


ebuchman:
This is a high level overview of potential plans for this repo over the next few months:

## Priorities
- [x] Wrap up the light-client release - https://github.com/informalsystems/tendermint-rs/milestone/1 (any other issues?)~~
- [x]  Wrap up the light-client verification spec (support with IBC)
- [x] Integrate model-based testing for the light client and more comprehensive test suite - #320, #330, #393 
- [ ] Support IBC effort with anything it might need from light-client or tendermint types - https://github.com/informalsystems/ibc-rs/issues/60
- [ ] Fork accountability spec and support for the Go implementation 
- [ ]  Light Client Evidence Submission - #415
- [x] Upgrade to be compatible with Tendermint v0.34 (ie. protobuf migration!) - #305 

## Common Goals
- [x] Wrap up the fork detection English spec (not sure if we should focus on the TLA+ for it now)
- [x] Absorb rust-abci repo (and potentially build out a rust abci client) - #29 
  - [ ]  Potentially also support some very basic application development framework
- [x] Improve the websocket connection - #311, #313 
- [x]  Release 0.16.0 - https://github.com/informalsystems/tendermint-rs/milestone/3
- [ ] Type Generalizations - #416
- [x]  Productionization Efforts  (documentation, crate organization, release process) - #323 
- [x] Compiling to WASM - #463

## Stretch Goals
- [ ] Implement the Tendermint MConnection, bring back SecretConnection from tmkms, and run a simple node that can make p2p connections to tendermint full nodes and exchange PEX messages (without any logic or address book ...) - to discuss and open issues about

tomtau:
I may suggest absorbing @devashishdxt https://github.com/devashishdxt/abci-rs instead, as 1) it works with rust async (so doesn't bring in legacy tokio 0.1 deps), 2) it was planned to replace the current rust-abci implementation:
https://github.com/tendermint/rust-abci/pull/112 but delayed for the moment.

So I think it's a better starting point -- instead of the custom wrapper types, it can perhaps reuse parts of tendermint-rs?

marbar3778:
> I may suggest absorbing @devashishdxt's rearchitected [devashishdxt/abci-rs](https://github.com/devashishdxt/abci-rs) instead, as 1) it works with rust async, 2) has some desired features in place (e.g. unix domain socket support), 3) it was planned to replace the current rust-abci implementation:
[...]

Issue #29, opened 2019-09-07
Absorb rust-abci


marbar3778:
Discuss absorbing the rust-abci repo as its own crate within this repo.

tomtau:
that could be good for unifying some of the amino/protobuf types

ebuchman:
seems like a good idea - we should add something about this to the ADR-001: https://github.com/interchainio/tendermint-rs/pull/28

davebryson:
It *feels* like pulling ABCI into this repo would just complicate things a bit.  Unless this repo is slowly moving from just a client to a full blown tendermint core implementation...which would be really cool btw...

tomtau:
Yeah, it can be a bit complicated at the moment, as the protobuf generated definitions versus the nice hand-written RPC types don't quite match up

tarcieri:
For what it's worth, I'd prefer generated RPC types for use in the KMS. I think the current handwritten ones are a bit overcomplicated, and using protos means they won't diverge.

tomtau:
I think that could work -- one would need to modify https://github.com/tendermint/rust-abci/blob/develop/build.rs

so that the generated ABCI types derive serde deserialisers etc.
https://docs.rs/protobuf-codegen/2.8.1/protobuf_codegen/struct.Customize.html#structfield.serde_derive

One thing to consider would be separating out the ABCI types into a dedicated `rust-abci-types` crate (versus keeping it with the ABCI server implementation)

ebuchman:
Ideally we can use proto as the canonical definition for all the types, and re-use them both for ABCI and the RPC. 

This won't 100% work for everything yet as there are still some things that depend on amino encoding with prefixes in the RPC (namely pubkeys, but I think that's it). But we plan to transition even amino interfaces to proper proto3 wellknowntype Any soon, so this should be fully possible. Definitely something to work towards.

> Unless this repo is slowly moving from just a client to a full blown tendermint core implementation...which would be really cool btw...

That's the plan!

marbar3778:
Would like to revisit this issue. Should a new ADR be written or an amendment be made to adr-001?

Liamsi:
@marbar3778 an adr or even a draft PR instead (or in parallel with that adr) would be awesome.

ebuchman:
Can we clarify the relationship between rust-abci and abci-rs? Has the former absorbed the latter in some way? 

[...]

PR #489, opened 2020-07-28
Merge abci-rs in tendermint-rs repository

devashishdxt:
Merging `abci-rs` crate in `tendermint-rs` repository as suggested by @tomtau [here](https://github.com/informalsystems/tendermint-rs/issues/388#issuecomment-651680938). Also, continuing discussion on https://github.com/informalsystems/tendermint-rs/issues/29.

### DIFF
--- a/abci/src/server.rs
+++ b/abci/src/server.rs
@@ -1,0 +1,278 @@
+#[cfg(unix)]
+use std::path::PathBuf;
+use std::{io::Result, net::SocketAddr, sync::Arc};
+
+#[cfg(all(unix, feature = "use-async-std"))]
+use async_std::os::unix::net::UnixListener;
+#[cfg(feature = "use-async-std")]
+use async_std::{
+    io::{Read, Write},
+    net::TcpListener,
+    prelude::*,
+    sync::Mutex,
+    task::spawn,
+};
+use tendermint_proto::abci::{
+    request::Value as RequestValue, response::Value as ResponseValue, Request, Response,
+};
+#[cfg(all(unix, feature = "use-tokio"))]
+use tokio::net::UnixListener;
+#[cfg(feature = "use-tokio")]
+use tokio::{
+    io::{AsyncRead as Read, AsyncWrite as Write},
+    net::TcpListener,
+    spawn,
+    stream::StreamExt,
+    sync::Mutex,
+};
+use tracing::{debug, error, info, instrument};
+
+use crate::{
+    state::ConsensusStateValidator,
+    types::{decode, encode},
+    Consensus, Info, Mempool, Snapshot,
+};
+
+/// ABCI Server
+pub struct Server<C, M, I, S>
+where
+    C: Consensus + 'static,
+    M: Mempool + 'static,
+    I: Info + 'static,
+    S: Snapshot + 'static,
+{
+    /// Wrapping inner type in `Arc` so that it becomes clonable and can be shared between multiple
+    /// async tasks
+    pub(crate) inner: Arc<Inner<C, M, I, S>>,
+}
+
+/// Inner type that contains all the trait implementations
+pub(crate) struct Inner<C, M, I, S>
+where
+    C: Consensus + 'static,
+    M: Mempool + 'static,
+    I: Info + 'static,
+    S: Snapshot + 'static,
+{
+    consensus: C,
+    mempool: M,
+    info: I,
+    snapshot: S,
+    consensus_state: Mutex<ConsensusStateValidator>,
+}
+
+impl<C, M, I, S> Server<C, M, I, S>
+where
+    C: Consensus + 'static,
+    M: Mempool + 'static,
+    I: Info + 'static,
+    S: Snapshot + 'static,
+{
+    /// Creates a new instance of [`Server`](struct.Server.html)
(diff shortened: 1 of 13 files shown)


abci-rs/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+features = ["doc"]
+rustdoc-args = ["--cfg", "feature=\"doc\""]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+async-std = { version = "1.6", optional = true }
+async-trait = "0.1"
+bytes = "0.5"
+integer-encoding = { version = "1.1", optional = true }
```
tomtau, line 28: there's a small bug in the released integer-encoding: https://github.com/dermesser/integer-encoding-rs/pull/10
not sure of the impact. but it's a fairly small crate / code needed here, so could be possibly copied / vendored in here directly?

ebuchman: From https://github.com/tendermint/rust-abci/pull/134 it seems we're not reading the first byte? Was it just being ignored or? Trying to understand the state of things and how this bug was surfaced in the first place

tomtau: It seems it's being read (https://github.com/tendermint/rust-abci/blob/develop/src/codec.rs#L26 -- or @yihuang why it's not?) and under normal circumstances, TM would be writing terminated varints as lengths ?

ebuchman: what does terminated mean here? do we mean like a terminal byte in the varint encoding, ie. where the MSB is not set? I think that first "length of length" is always just one byte here so should always be terminal, and should probably also always be small (like 1-3, maybe max 10, since theres a limit on how big a msg tendermint would send).

tomtau: terminated varint -- the last byte has MSB set.
the code in rust-abci seems to match what's in Golang:
https://github.com/tendermint/tendermint/blob/master/abci/types/messages.go#L54

devashishdxt:
> That's great. Quick early feedback, as we this becomes part of the tendermint workspace we can safely drop the `-rs` suffix in the crate name.

Done.

tomtau:
@xla it'll also be needed for @zramsay to transfer https://crates.io/crates/abci ownership

xla:
> @xla it'll also be needed that @zramsay transfers https://crates.io/crates/abci ownership

Good point, maybe @ebuchman can follow up on this.

marbar3778:
This PR may need some reworking around the generated proto files after: https://github.com/informalsystems/tendermint-rs/pull/508 lands.

tomtau:
https://github.com/informalsystems/tendermint-rs/pull/510

tomtau:
#508 is merged -- @marbar3778 I guess the natural thing here will be to remove abci-rs's types + build.rs and just use the "tendermint-proto" crate?

devashishdxt:
> #508 is merged -- @marbar3778 I guess the natural thing here will be to remove abci-rs's types + build.rs and just use the "tendermint-proto" crate?

`tendermint-proto` is not yet released on crates.io

tomtau:
> > #508 is merged -- @marbar3778 I guess the natural thing here will be to remove abci-rs's types + build.rs and just use the "tendermint-proto" crate?
> 
> `tendermint-proto` is not yet released on crates.io



> > #508 is merged -- @marbar3778 I guess the natural thing here will be to remove abci-rs's types + build.rs and just use the "tendermint-proto" crate?
> 
> `tendermint-proto` is not yet released on crates.io

since this will be in the same repo, the dependency can be specified with `path = "....."`?

zramsay:
> > @xla it'll also be needed that @zramsay transfers https://crates.io/crates/abci ownership
> 
> Good point, maybe @ebuchman can follow up on this.

invites sent to @xla and @liamsi

devashishdxt:
> #508 is merged -- @marbar3778 I guess the natural thing here will be to remove abci-rs's types + build.rs and just use the "tendermint-proto" crate?

Pushed the changes. Also included changes for tendermint `StateSync`.

abci/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+                match request {
+                    Some(request) => {
+                        let response = inner.process(request).await;
+
+                        if let Err(err) = encode(response, &mut stream).await {
+                            error!(message = "Error while writing to stream", %err, ?peer_addr);
+                        }
+                    }
+                    None => debug!(message = "Received empty request", ?peer_addr),
+                }
```
tomtau, line 158: `await` may need some re-work to follow the async CheckTx/DeliverTx semantics when it could potentially processing another request while response to the previous one is being written out (in order): https://github.com/informalsystems/tendermint-rs/issues/29#issuecomment-672964983

devashishdxt: https://github.com/informalsystems/tendermint-rs/issues/29#issuecomment-715102710

thanethomson:
@devashishdxt, after much discussion the team decided to go with #794. We still, however, think that `abci-rs` is a great alternative option to that in #794, and look forward to seeing people use whichever framework they feel is better suited to their particular use case (we think it's a great thing for people to have multiple options).

Issue #491, opened 2020-07-30
When checking signatures, vote type is fixed to Precommit, irrespective of the actual type


andrey-kuprianov:
I've encountered  this problem while executing my model-based tests. In the [`testgen`](https://github.com/informalsystems/tendermint-rs/tree/master/testgen), the default for a vote is `Prevote` (should it be `Precommit` actually?), while when calculating the signatures in [`lite_impl/signed_header.rs::voting_power_in()`](https://github.com/informalsystems/tendermint-rs/blob/667fc8d8be4ac1745b0ecb2284e14f7d87f4b2f0/tendermint/src/lite_impl/signed_header.rs#L17), more specifically [here](https://github.com/informalsystems/tendermint-rs/blob/667fc8d8be4ac1745b0ecb2284e14f7d87f4b2f0/tendermint/src/lite_impl/signed_header.rs#L126), the vote type is fixed to be `Precommit`, that's why the validation is failing.

Here are the two alternatives that I currently see, and both mean that it's probably a bug:
* If only `Precommit` votes are allowed in a `Commit`, then the `Prevote` votes included there should probably be ignored, but the rest of the votes still counted
* If both types of votes are allowed, then the vote type should be copied 'as is' when verifying signatures.

Could someone please clarify the issue?

andrey-kuprianov:
or is it actually, that only `Precommit` votes can be used for making a commit? Because the `CommitSig` itself doesn't carry any vote type... Then there is no bug of course, and I should change the default vote type in `testgen` to `Precommit`.

ebuchman:
Only Precommits can be included in a Commit. If a commit contains a signature from a Prevote it would be treated as invalid.

andrey-kuprianov:
Ok, it's clear; thanks!

PR #493, opened 2020-07-30
#491 testgen: make Precommit default vote type

andrey-kuprianov:
Trivial changes to make `Precommit` default for votes

* [X] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/testgen/src/vote.rs
+++ b/testgen/src/vote.rs
@@ -20,8 +20,8 @@ pub struct Vote {
     pub index: Option<u64>,
     #[options(help = "header to sign (default: commit header)")]
     pub header: Option<Header>,
-    #[options(help = "vote type; 'precommit' if set, otherwise 'prevote' (default)")]
-    pub precommit: Option<()>,
+    #[options(help = "vote type; 'prevote' if set, otherwise 'precommit' (default)")]
+    pub prevote: Option<()>,
     #[options(help = "block height (default: from header)")]
     pub height: Option<u64>,
     #[options(help = "time (default: from header)")]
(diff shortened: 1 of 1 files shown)


romac:
What do you think of having a ‘—vote-type=precommit|prevote’ flag instead?

andrey-kuprianov:
@romac Thanks for the suggestion, but I think the current solution is better. With what there is currently, it's

Rust API:
```rust
let vote = Vote::new(...)  // if default Precommit is fine
let vote = Vote::new(...).prevote(true)  // if Prevote desired
```

CLI:
```bash
tendermint-testgen --stdin vote # if Precommit is fine
tendermint-testgen --stdin vote --prevote # if Prevote desired
```

With your proposal it would be 

Rust API:
```rust
let vote = Vote::new(...)  // if default Precommit is fine
let vote = Vote::new(...).vote_type(Prevote)  // if Prevote desired : this either needs a separate def, or parsing a string
```

CLI:
```bash
tendermint-testgen --stdin vote # if Precommit is fine
tendermint-testgen --stdin vote --vote-type=prevote # if Prevote desired: requires parsing a string
```
In short, if there is an obvious default, there is no need to introduce additional overhead of parsing a choice. I hope this is convincing:)

andrey-kuprianov:
It turned out that more changes are necessary for `testgen`, so I will add more commits here

romac:
I don't feel strongly about it. The one argument I can raise for the solution I proposed is that it would scale to cases where there are 3 or more choices, which can then not be expressed as a boolean flag/parameter. But in this case, if there will only ever be these two options I guess the current solution is fine as well :)

andrey-kuprianov:
OK, then let's leave it as is for now, and convert to another form if ever the third choice arises:)

PR #500, opened 2020-08-03
Consolidate and delete old implementation of the light client

### LINKED ISSUE #437: light-client: consolidate and delete old implementations
With the advent of the `light-client` crate, which was meant to superseed prior implementations we now have a large corpus of code that is outdated and should be sun-setted. Keep maintaining this parts like `lite` and `lite_impl` is gonna add significant overhead that is unnecessary at this stage of the project.

These modules should be carefully evaluated to extract knowledge, implementation details which are worth preserving w.r.t. #416.

romac:
Close #437

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


tendermint/src/block/header.rs
```diff
@@ -60,6 +62,62 @@ pub struct Header {
...
+fn encode_hash(hash: &Hash) -> Vec<u8> {
+    encode_bytes(hash.as_bytes())
+}
+
+fn encode_varint(val: u64) -> Vec<u8> {
+    let mut val_enc = vec![];
+    prost_amino::encoding::encode_varint(val, &mut val_enc);
+    val_enc
+}
+
```
liamsi, line 120: I assume there were no changes made and this was just copied here, right?

tendermint/src/trust_threshold.rs
```diff
@@ -0,0 +1,76 @@
...
+                numerator,
+                denominator,
+            })
+        } else {
+            None
+        }
+    }
+}
+
+// TODO: should this go in the central place all impls live instead? (currently lite_impl)
```
liamsi, line 58:
```suggestion
// TODO: should this go in the central place all impls live instead? (currently lite_impl)
```
Does this TODO still hold? particularly "currently lite_impl"?

tendermint/src/block/header.rs
romac: Yep!

tendermint/src/trust_threshold.rs
romac: Good catch, will remove this comment since this file now lives outside of the `lite` module.

light-client/tests/support
```diff
@@ -1 +0,0 @@
-../../tendermint/tests/support/lite
```
liamsi, line 1: Just to double check: these tests are called from their new location now, right?

liamsi: Ah I see, they were always called from the new location. Now they are actually moved there (instead of using the sim-link). All good.

romac: Yep, that's it!

light-client/src/operations/hasher.rs
```diff
@@ -16,7 +14,7 @@ pub trait Hasher: Send {
 }
 
 /// Default implementation of a hasher
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Default, Debug)]
```
xla, line 17:
```suggestion
#[derive(Clone, Copy, Debug, Default)]
```

light-client/src/predicates.rs
```diff
@@ -14,7 +14,7 @@ pub mod errors;
 
 /// Production predicates, using the default implementation
 /// of the `VerificationPredicates` trait.
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Default, Debug)]
```
xla, line 17:
```suggestion
#[derive(Clone, Copy, Debug, Default)]
```

Issue #499, opened 2020-07-31
Lightstore Invariants + Maintain Verification Information


josef-widder:
In order to generate "proof of fork" (#461 , https://github.com/tendermint/tendermint/issues/5149), it would be good to efficiently compute a sequence of lightblocks b(1), b(2), ... b(n), so that b(i+1) can be verified with b(i) in one step. If the light client does forward verification only, this property is maintained by the algorithm (I think @konnov checked that property already). However, if the light client has to fetch lightblocks whose height is less than `latestTrusted` (see [LCV-FUNC-IBCMAIN.1](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification/verification.md#lcv-func-ibcmain1)), this property might not be maintained. 

For instance, if b(1).NextValidators = b(2).Validators, but for a later added lightblock in-between, b(1.5).NextValidators and b(1.5).Validators may be piecewise disjunct from both. 

To compute the proof of fork, I don't think we should re-verify the blocks to generate a sequence; this is expensive. I wonder whether we should maintain in the light store, for each lightblock which lightblock was used to originally verify it. We would need to add recording this information within VerifyToTarget.

Alternatively, we just keep these additional blocks in the proof of fork, but then the complexity of the "proof of fork verification" at the nodes who receive it increases.

OStevan:
@josef-widder As a note to this issue at least in the forward verification case:
1) if the property is that there is a sequence of light blocks with increasing heights in the light store which maintains this property than it is ok
2) but if the property is that all stored light blocks represent such a sequence than this property is not currently maintained in TLA nor Rust (explanation below)

In the case of TLA, this is explicitly mentioned in [StoredHeadersAreVerifiedOrNotTrustedInv](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification/Lightclient_A_1.tla#L330). And in rust it is a bit more involved, if we look at the `Scheduler`, specifically the [bisecting scheduler](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/components/scheduler.rs#L57) and the invariants, it can ignore Unverified light blocks in the store and as such might, miss previously fetched light blocks leaving them in Unverified state.

As a result in both cases, the light store will have a mix of Verified and Unverified light blocks after finishing verification. It is easy to extract just the sequence of Verified blocks, but it is not true that all stored light blocks form this sequence.

josef-widder:
[...]

Issue #497, opened 2020-07-30
Relayer requirements for the Light Client


ancazamfir:
## Summary
Relayer requirements (shown in **bold**) for and interactions with the per-chain light clients.
## Proposal
Relayer interactions with light clients:

**`get_header(0)`** - the relayer asks for the latest header from the light client when an IBC client has to be created.
 
**`get_header(h)`**- the relayer requests the verified header at height `h` when it needs to locally verify proofs obtained at height `h-1`, before including them in the handshake or packet transactions.

**`get_minimal_set(e, h)`** - this is for the IBC client update. 
The relayer knows that header `e` is the latest state of the IBC client hosted on-chain. To ensure proofs also verify on this chain, the relayer has to update the client with header `h` and any supporting headers in-between. 
Things to consider in this case:
- if the on-chain client has been previously updated with `e` by a different relayer, then the relayer light client might not have header `e`. 
- it is also possible that the local client might have previously updated to a header `j` between `e` and `h`, header that might not be needed if bisection would start with a store that has `e` as latest trusted height.
- it is possible that the light client has updated to a header `n`, higher than `h`

_Here are the ordered heights mentioned above_
`(1, .., i,.., [e], .., j,..,[h], .., n]`

In addition to the APIs above. The relayer must make sure the on-chain consensus state does not expire. For this purpose:
1) the relayer runs a per-chain timer with duration some fraction of the trusting period and, when it fires, requests the latest header from the light client. The same API as above, `get_minimal_set(e, h)` is used.
This timer is reset every time the on-chain client consensus state is updated. However, if the on-chain consensus state is updated by another relayer that consistently gets ahead of this relayer then there is a danger that the local trusted state expires...unless:
 
**2) the local light client should make sure a new header is downloaded and verified such that the latest trusted state does not expire.** There is no need for a notification from the light client thread to the relayer thread in this case.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Let's move this to v0.18 as we will likely not have time to get it done for v0.17. Will tackle this once it's actually needed on the ibc-rs side.

thanethomson:
@romac, when do you anticipate this will be needed in ibc-rs?

romac:
Very soon :) I am actually going to start working on that this week.

ancazamfir:
[...]

Issue #494, opened 2020-07-30
spec: Detector refactor


josef-widder:
Following a discussion with @OStevan about https://github.com/informalsystems/tendermint-rs/pull/446, I am going to refactor the supervisor and detector spec in #479 so that the supervisor maintains a lightstore that only contains lightblocks that have been cross-checked, (or the once that have been verified with the primary that resulted in a cross-checked state).

To do so, we need to 
- pass  `auxPrimaryLightStore` into VerifyToTarget
- pass a trusted and the latest verified into the detector
- update the supervisor lightstore once we finished cross checking (insert the `auxPrimaryLightStore`)

This will allow us to ignore the state of lightblocks in the supervisor lightstore, as they all have been cross checked. This defines a cleaner interface: All lightblocks in the supervisor lightstore can be used by the user. All lightblocks in the supervisor lightstore are on a chain that lead to a cross-checked (trusted) lightblock.

ebuchman:
> This will allow us to ignore the state of lightblocks in the supervisor lightstore, as they all have been cross checked. This defines a cleaner interface: All lightblocks in the supervisor lightstore can be used by the user. All lightblocks in the supervisor lightstore are on a chain that lead to a cross-checked (trusted) lightblock.

I agree with being able to ignore "state" of headers and having a store that only has headers that can be used by the user. But I also think we're going to need a more nuanced view of what can be "trusted" / put into the supervisors light store / used by the user, and our thinking about the problem on the full node side is I think a bit muddled by having a static set of peers. The problem right now is that with a static set of peers and assuming at least one is correct, we don't even need to do verification, we could just be comparing headers (until there's a conflict).

I think we want to move towards a model that functions more like the on-chain IBC light client, where we assume that eventually, within the trusting period, we will hear from a correct node (relayer in the IBC case). The basic version is to always trust (ie. "expose to the user") the first thing that gets verified, but to still be on the lookout for potential conflicts, that you might only discover later.

Even if we start with a static set in the implementation, we should probably be formulating the problem as if there were a dynamic set, and with more alignment with how the on-chain ibc client operates.

[...]

Issue #487, opened 2020-07-28
light-node: does not recover from primary restart


ebuchman:
If I kill a witness process and then restart it, the light node will log some errors but recover gracefully. However if I kill the primary, the light client logs an error for ever and never recovers:

```
started RPC server: 127.0.0.1:8888
synced to block: 2391
synced to block: 2394
synced to block: 2397
error: sync failed: I/O error: error trying to connect: tcp connect error: Connection refused (os error 61) (code: 0)
error: sync failed: I/O error: error trying to connect: tcp connect error: Connection refused (os error 61) (code: 0)
synced to block: 2407
synced to block: 2410
synced to block: 2413
synced to block: 2416
synced to block: 2419
synced to block: 2423
synced to block: 2426
synced to block: 2429
error: sync failed: no witness left: no initial trusted state
error: sync failed: no witness left: no initial trusted state
error: sync failed: no witness left: no initial trusted state
error: sync failed: no witness left: no initial trusted state
```

Not sure how this would behave with three nodes but I'd think if the primary dropped and there were still two witnesses we'd be fine.

But since we can recover from a witness going down and coming back I'd think we could recover from the primary doing the same?

And if we're not going to recover here, probably better to exit than to keep throwing the error

thanethomson:
Light node has now been removed/archived.

PR #509, opened 2020-08-07
spec: Rust like pseudo code for Sequential supervisor and core verification

OStevan:
[rendered](https://github.com/informalsystems/tendermint-rs/blob/1351513ade78fd4e1864783742be5ae41f1d05f1/docs/spec/supervisor/supervisor.md)

As @josef-widder mentioned here is the pseudo code proposal for specification and code changes. I was not sure how to name this and where to put it so we can change it as others see fit. 

**Goals?**

The proposed pseudo code should solve the main issues of `LightStore` having an unbounded interface which makes code of skipping forward verification more complicated than needed. If implemented it should solve several issues which came out of it.

**Possible issues?**
- with this proposal the `LightStore` requires that when inserting new information it is always possible to build a verification chain with what is already in it
- the requirement above makes backward verification using just hashes impossible as the links with lower `LightBlocks `are impossible to recreate in this situation

**Proposal details**
- current `LightStore` is now split: 
        1.  what were once verified LightBlocks are now stored in `Trace`
        2. `Unverified` LightBlocks are now on the `FetchedStack`
        3. There are no longer `Failed` LightBlocks as there can be only one and that one is returned as an error reason in `verify_to_target`
        4. The notion of `Trusted` is removed, and with the sequential supervisor everything in what is now `LightStore` is what were previously trusted.
- History of how `LightBlocks` were verified can be retrieved from the `LightStore`
- `LightStore` can recover from possible forks in the future by removing `LightBlocks` which are now considered to be from a fork.

**Order in which to review for easier understanding**
- basically top to bottom
- first part is related to skipping verification until the `LightStore` (excluding)
- second part relevant for supervisor and possible `ForkDetection` and `Relayer` requirements

**Issue which prompted this PR**
- #494 

**Issues this approach should solve.**
- #499 : chain invariant is easily retrievable, and chain computation is built-in the `LightStore` now
- #487 : previous losses of trusted `LightBlock` are no longer possible as we always ask for the highest from the supervisor `LightStore`
- #434 : no longer relevant with refactors proposed here
- #428 : logic is very much split and the API is now easier to reason about
- #420 : same as for 487
- #333 : simplified by design

**Should help with easier implementation of**
- #415 : should be easier to collect evidence as verification trace is collected
- #497 : for all the mentioned height combinations it should be possible to easily get the light block for verification.

**Issues which should be reworded depending on the changes here**
- #334 : everything stored in the `LightStore` is assumed to have passed verification at some point

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/docs/spec/supervisor/supervisor.md
+++ b/docs/spec/supervisor/supervisor.md
@@ -4,9 +4,11 @@
 /// the same chain as the predecessor light blocks.
 type FetchedStack = Vec<LightBlock>;
 
-/// Pair of light block and height of the light block which was the 
-/// highest one for which we believe originated from the same chain
-type ChainPair = (LightBlock, Height);
+/// Chain of verification with a starting root.
+enum Chain {
+    Root(LightBlock),
+    Link(LightBlock, Box<Chain>)
+}
 
 /// Traces the history of verify_to_target.
 trait Trace {
@@ -21,7 +23,7 @@ trait Trace {
 
     /// Returns the list of all LightBlock which were inserted through 
     /// insert with associated previous current_height().
-    fn chain(&self) -> LinkedList<ChainPair>;
+    fn chain(&self) -> Chain;
 }
 
 
@@ -48,24 +50,13 @@ pub fn verify_to_target(
         let lowest_unverified = fetch_stack.peek().unwrap().clone();
         
         // do the necessary checks
-        let verdict = valid_and_verified(trace.current(), &lowest_unverified);
-
-        match verdict {
-            // validated LightBlock, improve the verified_state
-            Valid => {
-                fetch_stack.pop();
-                trace.insert(lowest_unverified);
-                continue;
-            },
-            // invalid block stop
-            Invalid(error) => {
-                return error
-            },
-            Untrusted => {
-                // we need an intermediate LightBlock
-                current_height = 
-                    compute_height(trace.current_height(), lowest_unverified.height());
-            }
+        let outcome = try_verify(&trace, &lowest_unverified).map_err(|e| (lowest_unverified, e))?;
+        
+        if outcome {
+            fetch_stack.pop();
+            trace.insert(lowest_unverified)        
+        } else {
+            fetched_stack.push(compute_height(trace.current_height(), lowest_unverified.height()));
         }
     }
 
@@ -74,6 +65,26 @@ pub fn verify_to_target(
 
     return Ok(trace);
 }
+
+/// Tries to validate the light_block given an existing trace.
+pub fn try_verify(trace: &Trace, light_block: &LightBlock) -> Result<(), VerificationError> {
+    let verdict = valid_and_verified(trace.current(), light_block);
+
+    match verdict {
+        // validated LightBlock, improve the verified_state
+        Valid => {
+            Ok(true)
+        },
+        // invalid block stop
+        Invalid(error) => {
+            error
+        },
+        Untrusted => {
+            Ok(false)
+        }
+    }
+}
+
 
 /// LightStore no longer has notions of Trusted, Verified, Unverified and Failed.
 /// The LightBlocks which are stored are all assumed to have originated from the
@@ -84,24 +95,24 @@ trait LightStore {
     /// for which this store is associated with.
     fn highest(&self) -> LightBlock;
 
-    /// Stores the list of ChainPair, possibly changing the highest LightBlock;
+    /// Stores the list of Chain, possibly changing the highest LightBlock;
     /// Precondition is the existence of Light Block for which the chain can be linked
-    /// with the already stored ChainPair. Failing if the chain can not be linked with
+    /// with the already stored Chain. Failing if the chain can not be linked with
     /// existing history or if there is conflicting information.
-    fn store_chain(&mut self, chain: LinkedList<ChainPair>) -> Result<(), Error>;
+    fn store_chain(&mut self, chain: Chain) -> Result<(), Error>;
     
     /// Stores one ChainLink, failing if there is conflicting information in the store.
     /// (forks, etc.). Can be used when light blocks are received out of order.
     fn store_pair(&mut self, link: ChainLink) -> Result<(), Error>;
 
     /// Get verification chain of LightBlock ending at height high and starting starting from low.
-    fn get_chain(&self, low: Height, high: Height) -> LinkedList<ChainPair>;
+    fn get_chain(&self, low: Height, high: Height) -> Chain;
 
     /// In a case a fork is detected allows the light store to recover for a specified boundary
     /// height. Erases all light blocks from the store with height greater or equal to boundary.
     /// Returning the chain of light blocks which are removed. To be used if a fork is discovered
     /// after insertions of light blocks.
-    fn recover(&mut self, boundary: Height) -> Result<LinkedList<ChainPair>, LightStoreError>;
+    fn recover(&mut self, boundary: Height) -> Result<Chain, LightStoreError>;
 
     /// Returns the firsts LightBlock above if it exists.
     fn above(&self, height: Height) -> Result<&LightBlock, ListStoreError>;
--- a/docs/spec/supervisor/supervisor.md
+++ b/docs/spec/supervisor/supervisor.md
@@ -1,10 +1,10 @@
 
 ```rust
-/// Stores the Light Blocks which could not be confirmed to have originated from
+/// Stores the LightBlocks which could not be confirmed to have originated from
 /// the same chain as the predecesor light blocks.
 type FetchedStack = Stack<LightBlock>;
 
-/// Pair of light block and height of the light block which was highest 
+/// Pair of light block and height of the light block which was the 
 /// highest one for which we believe originated from the same chain
 type ChainPair = (LightBlock, Height);
 
@@ -21,7 +21,7 @@ trait Trace {
 
     /// Returns the list of all LightBlock which were inserted through 
     /// insert with associated previous current_height().
-    fn verification_chain(&self) -> LinkedList<ChainPair>;
+    fn chain(&self) -> LinkedList<ChainPair>;
 }
 
 
@@ -34,7 +34,7 @@ pub fn verify_to_target(
     target_height: Height) -> Result<Trace, (LightBlock, Reason)> {
 
     // Verified state always has one starting LightBlock
-    let mut trace = Trace::from(startingLightBlock);
+    let mut trace = Trace::from(starting_light_block);
     let mut fetch_stack = FetchedStack::empty();
 
     // insert the light block of target height, simplifies the loop
@@ -86,7 +86,7 @@ trait LightStore {
 
     /// Stores the list of ChainPair, possibly changing the highest LightBlock;
     /// Precondition is the existence of Light Block for which the chain can be linked
-    /// with the already stored ChainPair. Failing if the chain can bot be linked with
+    /// with the already stored ChainPair. Failing if the chain can not be linked with
     /// existing history or if there is conflicting information.
     fn store_chain(&mut self, chain: LinkedList<ChainPair>) -> Result<(), Error>;
     
--- a/docs/spec/supervisor/supervisor.md
+++ b/docs/spec/supervisor/supervisor.md
@@ -102,6 +102,15 @@ trait LightStore {
     /// Returning the chain of light blocks which are removed. To be used if a fork is discovered
     /// after insertions of light blocks.
     fn recover(&mut self, boundary: Height) -> LinkedList<ChainPair>;
+
+    /// Returns the firsts LightBlock above if it exists.
+    fn above(&self, height: Height) -> Option<&LightBlock>;
+    
+    /// Returns the firsts LightBlock below if it exists.
+    fn below(&self, height: Height) -> Option<&LightBlock>;
+    
+    /// Tries to get a certain height.
+    fn get(&self, height: Height) -> Option<&LightBlock>;
 }
 
 /// Current understanding how a LightNode functions.
--- a/docs/spec/supervisor/supervisor.md
+++ b/docs/spec/supervisor/supervisor.md
@@ -82,9 +82,9 @@ pub fn verify_to_target(
 trait LightStore {
     /// Current highest LightBlock which we believe to be from a chain 
     /// for which this store is associated with.
-    fn highest_trusted(&self) -> LightBlock;
+    fn highest(&self) -> LightBlock;
 
-    /// Stores the list of ChainPair, possibly changing the highest_trusted LightBlock;
+    /// Stores the list of ChainPair, possibly changing the highest LightBlock;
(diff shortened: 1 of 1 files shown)


OStevan:
@milosevic @brapse @ebuchman @romac @xla

josef-widder:
Thanks a lot! Both for the proposal of the code and the awesome overview of the issues! 

I think we should discuss this here, and then incorporate either the proposed pseudo code (or a go variant of it) into a supervisor spec and the existing verification.md. (A go variant of a supervisor lives in detection.md #479, but the way the lightstore is treated and the parameters that are passes here in #509 seem much more to the point in the rust pseudo code in this PR).

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
+
+```rust
+/// Stores the Light Blocks which could not be confirmed to have originated from
+/// the same chain as the predecesor light blocks.
+type FetchedStack = Stack<LightBlock>;
+
+/// Pair of light block and height of the light block which was highest
```
xla, line 7:
```suggestion
/// Pair of light block and height of the light block which was the 
```

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
+
+```rust
+/// Stores the Light Blocks which could not be confirmed to have originated from
+/// the same chain as the predecesor light blocks.
+type FetchedStack = Stack<LightBlock>;
```
xla, line 5: What are the semantics of the stack? I don't believe there is a type definition in this file.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
+
+```rust
+/// Stores the Light Blocks which could not be confirmed to have originated from
```
xla, line 3:
```suggestion
/// Stores the LightBlocks which could not be confirmed to have originated from
```

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+/// the same chain as the predecesor light blocks.
+type FetchedStack = Stack<LightBlock>;
+
+/// Pair of light block and height of the light block which was highest 
+/// highest one for which we believe originated from the same chain
+type ChainPair = (LightBlock, Height);
```
xla, line 9: The naming is confusing as it indicates that a `Chain` is paired with something. What I'm also wondering doesn't the `LightBlock` contain its `Height`, what's the benefit of pairing it with the `Height`.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+
+    /// Returns the current highest inserted light block.
+    fn current(&self) -> &LightBlock;
+
+    /// Insert a new highest light block. assert(self.current_height() < light_block.height())
+    fn insert(&mut self, light_block: LightBlock) -> Result<(), Error>;
+
+    /// Returns the list of all LightBlock which were inserted through 
+    /// insert with associated previous current_height().
+    fn verification_chain(&self) -> LinkedList<ChainPair>;
```
xla, line 24:
```suggestion
    fn chain(&self) -> LinkedList<ChainPair>;
```

This seems expressive enough.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+type FetchedStack = Stack<LightBlock>;
+
+/// Pair of light block and height of the light block which was highest 
+/// highest one for which we believe originated from the same chain
+type ChainPair = (LightBlock, Height);
+
+/// Traces the history of verify_to_target.
+trait Trace {
+    /// Returns the height of the current highest inserted
+    fn current_height(&self) -> Height;
```
xla, line 14: What's the use-case for just getting the current height of the `Trace`, can't you obtain that through the `current` `LightBlock`?

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+}
+
+
+/// Given a primary and a starting_light_block confirm that the light block of target_height received
+/// from primary for target_height is from the same chain as the starting_light_block.
+/// Solution for forward skipping verification.
+pub fn verify_to_target(
+    primary: PeerID,
+    starting_light_block: LightBlock,
+    target_height: Height) -> Result<Trace, (LightBlock, Reason)> {
```
xla, line 34: The `Reason` type is not expanded anywhere. Would be helpful to see what it actually looks like. It's uncommon to not

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+
+    // Verified state always has one starting LightBlock
+    let mut trace = Trace::from(startingLightBlock);
```
xla, line 37:
```suggestion
    let mut trace = Trace::from(starting_light_block);
```

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+    let mut fetch_stack = FetchedStack::empty();
+
+    // insert the light block of target height, simplifies the loop
+    // we can say that at each iteration there should be something on stack
+    fetch_stack.push(fetch_light_block(primary, target_height));
```
xla, line 42: What does `fetch_light_block` do?

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+    assert_eq!(trace.current_height(), target_height);
+    assert!(fetch_stack.is_empty());
+
+    return Ok(trace);
+}
+
+/// LightStore no longer has notions of Trusted, Verified, Unverified and Failed.
+/// The LightBlocks which are stored are all assumed to have originated from the
+/// same chain, irrespective of the peer from which we received it. The store
+/// provides the invariant that there is a 1-1 mapping from height to a LightBlock.
```
xla, line 81: If we want to ensure the 1-1 invariant we should think in terms of a `Set` where we have a uniqueness constraint on the height, unless I miss-understand this part.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+    /// with the already stored ChainPair. Failing if the chain can bot be linked with
+    /// existing history or if there is conflicting information.
+    fn store_chain(&mut self, chain: LinkedList<ChainPair>) -> Result<(), Error>;
+    
+    /// Stores one ChainLink, failing if there is conflicting information in the store.
+    /// (forks, etc.). Can be used when light blocks are received out of order.
+    fn store_pair(&mut self, link: ChainLink) -> Result<(), Error>;
+
+    /// Get verification chain of LightBlock ending at height high and starting starting from low.
+    fn get_chain(&self, low: Height, high: Height) -> LinkedList<ChainPair>;
```
xla, line 98: This operation can fail and should likely return a `Result`.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+    fn recover(&mut self, boundary: Height) -> LinkedList<ChainPair>;
+
+    /// Returns the firsts LightBlock above if it exists.
+    fn above(&self, height: Height) -> Option<&LightBlock>;
+    
+    /// Returns the firsts LightBlock below if it exists.
+    fn below(&self, height: Height) -> Option<&LightBlock>;
+    
+    /// Tries to get a certain height.
+    fn get(&self, height: Height) -> Option<&LightBlock>;
```
xla, line 113: I'd model all of these as methods that return `Result`s.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,153 @@
...
+/// same chain, irrespective of the peer from which we received it. The store
+/// provides the invariant that there is a 1-1 mapping from height to a LightBlock.
+trait LightStore {
+    /// Current highest LightBlock which we believe to be from a chain 
+    /// for which this store is associated with.
+    fn highest(&self) -> LightBlock;
+
+    /// Stores the list of ChainPair, possibly changing the highest LightBlock;
+    /// Precondition is the existence of Light Block for which the chain can be linked
+    /// with the already stored ChainPair. Failing if the chain can bot be linked with
```
xla, line 89:
```suggestion
    /// with the already stored ChainPair. Failing if the chain can not be linked with
```

docs/spec/supervisor/supervisor.md
OStevan: There is no semantics as we could use a `Vec` for this, the goal was just to name it here and explain it is going to be used as a stack (push, pop, peek, is_empty) and not with random access.

docs/spec/supervisor/supervisor.md
OStevan: Height in this case is the height of the block which was used to check the light. For example if we inserted a light block `A` into the `Trace` after passing all the checks in verify_to_target using light block B of height `h`. The `ChainPair` for `A` would be (`A`, `h`).

docs/spec/supervisor/supervisor.md
OStevan: Just as a convenience method, can be removed.

docs/spec/supervisor/supervisor.md
OStevan: It's a mistake it should be [`Error`](https://github.com/informalsystems/tendermint-rs/blob/548d76d8b7025f1a3ad1ee9adf32a5e6d24e2197/light-client/src/errors.rs#L18).

docs/spec/supervisor/supervisor.md
OStevan: It's just copied from the current specifications. In current code it's from the [`Io`](https://github.com/informalsystems/tendermint-rs/blob/5e298723f6f60ccf59b60c5bc7c202b338238260/light-client/src/components/io.rs#L71) component.

docs/spec/supervisor/supervisor.md
OStevan: I think it would make more sense to have a `Map<Height, LightBlock>` where keys would be automatically unique.

OStevan:
@xla thanks for the comments. One thing which might not be clear just from the pseudo code and which I was not sure how to capture. This PR only addresses the case when we are doing forward verification without implying how the backward verification case will be done. For example lets say we start from height 1 and want to verify with height 10. One possible trace could be (1<- 5 <- 10) and after verifying 10 we get a request for height 8.

Now this case is not clear for me yet, so I didn't imply anything about the LightStore. For example one can imagine a case where for height 8 we just used the previously fetched light block of height 5. In this case the `LightStore` would basically be comprised of two chains: (this is currently what is expected)
1 <- 5 <- 10
        ^
         - - 8
It does not seem reasonable to just insert 8 between 5 and 10 (1 <- 5 <-8 <- 10) because in the case of validator set change such that skipping trust can not be established between 8 and 10 and somebody wants to replay the verification process just with the `LightBlocks` in the chain. They will not be able to replay the process.

To solve this we would need to also verify 10 starting from 8, however in this case the question is what do we do with the previous information that originally the `LightBlock` of height 10 was checked with 5, do we update it and make one chain 1 <- 5 <- 8 <- 10 or just try and verify as a sanity check.

docs/spec/supervisor/supervisor.md
xla: Is it always the height of `B` -1? In any case the term `Height` is so ubiquitous everywhere that we should encode what height we are referring to here, which in turn an help challenge how it can be constructed.

xla: Still the term `ChainPair` for a pairing of block with height is miss-leading, imho.

docs/spec/supervisor/supervisor.md
xla: For clarification, every height can only exist once? And there is only exactly one `LightBlock` per unique height?

docs/spec/supervisor/supervisor.md
xla: It would greatly benefit if this side-effect is not done in this function, it could be a parameter to be called with.

OStevan: This was an error. The structure is changed a but now.

docs/spec/supervisor/supervisor.md
OStevan: Yes, that was the intention. At each point in time, there can be only one `LightBlock` which can be considered to be from the valid blockchain. Conflict on a single height would mean there is a fork.

docs/spec/supervisor/supervisor.md
OStevan: Made a custom `enum VerificationChain`.

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,164 @@
...
+    loop {
+	    // get the next height
+        let target_height = input();
+        // light_store contains all (LightBlock, previous), for this supervisor all have passed
+        // fork detection and there will be no recoveries, etc.
+		
+		// Verify
+        let mut result = Err(NoPeer);
+        // try with primaries until you succeed or there are no more peers
+        while result.is_err() {
```
melekes, line 138: looks like this will never terminate, making the rest of the code unreachable

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,164 @@
...
+pub fn sequential_supervisor() -> Result<(), Forked> {
+    loop {
+	    // get the next height
+        let target_height = input();
+        // light_store contains all (LightBlock, previous), for this supervisor all have passed
+        // fork detection and there will be no recoveries, etc.
+		
+		// Verify
+        let mut result = Err(NoPeer);
+        // try with primaries until you succeed or there are no more peers
```
melekes, line 137: don't you have a single primary at any moment?

docs/spec/supervisor/supervisor.md
```diff
@@ -0,0 +1,164 @@
...
+    fn get_chain(&self, low: Height, high: Height) -> VerificationChain;
+
+    /// In a case a fork is detected allows the light store to recover for a specified boundary
+    /// height. Erases all light blocks from the store with height greater or equal to boundary.
+    /// Returning the chain of light blocks which are removed. To be used if a fork is discovered
+    /// after insertions of light blocks.
+    fn recover(&mut self, boundary: Height) -> Result<VerificationChain, LightStoreError>;
+
+    /// Returns the firsts LightBlock above if it exists.
+    fn above(&self, height: Height) -> Result<&LightBlock, ListStoreError>;
```
melekes, line 118: why do you need `above` and `below` when you already have `get`?

josef-widder:
The ideas in the PR are great. I tried to incorporate them into https://github.com/tendermint/spec/pull/159. 
I propose that we conclude the discussions there and close this PR.

josef-widder:
Thanks for the discussions. Let's continue that the other PR https://github.com/tendermint/spec/pull/159.

PR #516, opened 2020-08-12
rpc: event subscription management for RPC client

### LINKED ISSUE #313: Event subscription ADR and refactor event_listener module

thanethomson:
Closes #313 

In the spirit of our session today, I'm building and testing this alongside developing ADR #476 at the moment (it diverges slightly from the ADR right now, but I will update the ADR tomorrow).

After several iterations, I think I've come up with a subscription management interface **and** a transport abstraction layer that could mostly work for our purposes.

Since this is a pretty large PR, the quickest way to get into it is to first read the updated crate docs for the `tendermint-rpc` package:

```bash
cd ./rpc/
cargo doc --all-features --open
```

* [x] Wrote tests
* [x] Updated CHANGELOG.md


rpc/Cargo.toml
```diff
@@ -25,7 +25,7 @@ all-features = true
 
 [features]
 default = []
-client = [ "async-tungstenite", "futures", "http", "hyper", "tokio" ]
+client = [ "async-tungstenite", "futures", "http", "hyper", "tokio", "async-trait" ]
```
xla, line 28:
```suggestion
client = [ "async-trait", "async-tungstenite", "futures", "http", "hyper", "tokio" ]
```

rpc/Cargo.toml
```diff
@@ -42,4 +42,5 @@ async-tungstenite = { version="0.5", features = ["tokio-runtime"], optional = tr
 futures = { version = "0.3", optional = true }
 http = { version = "0.2", optional = true }
 hyper = { version = "0.13", optional = true }
-tokio = { version = "0.2", features = ["macros"], optional = true }
+tokio = { version = "0.2", features = ["macros", "fs"], optional = true }
+async-trait = { version = "0.1.36", optional = true }
```
xla, line 46: Let's keep some alphanumeric order, also newlines.

xla, line 46: We usually don't pin on patch versions, a `0.1` should siffice here.

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
 //! Tendermint RPC client
 
-use bytes::buf::ext::BufExt;
-use hyper::header;
+use tendermint::{
+    abci::{self, Transaction},
+    block::Height,
+    evidence::Evidence,
+    net, Genesis,
+};
```
xla, line 8: This import rewrites don't improve editability and distract from the core changes of the PR.

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
...
-use tendermint::block::Height;
-use tendermint::evidence::Evidence;
-use tendermint::net;
-use tendermint::Genesis;
-
-use crate::{endpoint::*, Error, Request, Response};
+use crate::{
+    client::{
+        subscription::SubscriptionManager,
+        transport::{http_ws::HttpWsTransport, Transport},
```
xla, line 13: Still quite suprised to find these nested imports, they make it hard to edit them and don't help with readability either. I'd like to stick to the import style the crate has followed so far.

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
...
+    endpoint::*,
+    Error, Request, Response,
+};
 
 pub mod event_listener;
+pub mod subscription;
+pub mod transport;
+
+#[cfg(test)]
+pub mod testing;
```
xla, line 24:
```suggestion
pub mod test;
```

Goes in line with the other concise mod namings.

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
...
 
 /// Tendermint RPC client.
 ///
 /// Presently supports JSONRPC via HTTP.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
```
xla, line 29: What prevents us to keep `Clone` on the client?

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
...
+    client::{
+        subscription::SubscriptionManager,
+        transport::{http_ws::HttpWsTransport, Transport},
+    },
+    endpoint::*,
+    Error, Request, Response,
+};
 
 pub mod event_listener;
+pub mod subscription;
```
xla, line 20: The `use` statements should be following the mod declarations, makes it easier to correlate them.

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,206 @@
...
+    collections::HashMap,
+    pin::Pin,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+use tokio::{stream::StreamExt, sync::mpsc, task::JoinHandle};
+
+use crate::{client::transport::EventConnection, events::Event, Error};
+
+/// The subscription manager is just an interface/handle to the subscription
+/// router, which runs asynchronously in a separate process.
```
xla, line 17:
```suggestion
/// The subscription manager is an interface to the subscription
/// router, which runs asynchronously in a separate process.
```

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,206 @@
...
+        id: SubscriptionId,
+        query: String,
+        event_tx: mpsc::Sender<Event>,
+    },
+    Unsubscribe(Subscription),
+    Terminate,
+}
+
+impl SubscriptionManager {
+    pub(crate) fn new(conn: EventConnection, cmd_buf_size: usize) -> Self {
```
xla, line 58:
```suggestion
    pub fn new(conn: EventConnection, cmd_buf_size: usize) -> Self {
```

Unless explicirely reexported this method would be visible outside of the client mod.

rpc/src/events.rs
```diff
@@ -0,0 +1,74 @@
+//! RPC subscription event-related data structures.
```
xla, line 1: All other mods are singular, `event` is just as fine.

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
...
 
 /// Tendermint RPC client.
 ///
 /// Presently supports JSONRPC via HTTP.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Client {
-    /// Address of the RPC server
-    address: net::Address,
+    transport: Box<dyn Transport>,
```
xla, line 31: Why do we need to `Box` this instead of having it as a [type parameter](https://doc.rust-lang.org/reference/types/parameters.html)?

rpc/src/client.rs
```diff
@@ -1,31 +1,45 @@
...
 }
 
 impl Client {
-    /// Create a new Tendermint RPC client, connecting to the given address
-    pub fn new(address: net::Address) -> Self {
-        Self { address }
+    /// Create a new Tendermint RPC client, connecting to the given address.
+    /// By default this uses the [`HttpWsTransport`] transport layer. This
+    /// transport lazily initializes subscription mechanisms when first
+    /// subscribing to events generated by a particular query.
```
xla, line 38: If we have a default config we should use [Default](https://doc.rust-lang.org/std/default/trait.Default.html) for that and make it possible to use the new constructor to pass a transport. This will help with the right amount of composability.

rpc/src/client.rs
```diff
@@ -159,41 +173,137 @@ impl Client {
         self.perform(evidence::Request::new(e)).await
     }
 
+    /// Creates a subscription management interface for this RPC client. This
+    /// interface facilitates subscribing and unsubscribing from receiving
+    /// events produced by specific RPC queries.
+    pub async fn new_subscription_manager(
```
xla, line 179: It's surpising to see a constructor for a "foreign" - as in from another mod - on this entity. It might be better to have a `TryFrom` implementation on `SubscriptionManager` for `Client`.

rpc/src/client.rs
```diff
@@ -159,41 +173,137 @@ impl Client {
...
+    }
+
     /// Perform a request against the RPC endpoint
     pub async fn perform<R>(&self, request: R) -> Result<R::Response, Error>
     where
         R: Request,
     {
         let request_body = request.into_json();
+        let response_body = self.transport.request(request_body).await?;
+        R::Response::from_string(response_body)
```
xla, line 194: What changed that we need to do this `from_string` call?

rpc/src/client.rs
```diff
@@ -159,41 +173,137 @@ impl Client {
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::Method;
+    use testing::matching_transport::{MethodMatcher, RequestMatchingTransport};
+
+    // TODO: Read from a fixture in the crate.
```
xla, line 210:
```suggestion
    // TODO(thane): Read from a fixture in the crate.
```

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,206 @@
...
+    cmd_tx: mpsc::Sender<RouterCmd>,
+    next_subs_id: AtomicUsize,
+}
+
+/// An interface that can be used to asynchronously receive events for a
+/// particular subscription.
+#[derive(Debug)]
+pub struct Subscription {
+    id: SubscriptionId,
+    query: String,
```
xla, line 30: Should this not be a native `Query` type?

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,206 @@
...
+// routing incoming events to their relevant subscribers.
+struct SubscriptionRouter {
+    conn: EventConnection,
+    cmd_rx: mpsc::Receiver<RouterCmd>,
+    // Maps queries -> (maps of subscription IDs -> channels)
+    subscriptions: HashMap<String, HashMap<SubscriptionId, mpsc::Sender<Event>>>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct SubscriptionId(usize);
```
xla, line 44:
```suggestion
pub struct SubscriptionId(usize);
```

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,206 @@
...
+    Subscribe {
+        id: SubscriptionId,
+        query: String,
+        event_tx: mpsc::Sender<Event>,
+    },
+    Unsubscribe(Subscription),
+    Terminate,
+}
+
+impl SubscriptionManager {
```
xla, line 57: Let's keep the `impl`s close to their types.

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    async fn route_event(&mut self, ev: Event) {
+        let subs_for_query = match self.subscriptions.get_mut(&ev.query) {
+            Some(s) => s,
+            None => return,
+        };
+        let mut disconnected = Vec::new();
+        for (subs_id, tx) in subs_for_query {
+            // TODO: Right now we automatically remove any disconnected or full
```
xla, line 156:
```suggestion
            // TODO(thane): Right now we automatically remove any disconnected or full
```

rpc/src/client/transport.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// An `EventConnection` allows us to send subscribe/unsubscribe requests via
+/// the transport layer, as well as receive incoming events from subscriptions.
+#[derive(Debug)]
+pub struct EventConnection {
+    // The `EventConnection` struct is a workaround for the fact that we need
+    // to use `Transport` as a trait object, but trait objects are not allowed
+    // to use generics in their method signatures.
+    pub(crate) transport: Box<dyn SubscriptionTransport>,
+    pub(crate) event_producer: EventProducer,
```
xla, line 50:
```suggestion
    pub transport: Box<dyn SubscriptionTransport>,
    pub event_producer: EventProducer,
```

thanethomson:
> Had a first pass, left a bunch of comments regarding structure and style inline.
> 
> Generally this shapes up to be quite nice. The main concern I'd like to raise is that it seems a rather big amount of abstractions. There are a wide array of indirections and things that have to be plugged together to make the machinery work, i.e. Manager, Router, Subscription, SubscriptionTransport, EventConnection, EventProducer and so on. Another indicator that it's not yet the entirely correct set of abstractions is if you follow the imports it looks like deeply nested modules import entities from their top-level module. It would be nice to come to a more concise and clear hierarchy. Happy to discuss in more detail what I think could be reduced. Generally it would help to think along the lines of what needs to be injected/configurable for different environments.

Thanks @xla! I actually had an idea this morning during our onboarding as to how to refactor this a bit to reduce the number of abstractions. For one thing, right now there are effectively two asynchronous processes (outside of the originating task) facilitating the subscription mechanism:

1. A `SubscriptionRouter`, whose only job is to listen for incoming `Event`s and route them to `Subscription`s.
2. A transport driver (in the standard case, a `WsDriver` for the WebSocket connection).

I've realized I can probably reduce this logic to just one async task (the `WsDriver`) by embedding the routing logic, and I've got some ideas as to how to do this elegantly.

rpc/src/client.rs
thanethomson: Sure, I'll change them. My IDE was doing weird things, mixing the two styles, so I just went with this style. I'll reformat them 👍

rpc/src/client.rs
thanethomson: I considered the implications of having it as a type parameter as opposed to a trait object and I'm not sure which of the two approaches provides the best ergonomics in the long run. Of course, using trait objects (which seem to be used pretty extensively in the Light Client module) results in simpler higher-level interfaces, they come with their own challenges.

I can certainly experiment with the type parameter-based approach instead. It would certainly help to be able to use generics in the `Transport` trait, for example.

rpc/src/client/subscription.rs
thanethomson: I actually had implemented a reasonable-looking query implementation in one of my previous iterations, so I'll bring it back in here. I just wanted some feedback on the overarching architecture first prior to adding in more details.

rpc/src/client.rs
thanethomson: We'd need to be able to clone the transport (trait object), which I thought may not make sense for all possible transports.

For example, what would the cloning implications be for a purely WebSocket-based client? Would it be useful to clone such a client, and what would it mean to do so? (i.e. do we open a new connection, or do we provide another interface to the same underlying connection?)

rpc/src/client.rs
thanethomson: The trouble with the `TryFrom` approach is that it's not `async`.

We need a way to have the `SubscriptionClient` (the new interface name for the `SubscriptionManager` now) be produced from the `Client` (since in certain transports the `SubscriptionClient` will be tightly coupled to the `Client`), and doing so involves creating a WebSocket connection if you use `HttpTransport` at present, which is `async`.

Is there a good `async` alternative to `TryFrom`? I'm still working on my `async`/Rust right now 😁

thanethomson:
While I work on updating #476, the quickest way to get a feel for the new architecture here is to clone this branch locally and take a look at the docs:

```bash
cd ./rpc/
cargo doc --all-features --open
```

thanethomson:
I realized that the reason the `test-integration-latest` was hanging had to do with the fact that my WebSocket client implementation wasn't doing any checks on the incoming serialization formats from the RPC. I've now pushed some changes that:

1. Check the response from the remote endpoint to see if a subscription request was actually successful.
2. Check that responses for a specific subscription can be deserialized, and if not, produce an error through the `Subscription` interface.

rpc/src/error.rs
```diff
@@ -196,6 +209,7 @@ impl From<Code> for i32 {
         match code {
             Code::HttpError => 0,
             Code::WebSocketError => 1,
+            Code::ClientError => 2,
```
liamsi, line 212: Suggestion: this error should probably have a doc comment as it is "special" in the sense as it does not have a counterpart in golang.

thanethomson: Yeah I was wondering whether or not to create a new error here. I see the `InternalError` code is reserved for RPC server internal errors, but needed to have something to represent an internal error specific to the client. I'd say it's better to differentiate between those two kinds of "internal" errors - what do you think @liamsi ?

thanethomson: If we keep this then I may rename it to `ClientInternalError` and I'll add a doc comment indicating it doesn't have a counterpart in the Golang client.

xla (dismissed):
Had another pass, this iteration is already a great improvement towards a simpler API. Besides the comment about the client traits, I still think there is too much machinery around, which tries to serve the use-case when people want to implement custom solutions. I'd be in favour of reducing the amount of moving parts even further and work mostly towards one specific implementation to see how the generic interfaces could look like.

rpc/src/client.rs
```diff
@@ -76,124 +89,133 @@ impl Client {
...
+    async fn close(self) -> Result<()>;
+}
+
+/// A `FullClient` is one that augments a [`MinimalClient`] functionality with
+/// subscription capabilities.
+///
+/// [`MinimalClient`]: trait.MinimalClient.html
+///
+#[async_trait]
+pub trait FullClient: MinimalClient {
```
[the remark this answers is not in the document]
thanethomson: I've now refactored this to split out the non-subscription functionality (into a `Client` trait) from the subscription functionality (into a `SubscriptionClient` trait). In the current implementation, `SubscriptionClient` does not require `Client` to be implemented.

Is that the kind of change you were looking for?

rpc/src/client.rs
```diff
@@ -76,124 +89,133 @@ impl Client {
...
+    /// `/subscribe`: subscribe to receive events produced by the given query.
+    ///
+    /// Allows for specification of the `buf_size` parameter, which determines
```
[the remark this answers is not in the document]
thanethomson: Do you think we should only be exposing unbounded channels?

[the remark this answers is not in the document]
thanethomson: I've actually removed the bounded channels now (it's just an implementation detail now). All subscriptions now use unbounded channels under the hood, and if anyone strongly desires a bounded channel we can implement them super easily (I've created a `ChannelTx` and `ChannelRx` struct that abstracts away the underlying channel implementation).

rpc/src/request.rs
```diff
@@ -38,13 +38,30 @@ impl<R> Wrapper<R>
...
+    ///
+    /// By default this sets the ID of the request to a random [UUIDv4] value.
+    ///
+    /// [UUIDv4]: https://en.wikipedia.org/wiki/Universally_unique_identifier
     pub fn new(request: R) -> Self {
+        Wrapper::new_with_id(Id::uuid_v4(), request)
+    }
+
+    /// Create a new request wrapper with a custom JSONRPC request ID.
+    pub fn new_with_id(id: Id, request: R) -> Self {
```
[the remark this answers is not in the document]
thanethomson: Since the ID of incoming events to the client correlates with the ID of the JSON-RPC wrapper for the initial subscription request, I initially needed this constructor to be able to control that ID.

I've since refactored this in a way that we no longer need this constructor.

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,591 @@
...
+            query,
+            event_rx,
+            terminate_tx,
+        }
+    }
+
+    /// Gracefully terminate this subscription.
+    ///
+    /// This can be called from any asynchronous context. It only returns once
+    /// it receives confirmation of termination.
```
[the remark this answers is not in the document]
thanethomson: It consumes the `Subscription`, so there's no way it can be called from two contexts simultaneously. I'll reword this to make it clearer - what I actually meant was that the `Subscription` can be moved to any async context and it can be terminated within that context.

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,591 @@
...
+    result_tx: ChannelTx<Result<()>>,
+}
+
+/// The current state of a subscription.
+#[derive(Debug, Clone, PartialEq)]
+pub enum SubscriptionState {
+    Pending,
+    Active,
+    Cancelling,
+    NotFound,
```
melekes, line 205: what does not found mean in this context? how's subscription state can be not found?

thanethomson: Good point! It probably makes more sense to make this method return an optional subscription state: https://github.com/informalsystems/tendermint-rs/pull/516/files#diff-89fca07be889575fefb9e77a8de9051dR394

romac:
@thanethomson Just pushed a fix for the last remaining WASM build error.

thanethomson:
> @thanethomson Just pushed a fix for the last remaining WASM build error.

Awesome, thanks @romac! 👍

ebuchman (commented):
Looks like great work. Finally got a chance to at least read through it all but still some pieces I want to look more closely at.

While the ADR diagram is helpful for outlining the components, I think an even more helpful diagram might be one of the concurrency, showing what the different threads are and how the channel connectivity works and who's sending msgs to whom

light-client/Cargo.toml
```diff
@@ -28,7 +28,7 @@ secp256k1 = ["tendermint/secp256k1", "tendermint-rpc/secp256k1"]
 
 [dependencies]
 tendermint = { version = "0.16.0", path = "../tendermint" }
-tendermint-rpc = { version = "0.16.0", path = "../rpc", default-features = false }
+tendermint-rpc = { version = "0.16.0", path = "../rpc", features = ["client", "transport_http"] }
```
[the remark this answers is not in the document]
thanethomson: Fixed this 👍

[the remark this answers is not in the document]
thanethomson: As for the `transport_http` feature, I'm not sure. Are there instances where someone may want to exclusively use WebSocket-based subscription, but not perform any other requests via the HTTP interface?

Also, when we implement a WebSocket-based RPC client for the rest of the RPC methods (i.e. those currently only supported via HTTP), would we want to include the HTTP client by default?

light-client/src/components/io.rs
```diff
@@ -172,10 +175,11 @@ mod prod {
             }
         }
 
+        // TODO(thane): Generalize over client transport (instead of using HttpClient directly).
         #[pre(self.peer_map.contains_key(&peer))]
-        fn rpc_client_for(&self, peer: PeerId) -> rpc::Client {
+        fn rpc_client_for(&self, peer: PeerId) -> Result<rpc::HttpClient, IoError> {
             let peer_addr = self.peer_map.get(&peer).unwrap().to_owned();
-            rpc::Client::new(peer_addr)
+            Ok(rpc::HttpClient::new(peer_addr).map_err(IoError::from)?)
```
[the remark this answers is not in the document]
thanethomson: Technically a Tendermint `Address` could contain either a TCP or a Unix socket address, and we need the host/port details in order to connect to the RPC endpoint.

We could do away with the need for an error here by requiring host/port details instead of a `peer_addr`, or we could just panic if a user supplies a Unix socket address?

light-client/src/components/io.rs
```diff
@@ -172,10 +175,11 @@ mod prod {
             }
         }
 
+        // TODO(thane): Generalize over client transport (instead of using HttpClient directly).
```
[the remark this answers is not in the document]
thanethomson: The idea here was to allow the developer to replace the HTTP client with the mock client and simulate responses.

Also, we should eventually implement support for fully WebSocket-based RPC interactions.

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,591 @@
...
+}
+
+/// Provides a mechanism for tracking [`Subscription`]s and routing [`Event`]s
+/// to those subscriptions.
+///
+/// [`Subscription`]: struct.Subscription.html
+/// [`Event`]: ./event/struct.Event.html
+#[derive(Debug)]
+pub struct SubscriptionRouter {
+    subscriptions: HashMap<String, HashMap<SubscriptionId, ChannelTx<Result<Event>>>>,
```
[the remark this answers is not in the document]
thanethomson: Yeah, it's the query string. I'll add a comment here 👍

romac:
For some reason, I cannot reply [inline in the PR](https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r488324084), so here it is:

> @ebuchman Do we have any use cases for non http transport? I'm not sure anything else is even supported by Tendermint. I suppose it could be https but not sure if that would need entirely new transport...

Yes, we might also want to have a [`Fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)-based transport for running the light client against a full node directly from the browser.

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,591 @@
...
+                Ok(pending_subscribe.result_tx.send(Ok(())).await?)
+            }
+            None => Ok(()),
+        }
+    }
+
+    /// Attempts to cancel the pending subscription with the given ID, sending
+    /// the specified error to the original creator of the attempted
+    /// subscription.
+    pub async fn cancel_add(&mut self, req_id: &str, err: impl Into<Error>) -> Result<()> {
```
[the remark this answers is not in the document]
thanethomson: This is due to the possibility of concurrency issues when managing multiple subscriptions.

You could create one subscription and start receiving events for it, and then create a second subscription whose success/failure response from the RPC may only come back after an event relating to the first subscription.

Since events and subscription responses could be interspersed, and because subscriptions could possibly fail (implying the response from the RPC is important and we can't discard it), we need to facilitate a two-phase subscribe and unsubscribe process.

I'll see if I can add this into the ADR in a separate PR 👍

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,591 @@
...
+        };
+        let mut disconnected = Vec::<SubscriptionId>::new();
+        for (id, event_tx) in subs_for_query {
+            if event_tx.send(Ok(ev.clone())).await.is_err() {
+                disconnected.push(id.clone());
+            }
+        }
+        let subs_for_query = self.subscriptions.get_mut(&ev.query).unwrap();
+        for id in disconnected {
+            subs_for_query.remove(&id);
```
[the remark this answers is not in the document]
thanethomson: My assumption here (not sure if it's totally correct) is that the only way a subscription will make it into the `disconnected` list is if we failed to send the event to it, implying that it's been disconnected on the other side of the channel (i.e. it's been dropped).

tendermint/src/abci/responses.rs
```diff
@@ -91,7 +91,7 @@ pub struct Event {
 ///
 /// <https://github.com/tendermint/tendermint/blob/develop/abci/types/types.proto>
 // TODO(tarcieri): generate this automatically from the proto
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
```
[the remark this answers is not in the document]
thanethomson: I'm using it for testing purposes (comparing expected events to actual ones).

rpc/src/client/transport/websocket.rs
```diff
@@ -0,0 +1,703 @@
...
+/// driver that continuously interacts with the actual WebSocket connection.
+/// The `WebSocketSubscriptionClient` itself is effectively just a handle to
+/// this driver. This driver is spawned as the client is created.
+///
+/// To terminate the client and the driver, simply use its [`close`] method.
+///
+/// ## Examples
+///
+/// ```rust,ignore
+/// use tendermint_rpc::{WebSocketSubscriptionClient, SubscriptionClient, ClosableClient};
```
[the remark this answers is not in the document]
thanethomson: I'd love to have a way to do away with these extra imports, just not sure how. @romac do you perhaps know of some kind of trick to consolidate these imports? (I can only think of hairy macro-based ones)

rpc/src/client/subscription.rs
```diff
@@ -0,0 +1,591 @@
...
+        match self.pending_subscribe.remove(req_id) {
+            Some(mut pending_subscribe) => {
+                self.add(
+                    &pending_subscribe.id,
+                    pending_subscribe.query.clone(),
+                    pending_subscribe.event_tx,
+                );
+                Ok(pending_subscribe.result_tx.send(Ok(())).await?)
+            }
+            None => Ok(()),
```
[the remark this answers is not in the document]
thanethomson: My thinking here was that there are two kinds of problems that could arise in this method given the nature of its input and where/how it's used:

1. We try to confirm a subscription but we fail to do so (a serious failure that needs handling).
2. We get a subscription confirmation for a non-existent subscription (something that could just be ignored).

rpc/src/client/transport/websocket.rs
```diff
@@ -0,0 +1,703 @@
...
+        let query = req.params().query.clone();
+        if let Err(e) = self
+            .send(Message::Text(serde_json::to_string_pretty(&req).unwrap()))
+            .await
+        {
+            let _ = result_tx.send(Err(e)).await;
+            return Ok(());
+        }
+        self.router
+            .pending_add(id.as_ref(), &id, query, event_tx, result_tx);
```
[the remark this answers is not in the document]
thanethomson: This was technically an implementation of the [`AsRef` trait](https://doc.rust-lang.org/std/convert/trait.AsRef.html) to provide a `&str` reference, but looking at it now it's probably somewhat ambiguous. I've now implemented an `as_str()` method to return a `&str` instead, which I think should be clearer.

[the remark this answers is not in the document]
ebuchman: But it's also a bit jarring that these two distinct Ids are actually the same. Is that intended? Maybe could be documented/clarified somewhere the subscription id will be drawn from the json-rpc id ?

rpc/src/client/event_listener.rs
```diff
@@ -1,216 +0,0 @@
-//! Tendermint Websocket event listener client
```
ebuchman, line 1: Maybe we should bring this back for now so folks can upgrade their websocket code at their own pace?

rpc/src/client/transport/websocket.rs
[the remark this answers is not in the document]
thanethomson: Originally I had them both be exactly the same, but then I ran into an issue where I needed to send unsubscribe requests with their own request ID (distinct from the subscription ID) to ensure that the response from the remote endpoint is, in fact, exclusively relevant to the unsubscribe request. Otherwise we may confuse an error meant for the subscription with one meant for the unsubscribe request.

So subscribe requests' IDs are the same as the subscription ID (to correlate incoming events, and possibly errors, relating to the subscription), but unsubscribe requests' IDs need to be distinct.

rpc/src/client/transport/websocket.rs
```diff
@@ -0,0 +1,701 @@
...
+            tokio::select! {
+                Some(res) = self.stream.next() => match res {
+                    Ok(msg) => self.handle_incoming_msg(msg).await?,
+                    Err(e) => return Err(
+                        Error::websocket_error(
+                            format!("failed to read from WebSocket connection: {}", e),
+                        ),
+                    ),
+                },
+                Some(cmd) = self.cmd_rx.recv() => match cmd {
```
ebuchman, line 194: I find it a bit confusing that subscriptions come over the cmd_rx channel but the unsubscriptions come over a separate terminate channel. Why not give the Susbcription objects the cmd_tx and let them send Unsubscribe commands over that?

light-client/Cargo.toml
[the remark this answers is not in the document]
thanethomson: So the `HttpClient` requires the use of the `hyper` and `http` crates, whereas the WebSocket-based subscription client doesn't.

If you'd like to simplify, can I suggest doing away with the `subscription` feature and shortening the feature names:

| Flag | Description | Additional Dependencies |
| ---- | ------------ | ------------------------- |
| `default` | Just the RPC types | None |
| `client` | All the client-related traits, but no implementations | `async-trait`, `futures` |
| `http` | `HttpClient` (implements `Client`) | `http`, `hyper`, `tokio` |
| `websocket` | `WebSocketSubscriptionClient`, and in future a `WebSocketClient`, which will implement the `Client` and `SubscriptionClient` traits and will supersede the `WebSocketSubscriptionClient` | `async-tungstenite`, `tokio` |
| `mock` | `MockClient` and `MockSubscriptionClient` | `tokio` |

Perhaps it's worth it here to rename the existing `WebSocketSubscriptionClient` to `WebSocketClient`, so its name doesn't change in future?

Also, we could do away with the mock clients if you want. They were useful in my testing, but it's up to downstream users (e.g. IBC) to say if they'd see value in such a mock client.

rpc/src/client/transport/websocket.rs
```diff
@@ -0,0 +1,701 @@
...
+        Self {
+            stream,
+            router: SubscriptionRouter::default(),
+            cmd_rx,
+            terminate_rx,
+        }
+    }
+
+    async fn run(mut self) -> Result<()> {
+        // TODO(thane): Should this loop initiate a keepalive (ping) to the
```
ebuchman, line 182: I believe so. At least the [Go client does this](https://github.com/tendermint/tendermint/blob/master/rpc/jsonrpc/client/ws_client.go#L421) and I think it's proper behaviour for a websocket client

rpc/src/client/transport/websocket.rs
```diff
@@ -0,0 +1,701 @@
...
+        //              server on a regular basis?
+        loop {
+            tokio::select! {
+                Some(res) = self.stream.next() => match res {
+                    Ok(msg) => self.handle_incoming_msg(msg).await?,
```
ebuchman, line 187: So I'm trying to understand what benefit we're get from using async/await, if any. It seems that pretty much every possibly async function call has an await immediately after, so everything in this `run` method is effectively happening synchronously. Is that true? It seems it may apply more broadly than just this run method, but since this runs in it's own thread and is relatively contained, it's easier to analyze.

I was also somewhat surprised to see the use of an explicit thread for this when we were using async/await for everything since I might have naively thought that the runtime would manage threads for us and we could specify through our use of .await where we wanted to block.

But it's possible I'm not thinking about this right, these are just initial thoughts.

ebuchman (commented):
A few more thoughts while I continue to dive into this.

If we bring back the old event listener, then I think it's fine to merge this (ie. it should be mostly non-breaking) and then we can continue to review and improve it from master

rpc/src/client/event_listener.rs
thanethomson: The trouble with bringing this back is that the types exposed by the event listener conflict with the types I've added in `rpc/src/event.rs`. As far as I can see in IBC it's just this code that needs to be updated: https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/event_monitor.rs

ebuchman: Ok fair enough. Let's merge this as is and once ibc is ready to upgrade to our master we can work on integration there.

ebuchman (approved):
Great work. Let's merge this and open a follow up issue to track some fixes (eg adding pings) and any future work (refactor of APIs or internals)

light-client/Cargo.toml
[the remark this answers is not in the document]
thanethomson: For posterity, we've simplified this even further now, as per #569

romac:
Great stuff @thanethomson! 💯

PR #522, opened 2020-08-13
Replace `signatory` with `ed25519-dalek` and `k256` crates

tony-iqlusion:
Native support for the `signature` traits used by the `signatory*` was recently added to `ed25519-dalek` in v1.0.0-pre.4, and the `k256` crate recently added a native pure Rust implementation of ECDSA/secp256k1.

Together these changes eliminate the need for the `signatory*` wrapper crates. Additionally, `k256` eliminates the need for the rust-secp256k1 C-based library, which has been problematic with e.g. WASM (#391).

This commit eliminates all `signatory` dependencies and uses `ed25519-dalek` and `k256` directly instead. Both of these were already dependencies, so it's not adding any additional dependencies, but instead removes several of them.

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -27,16 +27,17 @@ authors = [
 
 [package.metadata.docs.rs]
 all-features = true
-
-[badges]
-codecov = { repository = "..."}
+rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 anomaly = "0.2"
 async-trait = "0.1"
 bytes = "0.5"
 chrono = { version = "0.4", features = ["serde"] }
+ed25519 = "1"
+ed25519-dalek = { version = "= 1.0.0-pre.4", features = ["serde"] }
 futures = "0.3"
+k256 = { version = "0.4", optional = true, features = ["ecdsa"] }
 once_cell = "1.3"
 prost-amino = "0.6"
 prost-amino-derive = "0.6"
@@ -45,9 +46,7 @@ serde_json = "1"
 serde_bytes = "0.11"
 serde_repr = "0.1"
 sha2 = { version = "0.9", default-features = false }
-signatory = { version = "0.20", features = ["ed25519", "ecdsa"] }
-signatory-dalek = "0.20"
-signatory-secp256k1 = { version = "0.20", optional = true }
+signature = "1.2"
 subtle = "2"
 subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
 tai64 = { version = "3", features = ["chrono"] }
@@ -61,4 +60,4 @@ tendermint-rpc = { path = "../rpc", features = [ "client" ] }
 tokio = { version = "0.2", features = [ "macros" ] }
 
 [features]
-secp256k1 = ["signatory-secp256k1", "ripemd160"]
+secp256k1 = ["k256", "ripemd160"]
--- a/tendermint/src/account.rs
+++ b/tendermint/src/account.rs
@@ -1,19 +1,24 @@
 //! Tendermint accounts
 
-use crate::error::{Error, Kind};
-#[cfg(feature = "secp256k1")]
-use ripemd160::Ripemd160;
+use crate::{
+    error::{Error, Kind},
+    public_key::Ed25519,
+};
+
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 use sha2::{Digest, Sha256};
-#[cfg(feature = "secp256k1")]
-use signatory::ecdsa::curve::secp256k1;
-use signatory::ed25519;
 use std::{
+    convert::TryInto,
     fmt::{self, Debug, Display},
     str::FromStr,
 };
 use subtle::{self, ConstantTimeEq};
 use subtle_encoding::hex;
+
+#[cfg(feature = "secp256k1")]
+use crate::public_key::Secp256k1;
+#[cfg(feature = "secp256k1")]
+use ripemd160::Ripemd160;
 
 /// Size of an  account ID in bytes
 pub const LENGTH: usize = 20;
@@ -64,8 +69,8 @@ impl Debug for Id {
 
 // RIPEMD160(SHA256(pk))
 #[cfg(feature = "secp256k1")]
-impl From<secp256k1::PublicKey> for Id {
-    fn from(pk: secp256k1::PublicKey) -> Id {
+impl From<Secp256k1> for Id {
+    fn from(pk: Secp256k1) -> Id {
         let sha_digest = Sha256::digest(pk.as_bytes());
         let ripemd_digest = Ripemd160::digest(&sha_digest[..]);
         let mut bytes = [0u8; LENGTH];
@@ -75,12 +80,10 @@ impl From<secp256k1::PublicKey> for Id {
 }
 
 // SHA256(pk)[:20]
-impl From<ed25519::PublicKey> for Id {
-    fn from(pk: ed25519::PublicKey) -> Id {
+impl From<Ed25519> for Id {
+    fn from(pk: Ed25519) -> Id {
         let digest = Sha256::digest(pk.as_bytes());
-        let mut bytes = [0u8; LENGTH];
-        bytes.copy_from_slice(&digest[..LENGTH]);
-        Id(bytes)
+        Id(digest[..LENGTH].try_into().unwrap())
     }
 }
 
@@ -141,7 +144,7 @@ mod tests {
         let id_bytes = Id::from_str(id_hex).expect("expected id_hex to decode properly");
 
         // get id for pubkey
-        let pubkey = ed25519::PublicKey::from_bytes(pubkey_bytes).unwrap();
+        let pubkey = Ed25519::from_bytes(pubkey_bytes).unwrap();
         let id = Id::from(pubkey);
 
         assert_eq!(id_bytes.ct_eq(&id).unwrap_u8(), 1);
@@ -160,7 +163,7 @@ mod tests {
         let id_bytes = Id::from_str(id_hex).expect("expected id_hex to decode properly");
 
         // get id for pubkey
-        let pubkey = secp256k1::PublicKey::from_bytes(pubkey_bytes).unwrap();
+        let pubkey = Secp256k1::from_bytes(pubkey_bytes).unwrap();
         let id = Id::from(pubkey);
 
         assert_eq!(id_bytes.ct_eq(&id).unwrap_u8(), 1);
--- a/tendermint/src/amino_types/ed25519.rs
+++ b/tendermint/src/amino_types/ed25519.rs
@@ -1,8 +1,13 @@
 use super::compute_prefix;
-use crate::public_key::PublicKey;
+use crate::{
+    error,
+    public_key::{Ed25519, PublicKey},
+    Error,
+};
+use anomaly::format_err;
 use once_cell::sync::Lazy;
 use prost_amino_derive::Message;
-use signatory::ed25519::PUBLIC_KEY_SIZE;
+use std::convert::TryFrom;
 
 // Note:On the golang side this is generic in the sense that it could everything that implements
 // github.com/tendermint/tendermint/crypto.PubKey
@@ -24,13 +29,15 @@ pub struct PubKeyResponse {
 #[amino_name = "tendermint/remotesigner/PubKeyRequest"]
 pub struct PubKeyRequest {}
 
-impl From<PubKeyResponse> for PublicKey {
+impl TryFrom<PubKeyResponse> for PublicKey {
+    type Error = Error;
+
     // This does not check if the underlying pub_key_ed25519 has the right size.
     // The caller needs to make sure that this is actually the case.
-    fn from(response: PubKeyResponse) -> PublicKey {
-        let mut public_key = [0u8; PUBLIC_KEY_SIZE];
-        public_key.copy_from_slice(response.pub_key_ed25519.as_ref());
-        PublicKey::Ed25519(signatory::ed25519::PublicKey::new(public_key))
+    fn try_from(response: PubKeyResponse) -> Result<PublicKey, Error> {
+        Ed25519::from_bytes(&response.pub_key_ed25519)
+            .map(Into::into)
+            .map_err(|_| format_err!(error::Kind::InvalidKey, "malformed Ed25519 key").into())
     }
 }
 
@@ -49,7 +56,9 @@ impl From<PublicKey> for PubKeyResponse {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use ed25519_dalek::PUBLIC_KEY_LENGTH;
     use prost_amino::Message;
+    use std::convert::TryInto;
 
     #[test]
     fn test_empty_pubkey_msg() {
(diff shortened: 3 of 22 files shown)


tendermint/src/amino_types/ed25519.rs
```diff
@@ -140,21 +149,21 @@ mod tests {
...
     #[test]
     fn test_into() {
-        let raw_pk: [u8; PUBLIC_KEY_SIZE] = [
-            0x79, 0xce, 0xd, 0xe0, 0x43, 0x33, 0x4a, 0xec, 0xe0, 0x8b, 0x7b, 0xb5, 0x61, 0xbc,
-            0xe7, 0xc1, 0xd4, 0x69, 0xc3, 0x44, 0x26, 0xec, 0xef, 0xc0, 0x72, 0xa, 0x52, 0x4d,
-            0x37, 0x32, 0xef, 0xed,
+        let raw_pk: [u8; PUBLIC_KEY_LENGTH] = [
+            0xaf, 0xf3, 0x94, 0xc5, 0xb7, 0x5c, 0xfb, 0xd, 0xd9, 0x28, 0xe5, 0x8a, 0x92, 0xdd,
+            0x76, 0x55, 0x2b, 0x2e, 0x8d, 0x19, 0x6f, 0xe9, 0x12, 0x14, 0x50, 0x80, 0x6b, 0xd0,
+            0xd9, 0x3f, 0xd0, 0xcb,
```
tony-iqlusion, line 155: Ed(wards)25519 point decompression was failing for this? (the old implementation didn't attempt to decompress points)

tendermint/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 //! blockchain networks, including chain information types, secret connections,
 //! and remote procedure calls (JSONRPC).
 
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
tony-iqlusion, line 7: I wired up `doc_cfg`, which allows you to document which Cargo features need to be enabled to use certain types/methods in rustdoc.

Rendered example:

Screen Shot 2020-08-12 at 8 26 45 PM

tendermint/src/private_key.rs
```diff
@@ -36,60 +38,31 @@ impl PrivateKey {
     }
 }
 
-/// Ed25519 keypairs
-#[derive(Zeroize)]
-#[zeroize(drop)]
-pub struct Ed25519Keypair([u8; ED25519_KEYPAIR_SIZE]);
```
tony-iqlusion, line 42: I was able to replace this with ed25519-dalek's own `ed25519_dalek::Keypair` type, which validates keys.

tendermint/src/validator.rs
```diff
@@ -103,14 +101,22 @@ impl Info {
...
     /// public key.
-    pub fn verify_signature(&self, sign_bytes: &[u8], signature: &[u8]) -> bool {
-        if let Some(pk) = &self.pub_key.ed25519() {
-            let verifier = Ed25519Verifier::from(pk);
-            if let Ok(sig) = ed25519::Signature::from_bytes(signature) {
-                return verifier.verify(sign_bytes, &sig).is_ok();
-            }
-        }
-        false
+    pub fn verify_signature(&self, sign_bytes: &[u8], signature: &Signature) -> Result<(), Error> {
```
tony-iqlusion, line 104: I made `signature` typed here, and changed it from returning a `bool` to a `Result`. I very strongly suggest doing this for a few reasons:

- `Result` is `#[must_use]`, so it significantly lowers the risk of ignoring signature validation by making most cases of it a compile-time error
- `Result` provides better type safety. Here's an example of a real-world signature validation bug which resulted from the use of combinators + `bool` instead of `Result`: https://github.com/libp2p/rust-libp2p/pull/1127/files

tendermint/src/validator.rs
```diff
@@ -270,18 +277,18 @@ mod tests {
             158_095_448_483_785_107,
         );
         let v3 = make_validator(
-            "EB6B732C4BD86B5FA3F3BC3DB688DA0ED182A7411F81C2D405506B298FC19E52",
+            "76A2B3F5CBB567F0D689D9DF7155FC89A4C878F040D7A5BB85FF68B74D253FC7",
             770_561_664_770_006_272,
         );
```
tony-iqlusion, line 282: This is *another* example of an Ed(wards)25519 point which failed to validate. It looks like it might've been munged slightly from the similar looking point below (L284).

It looks like these vectors were tested against the Go implementation, so they need to be rechecked for compatibility.

testgen/src/validator.rs
```diff
@@ -32,7 +36,7 @@ impl Validator {
     set_option!(proposer_priority, i64);
 
     /// Get a signer from this validator companion.
-    pub fn get_signer(&self) -> Result<Ed25519Signer, SimpleError> {
+    pub fn get_signer(&self) -> Result<Ed25519Keypair, SimpleError> {
```
tony-iqlusion, line 39: Perhaps this should be renamed to:

```suggestion
    pub fn get_keypair(&self) -> Result<Ed25519Keypair, SimpleError> {
```

testgen/src/validator.rs
```diff
@@ -45,19 +49,17 @@ impl Validator {
...
+        Ok(Ed25519Keypair { secret, public })
     }
 
     /// Get a verifier from this validator companion.
-    pub fn get_verifier(&self) -> Result<Ed25519Verifier, SimpleError> {
-        let signer = self.get_signer()?;
-        let public_key = try_with!(signer.public_key(), "failed to get public key");
-        let verifier = Ed25519Verifier::from(&public_key);
-        Ok(verifier)
+    pub fn get_verifier(&self) -> Result<Ed25519PublicKey, SimpleError> {
```
tony-iqlusion, line 61: Likewise, perhaps rename this to:

```suggestion
    pub fn get_verifier(&self) -> Result<Ed25519Verifier, SimpleError> {
```

tendermint/src/public_key.rs
```diff
@@ -1,73 +1,73 @@
...
     /// Secp256k1 keys
     #[cfg(feature = "secp256k1")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "secp256k1")))]
     #[serde(
         rename = "tendermint/PubKeySecp256k1",
         serialize_with = "serialize_secp256k1_base64",
         deserialize_with = "deserialize_secp256k1_base64"
     )]
-    Secp256k1(secp256k1::PublicKey),
+    Secp256k1(Secp256k1PublicKey),
```
tarcieri, line 34: If you want to keep this gated under a `secp256k1` feature, it might make sense to make `pub enum PublicKey` a `#[non_exhaustive]` enum. Otherwise clients might assume that variants do or do not exist depending on what features are enabled, and if they try to exhaustively match, they'll break if things don't match up to their expectations.

Or, perhaps the (lightweight, pure Rust) `k256` crate is a less onerous dependency than `signatory-secp256k1` (C library wrapper) to the point that the `secp256k1` feature can be removed.

tony-iqlusion, line 34: If you want to keep this gated under a `secp256k1` feature, it might make sense to make `pub enum PublicKey` a `#[non_exhaustive]` enum. Otherwise clients might assume that variants do or do not exist depending on what features are enabled, and if they try to exhaustively match, they'll break if things don't match up to their expectations.

Or, perhaps the (lightweight, pure Rust) `k256` crate is a less onerous dependency than `signatory-secp256k1` (C library wrapper) to the point that the `secp256k1` feature can be removed.

tony-iqlusion:
The clippy failures seem unrelated?

tendermint/src/amino_types/ed25519.rs
liamsi: I need to double check compatibility with the go-code here. 

I assume it's OK to check with the upcoming 0.34 / master instead of 0.33 (cc @brapse @ebuchman).

tendermint/src/validator.rs
liamsi: I'll recheck them during the weekend.

tendermint/Cargo.toml
```diff
@@ -27,16 +27,17 @@ authors = [
...
 
 [dependencies]
 anomaly = "0.2"
 async-trait = "0.1"
 bytes = "0.5"
 chrono = { version = "0.4", features = ["serde"] }
+ed25519 = "1"
+ed25519-dalek = { version = "= 1.0.0-pre.4", features = ["serde"] }
 futures = "0.3"
+k256 = { version = "0.4.2", optional = true, features = ["ecdsa"] }
```
xla, line 40: Are we dependent on this specific patch version? Or can we loosen the constraint?

tendermint/src/public_key.rs
```diff
@@ -1,73 +1,73 @@
 //! Public keys used in Tendermint networks
 
+pub use ed25519_dalek::PublicKey as Ed25519PublicKey;
```
xla, line 3:
```suggestion
pub use ed25519_dalek::PublicKey as Ed25519;
```

Can we drop the suffix as it is already encoded in the module name? So callers can use it as `public_key::Ed25519`.

tendermint/src/public_key.rs
```diff
@@ -1,73 +1,73 @@
...
+
+#[cfg(feature = "secp256k1")]
+pub use k256::PublicKey as Secp256k1PublicKey;
```
xla, line 6:
```suggestion
pub use k256::PublicKey as Secp256k1;
```

Same as above.

tendermint/src/account.rs
```diff
@@ -75,12 +80,10 @@ impl From<secp256k1::PublicKey> for Id {
...
 // SHA256(pk)[:20]
-impl From<ed25519::PublicKey> for Id {
-    fn from(pk: ed25519::PublicKey) -> Id {
+impl From<Ed25519PublicKey> for Id {
+    fn from(pk: Ed25519PublicKey) -> Id {
         let digest = Sha256::digest(pk.as_bytes());
-        let mut bytes = [0u8; LENGTH];
-        bytes.copy_from_slice(&digest[..LENGTH]);
-        Id(bytes)
+        Id(digest[..LENGTH].try_into().unwrap())
```
xla, line 86: If this is fallible now, should we try to rework it into a `TryFrom`?

tendermint/src/public_key.rs
xla: All for the `enum` with non_exhaustive - how much would it take to remove the `secp256k1` feature?

tendermint/src/validator.rs
xla: Thanks for this one!

testgen/src/validator.rs
xla: :+1:

testgen/src/validator.rs
xla: :+1:

tendermint/Cargo.toml
liamsi: This does not pin the specific patch version. It's equivalent to "^0.4.2".

tendermint/src/public_key.rs
tarcieri: Remove it as a feature, or remove all secp256k1-related code?

For what it's worth, the KMS still relies on it for the transaction signing feature.

tendermint/Cargo.toml
tarcieri: It could be relaxed to "0.4" without issue. Like @liamsi said, this is a SemVer baseline.

I think I originally went with such a specific version because this patch started as a Signatory upgrade and that was the first version it would've supported, but then I decided to rip Signatory out as it's not useful for this crate anymore and the other crates have subsumed much of its functionality and common signing/verification APIs.

tendermint/src/account.rs
tarcieri: The implementation is infallible. It's doing a slice to byte array conversion, and the potential failure case would be the length of the slice doesn't match the length of the destination array. Since both the slice and the array are sized using the `LENGTH` constant, I don't foresee this being a problem.

Note that `copy_from_slice` will panic in this case as well, it's just not as visible.

tendermint/src/public_key.rs
tarcieri: Sure, I can push up that change, but I'd suggest doing it to all of the types (`secret_key::Ed25519Keypair`, `signature::Ed25519Signature`) for consistency.

tarcieri:
@liamsi @xla hopefully I've addressed all of the line notes.

The major remaining concern to me is the point decompression-related issues. I'll try to verify the behaviors on both Rust and Go, as if there's an actual discrepancy there that sounds potentially consensus-critical.

tony-iqlusion:
@liamsi @xla hopefully I've addressed all of the line notes.

The major remaining concern to me is the point decompression-related issues. I'll try to verify the behaviors on both Rust and Go, as if there's an actual discrepancy there that sounds potentially consensus-critical.

liamsi:
Yeah, it's a bit weird as the go library also does point decompression as far as I remember. And the [test vectors](https://github.com/informalsystems/tendermint-rs/pull/522#discussion_r469674391) surely were generated using the go implementation (I wish we documented how exactly these were generated). I guess as long as we can "roundtrip" pubkeys between the go and rust implementation we should be fine. 

A while ago @valardragon suggested to cache these decompressions and add a verification method directly on the uncompressed pk as an optimization: https://github.com/tendermint/tendermint/issues/1784. But as far as I know this never materialized.

liamsi:
Actually, we did document were these test-vectors are coming from, besides, how the raw bytes came about:
https://github.com/informalsystems/tendermint-rs/blob/48a11d1aea190a3f9eaa8582c37b82e1395d5858/tendermint/src/amino_types/ed25519.rs#L106-L110

Note that the go code does not do any checks on these bytes but I would be surprised that the existing test-vectors were not generated from a priv-key. In that case the pubkey should have been compressed, as this is called while generating a key(pair): https://github.com/golang/crypto/blob/123391ffb6de907695e1066dc40c1ff09322aeb6/ed25519/internal/edwards25519/edwards25519.go#L708

My suggestion is to generate a fresh key pair in go and use this as a new test-vector (maybe this is what you did anyways?) and verify everything works as expected.

tendermint/src/amino_types/ed25519.rs
liamsi: OK, I've tested against a few freshly generated pub-keys and they all pass! I think that `raw_pk` wasn't generated using `ed25519.GenPrivKey().PubKey()` and we didn't notice before as the old lib didn't attempt to decompress points.

liamsi:
https://github.com/informalsystems/tendermint-rs/pull/522/files#r472102039
The bytes that were used were not generated using github.com/tendermint/tendermint/crypto/ed25519.ed25519.GenPrivKey()

This is not a compatibility issue!

tendermint/src/validator.rs
liamsi: I suspect this is a similar problem as above and the key was simply not valid. On the other hand I'm surprised that the other validator keys here (v1, v2) are valid 🤔 As we probably generated them in the same way. From looking at the `decompress` code in the rust lib I don't see any discrepancies though. 

I would assume that both libraries are properly tested against many test-vectors? https://ed25519.cr.yp.to/python/sign.input (true for the go lib not sure about the rust impl)

A simple way to confirm that this pub-key would also fail the go-implementation is to feed that old bytes into the internal package `edwards25519`:
```go
var A edwards25519.ExtendedGroupElement
var publicKeyBytes [32]byte
copy(publicKeyBytes[:], publicKey)
asster.False(t, A.FromBytes(&publicKeyBytes))
```
I'll give that a try shortly.

liamsi: OK, so this key won't work in the go implementation either. Tested via:
```go
func TestKey(t *testing.T) {
	publicKey, err := hex.DecodeString("EB6B732C4BD86B5FA3F3BC3DB688DA0ED182A7411F81C2D405506B298FC19E52")
	if err != nil {
		t.Fatalf("%v", err)
	}
	var A edwards25519.ExtendedGroupElement
	var publicKeyBytes [32]byte
	copy(publicKeyBytes[:], publicKey)
	if A.FromBytes(&publicKeyBytes) {
		panic("You shall not pass")
	} else {
	 	t.Log("OK")
	}
}
```

tendermint/src/private_key.rs
```diff
@@ -1,95 +1,67 @@
...
 #[serde(tag = "type", content = "value")]
 pub enum PrivateKey {
     /// Ed25519 keys
-    #[serde(rename = "tendermint/PrivKeyEd25519")]
-    Ed25519(Ed25519Keypair),
+    #[serde(
+        rename = "tendermint/PrivKeyEd25519",
+        serialize_with = "serialize_ed25519_keypair",
+        deserialize_with = "deserialize_ed25519_keypair"
+    )]
```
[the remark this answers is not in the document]
tony-iqlusion: The `Ed25519` struct this enum variant wraps is now from a different crate (`ed25519_dalek::Keypair`), but the serde serialization logic is intended to match `priv_validator.json` (or the bare Base64 encoded format used by the KMS)

tendermint/src/private_key.rs
tony-iqlusion: Yeah, this type previously didn't check that the associated point would decompress correctly (or was computed correctly from the Ed25519 "seed" in the keypair)

tony-iqlusion:
@xla looks like this is blocked on your suggested changes, but I think they've all been done.

Any ideas on the clippy error? Will it block merging?

tony-iqlusion:
After some investigation the clippy error seems to be related to the enum variants the `PublicKey` type wraps.

Specifically the `VerificationError::InvalidSignature` variant stores the entire `Validator` struct, which I guess is now large enough to push it over the clippy size limit.

tendermint/src/validator.rs
ebuchman: What was the take away from this?

liamsi: We should make sure that the test-vectors we use are generated properly, i.e. they don't just satisfy the properties needed for the test-case (here: 32 random bytes did the job) but rather they should be valid domain objects (here: a valid pub-key) - unless of course we want to test that an invalid case.

Issue #524, opened 2020-08-17
light-client: factor out test functions from integration tests


andrey-kuprianov:
Current tests in the `light-client` crate, in the integration tests [light_client.rs](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/tests/light_client.rs), [supervisor.rs](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/tests/supervisor.rs) contains a lot of duplicate code for reading tests files, iterating over directories with tests, etc. This code can be factored out into commonly usable functions, which is probably useful also for other crates. 

Another advantage of such a refactoring is that it will fit better with the test files organization, and will be able to provide better overview of the successful/failing tests.

In the near future this will provide also the basis for automating running of the model-based tests, and their integration with the CI (see #414 )

ebuchman:
Believe this was done in #525?

Issue #414, opened 2020-07-07
Develop a set of model-based tests for the LightClient


andrey-kuprianov:
Hi everyone!

Model-based testing is coming into shape: [Apalache](https://github.com/informalsystems/apalache) counterexamples can now be automatically translated into LightClient integrations tests: e.g., [this spec](https://github.com/informalsystems/jsonatr/blob/master/tests/support/counterexample_to_test.json) transforms [this counterexample](https://github.com/informalsystems/jsonatr/blob/master/tests/support/counterexample.json) into [this test](https://github.com/informalsystems/jsonatr/blob/master/tests/support/counterexample_to_test_output.json), which then can be used for testing the LightClient implementation using [this driver](https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/tests/lite.rs).

The current MBT handles only single-step test cases, but now, when most of the infrastructure is in place, it should be relatively easy to extend it to bisection cases, for example. Besides, the above counterexample was produced from one of the earlier specs, but it should be also relatively easy to adapt the process to [the most recent spec](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification/Lightclient_A_1.tla).

Now comes the creative part: we need to come up with testing scenarios, that can be encoded as spec invariants, model-checked with [Apalache](https://github.com/informalsystems/apalache) to get counterexamples, which then will be translated into tests. For a motivation about possible invariants, you can look [here](https://github.com/informalsystems/tendermint-rs/blob/142796b0cc37319f91af55cf2703b96a0cb80230/docs/spec/lightclient/verification/Lightclient_A_1.tla#L277).

Formulating the invariants can be not an easy task, but if you come up with an interesting scenario to test, more precisely those that potentially can reveal some bugs in the code, please post your ideas here. We will take care of turning plain text ideas into invariants, counterexamples, and, eventually, tests.

Looking forward to your input!

andrey-kuprianov:
Here is [the current set of LightClient tests](https://github.com/informalsystems/tendermint-rs/tree/master/tendermint/tests/support/lite)

PR #525, opened 2020-08-17
Testgen tester

andrey-kuprianov:
See #524 

* [X] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/testgen/src/tester.rs
+++ b/testgen/src/tester.rs
@@ -1,0 +1,340 @@
+use crate::{helpers::*};
+use std::{
+    panic,
+    sync::{Arc, Mutex},
+};
+use serde::{de::DeserializeOwned};
+use std::{fs, path::PathBuf};
+use std::panic::{UnwindSafe};
+use crate::tester::TestResult::{Success, Failure, ParseError, ReadError};
+
+#[derive(Debug, Clone)]
+pub struct TestEnv {
+    root_dir: String,
+    logs: Vec<String>
+}
+
+impl TestEnv {
+    pub fn add_log(&mut self, log: &str) {
+        self.logs.push(log.to_string());
+    }
+
+    /// Read a file from a path relative to the environment root into a string
+    pub fn read_file(&mut self, path: &str) -> Option<String> {
+        match self.full_path(path) {
+            None => None,
+            Some(full_path) => match fs::read_to_string(&full_path) {
+                Ok(file) => Some(file),
+                Err(_) => None
+            }
+        }
+    }
+
+    /// Convert a relative path to the full path from the test root
+    /// Return None the full path can't be formed
+    pub fn full_path(&self, rel_path: &str) -> Option<String> {
+        let full_path = PathBuf::from(&self.root_dir).join(rel_path);
+        match full_path.to_str() {
+            None => None,
+            Some(full_path) => Some(full_path.to_string())
+        }
+    }
+
+    /// Convert a full path to the path relative to the test root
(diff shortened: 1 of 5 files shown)


testgen/src/tester.rs
```diff
@@ -0,0 +1,337 @@
...
+            None => None,
+            Some(full_path) => match fs::read_to_string(&full_path) {
+                Ok(file) => Some(file),
+                Err(_) => None,
+            },
+        }
+    }
+
+    /// Convert a relative path to the full path from the test root
+    /// Return None the full path can't be formed
```
Shivani912, line 34:
```suggestion
    /// Return None if the full path can't be formed
```

testgen/src/tester.rs
```diff
@@ -0,0 +1,337 @@
...
+                None => None,
+                Some(rel_path) => Some(rel_path.to_string()),
+            },
+        }
+    }
+}
+
+pub struct Tester {
+    root_dir: String,
+    tests: Vec<(String, Box<dyn Fn(&str) -> TestResult>)>,
```
Shivani912, line 58: There's a clippy warning here for type complexity. I'm OKAY with this one being complex because it's not that difficult to read here but i'm not sure how much we adhere to clippy warnings

testgen/src/tester.rs
```diff
@@ -0,0 +1,337 @@
...
+            }),
+            Err(_) => ParseError,
+        };
+        self.tests.push((name.to_string(), Box::new(test_fn)));
+    }
+
+    fn add_result(&mut self, name: &str, path: &str, result: TestResult) {
+        self.results
+            .entry(name.to_string())
+            .or_insert(Vec::new())
```
Shivani912, line 183: clippy suggestion here again

testgen/src/tester.rs
```diff
@@ -0,0 +1,337 @@
...
+            .push((path.to_string(), result))
+    }
+
+    fn read_error(&mut self, path: &str) {
+        self.results
+            .entry("".to_string())
+            .or_insert(Vec::new())
```
Shivani912, line 190: also here (clippy)

testgen/src/tester.rs
```diff
@@ -0,0 +1,337 @@
...
+            .push((path.to_string(), TestResult::ReadError))
+    }
+
+    fn parse_error(&mut self, path: &str) {
+        self.results
+            .entry("".to_string())
+            .or_insert(Vec::new())
```
Shivani912, line 197: and here (clippy)

testgen/src/tester.rs
Shivani912: Oh, the warning is about this being complex

andrey-kuprianov:
Thanks for your suggesstions, @Shivani912! I have addressed them. Also, it's quite true that the source-level comments are very sparse right now, but what you see here is only a preliminary version of the Tester. I am already extending it substantially further under #414 , and there I will add more documentation. If there are no objections, I would like to merge this PR now, and the PR for #414 will follow shortly.

testgen/src/tester.rs
andrey-kuprianov: done

testgen/src/tester.rs
andrey-kuprianov: done; the types will also change in further PRs

Shivani912 (approved):
Thanks @andrey-kuprianov for explaining. 
This looks really good, we have much cleaner and organized code now. Let's merge it !

PR #529, opened 2020-08-21
Light Client model-based testing

andrey-kuprianov:
See #414. Also see [Light Client model-based testing guide](https://github.com/informalsystems/tendermint-rs/tree/andrey/lite-model-based/light-client/tests/support/model_based) for the high-level description how model-based testing will be working.

This PR touches quite a lot of the repository, so I will be grateful for multiple reviews. It is mostly about building a lot of infrastructure that enables model-based testing, so it should be easy to extend now to other types of tests (bisection), or other crates/repos when needed.

One particular issue is that it uncovered a number of discrepancies between the model and the implementation of the LIghtClient; please see the changes to the `Lightclient_A_1.tla` / `Blockchain_A_1.tla` I have fixed them by adjusting the model, but it would be nice if @josef-widder or @konnov take a look at it.

In general, I hope, this is the first step in the right direction of keeping the specs live and in sync with the implementation, and of thorough / to some extent exhaustive testing of the implementation.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


docs/spec/lightclient/verification/Blockchain_A_1.tla
```diff
@@ -75,7 +75,7 @@ LBT == [header |-> BT, Commits |-> {NT}]
 
 (* the header is still within the trusting period *)
 InTrustingPeriod(header) ==
-    now <= header.time + TRUSTING_PERIOD
+    now < header.time + TRUSTING_PERIOD
```
josef-widder, line 78: Does this change come from the implementation or the English spec? Are those consistent?

docs/spec/lightclient/verification/Lightclient_A_1.tla
```diff
@@ -115,8 +135,8 @@ ValidAndVerified(trusted, untrusted) ==
     THEN "FAILED_TRUSTING_PERIOD" 
     ELSE IF untrusted.header.height = trusted.header.height + 1
              \/ SignedByOneThirdOfTrusted(trusted, untrusted)
-         THEN "OK"
-         ELSE "CANNOT_VERIFY"
+         THEN "SUCCESS"
+         ELSE "NOT_ENOUGH_TRUST"
```
josef-widder, line 139: I think OK and CANNOT_VERIFY are used in the English spec. Are SUCCESS and NOT_ENOUGH_TRUST used in the implementation? If yes, we should adapt I guess the English spec.

docs/spec/lightclient/verification/Blockchain_A_1.tla
```diff
@@ -110,7 +110,7 @@ FaultAssumption(pFaultyNodes, pNow, pBlockchain) ==
 (* Can a block be produced by a correct peer, or an authenticated Byzantine peer *)
 IsLightBlockAllowedByDigitalSignatures(ht, block) == 
     \/ block.header = blockchain[ht] \* signed by correct and faulty (maybe)
-    \/ block.Commits \subseteq Faulty /\ block.header.height = ht \* signed only by faulty
+    \/ block.Commits \subseteq Faulty /\ block.header.height = ht /\ block.header.time > 0 \* signed only by faulty
```
josef-widder, line 113: I guess faulty processes may be allowed to write any time, also negative. Or is the time stored in some unsigned domain in the implementation?

josef-widder (commented):
I only looked that the blockchain and lightclient TLA files. It looks that some changes (I commented on them) are made to match the code, but as a result the TLA+ files may differ from the English spec. We should create an issue that highlights the changes needed in the English spec to keep everything consistent.

konnov (commented):
Nice trick with the history variable. However, instead of directly extending `Lightclient_A_1.tla`, I would recommend to extend Lightclient_A_1 with another module that monitors the light client with the history variable.

docs/spec/lightclient/verification/Blockchain_A_1.tla
andrey-kuprianov: this comes from the implementation, as otherwise I get a testing error: spec behavior is different from what we have in the implementation. And the English spec contains this `trusted.Header.Time > now - trustingPeriod`, so I guess this means that the English spec and the implementation agree, but the TLA+ spec is wrong in that respect; could you please verify that it's true?

docs/spec/lightclient/verification/Lightclient_A_1.tla
andrey-kuprianov: I've taken the verdicts from the implementation. There is another issue I've got, namely that according to the TLA+ constraints the verdict FAILED_TRUSTING_PERIOD cannot ever be produced. At least it's impossible to obtain the a test with such verdict from Apalache, and my quick manual check on the TLA+ constraints tells me it's indeed impossible. Could you please verify that? In case it's true, it may be makes sense to remove such a verdict. And I would actually update the english spec as well, as I find the implementation names a bit more informative

docs/spec/lightclient/verification/Blockchain_A_1.tla
andrey-kuprianov: it is indeed stored and serialized/deserialized as an unsigned integer, so signed integers do no make sense and only add confusion, I am afraid

josef-widder: I guess in the English spec we implicitly also assume positive times. Perhaps we should clarify it there explicitly.

docs/spec/lightclient/verification/Lightclient_A_1.tla
josef-widder: I will have a look with @konnov

docs/spec/lightclient/verification/Blockchain_A_1.tla
josef-widder: will do.

josef-widder: So now we do model-based testing as conformance test between English spec, TLA+, and the implementation. Amazing 😉

andrey-kuprianov:
It probably does make a lot of sense to split this PR into multiple ones, as @shonfeder proposed; I will do that when I am back from my vacation on Sept. 2. Just write it here so please do not spend your time reviewing this one.

ebuchman:
> It probably does make a lot of sense to split this PR into multiple ones, as @shonfeder proposed; I will do that when I am back from my vacation on Sept. 2. Just write it here so please do not spend your time reviewing this one.

Let's close this in the meantime then. Thanks for breaking it up!

shonfeder:
Thank you for the PR refactor! Will dig into reviewing whatever is still open ASAP. :pray:

docs/spec/lightclient/verification/Lightclient_A_1.tla
konnov: I don't see why `FAILED_TRUSTING_PERIOD` cannot be produced. It is produced in line 115. However, it is not propagated into `state`. Do you like to see it in `state`?

andrey-kuprianov: this PR is closed; let's discuss on another PR: [546](https://github.com/informalsystems/tendermint-rs/pull/546)

PR #537, opened 2020-08-26
Domain types for protobuf structs

### LINKED ISSUE #535: Implement domain types for protobuf-encoded structs
After merginf #527 to close #504 , the agreed domain types are missing from tendermint-rs.

* Implement domain types with TryFrom/From protobuf types. This should ease any conversion issues in the long run.
* Possibly implement a derive macro that indicates for each domain type what their protobuf type is and implements direct `.encode` and `.decode` functionality.

Check the notes in #527 for the requirements for domain types.

greg-szabo:
Closes #535 

* domain types with TryFrom(protobuf struct) and From(domaintype) for the protobuf structs
* DomainType trait for direct `.encode`/`.decode` functionality from the protobuf types
* DomainType derive macro

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


tarcieri:
> possibly a trait should be added for direct .encode/.decode functionality from the Raw type for easier handling.

Would definitely recommend an "extension trait" with a blanket impl for `T: prost::Message` here.

tendermint/Cargo.toml
```diff
@@ -39,8 +39,9 @@ ed25519-dalek = { version = "= 1.0.0-pre.4", features = ["serde"] }
 futures = "0.3"
 k256 = { version = "0.4", optional = true, features = ["ecdsa"] }
 once_cell = "1.3"
-prost-amino = "0.6"
-prost-amino-derive = "0.6"
+prost = "0.6"
+prost-derive = "0.6"
+prost-types = "0.6"
```
liamsi, line 44: ❤️

liamsi, line 44: Is this due to using the well known type for time?

greg-szabo: I'm not sure what you mean. If you refer to me removing TimeMsg, it's because its implementation was the same or very similar to the built-in prost::Timestamp implementation. We might reintroduce it as a domain-type, though.

greg-szabo: Ah, I think I understand your question! You are asking why the `prost-types` crate was added!?

It's for `Timestamp`, but I think we might not need it. I'll double-check and remove before we merge this.

liamsi: Yes, I was wondering why the Prost-types dependency was added. Thanks for the clarification.

PR #539, opened 2020-08-26
sled 0.34 upgrade

### LINKED ISSUE #490: upgrade sled to 0.34.1
https://github.com/spacejam/sled/blob/master/CHANGELOG.md#0341

greg-szabo:
Closes #490 

I'm not the most proficient in the light-client and light-node packages and those are the two affected by sled. It seems to me based on my understanding of the code, that the below two breaking changes in sled does not affect our code.

sled 0.34 breaking changes:
    #1131 transactions performed on Trees from different Dbs will now safely fail.
    #1131 transactions may now only be performed on tuples of up to 14 elements. For higher numbers, please use slices.

* [ ] Wrote tests
* [X] Updated CHANGELOG.md


light-client/Cargo.toml
```diff
@@ -31,7 +31,7 @@ prost-amino = "0.6.0"
 serde = "1.0.106"
 serde_cbor = "0.11.1"
 serde_derive = "1.0.106"
-sled = "0.33.0"
+sled = "0.34.3"
```
thanethomson, line 34: I see in the Light Client we've pinned the patch version numbers of our dependencies. Should we be? See [this comment](https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r469327310). (This is more of a general question than specific to this PR).

Does `sled` release breaking changes with patch version increments? Doesn't seem like it to me.

liamsi: It's not really pinning to a specific version, it's equivalent to "^0.34.3" which means >=0.34.3, <0.35.0. I agree that we should handle this consistently across crates though.

greg-szabo: I agree too and at first I wanted to just use "0.34". I kept this to keep it consistent with all other dependencies in the file.

I think opening a PR to fixing this in all `Cargo.toml`s in the repo makes more sense.

greg-szabo:
I think we should open a PR regarding datastore testing and keep it separate from the upgrade.

I'm somewhat pushing back on doing this in this PR, solely because the scope of the PR was never properly established. `sled` is a "low-class" dependency here and it's easier to decide to _not_ upgrade than to start implementing features _because of_ an upgrade. Especially because the benefits of upgrading (maybe safely failing somewhere?) are not clear while the disadvantages (breaking) are clear.

I completely agree, that if we implement a data store, we have to have decent testing around it. If testing is a strong suggestion before we accept the upgrade, then I propose we don't upgrade and postpone this issue into the v0.17 release. In parallel let's open an issue to create database testing, also slated for v0.17. Then we can upgrade properly.

The assumption from the v0.16 release perspective was that this is a quick version change and this is not going to be blocking. Unfortunately, I'm not familiar with the light-client implementation here enough to make an educated decision on the necessity of the upgrade and the issue didn't clarify it either. Also, closing out the v0.16 already seems to run into people-not-having-enough-time issues so I can't, with a good conscience, ask for someone to jump in and do it.

My 2 cents: let's drop this from v0.16 and open new issues detailing the upgrade.

ebuchman (approved):
Now that v0.17 is a significant breaking change lets just merge this. We can standardize the dependency  pinning to leave out the patch numbers in another PR

Issue #526, opened 2020-08-18
spec: full node proof of fork handling


milosevic:
After proof of fork is submitted by the light client to a full node, the full node needs to verify the PoF and if valid to create evidences of misbehaviour that should then be gossiped through the network and finally processed on chain in order to punish
misbehaving nodes. 

This task should specify different evidence types and it's on chain processing.

milosevic:
Done in tendermint/spec repo. See https://github.com/tendermint/spec/pull/164.

PR #544, opened 2020-08-31
Light client evidence handling

milosevic:
Addresses https://github.com/informalsystems/tendermint-rs/issues/526.


docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+        if current.Header == trace[i].Header continue
+
+        // we have identified a conflicting header
+        attackType = nil
+        trustedBlock = trace[i-1]
+        conflictingBlock = trace[i]
+        
+        if !isPossibleBlock(trustedBlock, conflictingBlock) { 
+            attackType = LunaticAttack
+        } else if trustedBlock.Header.Round == conflictingBlock.Header.Round {
```
cmwaters, line 157: This should be `current` instead of `trustedBlock` because `current` is at the same height as the `conflictingBlock` which is required for this verification logic.

Perhaps better terminology would be to call `trustedBlock` `commonBlock` because it is the last block that the trace and the peer have in common

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+
+```go
+func detectMisbehavingProcesses(lcAttackEvidence LightClientAttackEvidence, bc Blockchain) []ValidatorAddress {
+   assume lcAttackEvidence.Type == LunaticAttack
+   misbehavingValidators = []ValidatorAddress
+
+   conflictingHeight = lcAttackEvidence.ConflictingBlock.Header.Height
+   
+   bondedValidators = bc.GetValidators(lcAttackEvidence.CommonHeight) union 
+                      bc.GetNextValidators(lcAttackEvidence.CommonHeight) union
```
cmwaters, line 185:
```suggestion
                      bc.GetNextValidators(lcAttackEvidence.CommonHeight) intersection
```

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+                      bc.GetValidators(conflictingHeight))
```
cmwaters, line 186: Also for further clarification, this should be the validators in the set that actually voted for the conflicting header not just all validators

josef-widder (commented):
I will continue reading later today.

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+    return primary_lca_evidence, witness_lca_evidence
+}
+
+func DetectLightClientAttack(trace []LightBlock, peer PeerID) (LightClientAttackEvidence, []LightBlock) {
+
+    trusted = trace[0]
+    lightStore = new LightStore().Update(trusted, StateTrusted)
+
+    for i in 1..len(trace)-1 {
+        lightStore, result = VerifyToTarget(peer, lightStore , i)
```
josef-widder, line 141:
```suggestion
        lightStore, result = VerifyToTarget(peer, lightStore , trace[i].Header.Height) 
```

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+
+    trusted = trace[0]
+    lightStore = new LightStore().Update(trusted, StateTrusted)
+
+    for i in 1..len(trace)-1 {
+        lightStore, result = VerifyToTarget(peer, lightStore , i) 
+   
+        if result == ResultFailure then return (nil, nil)
+        
+        current = lightStore.Get(i)
```
josef-widder, line 145:
```suggestion
        current = lightStore.Get(trace[i].Header.Height)
```

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+        
+        case EquivocationAttack: return isPossibleBlock(trustedBlock, lcaEvidence.ConflictingBlock) and 
+                                        trustedBlock.Header.Round == lcaEvidence.ConflictingBlock.Header.Round
+
+        case AmnesiaAttack: return isPossibleBlock(trustedBlock, lcaEvidence.ConflictingBlock) and 
+                                   trustedBlock.Header.Round != lcaEvidence.ConflictingBlock.Header.Round
+    } 
+}
+
+func isPossibleBlock(trusted Header, conflicting Header) boolean {
```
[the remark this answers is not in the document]
milosevic: Good point. It should probably be isValidBlock?

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+    - amnesia attack. 
+
+TODO: Add references to attack definitions.     
+TODO: Add Callum's figures.
+    
+```go
+type LightClientAttackEvidence struct {
+    ConflictingBlock   LightBlock
+    CommonHeight       int64
+    Type               AttackType
```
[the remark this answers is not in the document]
josef-widder: Also I wonder if we should not add peerID of the node to which we are going to submit LightClientAttackEvidence to. At least internally. The point is that the evidence is specific to the peer for which we generate it.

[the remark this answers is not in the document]
milosevic: Given light client attack, there seems to be a simple way to differentiate between attack types. Keep in mind that the goal is to detect and punish +1/3 faulty processes. In case of lunatic attack (invalid block), we don't need to check for any other attack type as we have +1/3 of processes that precommitted invalid block. In case both headers are for valid blocks, we first check for pure equivocation attack. This is the case when commit round is the same. In this case we have enough information to punish +1/3 of equivocating processes. If commit round is different, we need to run full fork accountability protocol. In this case we could have mix of attacks: amnesia + equivocation, but we always have amnesia. Therefore we call such attack amnesia attack. Does this answer your question?

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+    witness_lca_evidence = nil
+    if witness_trace != nil {
+        witness_lca_evidence, _ = DetectLightClientAttack(witness_trace, primary)
+    }
+    return primary_lca_evidence, witness_lca_evidence
+}
+
+func DetectLightClientAttack(trace []LightBlock, peer PeerID) (LightClientAttackEvidence, []LightBlock) {
+
+    trusted = trace[0]
```
josef-widder, line 137:
```suggestion
    trusted = trace[0]
```
I do not see where trusted is later used

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+        
+        // if obtained header is the same as in the trace we continue with a next height
+        if current.Header == trace[i].Header continue
+
+        // we have identified a conflicting header
+        attackType = nil
+        trustedBlock = trace[i-1]
+        conflictingBlock = trace[i]
+        
+        if !isPossibleBlock(trustedBlock, conflictingBlock) {
```
josef-widder, line 155: I don't understand the idea how attacks are detected based on these conditions. We should discuss.

docs/spec/lightclient/attacks/evidence-handling.md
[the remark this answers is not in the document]
josef-widder: I guess lunatic is clear.
Regarding the other two I am not sure. Even if both blocks are from round 5, it might still be that some faulty nodes did amnesia, and other faulty nodes did equivocation. Do I miss something?

docs/spec/lightclient/attacks/evidence-handling.md
josef-widder, line 186: How expensive is it to compute this union? In principle we could check between CommonHeight and conflictingHeights where the validator set changed and do the union there.

docs/spec/lightclient/attacks/evidence-handling.md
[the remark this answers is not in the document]
milosevic: In case of pure equivocation attack the following holds: conflicting blocks are valid and commit round is the same. Note that the as conflicting blocks are valid that implies that validator sets are the same. Therefore, in order to have two commits, there need to be conflicting signatures of at least +1/3 validators. We could have amnesia attack in addition, but we don't need to detect it as we catch at least +1/3 of equivocations.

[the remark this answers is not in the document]
josef-widder: Right!! Good point.  
I guess then we should just make it clear in the final write-up what precisely we are required to catch. I guess for every attack we have to catch at least *f + 1* faulty voting power. Perhaps we should also mention that in the future we might try to catch more, e.g., these additional amnesia scenarios.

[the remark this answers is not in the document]
melekes: moreover, can't we deduce the attack type from ConflictingBlock?

docs/spec/lightclient/attacks/evidence-handling.md
```diff
@@ -0,0 +1,250 @@
...
+Detecting faulty processes in case of the amnesia attack is more complex and cannot be inferred 
+purely based on attack evidence data. In this case, in order to detect misbehaving processes we need
+access to votes processes sent during the conflicting height. Therefore, amnesia handling assumes that
+validators persist all votes received and sent during multi-round heights (as amnesia attack 
+is only possible in heights that executes over multiple rounds, i.e., commit round > 0).  
+
+To simplify description of the algorithm we assume existence of the trusted oracle called monitor that will 
+drive the algorithm and output faulty processes at the end. Monitor can be implemented in a
+distributed setting as on-chain module. The algorithm works as follows:
+    1) Monitor sends votesets request to validators of the conflicting height. Validators
```
melekes, line 227: do we really need to make requests? is it possible to, when amnesia attack evidence is committed, to say that suspected nodes must gossip their votesets. If they don't, we punish them.

docs/spec/lightclient/attacks/evidence-handling.md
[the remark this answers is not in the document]
milosevic: You need trusted block and conflicting block and then you can figure out the nature of attack. If this is your question. Common height is needed as part of attack evidence verification.

docs/spec/lightclient/attacks/evidence-handling.md
milosevic: Should be straightforward.

docs/spec/lightclient/attacks/evidence-handling.md
milosevic: Sure. Making requests is more conceptual step here; there are probably more efficient version to implement it.

josef-widder:
It is a cool idea to minimize the evidence! I think we should highlight the interfaces to the other light client protocols as follows:

I am in parallel sketching a supervisor. This basically is the **initialization protocol** (from genesis or block) followed by a loop that consists of

1. **verification** with the primary (outputs a verification trace, i.e., a sequence of lightblocks where one can be used to verify the next). We will need to slightly change `VerifyToTarget` and make a new version, that is, [LCV-FUNC-MAIN.1] -> [LCV-FUNC-MAIN.2]
2. **fork detection** (outputs proof-of-forks; i.e., two conflicting verification traces, with the information from which peers these traces where obtained)
3. a call to a function now called `MinimizeAndSubmitProofOfFork(p)` where `p` is a proof of fork. To the best of my understanding, this PR basically implements this function (modulo inconsistencies with parameters).

I suggest that we keep the structure of the supervisor, with init, verification, and the detector, and "minimize and submit". That also means, within the light client we keep the proof of fork data structure. And then the functions described here get as input a proof of fork, and as output generate `LightClientAttackEvidence` and submit it to the peer. This would mean the parts of the light client specs we will need to write are:

0. supervisor + init
1. verification
2. fork detections
3. evidence generation and submission

I am not sure where the specification for the functions that will run on a full node should go?

docs/spec/lightclient/attacks/evidence-handling.md
milosevic: On the next line, but I inlined it in the new version.

josef-widder (dismissed):
This PR documents great progress in understanding of the problem and great algorithmic ideas. We can merge it, and incorporate the approach into
1. a VDD spec for light client attack detection (we do not want to use the term fork in this context), and
2. a VDD accountability spec that describes evidence processing at a full node

josef-widder (approved):
Thanks all for contributing and discussing. We will merge this, and use it as base for VDD specs as discussed.

PR #545, opened 2020-09-01
Fixes dependency loop

Shivani912:
This particular commit (#534 ) introduced the circular dependency (my bad!) between tendermint-rs, testgen and light-client and I'm reverting that here.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


Shivani912:
hmm, not sure why the integration test fails. Looking at it.

Shivani912:
Okay, it's an `IoError` and looks like it is having an issue with establishing a connection. Since you've worked on this most @greg-szabo , do you know what is breaking it?

liamsi:
Just for my understanding: isn't creating a valset and a light-block one of the main features of the testgen library?

liamsi (approved):

greg-szabo:
Possibly. The one-man decision from me is that we're not publishing testgen yet. Maybe with the next release, when we don't need to make last-minute changes. 0.16.0 release should finish up.

Shivani912:
> Just for my understanding: isn't creating a valset and a light-block one of the main features of the testgen library?

It is and that's what I thought initially, but `LightBlock` is introduced in `light-client` and so testgen will have a dependency on `light-client`. This creates a dependency loop between light-client, testgen and tendermint. So, will have to discuss where to put LightBlock generation. As with validator set, yes it should be part of testgen but it was easier and quicker to simply revert this commit. Since I'm using it only locally right now, It wouldn't break anything and will be added in a later PR

PR #547, opened 2020-09-02
Testgen tester -- utilities to run multiple tests with logs and reports

andrey-kuprianov:
Part 2 or #529. See also #414.

This PR introduces Testgen Tester. The preliminary version of it appeared in #525, but as that PR has not received sufficient number of reviews, please consider the whole code of `tester.rs` as new.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
-    logs: Vec<String>,
+    /// Directory where the test is being executed
+    current_dir: String,
 }
 
 impl TestEnv {
-    pub fn add_log(&mut self, log: &str) {
-        self.logs.push(log.to_string());
+    pub fn new(current_dir: &str) -> Option<Self> {
+        fs::create_dir_all(current_dir).ok().map(|_| TestEnv {
```
romac, line 23: I wonder if we might need a sanity check here to prevent users passing `/`, `~`, or `.`, and then end up with their eg. home directory erased in the `cleanup` step?

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+    /// Returns the relative path of the file, or None if copying was not successful
+    pub fn copy_file_from_env(&self, other: &TestEnv, path: &str) -> Option<String> {
+        other
+            .full_path(path)
+            .and_then(|full_path| self.copy_file_from(&full_path))
     }
 
     /// Convert a relative path to the full path from the test root
     /// Return None if the full path can't be formed
     pub fn full_path(&self, rel_path: &str) -> Option<String> {
```
romac, line 111: Why not return a `PathBuf` instead of an `Option<String>`?

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+
+    /// Parse a file from a path relative to the environment current dir as the given type
+    pub fn parse_file_as<T: DeserializeOwned>(&self, rel_path: &str) -> Option<T> {
+        self.read_file(rel_path)
+            .and_then(|input| serde_json::from_str(&input).ok())
+    }
+
+    /// Copy a file from the path outside environment into the environment current dir
+    /// Returns the relative path of the file, or None if copying was not successful
+    pub fn copy_file_from(&self, path: &str) -> Option<String> {
```
romac, line 88: How about working directly with paths, ie.

```rust
pub fn copy_file_from(&self, path: &Path) -> Option<PathBuf>
```

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+    }
+
+    /// Write a file to a path relative to the environment current dir
+    pub fn write_file(&self, rel_path: &str, contents: &str) -> Option<()> {
+        self.full_path(rel_path)
+            .and_then(|full_path| fs::write(full_path, contents).ok())
+    }
+
+    /// Parse a file from a path relative to the environment current dir as the given type
+    pub fn parse_file_as<T: DeserializeOwned>(&self, rel_path: &str) -> Option<T> {
```
romac, line 81: How about working directly with a path, ie.

```rust
pub fn parse_file_as<T: DeserializeOwned>(&self, rel_path: &Path) -> Option<T>
```

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+    }
+
+    /// Read a file from a path relative to the environment current dir into a string
+    pub fn read_file(&self, rel_path: &str) -> Option<String> {
+        self.full_path(rel_path)
+            .and_then(|full_path| fs::read_to_string(&full_path).ok())
+    }
+
+    /// Write a file to a path relative to the environment current dir
+    pub fn write_file(&self, rel_path: &str, contents: &str) -> Option<()> {
```
romac, line 75: How about working directly with a path, ie.

```rust
pub fn write_file(&self, rel_path: &Path, contents: &str) -> Option<()>
```

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+            fs::OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(full_path)
+                .ok()
+                .and_then(|mut file| file.write_all((String::from(msg) + "\n").as_bytes()).ok())
+        })
+    }
+
+    pub fn logln_to(&self, msg: &str, rel_path: &str) -> Option<()> {
```
romac, line 56: How about working directly with a path, ie.

```rust
pub fn logln_to(&self, msg: &str, rel_path: &Path) -> Option<()>
```

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+
+    pub fn logln(&self, msg: &str) -> Option<()> {
+        println!("{}", msg);
+        self.full_path("_log").and_then(|full_path| {
+            fs::OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(full_path)
+                .ok()
+                .and_then(|mut file| file.write_all((String::from(msg) + "\n").as_bytes()).ok())
```
romac, line 52: You might be able to simplify this a bit and avoid the String allocation:

```rust
.and_then(|mut file| write!(file, "{}\n", msg))
```

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+
+    pub fn logln_to(&self, msg: &str, rel_path: &str) -> Option<()> {
+        println!("{}", msg);
+        self.full_path(rel_path).and_then(|full_path| {
+            fs::OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(full_path)
+                .ok()
+                .and_then(|mut file| file.write_all((String::from(msg) + "\n").as_bytes()).ok())
```
romac, line 64: You might be able to simplify this a bit and avoid the String allocation:

```rust
.and_then(|mut file| write!(file, "{}\n", msg))
```

testgen/src/tester.rs
```diff
@@ -1,77 +1,196 @@
...
+        let path = Path::new(path);
+        if !path.is_file() {
+            return None;
         }
+        path.file_name().and_then(|name| {
+            name.to_str().and_then(|name| {
+                self.full_path(name)
+                    .and_then(|dest| fs::copy(path, dest).ok().map(|_| name.to_string()))
+            })
+        })
```
romac, line 98: We could perhaps avoid the nesting here by using the `?` operator:

```suggestion
        let name = path.file_name()?.to_str()?;
        let dest = self.full_path(name)?;
        fs::copy(path, dest).ok()?;
        Some(name.to_string())
```

testgen/src/tester.rs
```diff
@@ -88,24 +207,29 @@ impl TestResult {
...
-        TestEnv {
-            root_dir: self.root_dir.clone(),
-            logs: vec![],
-        }
+    pub fn env(&self) -> Option<TestEnv> {
+        TestEnv::new(&self.root_dir)
+    }
+
+    pub fn output_env(&self) -> Option<TestEnv> {
+        fs::create_dir_all(self.root_dir.clone() + "/_" + &self.name)
```
romac, line 225: How about using the `PathBuf` API being to build/concatenate paths?

testgen/src/tester.rs
```diff
@@ -88,24 +207,29 @@ impl TestResult {
...
+            .ok()
+            .and(TestEnv::new(&(self.root_dir.clone() + "/_" + &self.name)))
     }
 
-    fn capture_test<F>(env: TestEnv, test: F) -> TestResult
+    fn capture_test<F>(test: F) -> TestResult
```
romac, line 230: Cool stuff!

testgen/src/tester.rs
```diff
@@ -140,60 +263,79 @@ impl Tester {
...
         };
-        self.tests.push((name.to_string(), Box::new(test_fn)));
+        self.tests.push(Test {
+            name: name.to_string(),
+            test: Box::new(test_fn),
+        });
     }
 
-    pub fn add_test_with_env<T>(&mut self, name: &str, test: fn(T, &mut TestEnv))
+    pub fn add_test_with_env<T>(&mut self, name: &str, test: fn(T, &TestEnv, &TestEnv, &TestEnv))
```
romac, line 278: Why the `fn(...)` type for `test` instead of `impl FnOnce(...)`? IIRC the former denotes a function pointer and would thus not allow passing a closure here.

testgen/src/tester.rs
```diff
@@ -140,60 +263,79 @@ impl Tester {
...
         };
-        self.tests.push((name.to_string(), Box::new(test_fn)));
+        self.tests.push(Test {
+            name: name.to_string(),
+            test: Box::new(test_fn),
+        });
     }
 
-    fn add_result(&mut self, name: &str, path: &str, result: TestResult) {
+    pub fn add_test_batch<T>(&mut self, batch: fn(T) -> Vec<(String, String)>)
```
romac, line 302: Same question as above.

testgen/src/tester.rs
```diff
@@ -305,9 +431,17 @@ impl Tester {
...
                     for path in paths {
                         if let Ok(entry) = path {
+                            // ignore path components starting with '_'
+                            if let Some(last) = entry.path().iter().rev().next() {
+                                if let Some(last) = last.to_str() {
+                                    if last.starts_with('_') {
+                                        continue;
+                                    }
+                                }
+                            }
```
romac, line 441: This code could be extracted into a helper function to improve readability.

romac (changes requested):
Looks good, I only have a few minor comments (mostly related to the use of String vs Path types). Great job :)

testgen/src/tester.rs
andrey-kuprianov: That's true; I will add the checks

testgen/src/tester.rs
andrey-kuprianov: The problem with passing and returning `Path` / `PathBuf` is that this shifts the complexity from the library to its user. E.g., in another PR I have the following code:

```
    env.copy_file_from_env(root_env, "Lightclient_A_1.tla");
    env.copy_file_from_env(root_env, "Blockchain_A_1.tla");
    env.copy_file_from_env(root_env, "LightTests.tla");
    env.copy_file_from_env(root_env, &test.model);
```

With `&str` this is easy; If the second parameter is `Path`, it becomes a nightmare...

testgen/src/tester.rs
romac: Perhaps just appending a directory to the given path is sufficient and even more reliable. This way, whatever path we are given, we will always store temp files one directory below and will thus not touch/override/remove whatever content there might already be within the given directory.

testgen/src/tester.rs
romac: Right, that makes sense.

Perhaps we can go with `path: impl AsRef<Path>` then? This allows both string-like types (&str, String) and path-like types (Path, PathBuf) and is thus a bit more flexible. It is also a common pattern in the standard library: https://users.rust-lang.org/t/when-to-use-asref-t-vs-t/9312/3

testgen/src/tester.rs
romac: Or perhaps:

```rust
pub fn copy_file_from(&self, path: impl AsRef<Path>) -> Option<PathBuf>
```

testgen/src/tester.rs
andrey-kuprianov: the last solution will not work, as I need this also to refer to and copy existing files from the test "root" directory. I think the better solution is not to check at all, and move the `cleanup` function away from the `TestEnv`, so no destructive actions will be available to the user. Instead I will do cleanup exclusively in the `Tester` itself, and only for the temporary dir it creates

testgen/src/tester.rs
andrey-kuprianov: That's indeed a good idea:) I have quickly tested, it seems to work. Thanks!

As a side-note: I find the current Rust APIs around Path/PathBuf/OsStr a bit over-complicated. Can't make a clear mental picture of it...

testgen/src/tester.rs
romac: Sounds like the best solution altogether!

testgen/src/tester.rs
romac: Do you mean the API exposed by the types or the need for these types themselves?

If the latter: the gist of it is:

- `PathBuf` is an owned path (similar to what `String` is to strings, or `Vec<u8>` to byte arrays)
- `Path` is a borrowed reference to a path (similar to what `&str` is to strings, or `&[u8]` is to byte arrays)
- `OsString` is an owned platform-native path string (~ `String`)
- `OsStr` is a borrowed reference to a platform-native string (~ `&str`)

For the latter two, I feel like the doc of [`OsString`](https://doc.rust-lang.org/stable/std/ffi/struct.OsString.html) gives a good explanation for why the type is needed. [This post](https://fasterthanli.me/articles/working-with-strings-in-rust) is another good resource worth reading.

andrey-kuprianov: Thank you for the explanation and the references, very useful. It's just the feeling that there are too many atoms in this string universe. The number of atoms, multiplied by the number of their interconnections is even bigger; hard to navigate;)  And also that feeling that this number could be reduced. Probably it will get easier as I learn more of the Rust standard libs, may be not. At least I am not alone:
* [pathbuf-and-path-why-not-string](https://users.rust-lang.org/t/pathbuf-and-path-why-not-string/28777)
* [is_working_with_paths_always_this_painful](https://www.reddit.com/r/rust/comments/7mu7q1/is_working_with_paths_always_this_painful/)

E.g., taking into account [PathBuf::into_os_str()](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.into_os_string) would it be a good idea just to merge the types `PathBuf`/`OsString`, and `Path`/`OsStr`? I don't know.

testgen/src/tester.rs
andrey-kuprianov: this works, thanks!

testgen/src/tester.rs
andrey-kuprianov: changed to `impl AsRef<Path>`

testgen/src/tester.rs
andrey-kuprianov: changed to `impl AsRef<Path>`

testgen/src/tester.rs
andrey-kuprianov: changed to `impl AsRef<Path>`

testgen/src/tester.rs
andrey-kuprianov: very nice; thanks! More precisely, this works: `.and_then(|mut file| write!(file, "{}\n", msg).ok())`

testgen/src/tester.rs
andrey-kuprianov: very nice; thanks! More precisely, this works: `.and_then(|mut file| write!(file, "{}\n", msg).ok())`

testgen/src/tester.rs
andrey-kuprianov: yes, this works, nice:)

testgen/src/tester.rs
andrey-kuprianov: Well, I've struggled a lot experimenting with different `FnOnce`, `Box<dyn FnOnce>`, etc. you may try yourself and follow the compiler errors;) The problem is that I need to store a function to call it later in `panic::catch_unwind`. This works fine with function pointers, but storing the closure is not that simple -- at least I have not managed to find a way to store a closure, and at the same time to satisfy the requirement that it can be used across unwind boundaries.

testgen/src/tester.rs
andrey-kuprianov: see above answer

testgen/src/tester.rs
romac: I see. Perhaps if you require a `Fn() + UnwindSafe + RefUnwindSafe`? Seems to work in [this simplified version](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b7643e64b74816d1b6841bd9312cb841) (which hopefully reproduces the same requirements).

testgen/src/tester.rs
andrey-kuprianov: done

testgen/src/tester.rs
andrey-kuprianov: Wow! THIS indeed works without any problems:)) The peace I was missing is `RefUnwindSafe`. Thanks a ton!

andrey-kuprianov:
@romac The integration tests are failing, but that's probably because I've merged master. Also, of everything is OK from your point of view, could you may be merge, and review the [549](https://github.com/informalsystems/tendermint-rs/pull/549)?

Thanks a lot!

romac (approved):

PR #555, opened 2020-09-07
spec: Light Client Supervisor

josef-widder:
This is the first draft ([rendered](https://github.com/informalsystems/tendermint-rs/blob/551b44f675a0cf4490c496071591135d87265ef1/docs/spec/lightclient/supervisor/supervisor.md)) for the lightclient supervisor. It addresses (partly) 
- the new lightstore semantics #499, 
- the proposed changes to the parameter handling from #509, 
- initialization from Genesis https://github.com/tendermint/spec/issues/131

In the spec, we call a function `AttackDetector` which will eventually implement the latest ideas to compute [evidence](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/attacks/evidence-handling.md) to address #461 (although proof of fork will only survive for the IBC use case). 

There are still many TODOs. I hope it is understandable nonetheless. I would prefer to start to work on the AttackDetector Spec now, before polishing this one.


josef-widder:
Opened a PR at #159 as we move all spec work to the tendermint/spec repo.

PR #557, opened 2020-09-08
Empty Merkle tree hash fix

### LINKED ISSUE #540: Revert the temporary revert in #538
#538: merkle: temporarily revert #514
The change introduced in #514 is necessary but we reverterd it to publish Tendermint v0.16. After that publish, the change should be included on master again so it is included with Tendermint v0.17.

greg-szabo:
Closes #540 .

Re-introduces the fix for empty Merkle tree hashing. This was created by Ismail before but it was pulled for the v0.16.0 release. We need it for 0.17.

* [X] Wrote tests
* [X] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## Unreleased
 
 - Add spec for the light client attack evidence handling ([#544])
+- return rfc6962 hash for empty merkle tree ([#498])
 
 [#544]: https://github.com/informalsystems/tendermint-rs/pull/544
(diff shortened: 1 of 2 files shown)


liamsi (approved):
LGTM 👍

ebuchman:
It bothers me that there's no integration test in the repo that would care about this. I guess that's because we don't have any tests with blocks that have txs or evidence? might be good to add a simple one for a block with some txs where we validate the header.

PR #558, opened 2020-09-08
Cargo option: Do not publish proto-compiler

### LINKED ISSUE #518: Explicitly mention that proto-compiler should not be published
It would be good to explicitly mention that this is not meant to be published (yet) via: https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

_Originally posted by @liamsi in https://github.com/informalsystems/tendermint-rs/pull/508_

greg-szabo:
Closes #518 .

This is an internal fix, no mention of it in the changelog.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


liamsi (approved):
👍

PR #563, opened 2020-09-11
Tendermint JSON test case fixes

### LINKED ISSUE #505: Update JSON formatting to 0.34

greg-szabo:
Closes #505 

* [X] Wrote tests
* [ ] Updated CHANGELOG.md


greg-szabo:
All right, this is ready for review. Some of the light-client tests had to be disabled that were affected because of the change in voting power ordering or protobuf-encoding. The light-client integration test was also lowered to one iteration (essentially disabled) until the voting power change is implemented. (#556 )
All other tests were fixed but it would be better to regenerate the JSON files.

light-client/tests/light_client.rs
```diff
@@ -136,6 +136,7 @@ fn bisection_test(tc: TestBisection<AnonLightBlock>) {
         None => false,
     };
 
+    dbg!(&tc.description);
```
greg-szabo, line 139:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
+name: Build
```
liamsi, line 1: What was the motivation to separate Build and test?

liamsi (approved):
IMO it could have been better to disable single test cases in some kind of blacklist. That way the JSON files could have served as a source for test-cases (s.t. we can guarantee the new generated tests to come will contain at least the cases that we already had).
On the other hand, deleting the failing cases is a bit cleaner. 

FYI @Shivani912 and @andrey-kuprianov

light-client/tests/support/single_step/sequential/commit/_less_than_one_third_nil_votes.json
```diff
@@ -0,0 +1,287 @@
+{
```
liamsi, line 1: Isn't this supposed to be ignored given the line in the gitignore above? `_*/*`

light-client/tests/support/single_step/sequential/commit/_more_than_two_third_vals_sign.json
```diff
@@ -0,0 +1,287 @@
+{
```
liamsi, line 1: Same question about gitgnore.

greg-szabo:
No test cases were deleted. The test harness reads a folder and ignores test JSON files that start with an underscore. I just renamed the test cases that failed with this underscore.

.github/workflows/build.yml
greg-szabo: If you look at GitHub reporting, now you get `Build / ...` and `Test / ...` separately. We don't need to build it for testing, so "Build" is a limited set of actions that don't run during testing. For example the WASM build has no test cases just a special build action.

light-client/tests/support/single_step/sequential/commit/_less_than_one_third_nil_votes.json
greg-szabo: No, the `.gitignore` only ignores the report folders that the test harness creates in case of failing tests. They look like this: `_bisection/report`, `_simple/report`, etc.

light-client/tests/support/single_step/sequential/commit/_more_than_two_third_vals_sign.json
greg-szabo: Same response. Gitignore only activates with an underscore in the folder name. Here, there is an underscore in the file name.

light-client/tests/support/single_step/sequential/commit/_less_than_one_third_nil_votes.json
andrey-kuprianov: In fact, because the tests for the light client are now [run with Testgen](https://github.com/informalsystems/tendermint-rs/blob/a1d51fa28eb18dcf8b5935dd6920fb29842c90f1/light-client/tests/light_client.rs#L189), it will ignore any files/folders starting with `_`

andrey-kuprianov: > IMO it could have been better to disable single test cases in some kind of blacklist. That way the JSON files could have served as a source for test-cases (s.t. we can guarantee the new generated tests to come will contain at least the cases that we already had).
> On the other hand, deleting the failing cases is a bit cleaner.
> 
> FYI @Shivani912 and @andrey-kuprianov

As I wrote above, Testgen will ignore any files or folders starting with `_`. So the easy way to disable a test or a set of tests, but to keep them at the same time, is to rename with `_` in front.

light-client/tests/support/single_step/sequential/commit/_more_than_two_third_vals_sign.json
liamsi: Ah OK. I see. Thanks for the clarification.

Issue #568, opened 2020-09-16
WebSockets follow up


ebuchman:
Websocket client was upgraded per #516. Lots of discussion there about minor follow ups and design improvements. Let's do the followup and capture the rest in the ADR:

- [x] Test it with transactions / ensure its working for the ibc relayer
- [x] Simplify/reduce number of features - #569 
- [ ] Send pings (keepalive) - https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r488887940
- Update ADR:
  - [x] Add concurrency diagram to clarify current concurrency model and how the components communicate with one another
  - [x] Clarify the `pending` states - https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r488332911
  - [x] Add notes about possible alternative/future approaches to concurrency model eg. 
    - Dropping the extra terminate channel and just using the cmd channel to communicate unsubsrciptions - https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r488884183 - Implemented in #569!
- [ ] Possibly eliminate async/await from the core, use a synchronous websocket with read/write routines, and wrap it in an async API for consumers that need it - https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r488892954

ebuchman:
ADR updates addressed in #576 but see https://github.com/informalsystems/tendermint-rs/pull/576#discussion_r493159561 re the need for pending.

ebuchman:
Closing for https://github.com/informalsystems/tendermint-rs/issues/590 and #582

PR #569, opened 2020-09-17
rpc: Simplify client interface

thanethomson:
Follow-up on #516, and in partial fulfillment of #568.

This PR:

1. Removes the `ClosableClient` trait. We now only expose a `close` method on clients that need it, and not in the traits. If we need it in one of the traits in future we can reintroduce it.
2. Reduces the number of RPC-related features down to 2 (see the table below).
3. Simplifies the internal communication mechanism with subscription drivers, as per [this comment](https://github.com/informalsystems/tendermint-rs/pull/516#discussion_r488884183).

The new features exposed by the `tendermint-rpc` crate are as follows:

| Feature | Description |
| ------- | ------------ |
| `default` | Only provides RPC types, no client(s) |
| `http-client` | Provides an RPC client for interacting with all of the Tendermint RPC HTTP endpoints (no subscription) |
| `websocket-client` | Provides an RPC client for managing event subscription via a WebSocket connection |

Some notes on these features:

* I couldn't call them simply `http` and `websocket` because `http` conflicts with our [`http`](https://crates.io/crates/http) dependency (and Rust won't let you use feature names that are the same as any of your dependencies).
* If you use either the `http-client` or `websocket-client` features, you also get access to the mock clients.

For more details, check out this branch locally and run:

```bash
cargo doc --all-features --open
```

* [ ] ~Referenced an issue explaining the need for the change~
* [ ] ~Wrote tests~
* [ ] ~Updated CHANGELOG.md~


rpc/src/client/subscription.rs
```diff
@@ -108,16 +109,37 @@ impl Subscription {
...
+        event_tx: ChannelTx<Result<Event>>,
+        /// Where to send the result of this subscription command.
+        result_tx: ChannelTx<Result<()>>,
+    },
+    /// Terminate an existing subscription.
+    Unsubscribe {
+        /// The ID of the subscription to terminate.
+        id: SubscriptionId,
+        /// The query associated with the subscription we want to terminate.
+        query: String,
```
ebuchman, line 137: A bit weird that we'd need the query in here, you'd think the id is enough to unsubscribe, but I guess given the query->(id->sub) map in the router its easier to lookup the sub if we have both.

ebuchman (approved):
Awesome, thanks!

rpc/src/client/subscription.rs
thanethomson: Yeah exactly. Either approach (supplying the query, or only supplying the ID) is an optimization of some kind, and given that we don't expect super high rates of unsubscribe requests, neither approach should have any meaningful impact.

This approach of supplying the query (which we have stored in the `Subscription` anyways) does, however, allow me to avoid writing another little helper method in the `SubscriptionRouter` to look up a query by its subscription ID.

PR #571, opened 2020-09-17
Blanket implementation for DomainType

greg-szabo:
Closes #537 again.

This change implements default functions for the DomainType trait, instead of implementing the functionality in the DomainType macro. The macro becomes syntactic sugar for `impl DomainType<RawType> for MyDomainStruct {}`.

It also fixes a "feature" that wasn't correctly implemented: all functions in the trait that relate to Prost::Message (encode, decode, encode_len, etc) now work exactly as intended and they don't consume the DomainType struct.

The trait now requires additional traits to be implemented on the struct (notably `Clone`) but it's a small price to pay.

It also enhances the DomainType trait with direct encoding/decoding using `Vec<u8>`. This functionality was available in the AminoMessage trait before.

Thanks to Ismail for pointing out that this can be done.

* [X] Wrote tests
* [ ] Updated CHANGELOG.md


proto-derive/src/lib.rs
```diff
@@ -55,41 +55,9 @@ fn expand_domaintype(input: &syn::DeriveInput) -> TokenStream {
         }
     };
 
+    // This is what all the hubbub is about.
```
liamsi, line 58: Not entirely. I was also wondering if we need the whole proc_macro at all. 
You can replace all these two lines that look like this
```rust
#[derive(DomainType)]
#[rawtype(RawPartSetHeader)]
```
with:
`impl DomainType<MyRawType> for ActualType {}`.
As a consequence you can delete the whole macro code:

https://github.com/informalsystems/tendermint-rs/blob/e89206774e5c1939c0c03c34782735e91621fef4/proto-derive/src/lib.rs#L10-L56


And in fact the whole proto-derive crate can be deleted, without losing any functionality or any further consequences. I would argue whole code becomes easier to navigate as the macro creates some unnecessary indirection for a single line that explains what is happing.

greg-szabo: Yes, true. Annotations seem cleaner to me, because I don't expect to look at the derive function. Serde does the same thing and it worked out pretty well there.

Also, we need to do some kind of JSON serialization for these domain types too and I didn't have time to get into the details of how we're going to do it. Maybe the derive macro helps, maybe not. Well, currently not really, it's just syntactic sugar.

I'll open this up to the team, I'm happy to do it either way, I just really wanted to do annotations. :)

liamsi (approved):
Great work 👍

proto/src/domaintype.rs
```diff
@@ -33,6 +33,13 @@
 //! Note that the Prost library and the TryFrom method have their own set of errors. These are
 //! merged into a custom Error type defined in this crate for easier handling.
 //!
+//! Requirements:
+//! * The DomainType trait requires the struct to implement the Clone trait.
+//! * The DomainType trait requires the struct to have a #[rawtype(MYRAWTYPE)] attribute set where
+//!   the MYRAWTYPE is a structure that has the prost::Message trait implemented. (protobuf type)
+//! * The DomainType trait requires that the TryFrom<RawType> implemented on the structure has an
+//!   error type that implements Into<BoxError>. (The current implementations with anomaly are
+//!   fine.)
```
liamsi, line 42: Actually some of the comments need updating now.

liamsi (changes requested):
I was too excited. It needs some cleanup in the documentation but everything else looks 👌

proto/src/domaintype.rs
greg-szabo: I'll do that.

greg-szabo: Minor updates in the documentation: removed references to the DomainType macro. I hope this is it. :)

liamsi (approved):
LGTM

PR #572, opened 2020-09-17
rpc: Fix and test WebSocket Tx event subscription

thanethomson:
In partial fulfillment of #568.

This adds an integration test to ensure that transaction subscription works (WebSocket-based subscription with query=`tm.event='Tx'`).

I had to make some fields in the relevant data structures optional, since they aren't always returned by the RPC remote endpoint.

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/rpc/src/client/subscription.rs
+++ b/rpc/src/client/subscription.rs
@@ -1,5 +1,8 @@
 //! Subscription- and subscription management-related functionality.
 
+#[cfg(feature = "websocket-client")]
+pub use two_phase_router::{SubscriptionState, TwoPhaseSubscriptionRouter};
+
 use crate::client::sync::{unbounded, ChannelRx, ChannelTx};
 use crate::event::Event;
 use crate::{Error, Id, Result};
@@ -205,30 +208,6 @@ impl SubscriptionId {
     }
 }
 
-#[derive(Debug)]
-struct PendingSubscribe {
-    id: SubscriptionId,
-    query: String,
-    event_tx: ChannelTx<Result<Event>>,
-    result_tx: ChannelTx<Result<()>>,
-}
-
-#[derive(Debug)]
-struct PendingUnsubscribe {
-    id: SubscriptionId,
-    query: String,
-    result_tx: ChannelTx<Result<()>>,
-}
-
-/// The current state of a subscription.
-#[cfg(feature = "websocket-client")]
-#[derive(Debug, Clone, PartialEq)]
-pub enum SubscriptionState {
-    Pending,
-    Active,
-    Cancelling,
-}
-
 /// Provides a mechanism for tracking [`Subscription`]s and routing [`Event`]s
 /// to those subscriptions.
 ///
@@ -240,14 +219,6 @@ pub struct SubscriptionRouter {
     // their result channels. Used for publishing events relating to a specific
     // query.
     subscriptions: HashMap<String, HashMap<SubscriptionId, ChannelTx<Result<Event>>>>,
-    // A map of JSON-RPC request IDs (for `/subscribe` requests) to pending
-    // subscription requests.
-    #[cfg(feature = "websocket-client")]
-    pending_subscribe: HashMap<String, PendingSubscribe>,
-    // A map of JSON-RPC request IDs (for the `/unsubscribe` requests) to pending
-    // unsubscribe requests.
-    #[cfg(feature = "websocket-client")]
-    pending_unsubscribe: HashMap<String, PendingUnsubscribe>,
 }
 
 impl SubscriptionRouter {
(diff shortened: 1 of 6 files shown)


rpc/src/client/subscription.rs
```diff
@@ -241,9 +242,11 @@ pub struct SubscriptionRouter {
     subscriptions: HashMap<String, HashMap<SubscriptionId, ChannelTx<Result<Event>>>>,
     // A map of JSON-RPC request IDs (for `/subscribe` requests) to pending
     // subscription requests.
+    #[cfg(feature = "websocket-client")]
```
ebuchman, line 245: Can't we just put these on the struct instead of the fields?

rpc/src/client/subscription.rs
```diff
@@ -411,6 +421,7 @@ impl SubscriptionRouter {
 
     /// Obtain a mutable reference to the subscription with the given ID (if it
     /// exists).
+    #[cfg(feature = "websocket-client")]
```
ebuchman, line 424: Do we need these on every method or can we just put them on the struct def?

tendermint/tests/integration.rs
```diff
@@ -171,4 +174,79 @@ mod rpc {
...
+                    .await
+                    .unwrap();
+            }
+        });
+
+        println!(
+            "Attempting to grab {} transaction events",
+            expected_tx_values.len()
+        );
+        // We reverse the expected tx values because we're popping them off the
```
ebuchman, line 209: This is confusing. Aren't they being received in the original order? Reversing so we can use `pop` because it cuts from the end of the array?

ebuchman (commented):
Nice test!

rpc/src/client/subscription.rs
thanethomson: So the need for this comes from all those annoying clippy warnings we keep getting, and they only come up when the `default` or `http-client` features are active (but they go away when the `websocket-client` feature is present).

I'll experiment with having two separate `impl` blocks for the same struct and adding the feature attribute to the WebSocket client-specific `impl` block. Haven't tried that yet.

tendermint/tests/integration.rs
thanethomson: Yeah, `pop` cuts elements from the end of the array. Actually, I just found that I can use `expected_tx_values.remove(0)` instead (see [`remove`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove)). I'll change that 👍

rpc/src/client/subscription.rs
thanethomson: I've refactored this a bit, separating out standard `SubscriptionRouter` functionality from `TwoPhaseSubscriptionRouter` functionality.

The two-phase subscription router is a composition of a `SubscriptionRouter` that enhances it with two-phase addition/removal of subscriptions. This makes it simpler to delineate which parts are only relevant for the `websocket-client` feature.

rpc/src/client/subscription.rs
thanethomson: See comment above

thanethomson:
The integration test will necessarily fail until https://github.com/tendermint/tendermint/pull/5378 is merged and available in the Docker image.

rpc/src/client/subscription.rs
```diff
@@ -358,91 +270,221 @@ impl SubscriptionRouter {
...
+    /// A subscription router that can manage pending subscribe and unsubscribe
+    /// requests, as well as their confirmation/cancellation.
+    ///
+    /// This is useful in instances where the underlying transport is complex,
+    /// e.g. WebSocket connections, where many messages are multiplexed on the
+    /// same communication line. In such cases, a response from the remote
+    /// endpoint immediately after a subscribe/unsubscribe request may not be
+    /// relevant to that request.
+    #[derive(Debug)]
+    pub struct TwoPhaseSubscriptionRouter {
```
ebuchman, line 296: Don't love the name. Not sure what a better one would be but probably something more descriptive of the functionality. I guess it's kind of like a cache? I also still cant help but wonder if we can get rid of it per https://github.com/informalsystems/tendermint-rs/pull/576#discussion_r493159561.

rpc/src/event.rs
```diff
@@ -46,17 +47,17 @@ pub enum EventData {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct TxInfo {
     pub height: String,
-    pub index: i64,
+    pub index: Option<i64>,
```
ebuchman, line 50: Weird that this could be empty I'd expect it to just be 0, I guess it's an `omitempty` on the Go side - https://github.com/tendermint/tendermint/blob/master/abci/types/types.pb.go#L2764

ebuchman (approved):

PR #576, opened 2020-09-21
rpc: Update and expand ADR-008

thanethomson:
Partially addresses #568.

This PR updates ADR-008 to include:

1. A description of why two-phase subscribe/unsubscribe requests are necessary for the WebSocket subscription client.
2. A slight expansion to the concurrency model description to illustrate the various different components involved in facilitating the "happy path" of a simple event subscription process.

[Rendered version of the ADR in this PR](https://github.com/informalsystems/tendermint-rs/blob/adr/subscription-update/docs/architecture/adr-008-event-subscription.md)

* [ ] ~Wrote tests~
* [ ] ~Updated CHANGELOG.md~


docs/architecture/adr-008-event-subscription.md
```diff
@@ -246,6 +246,12 @@ In cases where a driver is necessary, the client implementation would have to
 become a **handle** to the driver, facilitating communication with it across
 asynchronous tasks.
 
+A rough sketch of the interaction model between the different components
+envisaged to make up the subscription subsystem is shown in the following
+sequence diagram.
+
+![RPC client concurrency model](./assets/rpc-client-concurrency.png)
```
ebuchman, line 253: This image is awesome!

docs/architecture/adr-008-event-subscription.md
```diff
@@ -347,6 +353,41 @@ pub enum Operand {
...
+particular event in the JSON-RPC message, and so this data structure is optimal
+for such a configuration. This will necessarily change if Tendermint's WebSocket
+server [drops subscription IDs from events][tendermint-2949], which is likely if
+we want to conform more strictly to the [JSON-RPC standard for
+notifications][jsonrpc-notifications].
+
+#### Two-Phase Subscribe/Unsubscribe
+
+Due to the fact that a WebSocket connection lacks request/response semantics,
+when managing multiple subscriptions from a single client we need to implement a
```
ebuchman, line 378: So, it's true that events from previous subscriptions can be interleaved between the request and response for a new subscription, but when is that a problem? Is it only if there is an error in the new subscription? Are there cases where a new subscription might error but everything else keeps working? Or why couldn't we just optimistically confirm the new subscription and then just remove it if it turns out we receive an error?

ebuchman (approved):

docs/architecture/adr-008-event-subscription.md
thanethomson: That'd change the developer ergonomics. Here's what the two-phase subscription mechanism allows us:

```rust
let mut subs = match client
    .subscribe("tm.event='NewBlock'".to_string())
    .await {
    Ok(s) => s,
    Err(e) => { /* handle subscription error */ },
};
// ^^^ Ok(Subscription) means here that we have a well-formed
// subscription, accepted by the server, which is intuitively what I'd
// expect from a successful subscription request.

// Here, `subs` might be moved into a different async context with a certain
// degree of confidence since the subscription was successful.
while let Some(res) = subs.next().await {
    match res {
        Ok(event) => { /* handle event */ },
        Err(e) => {
            // Errors here can only be caused by network failures or internal
            // server errors, but we know already our subscription was accepted
            // by the server.
        }
    }
}
```

as opposed to, if we remove the two-phase subscription mechanism:

```rust
let mut subs = match client
    .subscribe("tm.event='NewBlock'".to_string())
    .await {
    Ok(s) => s,
    Err(e) => { /* handle networking error (caused by WebSocket send failure) */ },
};
// ^^^ The meaning of getting an Ok(Subscription) here changes now to
// something more like MaybeOk(Subscription).

// Here, `subs` might be moved into a different async context, and
// you have no guarantee it's been accepted by the server.
while let Some(res) = subs.next().await {
    match res {
        Ok(event) => { /* handle event */ },
        Err(e) => {
            // Depending on the use case, we may now need to handle
            // subscription failures here, which could be tricky if we're trying
            // to manage multiple subscriptions simultaneously or if `subs`
            // has been moved into a different async context.
        }
    }
}
```

To me, it seems like doing away with the two-phase subscription just pushes the additional complexity onto the developer and muddies the API.

ebuchman: I duno. This makes sense on the surface but not sure how negative it would really be. Not sure I understand the implications of "moving into a different async context". But the other thing that's nagging me more is when would a subscription even fail ? Are there really cases where a new subscription could fail but everything else just keeps working normally? I guess we'd have to dive into the server more to figure it out.

Anyways, not a priority to deal with this, it's already written and merged and much more robust this way, just thinking aloud.

PR #578, opened 2020-09-23
Bump `k256` crate dependency to v0.5

tony-iqlusion:
Release notes: https://github.com/RustCrypto/elliptic-curves/pull/188

### DIFF
--- a/tendermint/src/public_key.rs
+++ b/tendermint/src/public_key.rs
@@ -3,7 +3,7 @@
 pub use ed25519_dalek::PublicKey as Ed25519;
 
 #[cfg(feature = "secp256k1")]
-pub use k256::PublicKey as Secp256k1;
+pub use k256::EncodedPoint as Secp256k1;
 
 use crate::{
     error::{self, Error},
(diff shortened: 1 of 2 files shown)


tendermint/src/public_key.rs
```diff
@@ -3,7 +3,7 @@
 pub use ed25519_dalek::PublicKey as Ed25519;
 
 #[cfg(feature = "secp256k1")]
-pub use k256::PublicKey as Secp256k1;
+pub use k256::EncodedPoint as Secp256k1;
```
tony-iqlusion, line 6: Alternatively this could now use the [`k256::ecdsa::VerifyKey`](https://docs.rs/k256/0.5.2/k256/ecdsa/struct.VerifyKey.html) type, which is guaranteed to be a valid ECDSA/secp256k1 verification key.

liamsi: Is there any downside to using `VerifyKey`? It seems like the right choice here (unless EncodedPoint is also guaranteed to be a valid pubkey too).

tony-iqlusion: It's definitely better for this use case, IMO. The only drawback is it's (more of) a breaking change for downstream consumers, although I think at this point the only notable one is the KMS.

tony-iqlusion: Just tried switching it over. One current problem is `k256::ecdsa::VerifyKey` isn't a `Copy` type, although that's corrected easily enough.

tony-iqlusion: Another issue is there isn't a simple reference conversion between `&k256::ecdsa::VerifyKey` and `&[u8]`, which means the `PublicKey::as_bytes()` method can't be implemented (instead `to_bytes()` would be required).

That doesn't necessarily seem like the end of the world to me, but it is a notable problem.

ebuchman (approved):
Cool I used this version to get tx signing working for stargate in https://github.com/informalsystems/ibc-rs/pull/242

Issue #486, opened 2020-07-28
light-node: Error Handling and Exiting


ebuchman:
Breakout issue from #442. 

- [ ] consider `process::exit` alternatives: https://github.com/informalsystems/tendermint-rs/pull/430/files#r452658991
- [ ] include peer address in error logs. ie. `error: sync failed: I/O error: Internal error: height 194 must be less than or equal to the current blockchain height 0 (code: -32603)` doesn't tell me what peer this issue is with.

ebuchman:
@romac which part did #583 address ?

thanethomson:
Light node was archived.

Issue #484, opened 2020-07-28
light-node: Initialization Follow up


ebuchman:
Original issue https://github.com/informalsystems/tendermint-rs/issues/293#issue-629635994 was closed with #430. This is a breakout issue from #442.

Two main pieces:

- [ ] Specification needs to be clarified. Subjective init is currently under specified and this needs to be fixed. Eg. we may need to check that the initial trusted block is corroborated by witnesses (https://github.com/informalsystems/tendermint-rs/pull/430#discussion_r452789386)
- [x] `light-client` crate needs to expose functions for initialization. Currently the caller has to do some IO and set things up itself (https://github.com/informalsystems/tendermint-rs/pull/430#discussion_r452791199)

brapse:
This is related to some specification work https://github.com/tendermint/spec/issues/131.

ebuchman:
Per #422, consider using the [builder pattern](https://www.ameyalokare.com/rust/2017/11/02/rust-builder-pattern.html) to streamline creation of the light client instance

ebuchman:
hey @romac, does #583 sufficiently address the second bullet point in this issue? 

@milosevic do we need to do anything more re the spec for initialization?

romac:
> hey @romac, does #583 sufficiently address the second bullet point in this issue?

Yeah it does (even though we could still improve the API further)

thanethomson:
Since the second point is marked as done and the specs have moved to https://github.com/tendermint/spec, I'm going to close this. Feel free to reopen if there's anything still needed for this issue that's relevant to this repo.

PR #583, opened 2020-09-24
Implement builder DSL for light client initialization, and other things

romac:
Apologies for this huge, many-headed, multi-faced PR which would deserve to be split up but that has become quite difficult as I have been evolving it alongside the work on the ibc-rs side. Let me know if it's too much to handle, and I'll try to do it by the end of the week.

---

Here's what it does:

- Expose a builder DSL for initializing light client instances and their supervisor (partially addresses #484)
- Adapt the light node commands to make use of the new DSL
- Refactor the `Io` component a bit to enable the aforementioned DSL
- Improves error handling in the light node, just a bit (partially addresses #486)
- Improves the definition of the supervisor `Handle` for use within ibc-rs

---

**Disclaimer 0:** Documentation for the builder DSL is really minimal at the moment.

**Disclaimer 1:** This PR does **NOT** address #497.

**Disclaimer 2:** This PR is lacking unit tests for the new builder DSL.

**Disclaimer 3:** I have not tested whether the light node CLI still works.

**Disclaimer 4:** The builder DSL could definitely still be improved, but I'd first like to use it within ibc-rs before making other changes.

---

Given all this, I think that the easiest way to review this PR is to:

- Start with the first 6 commits (up to, and including, 07a2427).
- Then review the builders themselves in `light-client/src/builder.rs` and `light-client/src/builder/`.
- Then focus on the rest of the changes within the `light-client` directory.
- And finally review the changes within the `light-node` directory.

I won't assign any reviewers given the poor state of the PR, so feel free to assign yourselves.

---

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /// Set the given light block as the initial trusted state.
+    pub fn trust_light_block(
+        mut self,
+        trusted_state: LightBlock,
+    ) -> Result<LightClientBuilder<HasTrustedState>, Error> {
+        self.validate(&trusted_state)?;
+
+        // TODO(liamsi, romac): it is unclear if this should be Trusted or only Verified
```
melekes, line 125: it may be a good idea to cross-verify the first header to check if all witnesses have the same view.

but in the end, you have to trust the state I think.

light-client/src/builder/supervisor.rs
```diff
@@ -0,0 +1,114 @@
...
+        address: net::Address,
+        instance: Instance,
+    ) -> SupervisorBuilder<Done> {
+        self.instances.witness(peer_id, instance);
+        self.addresses.witness(peer_id, address);
+
+        self.with_state(Done)
+    }
+
+    /// Add mulitple witnesses at once.
```
thanethomson, line 81:
```suggestion
    /// Add multiple witnesses at once.
```

light-client/src/utils/block_on.rs
```diff
@@ -0,0 +1,29 @@
+use std::{future::Future, time::Duration};
+
+use crate::{components::io::IoError, types::PeerId};
+
+/// Run a future to completion on a new thread, with the given timeout.
+///
+/// This function will block the caller until the given future has completed.
+pub fn block_on<F>(f: F, peer: PeerId, timeout: Option<Duration>) -> Result<F::Output, IoError>
```
thanethomson, line 8: I don't think we're going to solve the sync/`async` conundrum in the short term (and definitely not in this PR), but to get the two to play nicely together we should try to figure out a simple, low-effort way to spawn a single Tokio runtime in a separate thread to which we can submit `async` tasks from synchronous code.

thanethomson (dismissed):
I'd personally say that this looks good enough to merge if it works for `ibc-rs` 👍

light-client/src/builder/light_client.rs
romac: > It may be a good idea to cross-verify the first header to check if all witnesses have the same view.

Sounds like a good idea, indeed!

@ebuchman What do you think? Should we address this in a follow-up PR?

light-client/src/builder/supervisor.rs
```diff
@@ -0,0 +1,114 @@
...
+        self.instances.primary(peer_id, instance);
+        self.addresses.primary(peer_id, address);
+
+        self.with_state(HasPrimary)
+    }
+}
+
+impl SupervisorBuilder<HasPrimary> {
+    /// Add a witness [`Instance`].
+    pub fn witness(
```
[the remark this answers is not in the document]
romac: We actually use this method on the IBC side, and it's quite useful when we statically know how many witnesses we have, in contrast with the `witnesses` method which should return an error if the provided witness list is empty (as of c95ebcd8).

I therefore suggest we keep it, at least for the moment.

ebuchman:
@romac what about the failing integration test?

romac:
Weird! Will take a look asap

light-client/src/builder/light_client.rs
ebuchman: Yup, if anyone disagrees should probably throw an error and refuse to start

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+        let trusted_state = self
+            .light_store
+            .latest_trusted_or_verified()
+            .ok_or_else(|| error::Kind::NoTrustedStateInStore)?;
+
+        self.trust_light_block(trusted_state)
+    }
+
+    /// Fetch and set the latest block from the primary peer as the trusted state.
+    pub fn trust_primary_latest(self) -> Result<LightClientBuilder<HasTrustedState>, Error> {
```
ebuchman, line 143: This is a bit dangerous and should either come with a big warning or be removed all together

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+            clock,
+            scheduler,
+            options,
+            predicates,
+            state: NoTrustedState,
+        }
+    }
+
+    /// Set the given light block as the initial trusted state.
+    pub fn trust_light_block(
```
ebuchman, line 119: Should this be public?

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        self.trust_light_block(trusted_state)
+    }
+
+    fn validate(&self, light_block: &LightBlock) -> Result<(), Error> {
+        // TODO(ismail, romac): actually verify more predicates of light block?
+
+        self.predicates
+            .validator_sets_match(light_block, &*self.hasher)
```
ebuchman, line 186: Important thing is actually the next_validators!

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+                given: trusted_hash,
+                found: header_hash,
+            });
+        }
+
+        self.trust_light_block(trusted_state)
+    }
+
+    fn validate(&self, light_block: &LightBlock) -> Result<(), Error> {
+        // TODO(ismail, romac): actually verify more predicates of light block?
```
ebuchman, line 183: We should probably check:
- is_within_trust_period
- is_header_from_past
- validator_sets_match (though this doesn't really matter since we will never use these vals)
- next_validators_match

Everything else depends on the commit and/or previous blocks. We could verify the commit but it's not really necessary.

light-client/tests/integration.rs
```diff
@@ -91,16 +86,21 @@ fn sync() {
...
 
-    let peer_list = PeerList::builder()
-        .primary(primary, primary_instance)
-        .witness(witness, witness_instance)
-        .build();
+    let peer_list = {
+        let mut builder = PeerList::builder();
+        builder.primary(primary, primary_instance);
+        builder.witness(witness, witness_instance);
+        builder.build()
```
ebuchman, line 100: Can't these still be chained?

ebuchman (approved):
Awesome. A bunch of comments but don't think any of them need to block, just need to figure out / fix the failing test

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+    io: Box<dyn Io>,
+    clock: Box<dyn Clock>,
+    hasher: Box<dyn Hasher>,
+    verifier: Box<dyn Verifier>,
+    scheduler: Box<dyn Scheduler>,
+    predicates: Box<dyn VerificationPredicates>,
+    light_store: Box<dyn LightStore>,
+
+    #[allow(dead_code)]
+    state: State,
```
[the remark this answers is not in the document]
romac: We need a field here because in Rust type parameters of a struct have to appear in its fields.

We could use `PhantomData<State>`, but since the two type states are zero-sized anyway (same as `PhantomData`), this saves us an import here and some potential confusion wrt to `PhantomData`.

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+impl<Current> LightClientBuilder<Current> {
```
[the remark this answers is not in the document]
romac: This pattern can definitely [be expressed both ways](http://cliffle.com/blog/rust-typestate/). I think it's just a matter of preference. I prefer the type param approach as it to me it expresses more clearly that the purpose of the struct stays the same, it's just in a different state. Perhaps it's just because I am more used to this pattern from Haskell or Scala, but in the end I don't feel that strongly either way. So if you feel like the wrapper structs would be more readable, I am happy to make the change.

light-client/src/builder/light_client.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// Private method to move from one state to another
+    fn with_state<Next>(self, state: Next) -> LightClientBuilder<Next> {
+        LightClientBuilder {
+            peer_id: self.peer_id,
```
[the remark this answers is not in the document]
romac: Unfortunately not, using `..self` yields the following error because `self` does not have the right type:

```
error[E0308]: mismatched types
  --> light-client/src/builder/light_client.rs:53:39
   |
50 | impl<Current> LightClientBuilder<Current> {
   |      ------- found type parameter
51 |     /// Private method to move from one state to another
52 |     fn with_state<Next>(self, state: Next) -> LightClientBuilder<Next> {
   |                   ---- expected type parameter
53 |         LightClientBuilder { state, ..self }
   |                                       ^^^^ expected type parameter `Next`, found type parameter `Current`
   |
   = note: expected struct `builder::light_client::LightClientBuilder<Next>`
              found struct `builder::light_client::LightClientBuilder<Current>`
   = note: a type parameter was expected, but a different one was found; you might be missing a type parameter or trait bound
   = note: for more information, visit https://doc.rust-lang.org/book/ch10-02-traits.html#traits-as-parameters
```

light-client/src/builder/light_client.rs
romac: Let's remove it since we don't have any use for it at the moment.

light-client/tests/integration.rs
romac: Not since the API change in 26d1efe4c3ce5bbc6436c0deb8f2e83bbfb00ef9.

I figured I would follow the [current recommendations](https://docs.rs/derive_builder/0.9.0/derive_builder/#mutable-aka-non-consuming-recommended) of the popular `derive_builder` crate. I agree that the chained version looks a lot better, so I would definitely not mind going back on that.

light-client/src/builder/light_client.rs
romac: Good catch. It should not, as it is quite dangerous and is only currently used in the light client example. The latter could be rewritten to use `trust_from_store` instead so it's not needed at all. Will remove it.

romac:
@thanethomson @ebuchman Thanks a lot for your inputs!

Just pushed a few commits which address some of the comments above. Will address the rest + CI failure tomorrow morning.

light-client/src/builder/light_client.rs
romac: 3f496fd

light-client/src/builder/light_client.rs
romac: 3f496fd

light-client/src/builder/light_client.rs
romac: 2705247

light-client/src/builder/light_client.rs
romac: 0e1709d

light-client/src/builder/light_client.rs
cmwaters: yeah I agree. We currently just check that the chain ID's are the same but it would make more sense to cross-verify the first header. This way we can also check for responsiveness i.e. if a witness doesn't respond we drop them, and for correctness i.e. if one of the headers are different we halt and return the error

josef-widder: I have collected the outcome of the discussions regarding initialization into
https://github.com/tendermint/spec/pull/159

We have to trust the given initial header. Cross-check doesn't do anything useful. It only opens the possibility that one fault peer can prevent the light node to start.

romac:
@ebuchman @thanethomson This PR is ready for merge imho, now that we have figured out where the CI failure came from (https://github.com/informalsystems/tendermint-rs/issues/600).

light-client/src/builder/light_client.rs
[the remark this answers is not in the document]
ebuchman: Lol ok. I think this and the need to have a dummy state field both motivate using the wrapper approach as I think that would solve both, but no biggie

ebuchman (approved):
Great!

light-client/src/builder/light_client.rs
ebuchman: Might as well catch that error at startup and get the user to provide a different witness rather than catch it later and just drop the witness without replacement, no ?

josef-widder: > Might as well catch that error at startup and get the user to provide a different witness rather than catch it later and just drop the witness without replacement, no ?

We can do that. Still, the user might need to do that several times, if the user is unlucky and always chooses another faulty peer. And these faulty peers don't even need to be clever but just return garbage with the right height. Still, if it makes sense in the practical setting we can do that. It is definitely safe.

Issue #582, opened 2020-09-24
rpc: Further improvements


thanethomson:
Building on #568, I'd like to recommend we implement the following improvements over time (this is a placeholder issue to keep track of what still needs improvement, from what I can currently see):

- [x] A structured query interface, [as per the ADR](https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-008-event-subscription.md#query).
- [x] Do an audit of the types employed by the event and query mechanisms to see if they need updating as per our recent domain types work.
- [ ] Implement support for more event types than just `NewBlock` and `Tx`.
- [x] Implement the full `Client` trait for the `WebSocketClient`.
- [x] Resolve the question around async vs sync I/O and, if sync is the way forward with async wrappers, submit a new ADR to outline how to go about migrating from async to sync (see #318).

tarcieri:
> Resolve the question around async vs sync I/O and, if sync is the way forward with async wrappers

I thought this was only for ABCI?

Much of the RPC client is factored in such a way that supporting both sync and async shouldn't be that difficult, so I think it'd be weird to write an async wrapper for a sync client.

As it were, the KMS transaction signer makes pretty extensive use of async RPC already...

thanethomson:
> I thought this was only for ABCI?

My understanding, given conversations with @ebuchman recently and issues like #318, is that we need to eventually settle on a broad approach to I/O that facilitates both async and sync access (beyond just ABCI and the RPC even). If we eventually want to facilitate both, it seems like the easiest way right now to do so is to write mostly synchronous code and wrap it in async code. Not sure how that plays out with the use of event-driven networking though.

The tickbox in this issue, however, relates purely to discussing and making a decision along these lines.

tarcieri:
I left more or less the same comment about async RPC on #318. See this one in particular:

https://github.com/informalsystems/tendermint-rs/issues/318#issuecomment-644808670

> If we eventually want to facilitate both, it seems like the easiest way right now to do so is to write mostly synchronous code and wrap it in async code. Not sure how that plays out with the use of event-driven networking though.

That makes sense for ABCI, where you can expose a synchronous API for running 3 threads, and then wrap each thread in channels. That's particularly advantageous given there are a number of different channel types to choose from depending on needs, and multiple ways to wrap up those channel-based APIs (tokio, tower, async-std).

[...]

PR #584, opened 2020-09-24
rpc: Add structured querying for event subscriptions

thanethomson:
Partially addresses #582.

This PR adds structured querying to RPC event subscription, which replaces the `String`-based interface for subscription.

No query parsing is implemented yet, as it's not clear yet whether we need it at the moment and we need to figure out how to translate the [query PEG](https://github.com/tendermint/tendermint/blob/master/libs/pubsub/query/query.peg) from the Go version of Tendermint.

* [x] Wrote tests
* [x] Updated CHANGELOG.md


romac:
@thanethomson Great stuff!

I took the liberty pushing a few non-controversial changes (imho):

- Add missing doc comments to some methods of `Query` (27f2de9ecfc1761bbfc2d335c5ff0ce1204d0475)
- Use `Display` instead of `ToString` to pretty-print queries, which is [more](https://doc.rust-lang.org/std/string/trait.ToString.html) [idiomatic](https://stackoverflow.com/a/27770058) (568abaf)
- Impement `From` instead of `Into` for `Operand`, which again is [more idiomatic](https://doc.rust-lang.org/std/convert/trait.Into.html) (345f281)

The last commit (27fe514) merges the `Operation` and `Condition` structs together, which I personally find simpler even though it distributes the key into each variant of the enum. I don't feel strongly about this and only did this as a suggestion so let me know if you'd rather keep the original version and I'll revert the commit.

romac:
Looks like the build [is failing on Nightly](https://github.com/informalsystems/tendermint-rs/runs/1219272754).

```
2020-10-07T08:16:01.4214668Z    Compiling ecdsa v0.8.3
2020-10-07T08:16:01.8028615Z error[E0277]: cannot multiply `&NonZeroScalar<C>` to `<C as ProjectiveArithmetic>::ProjectivePoint`
2020-10-07T08:16:01.8035228Z ##[error]  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/ecdsa-0.8.3/src/sign.rs:90:58
2020-10-07T08:16:01.8043918Z    |
2020-10-07T08:16:01.8044748Z 90 |             public_key: (C::ProjectivePoint::generator() * &self.secret_scalar).to_affine(),
2020-10-07T08:16:01.8045596Z    |                                                          ^ no implementation for `<C as ProjectiveArithmetic>::ProjectivePoint * &NonZeroScalar<C>`
2020-10-07T08:16:01.8046174Z    |
2020-10-07T08:16:01.8046773Z    = help: the trait `Mul<&NonZeroScalar<C>>` is not implemented for `<C as ProjectiveArithmetic>::ProjectivePoint`
2020-10-07T08:16:01.8047526Z help: consider further restricting the associated type
2020-10-07T08:16:01.8047864Z    |
2020-10-07T08:16:01.8048427Z 87 |         AffinePoint<C>: Clone + Debug, <C as ProjectiveArithmetic>::ProjectivePoint: Mul<&NonZeroScalar<C>>
2020-10-07T08:16:01.8049018Z    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2020-10-07T08:16:01.8049204Z 
2020-10-07T08:16:01.8476530Z error: aborting due to previous error
2020-10-07T08:16:01.8477619Z ##[error]aborting due to previous error
2020-10-07T08:16:01.8479396Z For more information about this error, try `rustc --explain E0277`.
2020-10-07T08:16:01.8636042Z error: could not compile `ecdsa`
```

thanethomson:
Excellent, thanks @romac! Looks great - I like the simplification 👍

Are there any clippy lints we can enable that can warn us about there being more idiomatic ways to implement specific pieces of code?

romac:
Not to my knowledge. Closest lint I could find is https://rust-lang.github.io/rust-clippy/master/#inherent_to_string but it would not have helped in this case.

I guess there are valid use cases for implementing ToString or defining a From instance directly which is why Clippy does not warn about that, or perhaps those lints are just currently missing.

romac (approved):

tony-iqlusion:
My best guess for the `ecdsa` breakage on nightly is something broken with `Deref` rules.

I can try to repro it locally and see if there's a quick fix to work around the issue. This might also be worth reporting upstream.

tony-iqlusion:
Plausible cause of the nightly regression: https://github.com/rust-lang/rust/issues/77638

PR #588, opened 2020-09-29
Add integration test for catching empty merkle root bug

### LINKED ISSUE #562: Add a test to catch the empty merkle root bug
Not totally clear why this wasn't being caught by integration tests but we should have some simple integration test that would catch #498. I guess there's nothing currently checking the header field hashes or even the hash of the header itself based on the block data? It's a bit of block validation functionality which we don't really have yet but we should start with enough to catch this so at least our Merkle tree is being tested in real context of being in a block.

The bug was fixed in #557 and has a unit test but no block-level test.

Shivani912:
Closes #562

P.S.: Is it okay to have the check inside the test for `/block` endpoint? Since this is part of block validation, I placed it there.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/tendermint/tests/integration.rs
+++ b/tendermint/tests/integration.rs
@@ -14,7 +14,7 @@ mod rpc {
     use tendermint_rpc::{Client, HttpClient, Id, SubscriptionClient, WebSocketClient};
 
     use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


ebuchman (approved):
Thanks!

PR #589, opened 2020-09-29
Remove public key field from DuplicateVoteEvidence

### LINKED ISSUE #502: Remove DuplicateVoteEvidence PublicKey
I just noted that in the go versions there is no pubkey field (anymore?). We need to remove this for compatibility. Tracking here: https://github.com/informalsystems/tendermint-rs/issues/300

_Originally posted by @liamsi in https://github.com/informalsystems/tendermint-rs/pull/242_

thanethomson:
Closes #502 

Doesn't seem like the `pub_key` field was used anywhere at all.

* [ ] ~Updated all relevant documentation in docs~
* [x] ~Wrote~ Updated tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/rpc/tests/support/block_with_evidences.json
+++ b/rpc/tests/support/block_with_evidences.json
@@ -43,7 +43,7 @@
           {
             "type": "tendermint/DuplicateVoteEvidence",
             "value": {
-              "VoteA": {
+              "vote_a": {
                 "type": 1,
                 "height": "21",
                 "round": "0",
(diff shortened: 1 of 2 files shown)


ebuchman (commented):
Looks good but we should fix the JSON too: https://github.com/tendermint/tendermint/blob/master/types/evidence.go#L37

thanethomson:
> Looks good but we should fix the JSON too: https://github.com/tendermint/tendermint/blob/master/types/evidence.go#L37

Done!

ebuchman (approved):

Issue #3, opened 2019-08-01
Tendermint client does not support batched RPC requests


tarcieri:
Copying/pasting this from https://github.com/tendermint/kms/issues/301 as reported by @devashishdxt

> Batched RPC requests are supported by tendermint since v0.31.6
>
> tendermint/tendermint#3534

thanethomson:
Do we want similar semantics to how the Tendermint RPC client batch API works? See [github.com/tendermint/tendermint/rpc/client/examples_test.go](https://github.com/tendermint/tendermint/blob/master/rpc/client/examples_test.go).

Essentially these semantics require that we:
1. Use the client to create a batch (with similar methods as the client; will derive the target address from the client)
2. Call methods on the batch that correspond to requests (e.g. `broadcast_tx_async`). This caches the calls in the batch.
3. Call a `send` method that serializes the requests and returns an array of results, with each result corresponding to a cached request.

Once we agree on semantics, I can look at coding this in the coming weeks.

tarcieri:
@thanethomson I don't know the answers to your questions, but will say that the current client probably needs some refactoring in general, especially with the recent Rust developments in `async`/`await` support. I'm happy to field questions about that, and it seems like you may want to use futures to implement some of what you're describing.

romac:
This was likely closer by mistake because a commit in 592 contained `#1`.

Issue #2, opened 2019-08-01
Secret Connection network timeouts


tarcieri:
Copying/pasting this from the original issue at https://github.com/tendermint/kms/issues/310

The Secret Connection implementation presently has no notion of timeouts for things like connecting/"dialing" validators or reading from/writing to sockets.

Without timeouts, there are many cases which could potentially deadlock indefinitely inside of blocking I/O operations, and at least one user has experienced this.

Here are a few recommendations:

- `futures`/`tokio`: this is the up-and-coming ecosystem solution to this general problem, and the one I'd recommend. It's a full asynchronous event model which solves, among other things, timeouts. When `async`/`await` support ships in Rust 1.38 (scheduled to be released in August), migrating from blocking I/O should be fairly straightforward.
- `libc` crate + `poll(2)` system call: if we wanted to stick with blocking I/O, the `poll(2)` system call, as invoked through the `libc` crate, can be used to determine I/O readiness prior to performing a blocking I/O call, and also takes a timeout as an argument. This would probably be the lowest impact way to implement timeouts as it wouldn't involve switching away from blocking I/O.

I would suggest migrating to `futures-rs`/Tokio to solve this problem, and leveraging the `async`/`await` syntax when it lands later this month.

tarcieri:
If anyone wants to take a crack at switching Secret Connection to use Tokio, they have just migrated to `std::future` upstream and released an alpha that supports it:

https://tokio.rs/blog/2019-08-alphas/

Using it will require nightly (soon to be 1.38 beta) Rust

tarcieri:
Another option (announced yesterday) is `async-std`, which is a sort of drop-in async replacement for `std::io`:

- https://async.rs/blog/announcing-async-std/
- https://github.com/async-rs/async-std

It allows easily adding timeouts which leverage futures while using an API that looks nearly identical to `std::io` (with the addition of `async`/`await`):

```rust
async fn get() -> io::Result<Vec<u8>> {
    let mut stream = TcpStream::connect("example.com:80").await?;
    stream.write_all(b"GET /index.html HTTP/1.0\r\n\r\n").await?;

    let mut buf = vec![];

    io::timeout(Duration::from_secs(5), async {
        stream.read_to_end(&mut buf).await?
        Ok(buf)
    })
}
```

Liamsi:
@xla want to pair up on this soon?

tarcieri:
[...]

Issue #585, opened 2020-09-25
Types for positive i64


greg-szabo:
The problem with using int64 (i64) in protobuf:

most of the values where we use i64, can't be negative. (Vote.Height, etc) We can convert them to u64 with a check when they come in from the wire and carry on with life...
.
..until, we need to convert things back to protobuf on the wire.

Currently, we expect our internal structures to always be valid so we can use From/Into for converting domain types onto the wire. But u64->i64 can fail when the value is above the maximum of i64. The current internal type would be something like "u63" that can be directly converted to i64.

One option is exactly that: implement a "u63" type that has checks and conversion to i64 and use it internally for types that don't have their own domaintype.

Another option is to loosen our requirement for internal types to be correct, but I feel that's a very good restriction to have.

Any other ideas?

tarcieri:
> Any other ideas?

Fallible serialization

greg-szabo:
No problem :D

That's definitely always an option. I really like the idea of "our internal structures are always valid" because it expects all structures to have a `TryFrom<wiretype> for struct` coming in from the wire and a `From<struct> for wiretype` going to the wire. This is easy to achieve with most of our structs and easy to remember.
The only exception is u64 which is not "ours" so it's harder to put limits on it. It breaks the expected pattern of being able to use From/Into when encoding to the wire and it's especially stressful when it happens within a From method of another type. The other type should be infallible, but because of a fallible serialization for u64, now that other type has to be fallible too.

The example I can bring is Vote.Height. Vote currently has an infallible conversion to encode it to the wire. It's awesome that we can trust our internal structure. Because of Height being u64 (and i64 on the wire), now either I panic if I have an invalid height (argh), or I make Vote fallible too. (There goes the nice neighbourhood.)

tarcieri:
> The example I can bring is Vote.Height. Vote currently has an infallible conversion to encode it to the wire. It's awesome that we can trust our internal structure. Because of Height being u64 (and i64 on the wire), now either I panic if I have an invalid height (argh), or I make Vote fallible too. (There goes the nice neighbourhood.)

There's a `block::Height` type which could enforce some kind of maximum:

https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/src/block/height.rs

Or at least, it could've until dda7afc0aa4bdfe77429d15d670e9851c49f5c90, which made the inner `u64` public.

greg-szabo:
[...]

PR #592, opened 2020-09-30
Removed amino folder

### LINKED ISSUE #536: Consolidate, deduplicate, reorganize remaining amino_types
After all amino encoding was changed to protobuf, rename the folder name (and all outstanding comments).

This is considered the last step because reviewing folder renames is cumbersome.

greg-szabo:
Closes: #536 #585 #600

Removed all amino_types structs and implemented DomainType for the relevant Tendermint structs.

* [X] Wrote tests
* [X] Updated CHANGELOG.md


greg-szabo:
Unfortunately one of the model-based tests (single_step_tests) in light-client started acting up since a merge from master: the commit_hash is invalid. I think those commit hashes might have been created with the old tendermint source code before Thane's fixes.

CHANGELOG.md
```diff
@@ -44,8 +46,12 @@
 [#463]: https://github.com/informalsystems/tendermint-rs/issues/463
 [#504]: https://github.com/informalsystems/tendermint-rs/issues/504
 [#535]: https://github.com/informalsystems/tendermint-rs/issues/535
+[#504]: https://github.com/informalsystems/tendermint-rs/issues/504
```
greg-szabo, line 49:
```suggestion
```

CHANGELOG.md
```diff
@@ -44,8 +46,12 @@
...
 [#506]: https://github.com/informalsystems/tendermint-rs/issues/506
 [#516]: https://github.com/informalsystems/tendermint-rs/pull/516
+[#535]: https://github.com/informalsystems/tendermint-rs/issues/535
```
greg-szabo, line 52:
```suggestion
```

tendermint/Cargo.toml
```diff
@@ -57,7 +57,7 @@ tendermint-proto = { path = "../proto" }
 toml = { version = "0.5" }
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
 ripemd160 = { version = "0.9", optional = true }
-
+base64 = "0.13.0"
```
greg-szabo, line 60:
```suggestion
```

tendermint/src/amino_types/signature.rs
```diff
@@ -1,22 +0,0 @@
...
-        sign_bytes: &mut B,
-    ) -> Result<bool, EncodeError>;
-
-    /// Set the Ed25519 signature on the underlying message
-    fn set_signature(&mut self, sig: &ed25519::Signature);
-    fn validate(&self) -> Result<(), validate::Error>;
-    fn consensus_state(&self) -> Option<consensus::State>;
-    fn height(&self) -> Option<i64>;
-    fn msg_type(&self) -> Option<SignedMsgType>;
-}
```
tomtau, line 22: Will there be an equivalent of this trait? Currently, some of the types have `to_signable_bytes`, so the trait implementations can just use that.
In tmkms, it's used for over different requests: https://github.com/iqlusioninc/tmkms/blob/develop/src/session.rs#L171

Or is it expected that the equivalent of this trait would be moved to tmkms?

greg-szabo: I'll be honest, I didn't see the need for a trait here. (Your link shows that this might be incorrect, but I didn't find that specific part of the KMS code before.)
Since the trait is only used in two types: Vote and Proposal, the trait implementation is a lot of extra code for very little benefit. (Unfortunately, traits are not interfaces.)

I'm going to go through that code to see what the expectations are. One option I would like to investigate is a message type enum that includes all types of messages. In that case the code can be replaced with a `match message {}` and call the functions in there. Since we're talking about two types of structs, this is not that overbearing. (If the number of structs can explode, then maybe the trait is a good call.)

If this doesn't pan out, we can always reintroduce the trait.

greg-szabo: Ah, I think I found it: https://github.com/iqlusioninc/tmkms/blob/3e3ae75fee35d1d81258ed61c286b2efb240be40/src/session.rs#L113

You can see that when a request is identified as a Vote or a Proposal, it calls the generic `.sign()` function to get the signable bytes. But if we already identified the structs, there's no need to generalize over them any more.

Instead of
```
Request::SignProposal(req) => self.sign(req)?,
```

it could be something like:
```
Request::SignProposal(req) => SignProposalRequest::try_from(req).unwrap().to_signable_bytes(),
```

where the try_from ensures that the incoming data is valid. (and errors out if not. be sure to handle the error instead of unwrapping)

This was both the trait and the generic `sign` function is removed and we got a cleaner understanding of what's happening when calling that code.

Does this way of thinking helps? Hoesnt question. I feel it's clearer to me but I understand it's a different concept than previous iterations.

greg-szabo: Extra note: I went further into the code and I see that the `req` is an amino_type (or as we call it in the new version, a Raw type). I have a feeling that it could be upgraded into the domaintype even and all checks of incoming data moved into the `read` function. The benefits of domain types are that there is a strict input check when creating the domain type and it can happen very early in the process. By the time the `.to_signable_bytes()` function has to be called, you might have a valid type already with no need to check for invalid data.

tendermint/src/block/id.rs
```diff
@@ -34,12 +40,92 @@ pub struct Id {
...
+            parts: match value.part_set_header {
+                None => None,
+                Some(p) => Some(p.try_into()?),
+            },
+        })
+    }
+}
+
+impl From<Id> for RawCanonicalBlockId {
+    // Todo: it's possible that an empty CanonicalBlockId is encoded differently. Test this.
```
romac, line 100: Just making sure we do not forget about this TODO.

tendermint/src/serializers/custom.rs
```diff
@@ -47,7 +47,7 @@ where
                     None
                 } else {
                     Some(block::parts::Header {
-                        total: tmp_id.parts.total,
+                        total: tmp_id.parts.total as u32, // Todo: manage overflow
```
romac, line 50: Just making sure we do not forget about this TODO.

testgen/src/header.rs
```diff
@@ -107,15 +108,16 @@ impl Generator<block::Header> for Header {
         let header = block::Header {
             version: block::header::Version { block: 0, app: 0 },
             chain_id,
-            height: block::Height(self.height.unwrap_or(1)),
+            height: block::Height::try_from(self.height.unwrap_or(1)).unwrap(), /* TODO: might
+                                                                                 * overflow */
```
romac, line 112: Just making sure we do not forget about this TODO.

testgen/src/header.rs
```diff
@@ -107,15 +108,16 @@ impl Generator<block::Header> for Header {
...
             time,
             last_block_id: None,
             last_commit_hash: None,
             data_hash: None,
             validators_hash: valset.hash(),
             next_validators_hash: next_valset.hash(),
             consensus_hash: valset.hash(), // TODO: currently not clear how to produce a valid hash
```
romac, line 119:
```suggestion
            consensus_hash: valset.hash(), // TODO: currently not clear how to produce a valid hash
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
     /// Signature invalid
     #[error("bad signature")]
     SignatureInvalid,
+
+    /// invalid Vote type
+    #[error("invalid Type")]
+    InvalidMessageType,
```
romac, line 46: Can we somehow unify the comment, message and variant name?

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    /// Negative block height
+    #[error("negative height")]
+    NegativeHeight,
+
+    /// Negative voting round
+    #[error("negative round")]
+    NegativeRound,
+
+    /// Negative POL round
+    #[error("negative pol round")]
```
romac, line 57:
```suggestion
    #[error("negative POL round")]
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    NegativePOLRound,
+
+    /// Negative validator index in vote
+    #[error("negative ValidatorIndex")]
```
romac, line 61:
```suggestion
    #[error("negative validator index")]
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    NegativeValidatorIndex,
+
+    /// Wrong hash size in part_set_header
+    #[error("Wrong hash: expected Hash size to be 32 bytes")]
```
romac, line 65:
```suggestion
    #[error("invalid hash: expected hash size to be 32 bytes")]
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    InvalidHashSize,
+
+    /// No timestamp in vote
+    #[error("no timestamp")]
+    NoTimestamp,
+
+    /// Invalid account ID length
+    #[error("invalid account ID length")]
+    InvalidAccountIDLength,
```
romac, line 74:
```suggestion
    InvalidAccountIdLength,
```

Convention in Rust is to use strict CamelCase (I know, it looks weird…)

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+
+    /// Invalid signature ID length
+    #[error("invalid signature ID length")]
+    InvalidSignatureIDLength,
```
romac, line 78:
```suggestion
    InvalidSignatureIdLength,
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+
+    /// Overflow during conversion
+    #[error("Integer overflow")]
```
romac, line 81:
```suggestion
    #[error("integer overflow")]
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    IntegerOverflow,
+
+    /// No Vote found during conversion
+    #[error("No vote found")]
```
romac, line 85:
```suggestion
    #[error("no vote found")]
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    NoVoteFound,
+
+    /// No Proposal found during conversion
+    #[error("No proposal found")]
```
romac, line 89:
```suggestion
    #[error("no proposal found")]
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    NoProposalFound,
+
+    /// Invalid AppHash length found during conversion
+    #[error("Invalid AppHash Length")]
```
romac, line 93:
```suggestion
    #[error("lnvalid app hash length")]
```

tendermint/src/proposal.rs
```diff
@@ -0,0 +1,267 @@
...
+            n => Some(Round::try_from(n)?),
+        };
+        Ok(Proposal {
+            msg_type: value.r#type.try_into()?,
+            height: value.height.try_into()?,
+            round: value.round.try_into()?,
+            pol_round,
+            block_id: match value.block_id {
+                None => None,
+                Some(raw_block_id) => Some(BlockId::try_from(raw_block_id).unwrap()),
```
romac, line 61: Can we document why the unwrap here is safe?

tendermint/src/proposal.rs
```diff
@@ -0,0 +1,267 @@
...
+            },
+            timestamp: match value.timestamp {
+                None => None,
+                Some(t) => Some(t.try_into()?),
+            },
```
romac, line 66: This can probably be rewritten as follows (though that might obscure the meaning a bit for people not familiar with [`Option::transpose`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.transpose) so feel free to disregard):

```suggestion
            timestamp: value.timestamp.map(|t| t.try_into()).transpose()?,
```

tendermint/src/error.rs
```diff
@@ -40,6 +40,58 @@ pub enum Kind {
...
+    #[error("negative height")]
+    NegativeHeight,
+
+    /// Negative voting round
+    #[error("negative round")]
+    NegativeRound,
+
+    /// Negative POL round
+    #[error("negative pol round")]
+    NegativePOLRound,
```
romac, line 58:
```suggestion
    NegativePolRound,
```

Convention in Rust is to use strict CamelCase (I know, it looks weird…)

tendermint/src/proposal.rs
```diff
@@ -0,0 +1,267 @@
...
+        let mut got = vec![];
+
+        let request = SignProposalRequest {
+            proposal,
+            chain_id: ChainId::from_str("test_chain_id").unwrap(),
+        };
+
+        let _have = request.to_signable_bytes(&mut got);
+
+        // the following vector is generated via:
```
romac, line 171: Such a great idea to include the code which was used to generate those vectors :)

tendermint/src/hash.rs
```diff
@@ -25,6 +27,28 @@ pub enum Hash {
...
+            return vec![];
+        }
+        match value {
+            Hash::Sha256(s) => s.to_vec(),
+        }
+    }
+}
+
 impl Hash {
     #[allow(clippy::new_ret_no_self)]
```
romac, line 53: I know this is not strictly part of this PR, but perhaps we could tackle this while we're at it and rename `Hash::new` to `Hash::from_bytes` like Clippy suggests and remove this attribute?

tendermint/src/proposal/sign_proposal.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl From<SignedProposalResponse> for RawSignedProposalResponse {
+    fn from(value: SignedProposalResponse) -> Self {
+        RawSignedProposalResponse {
+            proposal: match value.proposal {
+                None => None,
+                Some(proposal) => Some(proposal.into()),
+            },
```
romac, line 92: This can likely be simplified to:

```suggestion
            proposal: value.proposal.map(Into::into),
```

or perhaps like this if rustc fails to infer the appropriate `Into` instance:

```suggestion
            proposal: value.proposal.map(TargetType::from),
```

(after substituting the actual target type for `TargetType`)

tendermint/src/public_key.rs
```diff
@@ -39,6 +50,44 @@ pub enum PublicKey {
...
+    type Error = Error;
+
+    // This does not check if the underlying pub_key_ed25519 has the right size.
+    // The caller needs to make sure that this is actually the case.
+    fn try_from(value: RawPubKeyResponse) -> Result<Self, Self::Error> {
+        match &value
+            .pub_key
+            .ok_or_else(|| format_err!(error::Kind::InvalidKey, "empty pubkey"))?
+            .sum
+            .ok_or_else(|| format_err!(error::Kind::InvalidKey, "empty sum"))?
```
romac, line 65: Can we introduce a let binding here to help with readability a little bit?

tendermint/src/vote.rs
```diff
@@ -28,26 +42,69 @@ pub struct Vote {
...
+            height: value.height.try_into()?,
+            round: value.round.try_into()?,
+            block_id: match value.block_id {
+                None => None,
+                Some(b) => Some(b.try_into()?),
+            },
+            timestamp: match value.timestamp {
+                None => None,
+                Some(t) => Some(t.try_into()?),
+            },
```
romac, line 85: As above, these two expressions could be simplified using [`Option::transpose`](https://doc.rust-lang.org/beta/std/option/enum.Option.html#method.transpose).

romac (commented):
Wow, what an undertaking! Thank you so much for putting in all this work.

Just went through the Rust code, everything looks good to me, only left some minor comments.

Let's wait until someone can review the changes in the TLA models before approving.

tendermint/src/amino_types/signature.rs
tomtau: I'll try to look more into tmkms, but yes, I think using the domain types may make the code simpler there

tarcieri: I'll just say I'm not a fan of this trait either

greg-szabo:
Thanks Romain for the great feedback! Your observations are appreciated, I'll fix the code tomorrow. Most of the suboptimal code is because of late-night coding. Some of the suboptimal code is copy-pasted from the earlier amino_types folder, we can and should fix those too.

I was looking for that transpose function, thanks for pointing it out.

rpc/src/lib.rs
```diff
@@ -10,8 +10,13 @@
 //!
 //! | Feature | Description |
 //! | ------- | ----------- |
-//! | `http-client` | Provides [`HttpClient`], which is a basic RPC client that interacts with remote Tendermint nodes via **JSON-RPC over HTTP**. This client does not provide [`Event`] subscription functionality. See the [Tendermint RPC] for more details. |
-//! | `websocket-client` | Provides [`WebSocketClient`], which currently only provides [`Event`] subscription functionality over a WebSocket connection. See the [`/subscribe` endpoint] in the Tendermint RPC for more details. This client does not yet provide access to the RPC methods provided by the [`Client`] trait (this is planned for a future release). |
+//! | `http-client` | Provides [`HttpClient`], which is a basic RPC client that interacts with
```
thanethomson, line 13: Does this still render correctly when wrapped? I don't think it should be wrapped.

thanethomson (dismissed):

tendermint/src/block/id.rs
greg-szabo: I just looked into it and rearranged the BlockID to reflect better on the Go implementation and the differences between the protobuf message and the Go expectations. You will see it in the next commit, with comments.

tendermint/src/serializers/custom.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/proposal.rs
greg-szabo: It's not, good catch! Fixed in the next commit.

tendermint/src/error.rs
greg-szabo: Covered in the next commit.

tendermint/src/proposal.rs
greg-szabo: Thanks! It's not my idea, but I extended on it. I'll try to keep up the habit for easier upgrades in the future.

tendermint/src/hash.rs
greg-szabo: Absolutely! Covered in the next commit.

tendermint/src/proposal/sign_proposal.rs
greg-szabo: Yup, another example of late-night coding.
I exchanged it to the currently used `.map(|p| p.into())` for the next commit. I'll do another sweep where I want to change all of these into `.map(into::Into)` which is what I wanted to do in the first place, I just didn't know I have to explicitly call the trait to be able to do it...

tendermint/src/public_key.rs
greg-szabo: I tried, check in the next commit

rpc/src/lib.rs
greg-szabo: must have been fmt or me adding a typo. Fixed in the next commit.

testgen/src/header.rs
greg-szabo: Covered in commit fixes commit #2.

testgen/src/header.rs
greg-szabo: Covered in commit "Fixes #2".

tendermint/src/proposal/sign_proposal.rs
greg-szabo: Covered in the third installment of fixes.

tendermint/src/proposal.rs
greg-szabo: Covered in fixes #3 commit.

tendermint/src/vote.rs
greg-szabo: Covered in fixes #3 commit.

romac:
Great stuff, @greg-szabo!

Now it looks like the model-based tests are failing because `tendermint-testgen not found`. /cc @andrey-kuprianov

tendermint/src/public_key.rs
```diff
@@ -1,6 +1,7 @@
 //! Public keys used in Tendermint networks
 
 pub use ed25519_dalek::PublicKey as Ed25519;
+mod pub_key_request;
```
tomtau, line 5: I guess it'd be good to expose `PubKeyRequest`?
```suggestion
pub use pub_key_request::PubKeyRequest;
```

greg-szabo: Nice catch! Late night coding strikes again.

tendermint/src/vote/sign_vote.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    /// Create signable vector from Vote.
+    pub fn to_signable_vec(&self) -> Result<Vec<u8>, DomainTypeError> {
+        self.vote.to_signable_vec(self.chain_id.clone())
+    }
+}
+
+/// SignedVoteResponse is a response containing a signed vote or an error
+#[derive(Clone, PartialEq)]
```
tomtau, line 62:
```suggestion
#[derive(Clone, PartialEq, Debug)]
```

tendermint/src/proposal/sign_proposal.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// Create signable vector from Proposal.
+    pub fn to_signable_vec(&self) -> Result<Vec<u8>, DomainTypeError> {
+        self.proposal.to_signable_vec(self.chain_id.clone())
+    }
+}
+
+/// SignedProposalResponse is response containing a signed proposal or an error
+#[derive(Clone, PartialEq)]
```
tomtau, line 64:
```suggestion
#[derive(Clone, PartialEq, Debug)]
```

tendermint/src/proposal.rs
```diff
@@ -0,0 +1,261 @@
...
+        CanonicalProposal::new(self.clone(), chain_id).encode_length_delimited(sign_bytes)?;
+        Ok(true)
+    }
+
+    /// Create signable vector from Proposal.
+    pub fn to_signable_vec(&self, chain_id: ChainId) -> Result<Vec<u8>, DomainTypeError> {
+        CanonicalProposal::new(self.clone(), chain_id).encode_length_delimited_vec()
+    }
+
+    /// Consensus state from this proposal - This doesn't seem to be used anywhere.
```
tomtau, line 99: this is used in tmkms via that trait: https://github.com/iqlusioninc/tmkms/blob/develop/src/session.rs#L315

but I guess it could be simpler / clearer without it

tendermint/src/chain/id.rs
```diff
@@ -2,89 +2,100 @@
...
 
 /// Maximum length of a `chain::Id` name. Matches `MaxChainIDLen` from:
 /// <https://github.com/tendermint/tendermint/blob/develop/types/genesis.go>
 // TODO: update this when `chain::Id` is derived from a digest output
 pub const MAX_LENGTH: usize = 50;
 
 /// Chain identifier (e.g. 'gaia-9000')
-#[derive(Copy, Clone)]
-pub struct Id([u8; MAX_LENGTH]);
+#[derive(Clone)]
```
tomtau, line 20:
```suggestion
#[derive(Copy, Clone)]
```

tendermint/src/proposal.rs
greg-szabo: Yes, I think so too. It's a similar case to the previous. It seems it's generating a `State{}` struct and matching whatever the incoming request was. We can easily do that with a 
`consensus_state = State::from(vote);` or `consensus_state = State::from(proposal);` call. Which tells me that implementing this From trait would be a good idea. I'll check the KMS code a bit further, I'm interested how this state is used.

tendermint/src/chain/id.rs
greg-szabo: This, unfortunately won't work, because the implemented `String` does not have the `Copy` trait. If you need bytes, you can use `as_bytes()` or `as_str()`. (You can read about the issue with the Copy trait and String in the Rust book. It's a size-in-memory thing.)

greg-szabo: Alternatively, we can revert back to using [u8] internally but it seemed to me that `String` is a better representation with the added functions. Up to the team to decide.

tendermint/src/proposal/sign_proposal.rs
greg-szabo: Thanks, added to the next commit

tendermint/src/vote/sign_vote.rs
greg-szabo: Thanks, exactly the type of feedback I need to make the lib more usable. Added to the next commit.

ebuchman (approved):
Awesome, thanks Greg! This is just an initial review, haven't gone through it all, but let's merge and continue to address anything else in a follow up

rpc/src/lib.rs
ebuchman: Did this get fixed? Looks like it's still wrapping?

rpc/tests/integration.rs
```diff
@@ -208,7 +208,7 @@ fn broadcast_tx_commit_null_data() {
 fn commit() {
     let response = endpoint::commit::Response::from_string(&read_json_fixture("commit")).unwrap();
     let header = response.signed_header.header;
-    assert_eq!(header.chain_id.as_ref(), EXAMPLE_CHAIN);
+    assert_eq!(header.chain_id.as_ref(), "dockerchain");
```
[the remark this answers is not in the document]
greg-szabo: Uh, I'm not sure, it was completely fine as it was. I don't think it was changed by me, it seems it was changed when the commit.json was regenerated. (The previous one had `cosmoshub-2` as the chain ID and the current one is `dockerchain`.) I don't mind it this way because most other assertions use constants to check validity. But we can revert to a const too.

tendermint/src/block/height.rs
```diff
@@ -1,27 +1,62 @@
...
-pub struct Height(pub u64);
+pub struct Height(u64);
+
+impl DomainType<i64> for Height {}
+
+impl TryFrom<i64> for Height {
+    type Error = Error;
+
+    fn try_from(value: i64) -> Result<Self, Self::Error> {
+        Ok(Height(value.try_into().map_err(|_| Kind::NegativeHeight)?))
```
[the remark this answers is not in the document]
greg-szabo: Yes, it's a kind of Rust inherent magic: Height is defined as `Height(u64)`. Hence the value's `try_from()` method is going to try to convert the number to u64 from the original i64. This fails if the value is negative (if it's positive it fits without overflow) so the resultant error is because the height was negative. The TryInto trait's error is not compatible with the defined Self::Error, so it has to be converted.

This can fail miserably if/when we redefine the Height newtype. We can change it to `u64::try_from(value)` which is more explicit and resolves into the same code.

tendermint/src/block/id.rs
```diff
@@ -34,12 +40,92 @@ pub struct Id {
...
+        } else {
+            RawBlockId {
+                hash: value.hash.into(),
+                part_set_header: value.parts.map(|p| p.into()),
+            }
+        }
+    }
+}
+
+impl TryFrom<RawCanonicalBlockId> for Id {
```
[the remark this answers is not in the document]
greg-szabo: We don't use it. There are multiple weak reasons why it's there:
1. Symmetric implementation of all conversions: `Domain::TryFrom(Raw)` and `Raw::From(Domain)`. This sets a good precedent for the developer and standard expectations for all types and conversions. It's a weak reason because if we know that we don't use one conversion, we could simply omit it from the code. I kept it because this is a library and I wasn't sure who uses it for what.
2. The above symmetry is a prerequisite for the domaintype trait. It's a weak argument, because the trait is only available with exactly one raw type, and in this case, it's implemented with the raw BlockId and not the CanonicalBlockId. So, there's no real need for the same symmetric implementation because the DomainType trait doesn't really depend on it.

So it's there for good measure. We can remove it, if unnecessary.

tendermint/src/chain/id.rs
```diff
@@ -2,89 +2,100 @@
...
+use tendermint_proto::DomainType;
 
 /// Maximum length of a `chain::Id` name. Matches `MaxChainIDLen` from:
 /// <https://github.com/tendermint/tendermint/blob/develop/types/genesis.go>
 // TODO: update this when `chain::Id` is derived from a digest output
 pub const MAX_LENGTH: usize = 50;
 
 /// Chain identifier (e.g. 'gaia-9000')
-#[derive(Copy, Clone)]
-pub struct Id([u8; MAX_LENGTH]);
```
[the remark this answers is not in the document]
greg-szabo: Recording it for any outsiders:
We discussed this on Slack and the reason might be because of the `Copy` trait (which was removed when this became a String). The `Copy` trait is not available for Strings. But Strings are more convenient to use in the code and the Clone trait is good enough for copying the structure where necessary.

tendermint/src/consensus/state.rs
```diff
@@ -68,33 +64,34 @@ impl PartialOrd for State {
...
+            height: block::Height::try_from(1001u64).unwrap(),
+            round: block::Round::try_from(0).unwrap(),
             step: 0,
             block_id: None,
         };
 
         let oldest = State {
             height: block::Height::default(),
-            round: 0,
+            round: block::Round::try_from(0).unwrap(),
```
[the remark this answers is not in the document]
greg-szabo: LOL, we can change that to `Round::default()` and get on with life. (Maybe in the next PR.)

The longer answer for the try-unwrap dance for constants is this:
We made it a strict requirement that the domain structs we use are safe or we invite abuse of the structs and the possibility of bad code. If constants can add any value without consequences, invalid structs can be created. A Round is a positive value between 0 and i32::MAX. There's no primitive value that can store this in Rust, hence the safe struct that only accepts and emits values between these two. If There's a `Round(pub i32)` definition, then I can create rounds with negative numbers. If there's a `Round(pub u32)` definition, I can create rounds with bigger numbers than i32::MAX. Both create problems when we want to encode to the wire and the current expectation is that encoding to the wire is infallible.

Also, we're only going to deal with constants in our tests. In "real code", we get other variables coming in from the wire. The try-unwrap dance is warranted there. (And it's possibly a `try-?` dance.)

tendermint/src/error.rs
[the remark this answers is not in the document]
greg-szabo: This was copied from the amino_types/error.rs . Yeah, eventually this should be more general. Right now all our hashes that can have invalid length, are supposed to be 32 bytes.

(I might create issues tomorrow from these notes.)

light-client/src/components/scheduler.rs
```diff
@@ -122,5 +123,7 @@ pub fn valid_schedule(
 #[pre(low <= high)]
 #[post(low <= ret && ret <= high)]
 fn midpoint(low: Height, high: Height) -> Height {
-    (low.value() + (high.value() + 1 - low.value()) / 2).into()
+    (low.value() + (high.value() + 1 - low.value()) / 2)
+        .try_into()
+        .unwrap() // Will panic if midpoint is higher than i64::MAX
```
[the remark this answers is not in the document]
greg-szabo: Good idea, although in this case it's more of a theoretical possibility than a real one.
You can get the biggest number with `high.value()==i64::MAX` and midpoint is going to be `i64::MAX/2` in that case. So this is an unwrap case that can never fail, unless the developer changes the types without concern to the note.

Maybe it's better to remove it, since it's theoretical only.

tendermint/src/consensus/state.rs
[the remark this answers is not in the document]
greg-szabo: Obviously, after my long-winded explanation last night, Romain woke up this morning and asked me something that made me realize how overcomplicated my explanation is. (I always learn something from Romain.)

So, we implemented the From trait for Height in #628 and it will come for the Round struct too with types that will definitely fit into the struct. In the case of Round,  you can do for example: `Round::from(0_u16)`.

melekes (commented):
🚀

PR #593, opened 2020-09-30
Fix for running MBT on Macs

### LINKED ISSUE #587: Make model-based tests running on Macs
Currently, the model-based test driver fails to run on Mac's because of this small issue: informalsystems/jsonatr#10, namely that the `date` command works differently on Linux and Mac. As a result, the call `utc_timestamp` command of `jsonatr-lib` [here](https://github.com/informalsystems/tendermint-rs/blob/b5797e30b521ec764b5178c8b9e9eabd8700b7ac/testgen/jsonatr-lib/apalache_to_lite_test.json#L16) fails.

We need to fix that before the 0.17 release. Possible fixes could be:
 - modify the command arguments such that it works uniformly on Linux and Mac
 - use a branching command `ifelse`, and first detect the architecture, then run different commands based on that
 - sidestep the issue, and use `tendermint-testgen time` command together with some abstract time to get the desired result.

The generated timestamp is not used anywhere actually, so the value can be arbitrary. It only needs to be formatted properly to pass deserialization.

Shivani912:
closes #587 

As discussed with @andrey-kuprianov, this command isn't really used anywhere except just to get it parsed into the existing `TestCase struct` from Conformance Tests. This is because MBT currently uses it to run tests with the existing test driver. We should eventually clean that up but this is a quick fix for now.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/testgen/jsonatr-lib/apalache_to_lite_test.json
+++ b/testgen/jsonatr-lib/apalache_to_lite_test.json
@@ -13,7 +13,7 @@
         "signed_header": "$ | block_to_signed_header",
         "next_validator_set": "$ | block_next_validators | ids_to_validator_set",
         "trusting_period": "1400000000000",
-        "now": "$utc_timestamp"
+        "now": "2020-09-25T08:18:04.996061145Z"
       }
     },
     {
--- a/testgen/jsonatr-lib/unix.json
+++ b/testgen/jsonatr-lib/unix.json
@@ -9,8 +9,8 @@
     },
     {
       "name": "utc_timestamp",
-      "kind": "COMMAND",
-      "source": "date --utc +%FT%H:%M:%S.%NZ",
+      "kind": "INLINE",
+      "source": "2020-09-25T08:18:04.996061145Z",
       "stdin": false
     },
     {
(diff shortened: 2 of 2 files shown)


andrey-kuprianov (changes requested):
Actually, I think here it's better to put this constant here:

https://github.com/informalsystems/tendermint-rs/blob/941365899254c16a1db593c536b335cee6d405c0/testgen/jsonatr-lib/apalache_to_lite_test.json#L16

instead of changing the `unix.json`

Shivani912:
> Actually, I think here it's better to put this constant here:
> 
> https://github.com/informalsystems/tendermint-rs/blob/941365899254c16a1db593c536b335cee6d405c0/testgen/jsonatr-lib/apalache_to_lite_test.json#L16
> 
> instead of changing the `unix.json`

hmmm makes sense! 👍

andrey-kuprianov (approved):
Looks fine!

We should fix informalsystems/jsonatr#10 properly, so that it works correctly both on Linux and Mac, and then apply those changes here; but that's after the 0.17 release.

Shivani912:
> Looks fine!
> 
> We should fix [informalsystems/jsonatr#10](https://github.com/informalsystems/jsonatr/issues/10) properly, so that it works correctly both on Linux and Mac, and then apply those changes here; but that's after the 0.17 release.

Sure! Working on that 👍 Thanks! :)

Issue #594, opened 2020-09-30
spec: put links into old files


josef-widder:
To finish moving the specifications #561 we have to put links into the old files that point to the new location in the tendermint/spec repository.

josef-widder:
#599 resolved it

PR #599, opened 2020-09-30
links to tendermint/spec

josef-widder:
#594: I added a note to the files that they are not maintained and that the newer versions are in tendermint/spec.

@konnov : I did this for the main TLA+ files but not for each `MCx_y_bla.tla`. Is it OK?

### DIFF
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -1,11 +1,9 @@
 # Specification
 
-The primary English specifications for Tendermint data structures can be found in the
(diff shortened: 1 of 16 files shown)


konnov:
I would not introduce individual versions for the MC-files, as they are plenty. I would just keep the versions for the latest version that has been model checked.

konnov:
Right. We don't need comments like that in the MC files.

konnov (approved):
The disclaimers look good!

PR #602, opened 2020-10-01
MBT timestamp fix 

Shivani912:
Reverting the previous quick fix for #587 (PR #593) and propagating the solution here from jsonatr (see [here](https://github.com/informalsystems/jsonatr/pull/11))

**What was the fix?** - modified the command argument so that it works uniformly on Linux and Mac

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


andrey-kuprianov (approved):
Looks good, thanks for fixing this:)

PR #604, opened 2020-10-01
rpc: Add testing-related documentation overview

### LINKED ISSUE #597: docs: Document testing of rpc crate
We need comprehensive documentation for the ways in which we're testing the `rpc` crate in its [README](https://github.com/informalsystems/tendermint-rs/blob/master/rpc/README.md).

thanethomson:
Closes #597 

[Rendered README](https://github.com/informalsystems/tendermint-rs/blob/docs/rpc-testing/rpc/README.md)

* [ ] ~Wrote tests~
* [ ] ~Updated CHANGELOG.md~


melekes (dismissed):

rpc/README.md
```diff
@@ -43,6 +43,39 @@ subscriptions to events being generated by a Tendermint node.
 - RPC endpoints REST interface documentation:
 https://docs.tendermint.com/master/rpc/ 
 
+## Testing
+
+The RPC types are directly tested through the [integration
+tests](./tests/integration.rs). These tests use fixtures taken from running
```
ebuchman, line 49: Should add a bit more detail re:

- What should be tested here? Eg. we should have at least one fixture for each tendermint rpc endpoint, probably a couple to handle different cases (eg. ressponses with/without certain fields, which have tripped us up before)
- Link to the list of fixtures
- How do we regenerate fixtures?

ebuchman (commented):
Super helpful, especially for surfacing more questions, thanks!

rpc/README.md
thanethomson: So my understanding of what we want to achieve with #597 is just to document what's currently there. If we want to expand the scope of what's being tested we should do that in a separate issue (I've added #611 now).

As for the list of fixtures, that's accessible from the `integration.rs` file (linked in the `README.md`).

I can add a note on how we manually generated the fixtures, but if we want to automate that, again, for that we should create a separate issue (I've added #612 for this).

rpc/README.md
```diff
@@ -43,6 +43,39 @@ subscriptions to events being generated by a Tendermint node.
...
+Tendermint nodes to ensure compatibility without needing access to a running
+node during testing. To run these tests locally:
+
+```bash
+# From within the rpc crate
+cargo test --all-features
+```
+
+The RPC client is also indirectly tested through the [Tendermint integration
+tests](../tendermint/tests/integration.rs), which happens during
```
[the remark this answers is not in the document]
thanethomson: Same as the comment above

ebuchman (approved):

Issue #605, opened 2020-10-02
Cyclic dependency between light-client and testgen crates


adizere:
As part of testing the IBC relayer in `ibc-rs`, we would like to have the capability of generating light blocks. See, for instance, this issue https://github.com/informalsystems/ibc-rs/issues/158.

After discussing briefly with @andrey-kuprianov and @Shivani912, I understand that this is the current state of the light block generation testgen feature:

- light block generation feature was added (https://github.com/informalsystems/tendermint-rs/pull/534) but unfortunately it also introduced a circular dependency between crates `testgen` and `light-client`;
- the feature was subsequently reverted (https://github.com/informalsystems/tendermint-rs/pull/545), fixing the dependency loop;
- the cause for the dependency loop seems to be in the `LightBlock` struct, which `testgen` needs but it's defined in `light-client` (so the former crate would need to depend on the latter)

Possible solutions to avoiding the dependency loop:
- Option 1. Shivani has a branch https://github.com/informalsystems/tendermint-rs/pull/606 with the feature, and the temporary fix is to copy the `LightBlock` definition into `testgen`
- Option 2. Move the `LightBlock` outside of `light-client` crate and into some (new?) "common types" crate, so that both `light-client` and `testgen` would depend on this new crate.

Other solutions or thoughts?

PR #606, opened 2020-10-02
Testgen: LightBlock generation without cyclic dependency

adizere:
Discussion ground for reviewing the fix for closing #605 

EDIT: Also contains a minimal implementation of LightChain (which is eventually expected to support ibc-rs testing). 

* [x] Wrote tests
* [ ] Updated CHANGELOG.md


Shivani912:
Converted to draft PR because it needs some improvements in the way it is coded like `LightBlock` and `ValidatorSet` needs to implement the `Generator` trait and I think it's also good to add the chain generation code here.

testgen/src/validator.rs
```diff
@@ -27,6 +27,7 @@ impl Validator {
             proposer_priority: None,
         }
     }
+    // Question: Why do we need this option since we're already initializing id with fn new()??
```
Shivani912, line 30: @andrey-kuprianov

Shivani912:
Test / nightly-coverage (pull_request) fails on master as well and the error seems to be coming from outside the scope of our code

romac:
@Shivani912 Is this ready for review?

adizere:
> I need to bring some input to help progress on this, namely, to answer the question: which components of a light block are necessary (i.e., read and verified) as part of chain mocking to test the relayer. I'm still chewing on this question.

Coming back to this. Initially, the properties/interface we require from a light chain are:

1. Heights in blocks should be monotonically increasing when we call `advance_chain`. So each light block that is subsequently created upon calling this function, should have the field `.signed_header.header.height` larger by 1 compared to the header of the previous block.

2. We need the ability to create the first block (at Height 1). The function `default_with_length()` in this PR looks good for this purpose.

3. We need to fetch a block at a certain height, e.g., a function `LightChain::block(&self, target_height: u64) -> Option<LightBlock>` that may return none if the target height does not exist.

4. We need a getter for the latest light block, e.g., a function called `latest_block(&self) -> LightBlock`.

5. We need to be able to convert between a `LightBlock` and a `ConsensusState` type. This [seems to be possible](https://github.com/informalsystems/ibc-rs/blob/master/modules/src/ics07_tendermint/consensus_state.rs#L83) given the `signed_header` field inside the block.

Shivani912:
Thanks for a detailed description of requirements @adizere! I'm on it :))

adizere:
Scratch [my prior comment](https://github.com/informalsystems/tendermint-rs/pull/606#issuecomment-714367883). We can simplify if we forget about `LightChain` and instead stick to the lower-level `LighBlock` interface. What would be great is the following interface:

```rust
impl LightBlock {
    /// Constructs a default light block for a certain height: field `.signed_header.header.height`
    /// will correspond to input `height`.
    pub fn default_with_height(height: u64) -> Self {}

    /// Produces a light block "chained" to the input block (i.e., having height larger by 1). All
    /// other details of the block are currently ignored.
    pub fn next(&self) -> Self {}
}
```

testgen/src/light_block.rs
```diff
@@ -0,0 +1,252 @@
...
+        Self {
+            header: Some(header),
+            commit: Some(commit),
+            validators: Some(validators),
+            next_validators: Some(next_validators),
+            provider: Some(provider.to_string()),
+        }
+    }
+
+    pub fn new_default(validators: &[Validator], height: u64) -> Self {
```
adizere, line 88: Is it possible to have a function with the same signature as this one, except it's missing the input parameter `validators: &[Validator]`? I mean something of this type:

```rust
    pub fn new_default(height: u64) -> Self
```

testgen/src/validator.rs
Shivani912: Okay, I figured 👍

testgen/src/light_block.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub validators: ValidatorSet,
+    /// Validator set at the next block height
+    pub next_validators: ValidatorSet,
+    /// The peer ID of the node that provided this block
+    pub provider: PeerId,
+}
+
+impl TMLightBlock {
+    /// Constructs a new light block
+    pub fn new(
```
adizere, line 34: This function is not used anywhere. Should we keep it?

testgen/src/light_block.rs
```diff
@@ -0,0 +1,252 @@
...
+use tendermint::validator::Set as ValidatorSet;
+
+/// A light block is the core data structure used by the light client.
+/// It records everything the light client needs to know about a block.
+/// NOTE: This struct & associated `impl` below are a copy of light-client's `LightBlock`.
+/// The copy is necessary here to avoid a circular dependency.
+/// Cf. https://github.com/informalsystems/tendermint-rs/issues/605
+/// TODO: fix redundant code without introducing cyclic dependency.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct TMLightBlock {
```
adizere, line 21: It's not clear to me why do we have two types:

- `TMLightBlock` which is a copy of the same type from light-client crate
- and `LightBlock`.

The one which is necessary for testing/mocking is the former. Not sure what we use the latter for.

adizere (commented):
Some parts are not clear to me, maybe we can simplify?

testgen/src/light_block.rs
```diff
@@ -0,0 +1,249 @@
...
+        Err(e) => bail!("Failed to generate commit with error: {}", e),
+        Ok(c) => c,
+    };
+
+    Ok(SignedHeader { header, commit })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
adizere, line 235: Could we also have a test that describes the conversion from a `LightBlock` into a `TMLightBlock`? It can also include some calls to `next` and `new_default` like in the test below.

testgen/src/light_block.rs
Shivani912: done ✅

testgen/src/light_block.rs
Shivani912: I removed it for now :)

testgen/src/light_block.rs
Shivani912: done ✅

testgen/src/light_block.rs
```diff
@@ -0,0 +1,273 @@
...
+use tendermint::validator::Set as ValidatorSet;
+
+/// A light block is the core data structure used by the light client.
+/// It records everything the light client needs to know about a block.
+/// NOTE: This struct & associated `impl` below are a copy of light-client's `LightBlock`.
+/// The copy is necessary here to avoid a circular dependency.
+/// Cf. https://github.com/informalsystems/tendermint-rs/issues/605
+/// TODO: fix redundant code without introducing cyclic dependency.
+///
+/// To convert `TMLightBlock` to the Domain type `LightBlock` used in light-client crate
```
Shivani912, line 20: @adizere This might be helpful for you :))

testgen/src/light_block.rs
```diff
@@ -0,0 +1,273 @@
...
+            header: Some(header),
+            commit: Some(commit),
+            validators: Some(validators.to_vec()),
+            next_validators: Some(validators.to_vec()),
+            provider: Some(default_peer_id()),
+        }
+    }
+
+    /// Produces a subsequent, i.e. at (height+1), light block to the supplied one
+    // TODO: figure how to represent the currently ignored details in header
```
Shivani912, line 104: Left a few TODOs here but with the current state you should be able to use it for testing so long as you don't care about the ignored details.

testgen/src/light_block.rs
```diff
@@ -0,0 +1,273 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_light_block() {
+        let light_block_1 = LightBlock::new_default(1);
+        let light_block_2 = LightBlock::new_default(1);
+
+        assert_eq!(light_block_1.generate(), light_block_2.generate());
```
Shivani912, line 234: `.generate()` produces the `TMLightBlock` data structure from `LightBlock`

Shivani912 (commented):
@adizere I've addressed your comments above. I won't be able to add you as a reviewer here, so please review this PR at your convenience and let me know if this is okay!

testgen/src/light_block.rs
adizere: Indeed, this should not pose a problem at the moment.

adizere:
From the IBC-side this looks good. Thanks @Shivani912, much appreciated!

@andrey-kuprianov can you do a last pass to review this PR towards getting it merged?

andrey-kuprianov (approved):
Great work, @Shivani912! Glad to see the Testgen's capabilities growing:)

PR #607, opened 2020-10-02
Enable RPC integration tests in a new job on CI

greg-szabo:
I just realized that tendermint-rpc integration tests are not running in CI. This enables them.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


greg-szabo:
Setting this to draft until the aminoff changes make it into master.

greg-szabo:
An interesting question: one of the tests (block) fail because it receives an empty vector for merkle hash. Our implementation has the option to follow the RFC or use [0u8;32] as the interpretation of an empty vector. Or alternatively we can introduce an `Option<>` so `None` means the incoming hash string is empty.

What would be the correct approach?

romac:
It seems to me that these tests are already executed in the `tendermint-rpc` job since they are not ignored.

romac:
The tests in `tests/integration.rs` are more like conformance tests than integration tests since they do not run against a Tendermint node.

On the other hand, we have two tests which run against a node in:

```
test src/client/transport/http.rs - client::transport::http::HttpClient (line 17) ... ignored
test src/client/transport/websocket.rs - client::transport::websocket::WebSocketClient (line 72) ... ignored
```

romac:
I pushed a commit that runs all ignored tests as part of the `tendermint-rpc-integration` job rather than only those in the `integration` test (of which they are none).

Perhaps we should rename that file/module/test to `conformance`?

romac:
Just renamed it to `parse_response` which imho describes the tests better than `integration` or `conformance`.

romac (dismissed):

romac (approved):

PR #610, opened 2020-10-05
tendermint: impl fmt::UpperHex for abci::Transaction

tony-iqlusion:
Found myself wanting this today.

I imagine you could use similar impls throughout this crate, which allow you to do things like:

```rust
println!("tx: {:X}", tx);
```


greg-szabo (approved):
I love the idea and I would like to see this for other hash-based fields too.

PR #613, opened 2020-10-06
tendermint: improve fmt::UpperHex impl on Transaction

tony-iqlusion:
The previous implementation would print hex strings like:

```
[FF, 1, FE, 2]
```

This changes the implementation to print the following instead (with test):

```
FF01FE02
```


thanethomson (approved):
LGTM 👍

PR #617, opened 2020-10-07
rpc: Refactor WebSocketClient interface to expose driver

### LINKED ISSUE #616: rpc: WebSocketClient driver ownership
Right now the WebSocketClient encapsulates and manages the JoinHandle for the driver, which means the owner of the WebSocketClient doesn't need to worry about this (see [here](https://github.com/informalsystems/tendermint-rs/blob/dd5df7b01aacb46e6dbb8fdb8788ec2919959872/rpc/src/client/transport/websocket.rs#L110)). The problem I've recently realized with this is that you can't clone the WebSocketClient instance because you can't clone the JoinHandle.

A solution to this would be to pass ownership of the driver to whoever creates the WebSocketClient instance. This complicates things a little more from the developer experience perspective, but allows for:

1. Cloning of the client handle, allowing access to the driver from multiple concurrent contexts.
2. Control over where the driver runs.

The new WebSocketClient struct would look as follows:

```rust
// We'll now be able to clone the client
#[derive(Debug, Clone)]
pub struct WebSocketClient {
    // Technically we could get rid of the host and port fields too,
    // since they're only used in setting up the initial connection.
    host: String,
    port: u16,
    cmd_tx: ChannelTx<SubscriptionDriverCmd>,
}
```

and the API for creating/using the WebSocketClient would look something like:

```rust
impl WebSocketClient {
    pub async fn new(addr: net::Address) -> Result<(Self, WebSocketClientDriver)> {
        // ...
    }
}

async fn test_websocketclient(addr: net::Address) -> Result<()> {
    let (mut client, driver) = WebSocketClient::new(addr).await?;
    let driver_hdl = tokio::spawn(async move { driver.run().await });

    // ...

    // This now would just signal to the driver to terminate
    client.terminate().await?;
    // Wait for the driver to terminate
    driver_hdl.await
}
```

thanethomson:
Closes #616 

I'm submitting an example of what the code could look like if we expose the WebSocket client driver.

* [x] Wrote tests
* [ ] ~Updated CHANGELOG.md~


romac (approved):
Great stuff, it's much better indeed to not spawn and hold on to a new task from within the client itself.

PR #622, opened 2020-10-08
serialization: Supply default value for app version if field is missing in JSON

thanethomson:
This fixes the issue where deserialization of the header version fails when the `app` field is not supplied. We now explicitly supply the default value of 0 when the `app` field is not supplied.

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/tendermint/src/block/header.rs
+++ b/tendermint/src/block/header.rs
@@ -132,7 +132,7 @@ fn encode_varint(val: u64) -> Vec<u8> {
 /// application.
 ///
 /// When deserializing from JSON, if the `app` field is not supplied it is
-/// automatically to 0.
+/// automatically set to 0.
 ///
 /// <https://github.com/tendermint/spec/blob/d46cd7f573a2c6a2399fcab2cde981330aa63f37/spec/core/data_structures.md#version>
 #[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
--- a/tendermint/src/block/header.rs
+++ b/tendermint/src/block/header.rs
@@ -139,8 +139,18 @@ pub struct Version {
     pub block: u64,
 
     /// App version
-    #[serde(with = "serializers::from_str")]
+    #[serde(
+        with = "serializers::from_str",
+        default = "Version::default_app_version"
+    )]
     pub app: u64,
+}
+
+impl Version {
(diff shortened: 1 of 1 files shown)


tendermint/src/block/header.rs
```diff
@@ -139,10 +142,20 @@ pub struct Version {
     pub block: u64,
 
     /// App version
-    #[serde(with = "serializers::from_str")]
+    #[serde(
+        with = "serializers::from_str",
+        default = "Version::default_app_version"
```
ebuchman, line 147: Do we really need our own function here? Can't we specify something like "use the default for this type" which would be 0?

ebuchman (approved):
Plenty of bikeshedding to do here but lets merge for now

tendermint/src/block/header.rs
ebuchman: cc @greg-szabo

tendermint/src/block/header.rs
```diff
@@ -139,10 +142,20 @@ pub struct Version {
...
+    )]
     pub app: u64,
 }
 
+impl Version {
```
greg-szabo, line 152: Can you please implement the default trait instead?

greg-szabo (changes requested):

tendermint/src/block/header.rs
thanethomson: I'm not sure how implementing the `Default` trait would solve this issue? You'd actually need to implement `Default` for the `u64` type, because serde calls `Default::default()` for the field whose value was omitted in the JSON, not the overall struct, as far as I understand it.

tendermint/src/block/header.rs
thanethomson: I did this consciously to be explicit about the default value. The following has the same effect, but only coincidentally:

```rust
#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
pub struct Version {
    /// Block version
    #[serde(with = "serializers::from_str")]
    pub block: u64,

    /// App version
    #[serde(with = "serializers::from_str", default)]
    pub app: u64,
}
```

tarcieri: I think it's still clear if `#[serde(default)]` is used here.

PR #629, opened 2020-10-09
Height conversion support for u32,u16,u8

### LINKED ISSUE #628: Add From instance to Height for all unsigned integer types which fit in a i64
In ibc-rs, especially in tests, we often constructs instances of `Height` where we statically know that the value will fit in a `i64`.
Currently, such code looks like this:

```rust
let height: Height = 0_64.try_into().unwrap(); // SAFETY: 0 <= i64::MAX
```

It would be nice if we could leverage the fact that some integer types [entirely fit](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bdda71914fdeafa0cac1444baf1d1d58) in a `i64` (`u8`, `u16`, `u32`), such that if we had a `From` instance for all these types, we could rewrite the code above as:

```rust
let height: Height = 0_u32.into();
// or
let height = Height::from(0_u32);
// or perhaps even
let height = Height::from(0);
```

greg-szabo:
Closes #628 

Very cool idea from Romain that I should have thought about.

* [X] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/rpc/src/client/transport/mock.rs
+++ b/rpc/src/client/transport/mock.rs
@@ -116,7 +116,6 @@ impl MockRequestMethodMatcher {
 #[cfg(test)]
 mod test {
     use super::*;
-    use std::convert::TryFrom;
     use std::path::PathBuf;
     use tendermint::block::Height;
     use tendermint::chain::Id;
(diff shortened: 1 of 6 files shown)


romac (approved):

adizere (commented):
CooL!

PR #632, opened 2020-10-12
Trivial fuzzers for MBT tests

### LINKED ISSUE #627: Add trivial fuzzers to MBT tests
As the full-fledged fuzzer #564 is not yet ready, and we need to do some mutation testing for the LIghtClient before the 0.17, let's add the trivial fuzzing capabilities to model-based tests. This approach will not be general, but will for each model-based test mutate it in a set of predefined ways.

andrey-kuprianov:
closes #627 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


andrey-kuprianov (commented):
Great work, @Shivani912!

I am very happy to see that the static JSON test fixtures are now replaced with the fuzzers.

What would be great to see, additionally, is that our tests do indeed cover at least the cases of errors that the LightClient expects, which is unfortunately not yet the case for all of them.

For the list of errors please see [predicates/errors](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/predicates/errors.rs) and [operations/commit_validator](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/operations/commit_validator.rs). 

What we ideally want to see, is that the `_test_run/report` file has these errors in the output. As far as I see, there are a couple of errors that should be covered by fuzzers, but are not yet present:

 - [DuplicateValidator](https://github.com/informalsystems/tendermint-rs/blob/34a23860f62d3ad87d3b335fda1a99ba169181e8/light-client/src/predicates/errors.rs#L38)
 - [InvalidSignature](https://github.com/informalsystems/tendermint-rs/blob/34a23860f62d3ad87d3b335fda1a99ba169181e8/light-client/src/predicates/errors.rs#L42)

Could you please add fuzzers that capture those?
Thanks!

Shivani912 (commented):
Thanks Andrey! Yes that is true we do not cover all the error types yet. I'm happy to add the missing ones:) For that, I was rather thinking to get the replaced tests (both fuzzer and tla ones) merged first because I see that like this [PR](https://github.com/informalsystems/tendermint-rs/pull/657) we might be blocking other work. And I could incrementally keep working on adding more tests to it :)

testgen/src/vote.rs
```diff
@@ -29,6 +29,10 @@ pub struct Vote {
     pub time: Option<u64>,
     #[options(help = "commit round (default: from commit)")]
     pub round: Option<u32>,
+    #[options(
+        help = "to indicate if the vote is nil; produces a 'BlockIfFlagNil' if set, otherwise 'BlockIdFlagCommit' (default)"
```
romac, line 33:
```suggestion
        help = "to indicate if the vote is nil; produces a 'BlockIdFlagNil' if set, otherwise 'BlockIdFlagCommit' (default)"
```

testgen/src/vote.rs
```diff
@@ -83,9 +89,13 @@ impl Generator<vote::Vote> for Vote {
...
-        let block_id = block::Id {
-            hash: block_header.hash(),
-            part_set_header: PartSetHeader::new(1, block_header.hash()),
+        let block_id = if self.is_nil.is_some() {
+            None
+        } else {
+            Some(block::Id {
+                hash: block_header.hash(),
+                part_set_header: PartSetHeader::new(1, block_header.hash()),
+            })
```
romac, line 98: Do we want to generate `BlockIdFlagAbsent` votes as well?

light-client/Cargo.toml
```diff
@@ -45,7 +44,8 @@ thiserror = "1.0.15"
 tokio = "0.2.20"
 
 [dev-dependencies]
+tendermint-testgen = { path = "../testgen"}
```
Shivani912, line 47: This seemed to not work for me when using testgen data structure inside src/test.rs

romac (approved):

testgen/src/vote.rs
Shivani912: Good eye! Thanks :))

light-client/Cargo.toml
romac: Yes, but I moved the `From` instance to convert between `TMLightBlock` and `AnonBlock` to a simple function `tests/model_based.rs` to avoid introducing the dependency.

It's unfortunate that one cannot define a test-only module within a crate and then reuse it in standalone tests (ie. those in the `tests` folder).

Until then, that's the only way I saw to avoid adding `testgen` as a regular dependency.

andrey-kuprianov: I think @romac is very right to point out that testgen should be a dev-dependency. What this shows, is that a bit of refactoring is needed.

testgen/src/vote.rs
Shivani912: It does that automatically for validators who haven't voted ;) 
https://github.com/informalsystems/tendermint-rs/blob/3fb0cb9fcb0efae87dbcb813486992198e1aab8d/testgen/src/commit.rs#L146-L148

PR #635, opened 2020-10-13
Added empty option to signature

### LINKED ISSUE #626: Make the signature on Proposal optional field?
In TM: https://github.com/tendermint/tendermint/blob/master/consensus/state.go#L1089
`SignProposalRequest` is created without the signature (the raw type will have `vec![0]` as the signature), so https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/src/proposal/sign_proposal.rs#L24 will fail, because the "signature" has an incorrect size.
should this be patched here to account for this case, or should Tendermint be patched (change the protobuf type or "pad" the empty signature field to the expected length)? @greg-szabo @marbar3778

greg-szabo:
Fixes #626 .

This is a proposed fix for optional signatures.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/tendermint/src/signature.rs
+++ b/tendermint/src/signature.rs
@@ -30,8 +30,7 @@ impl TryFrom<Vec<u8>> for Signature {
     type Error = Error;
 
     fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
-        // Tendermint Go v0.34 compatibility: vec![0u8] is considered an empty Signature.
-        if value == vec![0u8] {
+        if value.is_empty() {
             return Ok(Self::default());
         }
         if value.len() != ED25519_SIGNATURE_SIZE {
@@ -88,7 +87,7 @@ impl AsRef<[u8]> for Signature {
     fn as_ref(&self) -> &[u8] {
         match self {
             Signature::Ed25519(sig) => sig.as_ref(),
-            Signature::None => &[0u8], // This is for Tendermint Go v0.34 compatibility.
+            Signature::None => &[],
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tendermint/src/signature.rs
```diff
@@ -27,6 +30,9 @@ impl TryFrom<Vec<u8>> for Signature {
     type Error = Error;
 
     fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        if value.is_empty() {
```
tomtau, line 33: The value sent from Go Tendermint over privval socket (for signing) in Vote/Proposal requests is actually not empty: https://github.com/informalsystems/tendermint-rs/issues/626#issue-717983039 -- it has one element which is `0u8`.
tmkms could do some preprocessing though and clear that value before creating a domain type

greg-szabo: Apologies, I misunderstood your comment in the issue. And, come on, a single `0u8`? Aaaargh.

I'll add that in the code. One more thing to raise to the Go team. Thanks!

tomtau: oops, my apologies -- the `vec![0u8]` is something I saw in a debugger when running the Go unit tests. I've just tried against the TM process (0.34.0-rc5) and it's indeed empty.
735c153faf19133305f8ab3adad9385220999a4d is OK
(9ed45f012a1d2d76b27d9d49858217f60d8281f9 can be reverted)

tomtau:
Tested ACK -- https://github.com/informalsystems/tendermint-rs/pull/635#discussion_r504477240

tendermint/src/signature.rs
greg-szabo: LOL, thanks for checking again! I'm happy to implement extra work if it leads to extra checks! I should have checked it myself...

thanethomson (approved):

PR #641, opened 2020-10-14
Replace LightClient single_step JSON tests with MBT

### LINKED ISSUE #640: Replace all LightClient single_step JSON tests with MBT
As a step towards full replacement of the current LC testing with model-based tests, we are going to replace all [single step JSON fixtures](https://github.com/informalsystems/tendermint-rs/tree/master/light-client/tests/support/single_step) with a combination of:
 * MBT tests from TLA+ assertions
 * mutations on top of those.

The result of this issue should be the complete removal of the above directory, while preserving the test coverage.

andrey-kuprianov:
closes #640 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


andrey-kuprianov:
Besides clippy warnings, looks good from my side:) Thanks @Shivani912 for finishing this!

light-client/tests/model_based.rs
```diff
@@ -440,7 +443,7 @@ fn single_step_test(
         tc.initial.signed_header.clone(),
         tc.initial.next_validator_set.clone(),
     );
-    let clock_drift = Duration::from_secs(1);
+    let clock_drift = Duration::from_secs(0);
```
romac, line 446: Just wondering what's the purpose of this change? Does it strengthen the tests somehow by making them not rely on any compensation for the clock drift?

romac (approved):

romac:
Also, could you point me at the changes in the TLA+ specs/models that accounts for the change in https://github.com/informalsystems/tendermint-rs/pull/657?

andrey-kuprianov:
> Also, could someone point me at the changes in the TLA+ specs/models that accounts for the change in #657?

@romac it is described in #659. We have disabled the tests that were checking this for now; will reenable them when the TLA+ model is updated.

romac:
Thanks!

light-client/tests/model_based.rs
andrey-kuprianov: We have disabled the clock drift, because the TLA+ model doesn't account for it at the moment. As a result, some tests were failing non-deterministically. As soon as the clock drift is introduced into the TLA+ model (@konnov and @josef-widder are working on that), we will also enable it here,

romac: Thanks for the clarification!

romac:
@Shivani912 @andrey-kuprianov Great job!

light-client/tests/model_based.rs
Shivani912: The model does not account for `clock_drift` and so several timing-related tests were failing where the difference was of 1 second between a "valid" and "invalid" header. So it just made sense to remove this param from the test as well

Shivani912:
Thanks @romac !

light-client/tests/model_based.rs
konnov: This PR introduces clock drift in the TLA+ spec: https://github.com/tendermint/spec/pull/216

andrey-kuprianov:
@romac thank you for reviewing the PRs so quickly:)

PR #642, opened 2020-10-15
docs: Update Tendermint/Rust versions in READMEs

thanethomson:
For the next v0.17.0-related release, we should probably bump these version numbers.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Build Status][build-image]][build-link]
 [![Audit Status][audit-image]][audit-link]
 [![Apache 2.0 Licensed][license-image]][license-link]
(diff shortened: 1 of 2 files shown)


greg-szabo (approved):
I'm ok with this but is there any reason why Rust 1.47 was specifically chosen? (I guess we only support the latest stable normally, is that it?)

thanethomson:
Yes, I was wondering if we should rather just label it as `stable` instead of specifying the exact version. I think I'll change it to `stable`.

greg-szabo (approved):

PR #646, opened 2020-10-20
rpc: Upgrade WebSocketClient to support full client functionality

thanethomson:
In partial fulfillment of #582.

This PR expands the `WebSocketClient`'s capabilities to be able to handle other types of requests other than just subscriptions (i.e. it implements the `Client` trait).

It also drastically simplifies some of the internal code for the RPC crate.

📖 [Rendered updated ADR (008)](https://github.com/informalsystems/tendermint-rs/blob/rpc/upgrade-websocket-client/docs/architecture/adr-008-event-subscription.md)

* [x] Wrote tests
* [x] Updated CHANGELOG.md


rpc/src/client.rs
```diff
@@ -32,13 +30,13 @@ use tendermint::Genesis;
 #[async_trait]
 pub trait Client {
     /// `/abci_info`: get information about the ABCI application.
-    async fn abci_info(&self) -> Result<abci_info::AbciInfo> {
+    async fn abci_info(&mut self) -> Result<abci_info::AbciInfo> {
```
thanethomson, line 33: The `Client` trait now requires self mutability because the `WebSocketClient` needs this to use its internal channel to communicate with its driver. Unfortunately this change becomes viral and affects the `HttpClient` too.

rpc/src/client/sync.rs
```diff
@@ -39,6 +39,7 @@ pub struct ChannelRx<T>(mpsc::UnboundedReceiver<T>);
 impl<T> ChannelRx<T> {
     /// Wait indefinitely until we receive a value from the channel (or the
     /// channel is closed).
+    #[allow(dead_code)]
```
thanethomson, line 42: This method is not used if only the `http-client` feature is enabled, but putting it behind a feature flag breaks things.

rpc/src/request.rs
```diff
@@ -14,10 +14,20 @@ pub trait Request: Debug + DeserializeOwned + Serialize + Sized + Send {
...
 
+/// Simple JSON-RPC requests which correlate with a single response from the
+/// remote endpoint.
+///
+/// An example of a request which is not simple would be the event subscription
+/// request, which, on success, returns a [`Subscription`] and not just a
+/// simple, singular response.
+///
+/// [`Subscription`]: struct.Subscription.html
+pub trait SimpleRequest: Request {}
```
thanethomson, line 29: This is the simplest way I could find to configure the type system in such a way that I could restrict access to the `Client::perform` method to only those kinds of requests that expect a single response (i.e. non-subscription-related requests), while still preserving the existing overall type architecture.

rpc/src/client.rs
thanethomson: This is no longer relevant as of https://github.com/informalsystems/tendermint-rs/pull/646/commits/56ce9f4ed202e663bcd761dbf0772ca4e507d0fd since @romac pointed out that we could use immutable channels for sending data.

greg-szabo:
The latest changes are awesome. When can we review and merge?

rpc/src/client/transport/router.rs
```diff
@@ -0,0 +1,182 @@
...
+        for (id, event_tx) in subs_for_query {
+            if let Err(e) = event_tx.send(Ok(ev.clone())) {
+                disconnected.insert(id.clone());
+                debug!(
+                    "Automatically disconnecting subscription with ID {} for query \"{}\" due to failure to publish to it: {}",
+                    id, ev.query, e
+                );
+            }
+        }
+        // Obtain a mutable reference because the previous reference was
```
melekes, line 44: interesting 🤔 i would expect Rust to figure out that the loop is over

rpc/src/client/transport/router.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub fn add(&mut self, id: impl ToString, query: impl ToString, tx: SubscriptionTx) {
+        let query = query.to_string();
+        let subs_for_query = match self.subscriptions.get_mut(&query) {
+            Some(s) => s,
+            None => {
+                self.subscriptions.insert(query.clone(), HashMap::new());
+                self.subscriptions.get_mut(&query).unwrap()
+            }
+        };
+        subs_for_query.insert(id.to_string(), tx);
```
melekes, line 69: what if the id already exists?

rpc/src/client/transport/router.rs
```diff
@@ -0,0 +1,182 @@
...
+            subscriptions: HashMap::new(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum PublishResult {
+    Success,
+    NoSubscribers,
+    AllDisconnected,
```
melekes, line 104: what is the intuition behind this type of result?

rpc/src/client/transport/websocket.rs
```diff
@@ -246,124 +340,146 @@ impl WebSocketClientDriver {
...
-                    // subscriptions.
-                    SubscriptionState::Active => {
-                        if let Some(event_tx) = self.router.get_active_subscription_mut(
-                            &SubscriptionId::from_str(&req_id).unwrap(),
-                        ) {
-                            let _ = event_tx.send(Err(e)).await;
-                        }
-                    }
-                },
+    async fn confirm_pending_command(
```
melekes, line 468: confirm here means sending response?

rpc/src/client/transport/router.rs
romac: `for x in y { b }` is syntactic sugar for `let mut iter = y.into_iter(); while let Some(x) = iter.next() { b }` and [`into_iter()`](https://doc.rust-lang.org/stable/std/iter/trait.IntoIterator.html#tymethod.into_iter) consumes `self`, therefore one cannot access the collection being iterated over (ie. `subs_for_query`) after the loop is over.

On the other hand, `for x in &y { b }` would desugar to a call to `iter()` which does not not consume `self` but instead yields reference to the underlying elements. Same with `for x in &mut y`.

I wonder though if we could use `for (id, event_tx) in &subs_for_query` here instead?

melekes: got it! thanks for explaining it @romac 🙏

thanethomson: So interestingly, the `for (id, event_tx) in &subs_for_query` didn't work previously, and I've checked again now, it results in the following error:

```
error[E0277]: `&&mut std::collections::HashMap<std::string::String, client::sync::ChannelTx<std::result::Result<event::Event, error::Error>>>` is not an iterator
  --> rpc/src/client/transport/router.rs:35:31
   |
35 |         for (id, event_tx) in &subs_for_query {
   |                               ^^^^^^^^^^^^^^^ `&&mut std::collections::HashMap<std::string::String, client::sync::ChannelTx<std::result::Result<event::Event, error::Error>>>` is not an iterator
   |
   = help: the trait `std::iter::Iterator` is not implemented for `&&mut std::collections::HashMap<std::string::String, client::sync::ChannelTx<std::result::Result<event::Event, error::Error>>>`
   = note: required by `std::iter::IntoIterator::into_iter`
```

`for (id, event_tx) in &mut subs_for_query` also doesn't compile.

The following code, however, does work, so I'll update the code accordingly:

```rust
for (id, event_tx) in subs_for_query.borrow_mut() {
    // ...
}
```

I feel as though if I manually expanded all the syntactic sugar, I'd understand this better 😁

rpc/src/client/transport/router.rs
thanethomson: It'd be overwritten.

My assumption here is that our UUIDs that we're automatically generating for subscriptions won't ever collide. I suppose that leaves a very small, albeit non-zero, non-deterministic probability of having a subscription dropped arbitrarily, but I'd imagine that networking failures leading to the same outcome would be far higher probability than that of a UUIDv4 collision.

rpc/src/client/transport/router.rs
thanethomson: It was one way of passing some additional information back to the `WebSocketClient` after it attempts to publish an event.

Since the `WebSocketClient` only sends an unsubscribe request once all subscribers for a particular query have dropped, I needed to be able to distinguish between the case where there were no subscribers to begin with (i.e. we somehow received a spurious event without having a subscription) and the case where we had subscribers, but since we published last to them they dropped their receiving ends. Both would publish the event to zero subscribers.

In the former case, we would consider this to be unexpected behaviour on the part of the node (I don't do anything about it right now though), and in the latter case it'd be a signal to send a single unsubscribe request.

rpc/src/client/transport/websocket.rs
thanethomson: Yes. I can rename it to something a bit clearer though.

melekes (approved):

PR #639, opened 2020-10-14
Re-built tendermint-proto with serialization annotations

greg-szabo:
Closes #634.

This is a research for #638 that adds Serialization support for protobuf structs.
 
* [X] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [X] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

PR #648, opened 2020-10-21
Renaming "parts" to "part_set_header"

Shivani912:
**Summary of changes:**
- Renamed `parts` to `part_set_header` in `block::Id` (Domain Type)
- Re-generated MBT test files to reflect this change
- Remaining: @thanethomson will update the RPC JSON test files

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


Shivani912:
Woohoo!! @thanethomson  Thanks! That's a lot of JSON files 😀

thanethomson (approved):
Awesome, thanks @Shivani912! 🍷 

I'm merging this now as an intermediate solution prior to the completion of #639, which is more holistic, but this PR at least solves the field name change from Tendermint's side in https://github.com/tendermint/tendermint/pull/5523 in the short-term to fix our integration tests.

Issue #650, opened 2020-10-22
Wrong validation of LightBlock's commits


andrey-kuprianov:
I have a failing MBT test for the LightClient verifier, with an empty commit in the LightBlock.

This is how the test fails:

```
    > step 2, expecting Invalid
      > lite: NotEnoughTrust(NotEnoughTrust(VotingPowerTally { total: 100, tallied: 0, trust_threshold: TrustThresholdFraction { numerator: 1, denominator: 3 } }))
```

The relevant test part is this:

```tla+
[current |->
    [Commits |-> {},
      header |->
        [NextVS |-> { "n1", "n2", "n3", "n4" },
          VS |-> { "n2", "n3" },
          height |-> 5,
          lastCommit |-> { "n1", "n3" },
          time |-> 6]],
  now |-> 1401,
  verdict |-> "INVALID",
  verified |->
    [Commits |-> { "n1", "n2", "n4" },
      header |->
        [NextVS |-> { "n2", "n3" },
          VS |-> { "n1", "n2", "n4" },
          height |-> 2,
          lastCommit |-> { "n1", "n2", "n3", "n4" },
          time |-> 3]]]
```

`current` here is the untrusted header, and `verified` is the trusted header, which are submitted to the `verifier.verify()` function. From the point of view of the spec, and also from my personal understanding, an untrusted signed header with an empty commit should be rejected as invalid. The implementation, on the other hand, returns the `NotEnoughTrust` verdict. My understanding is that this comes from the specific order of executing checks in [this code](https://github.com/informalsystems/tendermint-rs/blob/6be5ebf8dc673cebfbfeed67737f76e73e7d4eef/light-client/src/predicates.rs#L278-L293), probably for performance reasons, as verifying the signatures is costly.

We could address this from different angles:
* Change the spec, to allow more behaviors in this case. While this is doable, in my view a signed header with an empty commit still should be under the `Invalid` category, and not under `NotEnoughTrust`.
* Change the order of checks in the code; don't know whether this is doable or acceptable from the performance point of view. 
* Add some lightweight checks that would verify, e.g., that the commit contains more than 1/3 of votes from the validator set, without actually verifying the signatures. This is going to be lightweight from the performance point of view, but would make the code a bit more complicated.

I think @romac should probably decide here; I am open for discussions.

[...]

PR #652, opened 2020-10-22
Fix bug where a commit with only absent signatures would be deemed valid instead of invalid

romac:
Close: #650

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


light-client/src/operations/commit_validator.rs
```diff
@@ -53,22 +53,29 @@ impl CommitValidator for ProdCommitValidator {
...
     ) -> Result<(), VerificationError> {
-        // TODO: `self.commit.block_id` cannot be zero in the same way as in Go
-        //       Clarify if this another encoding related issue
-        if signed_header.commit.signatures.len() == 0 {
+        let signatures = &signed_header.commit.signatures;
+
+        // Check the the commit contains at least on non-absent signature.
+        // See https://github.com/informalsystems/tendermint-rs/issues/650
+        let has_non_absent_signatures = signatures.iter().any(|cs| !cs.is_absent());
+        if !has_non_absent_signatures {
```
thanethomson, line 61: My mind struggles a little to keep up with the number of logical negations going on over here 🤯 😄 

Could it perhaps be a bit more readable if it were worded like:

```suggestion
        let has_present_signatures = signatures.iter().any(|cs| !cs.is_absent());
        if !has_present_signatures {
```

or does the term "non-absent" hold special meaning above simply "present"?

light-client/src/operations/commit_validator.rs
```diff
@@ -53,22 +53,29 @@ impl CommitValidator for ProdCommitValidator {
         signed_header: &SignedHeader,
         validator_set: &ValidatorSet,
     ) -> Result<(), VerificationError> {
-        // TODO: `self.commit.block_id` cannot be zero in the same way as in Go
-        //       Clarify if this another encoding related issue
-        if signed_header.commit.signatures.len() == 0 {
+        let signatures = &signed_header.commit.signatures;
+
+        // Check the the commit contains at least on non-absent signature.
```
thanethomson, line 58:
```suggestion
        // Check the the commit contains at least one non-absent signature.
```

liamsi (commented):
So great to see that the MBT by @andrey-kuprianov found an actual bug in the implementation.

PR looks good to me. Two minor comments:
 - light-client/detailed.log seems like this file was committed accidentaly? (it's empty)
 - the TODO isn't relevant anymore I think and if it should be dealt with at the encoding level

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## Unreleased
+
+### BUG FIXES:
+
+- `[light-client]` Fix bug where a commit with only absent signatures would be
+  deemed valid instead of invalid ([#652])
+
+[#652]: https://github.com/informalsystems/tendermint-rs/pull/652
```
andrey-kuprianov, line 8: Isn't it that we should reference only the issues in the changelog, not PRs? 
Also, probably it makes sense to link this PR to the issue via "closes #650"?

andrey-kuprianov (changes requested):
Besides the comment about changelog, the PR looks good to me!

It's a bit like an interactive game -- MBT proposes counterexamples, developers propose changes, and the cycle repeats;)

andrey-kuprianov (approved):
Looks good, @romac, thanks for fixing this!

light-client/src/operations/commit_validator.rs
```diff
@@ -53,18 +53,22 @@ impl CommitValidator for ProdCommitValidator {
...
         validator_set: &ValidatorSet,
     ) -> Result<(), VerificationError> {
-        // TODO: `self.commit.block_id` cannot be zero in the same way as in Go
-        //       Clarify if this another encoding related issue
-        if signed_header.commit.signatures.len() == 0 {
+        let signatures = &signed_header.commit.signatures;
+
+        // Check the the commit contains at least one non-absent signature.
+        // See https://github.com/informalsystems/tendermint-rs/issues/650
+        let has_present_signatures = signatures.iter().any(|cs| !cs.is_absent());
```
ebuchman, line 60: Shouldn't this check `is_commit`? It might not be absent but it could still be nil ...

light-client/src/predicates.rs
```diff
@@ -261,6 +261,13 @@ pub fn verify(
         &trusted.signed_header.header,
     )?;
 
+    // Verify that more than 2/3 of the validators correctly committed the block.
```
ebuchman, line 264: I believe this is false security and uneconomical - it's trivial for a faulty node to make this pass now even though we might never be able to extend trust to this header in the first place. I think it also means we end up doing this verification twice in the light node? 

The performance issue isn't that big a deal to the light-node, but for IBC handlers it probably doesn't make sense to do this more expensive check before the +1/3 check.

ebuchman (commented):
I think this whole thing is a sign we want a better client-server protocol, so we can just ask full nodes for the next header we'll be able to trust based on our current trusted validator set, and then we can just reject them right away if they don't give us a good one, rather than doing the whole bisection scheduling on the client side. I think this would basically mean the full node does the bisection against its own db on behalf of the light client (though it doesn't have to verify signatures again since it already did so, just check which validators signed and if they overlap enough with those the client trusts).

light-client/src/predicates.rs
ebuchman: I still think we should revert this. It creates DoS attacks for light clients since full nodes can now spam us with massive commits (eg. 10k validators...) https://github.com/tendermint/tendermint/pull/5606

I guess it means we need to fix the MBT tests though to better describe why certain cases should fail?

romac: Ok, will open a new PR to revert this. We can perhaps do the required changes in MBT tests as part of that PR as well.

Issue #612, opened 2020-10-06
rpc: Automatically generate fixtures for testing


thanethomson:
Follows from https://github.com/informalsystems/tendermint-rs/pull/604#discussion_r499586326

Right now, it appears as though our fixtures with which we're testing the RPC are manually generated from a running Gaia or Tendermint instance. We need to automate this process, so that updating the fixtures is easy.

thanethomson:
Completed for the `kvstore` ABCI app in #653. If we want to support Gaia, we should open a separate issue about this.

PR #653, opened 2020-10-23
rpc: Add probe to capture responses from Tendermint RPC for use in testing

### LINKED ISSUE #611: rpc: Include more variations on fixtures for testing
Follows from discussion here: https://github.com/informalsystems/tendermint-rs/pull/604#discussion_r499586326

We need to evaluate the fixtures that we are testing for the RPC and see if we should add more variants for each case to ensure our serialization/deserialization is working as expected.

thanethomson:
Closes #611 and #612.

I'm adding a small RPC probe application here that specifically does not rely on the `tendermint-rpc` crate and allows for maximal flexibility in JSON-RPC response structure. The idea here is to be able to execute a bunch of queries/subscriptions against a running Tendermint node and capture the responses (and possibly the requests, if needed).

These responses we can then add into our text fixtures for local "integration" testing.

The probe doesn't depend on any other crates in the repo - this is by design, since its purpose is to capture responses potentially outside the purview of our serialization infrastructure.

📖 [Rendered README](https://github.com/informalsystems/tendermint-rs/blob/testing/fixture-probe/rpc-probe/README.md)

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


greg-szabo (approved):
In a follow up PR I would like to set the version number of the crate to the upcoming release and pin the Docker Hub image to the same version that the release is going to use.

thanethomson:
> In a follow up PR I would like to set the version number of the crate to the upcoming release and pin the Docker Hub image to the same version that the release is going to use.

Cool, we could do that in a separate PR or just in the v0.17.0-rc3 release PR.

PR #658, opened 2020-11-03
proto: re-export `prost_types` as `well_known_types`

tarcieri:
My main interest is being able to get to `prost_types::Timestamp` without having to include `prost_types` in addition to
`tendermint-proto`, however I think there might be more cases where it's helpful to have access to the well-known proto types in the future.

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -85,6 +85,7 @@ mod tendermint {
     }
 }
 
+pub use prost_types as well_known_types;
 pub use tendermint::*;
 
 mod domaintype;


liamsi (commented):
I think that makes sense! LGTM 👍 I'll let Greg have the final say on this PR as I'm bit out of loop on everything proto-related currently.

thanethomson (approved):
If it helps you @tony-iqlusion, it makes sense to me 😁

@greg-szabo, any objections from your side?

tarcieri:
Looks like this is also addressed in #653?

greg-szabo:
Apologies, I gave Tony the wrong PR number. In #639 we dissected prost_types and added our own implementation so they can be serialized as JSON objects. So, there is a `tendermint_proto::google::protobuf::Timestamp` and similar on `master` right now.

But we are trying to separate protobuf and JSON serialization in an upcoming RC and it's possible that we won't need to dissect `prost_types` (because it's cumbersome and hacky). In that case, this idea is nice.

To keep things manageable (although not the simplest), I suggest that you use the `tendermint_proto::google::protobuf` types instead of the prost_types in RC2 (soon to be released). IF we make the decision to not build our own prost_types, then the `tendermint_proto::well_known_types` is a good solution to keep life simple. (A possibility in RC3.)

But I would rather not add both to RC2 to avoid confusion.

If anyone has strong opinions in adding it to RC2, please raise it here. Maybe I'm the only one who would get confused. :)

tarcieri:
`tendermint_proto::google::protobuf` works for me!

PR #664, opened 2020-11-09
CanonicalVote and CanonicalProposal now correctly encode/decode BlockId

### LINKED ISSUE #663: Incorrect canonical block id
In Go tendermint: https://github.com/tendermint/tendermint/blob/master/types/canonical.go#L18
If block id hash is empty, `CanonicalizeBlockID` will return `nil` for CanonicalBlockID.

In current tendermint-rs, one will get an empty struct (as with non-canonical Block ID): https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/src/block/id.rs#L116

greg-szabo:
Closes #663.

Correctly encode/decode CanonicalBlockId for CanonicalVote and CanonicalProposal. See issue for more details.

* [ ] Wrote tests
* [X] Updated CHANGELOG.md


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
   deemed valid instead of invalid ([#650])
 - `[light-client]` Revert a change introduced in [#652] that would enable DoS attacks,
   where full nodes could spam the light client with massive commits (eg. 10k validators).
+- `[tendermint]` (Since RC1) CanonicalBlockId is now correctly decoded to `None` in `CanonicalVote` and `CanonicalProposal`.
```
thanethomson, line 18:
```suggestion
- `[tendermint]` (Since v0.17.0-rc1) CanonicalBlockId is now correctly decoded to `None`
  in `CanonicalVote` and `CanonicalProposal` when its hash is empty.
```

Is this perhaps more explicit about the change?

thanethomson (changes requested):
I'd recommend adding regression tests when fixing compatibility issues like this, but otherwise LGTM 👍

tomtau (approved):

tendermint/src/proposal.rs
```diff
@@ -210,6 +210,86 @@ mod tests {
         assert_eq!(got, want)
     }
 
+    #[test]
+    // Test proposal encoding with a malformed block ID which is considered null in Go.
+    fn test_encoding_with_emtpy_block_id() {
```
thanethomson, line 215:
```suggestion
    fn test_encoding_with_empty_block_id() {
```

greg-szabo:
All right, regression tests were added.
The first problem with regression tests is that there is supposed to be an original test case at the "before" state of the change that you compare to and shame on me: the CanonicalVote and CanonicalProposal structs didn't have tests at all.
So I added a few unit tests to check the domain knowledge about these types. As we implement more domain knowledge, hopefully this gets extended more and possibly used for regression testing.
The second problem was that regression tests are supposed to prove that the rest of the system behaves similarly as before and only the area of change are affected. CanonicalVote and CanonicalProposal are used in signing and although signing has tests (which still pass), but not the specific case where the block ID is present but with an empty hash. So I added this specific test to the signing tests. The proof is that all signing still passes and builds the sameprotobuf bytestream as they do in Go.

thanethomson (approved):
Nice tests! 😁

thanethomson (approved):

Issue #665, opened 2020-11-09
Extend Testgen to generate Validator sets


andrey-kuprianov:
After extending Tendermint Validator Set with the [total_voting_power field](https://github.com/informalsystems/tendermint-rs/blob/0b1962a6fb985830fd3fe610df54bd6aa2e0c097/tendermint/src/validator.rs#L21), the model-based tests stopped to work.

The reason for tha behavior is that validator sets are not first class citizens in the Testgen. Instead, when generating validator sets via Jsonatr transform, we [map Testgen validator generation](https://github.com/informalsystems/tendermint-rs/blob/0b1962a6fb985830fd3fe610df54bd6aa2e0c097/testgen/jsonatr-lib/apalache-tendermint.json#L106) over the array of abstract validator identifiers, as produced by Apalache. This was ok, until `total_voting_power`, that requires treating Validator Set as a datastructure on its own.

The solution is as follows:
 - add `ValidatorSet` to Testgen, and generate it in the same way as all other datastructures.
 - add `valset` command to `tendermint-testgen` binary.
 - in the Jsonatr transform above, invoke `tendermint_valset` for the array of identifiers, instead of mapping the `tendermint_validator`.

Shivani912:
Thanks @andrey-kuprianov , we definitely need to make this improvement to Testgen. On a second thought, does it make sense to use `validator::Set::new(validators, voting_power)` instead of the old valset constructor since we do not currently make use of `total_voting_power` in Testgen apart from construction. I think it'll be easier to quickly unblock things this way.

PR #666, opened 2020-11-09
rfc3339 direct ser/deser fix for protobuf Timestamp

greg-szabo:
A fix for an issue that bubbled up during #665 .

The Time domain type was incorrectly serialized by itself. (It was correctly serialized as part of another struct.)

This change implements correct serialization for tendermint::Time via tendermint_proto::Timestamp.
Timestamp needed a helper struct so it could be serialized/deserialized without custom serializers.
You can clearly see the dependencies on serializers in this change:
* The serializer to correctly ser/deser an RFC3339 string into a Timestamp is moved into `tendermint_proto::serializers::timestamp`.
* The Timestamp struct needs a helper struct so it can use those serializers. Otherwise Timestamp would have required a custom serializer.
* With the Timestamp struct directly serializable, there's no need for special serialization for `Option<Timestamp>`. All structs having that field can now be serialized by just defining serialization over `Option<>` with the `tendermint_proto::serializers::optional` module.
* Unfortunately, the tendermint::Time struct is directly serialized/deserialized in testgen structs (and possibly in IBC) in several places. But now that it is directly serializable through Timestamp (and from an RFC3339 string), there is no need for extra annotation when serializing/deserializing it. It's going to be interesting if/when someone comes to us that they need it serialized differently (in the `{seconds:x,nanos:x}` format, isntead of RFC3339).

* [X] Wrote tests
* [X] Updated CHANGELOG.md


CHANGELOG.md
```diff
@@ -25,7 +27,7 @@
 [#652]: https://github.com/informalsystems/tendermint-rs/pulls/652
 [#639]: https://github.com/informalsystems/tendermint-rs/pull/639
 [#660]: https://github.com/informalsystems/tendermint-rs/issues/660
-
+[#665]: https://github.com/informalsystems/tendermint-rs/issues/665
 ## v0.17.0-rc1
```
romac, line 31:
```suggestion
[#665]: https://github.com/informalsystems/tendermint-rs/issues/665

## v0.17.0-rc1
```

romac (approved):
I like the new `Rfc3339` struct :)

Just suggested a whitespace change in the README, but otherwise LGTM :)

romac (approved):

andrey-kuprianov:
@greg-szabo What I especially like about this PR is that it removes more code than it adds -- it simplifies things instead of complicating them :100:

Issue #654, opened 2020-10-28
Tendermint data structures and serialization


thanethomson:
After working through #639, we're pretty convinced that it's best to maintain 2 categories of data structures:

1. **Domain types**, where we should not be able to construct incorrect data structures.
2. **Serialization types**, which facilitate serialization/deserialization to/from other formats such as Protobuf (for future wire-level communication and signature generation) and JSON (for use in the RPC).

Previously we've mixed our domain types and serialization types when it comes to JSON, and we'd like to split them up. A rough plan to implement this would be the following:

- [x] Domain types, with constructors that do not allow for incorrect construction. These should continue to live in the `tendermint` crate.
- [ ] Serialization types, with `TryFrom` converters to corresponding domain types. These should probably be in a single crate together (we could rename the `proto` crate at some point to something more generic, and feature-guard each set of data structures).
   - [x] Protobuf (already present, automatically generated using [prost])
   - [ ] JSON (ideally auto-generated from something like the [Tendermint RPC OpenAPI spec][openapi])

[prost]: https://github.com/danburkert/prost
[openapi]: https://docs.tendermint.com/master/rpc/openapi.yaml

greg-szabo:
I was able to generate JSON structs (or Serialization types, as the issue calls it) from the OpenAPI definition in Tendermint Go. We need to decide how/where we want to store them.

Currently, I've created a `json` folder for the compiled structs and named the crate `tendermint-json` which is in line with the protobuf folder structure. But I'm happy to name them differently.

The OpenAPI generator creates a full crate for the json structs (including a `Cargo.toml` and `src` folder) so merging the JSON structs with the protobuf structs would be counterintuitive.

The generation is also done differently. Instead of a `json-compiler` crate, the generation can be done with a one-liner `docker` call. I propose we keep that and not write additional Rust code for JSON generation.

thanethomson:
> I was able to generate JSON structs (or Serialization types, as the issue calls it) from the OpenAPI definition in Tendermint Go. We need to decide how/where we want to store them.

Nice! 👍 

> Currently, I've created a json folder for the compiled structs and named the crate tendermint-json which is in line with the protobuf folder structure. But I'm happy to name them differently.

Wouldn't it be better, in the long run, to have a single crate (e.g. `tendermint-serialization` or `tendermint-wire`) that contains both the Protobuf and serde-compatible structs and put each set of structures behind a feature flag?

greg-szabo:
[...]

PR #672, opened 2020-11-12
proto: Rename DomainType trait to Protobuf

thanethomson:
As discussed on our call yesterday, and towards fulfilling #654, in this PR I've:

1. Renamed the `DomainType` trait to `Protobuf` to clarify its purpose.
2. Kept it in the `tendermint-proto` crate, because it makes the most sense there.
3. Added a testable example of how to make use of the trait.

* [x] Wrote tests
* [x] Updated CHANGELOG.md


tendermint/src/proposal.rs
```diff
@@ -17,7 +17,7 @@ use crate::{Error, Kind};
 use bytes::BufMut;
 use std::convert::{TryFrom, TryInto};
 use tendermint_proto::types::Proposal as RawProposal;
-use tendermint_proto::{DomainType, Error as DomainTypeError};
+use tendermint_proto::{Error as DomainTypeError, Protobuf};
```
greg-szabo, line 20:
```suggestion
use tendermint_proto::{Error as ProtobufError, Protobuf};
```
There are not many examples of this, but I guess it's worth renaming these too, for consistency.

greg-szabo (approved):
Looks good. Maybe we should also rename the cases where there are multiple Error crates imported and the one from tendermint_proto::Error is aliased as DomainTypeError. A quick search showed me this in vote.rs, proposal.rs, sign_proposal.rs and sign_vote.rs.

proto/src/lib.rs
```diff
@@ -73,10 +73,174 @@ mod tendermint {
...
+///
+/// // We expect a validation error here
+/// assert!(MyDomainType::decode(invalid_raw_bytes.as_ref()).is_err());
+/// ```
+pub trait Protobuf<T: Message + From<Self> + Default>
+where
+    Self: Sized + Clone + TryFrom<T>,
+    <Self as TryFrom<T>>::Error: Into<BoxError>,
+{
+    /// Encodes the DomainType into a buffer.
```
adizere, line 168: Nit:
Maybe we could stop referring to Self as "DomainType" in the comments for these methods. For example:

```suggestion
    /// Encodes the domain type (self) into a buffer.
```

thanethomson: Yes you're right, I forgot to check the comments. Will update 👍

tendermint/src/proposal.rs
thanethomson: Done 👍

greg-szabo (approved):
Good stuff.

I'm not sure if the word "equivalent" is the best to describe the similarities between the `Protobuf::encode()` and the `Prost::Message::encode()` functions. They have similar functionality that they implement differently. Maybe the word is part of some kind of documentation standard (you seem to have used the same template in multiple places) and I don't have a good enough vocabulary to give you a better alternative. Maybe "replacement"? `Protobuf::encode()` replaces the `Prost::Message::encode()` function with the added features of converting to a domain type. Or something.

If people understand "equivalent", then just ignore me.

greg-szabo (approved):

PR #676, opened 2020-11-16
State struct JSON serialization

### LINKED ISSUE #675: tendermint::consensus::State serde impls
We serialize `tendermint::consensus::State` to JSON in TMKMS for double signing prevention.

It looks like the `Serialize` and `Deserialize` impls were removed recently. Can we get them back?

greg-szabo:
Fixes #675.

Well, for now. Eventually, we will move serialization as per #654.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/tendermint/src/consensus/state.rs
+++ b/tendermint/src/consensus/state.rs
@@ -1,6 +1,7 @@
 //! Tendermint consensus state
 
 pub use crate::block;
+use serde::{Deserialize, Serialize};
 pub use std::{cmp::Ordering, fmt};
 
 /// Placeholder string to show when block ID is absent. Syntax from:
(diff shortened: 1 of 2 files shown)


thanethomson (approved):
LGTM 👍

tendermint/src/consensus/state.rs
```diff
@@ -20,6 +23,7 @@ pub struct State {
     pub step: i8,
 
     /// Block ID being proposed (if available)
+    #[serde(with = "tendermint_proto::serializers::optional")]
```
tarcieri, line 26: @greg-szabo @thanethomson does this not handle deserialization in the case a field is null?

The existing state files have data like...

```json
{"height":"3820001","round":"0","step":2,"block_id":null}
```

But that no longer parses:

```
invalid type: null, expected struct BlockId at line 1 column 50
```

tarcieri: see also: https://github.com/iqlusioninc/tmkms/pull/226

PR #680, opened 2020-11-17
consensus::State deserialization fixes

### LINKED ISSUE #679: consensus::State (de)serialization regressions
There are a couple problems with the serialization (or rather, deserialization) implemented in #676 which prevents existing consensus state JSON files from being read. (Note: we can support a new *serialization*, it's just the deserialization that needs to be backwards compatible)

### null `block_id` field

`tendermint_proto::serializers::optional` does not appear to handle the field being explicitly `null`, e.g.:

```json
{"height":"3820001","round":"0","step":2,"block_id":null}
```

This results in the following error:

```
invalid type: null, expected struct BlockId at line 1 column 50
```

### `total` field of `parts::Header`

This was previously serialized as a string. It now appears to be serialized as an integer. This causes the following error when deserializing an existing file:

```
 invalid type: string "1", expected u32 at line 1 column 146
```

greg-szabo:
Fixes #679.

This should make the `consensus::State` struct backwards compatible for JSON deserialization. Note that serialization will always use the new format which is not backwards compatible because of a rename from `parts` to `part_set_header` in one field.

* [X] Wrote tests
* [X] Updated CHANGELOG.md


greg-szabo:
That's interesting. These failures didn't show up in local testing. Checking...

greg-szabo:
All right, the test problem is my fault, I didn't run the full test suite after changing the `total` to a string.

proto/src/serializers/part_set_header_total.rs
```diff
@@ -0,0 +1,51 @@
...
+//!
+//! The deserializer is created for backwards compatibility: `total` was changed from a
+//! string-quoted integer value into an integer value without quotes in Tendermint Core v0.34.0.
+//! This deserializer allows backwards-compatibility by deserializing both ways.
+//! See also: https://github.com/informalsystems/tendermint-rs/issues/679
+use serde::{de::Error, de::Visitor, Deserializer, Serialize, Serializer};
+use std::convert::TryFrom;
+use std::fmt::Formatter;
+
+struct PartSetHeaderTotalStringOrU32;
```
romac, line 12: Do we really need to define a type + a visitor?

I wonder if this can be expressed instead with just the `serialize` and `deserialize` functions, and invoked via a  `#[serde(with = "crate::serializers::part_set_header_total")]` attribute?

romac: Instead of a visitor, you can use the `deserialize` method provided by the `Deserialize` trait: https://docs.rs/serde/1.0.117/serde/trait.Deserialize.html

eg. to deserialize a `u64`:

```rust
u64::deserialize(deserializer)
```

greg-szabo: Yes, that was the original solution. Unfortunately, the deserializer in the above example is not clonable and I need to be able to derive if the incoming input is an integer or a string.

Here's the outline of the original solution:

```
let try_string = String::deserialize(deserializer.clone());
if try_string.is_err() {
  u32::deserialize(deserializer)?
} else {
  try_string.unwrap()
}
```

The deserialize function consumes the deserializer.

So, you need to be able to figure out if the incoming data is an integer or a String _before_ trying to deserialize it. Short of writing our own deserializer, the Visitor seemed the right solution. The "either string or struct" example in the Serde documentation also uses this solution: https://serde.rs/string-or-struct.html

romac: Right, makes sense! That's an unfortunate limitation/design decision of `serde`. Thanks for pointing me to the serde doc, had completely forgotten about this example,  my bad.

All good then :)

greg-szabo: No worries, I'm grateful that you point out possible optimizations!

thanethomson (approved):

PR #681, opened 2020-11-17
Upgrade Tokio to version 0.3.0

romac:
Close: #683 

**Target:** v0.17.0-rc4

- [x] Upgrade Tokio to version 0.3.
- [x] Upgrade all dependencies which depends on Tokio 0.2 to newer versions for Tokio 0.3 support.
  - [x] `async-tungstenite`: 0.8 => 0.10
  - [x] `hyper`: 0.13 => 0.14.0-dev (development version)

---

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


tony-iqlusion:
It'd be really helpful if we could get another `-rc` release with #679 before this lands.

It's a breaking change to TMKMS to upgrade Tokio and we haven't done it yet, and we need to be able to test Stargate upgrades with production keys ASAP.

thanethomson:
> It'd be really helpful if we could get another `-rc` release with #679 fixed before this lands.
> 
> It's a breaking change to TMKMS to upgrade Tokio and we haven't done it yet, and we need to be able to test Stargate upgrades with production keys ASAP.

Sure - I'd say we should target v0.17.0-rc4 to upgrade Tokio 👍 (rc3 will release the fix for #679)

melekes (approved):
👍

rpc/src/client/subscription.rs
```diff
@@ -53,14 +54,19 @@ pub trait SubscriptionClient {
 /// ```
 ///
 /// [`Event`]: ./event/struct.Event.html
+#[pin_project]
```
thanethomson, line 57: Thanks for this! 👍  I was wondering when this was going to become necessary 😁

thanethomson (approved):
Looks awesome! 👍

rpc/src/client/subscription.rs
romac: FWIW, I initially did it the unsafe and manual way, but that unsafe block kept me most of the following night, so I figured I'd rather trust a real Rust dev and take the (admittedly very small) compile-time + code size penalty with `pin-project`.

romac: Before/after: https://github.com/informalsystems/tendermint-rs/pull/681/commits/7c8f4448e23ca79ce2a12eba87ac5012a98ac0c4

adizere:
As a temporary workaround, the PR https://github.com/informalsystems/ibc-rs/pull/364 depends on branch `romac/upgrade-tokio` here. We'll delete this branch later once it's not needed anymore.

PR #682, opened 2020-11-18
Add MessageExt trait with blanket implementation

thanethomson:
As discussed earlier with @tony-iqlusion, this would be useful in instances where users may want to convert a Prost `Message` directly to a `Vec<u8>`.

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -202,3 +202,18 @@ where
         Self::decode_length_delimited(v)
     }
 }
+
+/// Convenience extension to Prost's [`Message`] trait.
+///
+/// [`Message`]: https://docs.rs/prost/*/prost/trait.Message.html
+pub trait MessageExt: Message + Sized {
+    /// Attempt to encode the message directly to a `Vec<u8>`.
+    fn to_vec(&self) -> Result<Vec<u8>, Error> {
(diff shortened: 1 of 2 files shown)


greg-szabo (approved):
I still feel that this should go into the downstream projects instead because `tendermint-proto` or the `tendermint` crate doesn't seem to make use of it but if it keeps prost/protobuf-encoding-related functionality in one place, then this is as good as it gets.

greg-szabo:
Oh, one more thing! I think this should make it into the next RC release and it will need a short changelog entry too.

thanethomson:
> I still feel that this should go into the downstream projects instead because tendermint-proto or the tendermint crate doesn't seem to make use of it but if it keeps prost/protobuf-encoding-related functionality in one place, then this is as good as it gets.

Sure, we could defer this to the downstream projects. Technically, extensions to Prost should probably have their own standalone crate that's easy to import into other projects.

Let's discuss in today's meeting.

thanethomson:
So after our call today, we've decided that it's probably better for downstream projects to implement this kind of trait themselves because nothing in tendermint-rs or ibc-rs actually uses this trait.

PR #685, opened 2020-11-18
Add basic release script

thanethomson:
The current manual release process is a bit of a pain, so I thought I'd try to automate it just a tiny bit 😁 

From the documentation inside the script:

---

`release.sh` will hopefully allow us to publish all of the necessary crates in this repo in the right order, along with a few checks and balances to try to avoid mistakes. It is assumed that only one person will be releasing all crates at the same time.

For each crate, it will:

1. Run `cargo publish --dry-run` for that crate
2. List all files in the package with `cargo package --list`
3. Prompt the user as to whether to publish or not
4. Publish the package with `cargo publish` (no dry run)

It has a default set of crates it will publish, which can be overridden by
way of command line arguments:

```bash
# Release all packages, prompting for each package as to whether to publish
./release.sh

# Just release the proto and tendermint crates, but nothing else
./release.sh proto tendermint
```

Once it publishes a crate, it will create a file at `/tmp/tendermint-rs-release/${TODAY}/${CRATE}`, where `${TODAY}` is today's date and `${CRATE}` is the name of the crate that was successfully published.

Prior to publishing a crate, it checks whether this file is present before attempting to publish it. If it's present, it will ask if you really want to publish it again. Of course, this is pretty dumb, and doesn't cater for instances where multiple people could publish the crates on the same day, and instances where someone reboots their machine or wipes their `/tmp` folder between runs.

* [ ] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -e
+
+ARGS="$@"
+
+# A space-separated list of all the crates we want to publish, in the order in
+# which they must be published. It's important to respect this order, since
+# each subsequent crate depends on one or more of the preceding ones.
+DEFAULT_CRATES="proto tendermint rpc light-client light-node testgen"
+
+# Allows us to override the crates we want to publish.
+CRATES=${ARGS:-${DEFAULT_CRATES}}
+read -ra CRATES_ARR <<< "${CRATES}"
+
+publish() {
+  echo "Publishing crate $1..."
+  cargo publish --manifest-path "$1/Cargo.toml"
+  echo ""
+}
+
+publish_dry_run() {
+  echo "Attempting dry run of publishing crate $1..."
+  cargo publish --dry-run --manifest-path "$1/Cargo.toml"
+}
+
+list_package_files() {
+  cargo package --list --manifest-path "$1/Cargo.toml"
+}
+
+echo "Attempting to publish crate(s): ${CRATES}"
+
+for crate in "${CRATES_ARR[@]}"; do
+  publish_dry_run "${crate}"
+  list_package_files "${crate}"
+  echo ""
+  read -rp "Are you sure you want to publish \"${crate}\"? (type YES to publish, anything else to exit) " answer
+  case $answer in
+    YES ) publish "${crate}"; break;;
+    * ) echo "Terminating"; exit;;
+  esac
+done
--- a/release.sh
+++ b/release.sh
@@ -1,4 +1,36 @@
 #!/bin/bash
+
+# release.sh will hopefully allow us to publish all of the necessary crates in
+# this repo in the right order, along with a few checks and balances to try to
+# avoid mistakes. It is assumed that only one person will be releasing all
+# crates at the same time.
+#
+# For each crate, it will:
+# 1. Run `cargo publish --dry-run` for that crate
+# 2. List all files in the package with `cargo package --list`
+# 3. Prompt the user as to whether to publish or not
+# 4. Publish the package with `cargo publish` (no dry run)
+#
+# It has a default set of crates it will publish, which can be overridden by
+# way of command line arguments:
+#
+#   # Release all packages, prompting for each package as to whether to publish
+#   ./release.sh
+#
+#   # Just release the proto and tendermint crates, but nothing else
+#   ./release.sh proto tendermint
+#
+# Once it publishes a crate, it will create a file at
+# /tmp/tendermint-rs-release/${TODAY}/${CRATE}, where ${TODAY} is today's date
+# and ${CRATE} is the name of the crate that was successfully published.
+#
+# Prior to publishing a crate, it checks whether this file is present before
+# attempting to publish it. If it's present, it will ask if you really want to
+# publish it again. Of course, this is pretty dumb, and doesn't cater for
+# instances where multiple people could publish the crates on the same day, and
+# instances where someone reboots their machine or wipes their /tmp folder
+# between runs.
+
 set -e
 
 ARGS="$@"
--- a/release.sh
+++ b/release.sh
@@ -2,6 +2,9 @@
 set -e
 
 ARGS="$@"
+TODAY=$(date +%Y-%m-%d)
+TMP_RELEASE_FOLDER="/tmp/tendermint-rs-release/${TODAY}"
+mkdir -p "${TMP_RELEASE_FOLDER}"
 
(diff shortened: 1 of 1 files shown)


tony-iqlusion:
For some precedent, maybe check out:

- https://github.com/sunng87/cargo-release
- https://github.com/sagiegurari/cargo-make

release.sh
```diff
@@ -0,0 +1,88 @@
...
+# publish it again. Of course, this is pretty dumb, and doesn't cater for
+# instances where multiple people could publish the crates on the same day, and
+# instances where someone reboots their machine or wipes their /tmp folder
+# between runs.
+
+set -e
+
+ARGS="$@"
+TODAY=$(date +%Y-%m-%d)
+TMP_RELEASE_FOLDER="/tmp/tendermint-rs-release/${TODAY}"
```
melekes, line 38: alternative:

```suggestion
TMP_RELEASE_FOLDER="$(mktemp -d -t tendermint-rs-release)"
```

release.sh
```diff
@@ -0,0 +1,88 @@
...
+}
+
+echo "Attempting to publish crate(s): ${CRATES}"
+
+for crate in "${CRATES_ARR[@]}"; do
+  if [ -f "${TMP_RELEASE_FOLDER}/${crate}" ]; then
+    echo "Crate \"${crate}\" has already been published today."
+    read -rp "Do you want to publish again? (type YES to publish, anything else to skip) " answer
+    case $answer in
+      YES ) ;;
```
melekes, line 75:
```suggestion
      Y ) ;;
```

romac:
> For some precedent, maybe check out:
> 
> * https://github.com/sunng87/cargo-release
> * https://github.com/sagiegurari/cargo-make

One more: https://crates.io/crates/cargo-workspaces (written by and used in the [Clap](https://github.com/clap-rs/clap) project)

release.sh
greg-szabo: I'm also a fan of using `mktemp` for temporary folders, but in this case it takes away the feature of checking if there was a release already today. (A subsequent run would create a new temporary folder, never being able to check for older files.)

I'm more inclined to check on `crates.io` instead if a release has happened before, possibly using an API call. Which would also require a check in the source code for the version number of the crate. That can be done with `cargo metadata`.

I feel at this point this script is better suited for private use: it definitely shortens the release time but you need to be aware of our release process and a lot of assumptions have to be correct.

Probably I would like to see a script that does some analysis&print first. For example, use case:
- If I run the script without any parameters, it prints out all crates that are supposed to be published, their version numbers in the source code and a check if these version numbers are already published on crates.io. (Or possibly the latest version that's published on crates.io.)
- If I run the script with `--dry-run` or similar, it does the same and also runs the package list and publish with dry-run.
- If I run the script with `--publish`, it does the above and then publishes it.
- If I add package names on the command line, it filters down to those only.

I understand that adding all this might overcomplicate the script, so I think the most use we'd get from the first feature: gather and print as much info as possible. Even if we don't implement the actual publish or dry-run commands, it gives a much better handle on where we are in the publishing process.

With all that said, I'm happy to just use this script and continue improving this one instead. Just my two cents.

release.sh
greg-szabo: This would also require updating the printed text (YES -> Y). Just a minor thing.
In the sense of the Unix command-line, I would most probably make this an input parameter as well. Or even better, we had this idea with the rpc-probe script where the script accepts environment variables as input. They are easier to maintain in the script. For example:

If the script normally just prints, you can tell it to do the actual publishsing with:
```
PUBLISH=1 ./release.sh
```
If you want the dry-run but not the full publish, you can do:
```
VERBOSE=1 ./release.sh
```
(or `DRY_RUN=1` if you like that better)

greg-szabo:
Looking through Romain's links, I realized we should formalize our assumptions about the release process. Here are my extended assumptions, feel free to add/argue about them:

* The branch we publish from can be anything. (We have assumptions of `release/` branches but those do not need to be hard-coded for release.)
* The local source code is not dirty. All changes were committed.
* The local source code is already pushed or will be pushed to GitHub. (No check in place, just assumed.)
* The source code ran through testing, cargo fmt and clippy with success. (No check in place.)
* All version numbers are the same for packages to be released. (We can put a soft check in place with a warning maybe.)
* All html_root_url links were updated for documentation. (In tendermint, light-client, light-node and proto crates.) - This, I guess, is going to stay manual for a while.
* (The package versions do not exist on crates.io - this is checked by cargo.)
* After a successful release, the releaser will manually place a version on the Git source code and push it to GitHub. Also, possible merge the release to master, if it's on a separate branch.


So the script here really only need to do the actual process and possibly some checks around version numbers. The `cargo-make` project seems nice and with its toml configuration seems to have a standardized way of defining the release process, but it seems to be overkill for our needs right now. Most of the assumptions we make just makes the features of `cargo-make` irrelevant for the near term. (If we want nice, standardized and complex release process in the future, it might be worth using it.)

The `cargo-release` crate seems to have a smaller scope, but I'm a bit uncomfortable leaving some of the work to and external app: bumping version number, pushing to git, etc. Will it be able to update the html_root_urls too? Will it know what version we want to deploy? (rc or new release) Will it be able to automatically open my Yubikey to access my SSH key when it pushes to Git? (This should be a minor thing, but it's a problem with all apps, if I didn't use the key for a while.)

greg-szabo:
Apologies for rudely pushing to the branch but it seemed it's worth doing it here, instead of yet another PR pointing to this one.
I made a few changes:
* No more temporary folder use (slightly smaller footprint)
* Version checks are against crates.io
* Changed input to crate names instead of folder names (since we talk about crates everywhere)
* Slightly more verbose feedback (it tells you when a package was last published on crates.io)
* Slightly simpler code (removed arrays)
* New requirements: `jq`, `curl`. Apologies for that...

I didn't change the execution logic, because I expect Thane to be using this script the most for the foreseeable future and it fits his thinking. I would still check environment variables instead of reading input from stdin, but this is now just slicing hair.

I'm happy to approve the script as it is, but since I added code, I'd rather at least @thanethomson or others check first, what I changed.

greg-szabo (approved):
I'm going to quietly approve this PR so we can move forward with it. I think it's a useful tool during release, and I hope the additions I made are helping too.

thanethomson:
I'm merging this for now to help us in the short term, but have also opened #707 to implement a better, more robust solution in the short- to medium-term.

PR #686, opened 2020-11-19
Enable MBT on master

### LINKED ISSUE #567: Fix tests for v0.34 upgrade
Various JSON based tests have been disabled as we've been upgrading to v0.34. See #563 and others (?).

We should regenerate these tests with the new testgen tooling and otherwise work towards better documentation of whats being tested and how. 

As part of this work we should do a diligent sweep of all the JSON structs and read through #197 again to make sure we got everything :)

Shivani912:
This PR enables model based testing on master. Some tests had to be disabled as they fail to parse because of more restrictive data structure construction and this will be addressed in a follow up PR. All of the other tests were fixed simply by adding the new `total_voting_power` field to validator set in testing :)

And with this, I think we're in good shape to close issue #567
(closes #567 )
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


light-client/tests/model_based.rs
```diff
@@ -446,18 +447,33 @@ fn single_step_test(
     let trusting_period: Duration = tc.initial.trusting_period.into();
     for (i, input) in tc.input.iter().enumerate() {
         output_env.logln(&format!("    > step {}, expecting {:?}", i, input.verdict));
+
+        // ------------------->
+        // Below is a temporary work around to get rid of bug-gy validator sorting
```
melekes, line 452: Is there an issue for this sorting bug? Could you point us to any specific line where the bug occurs or you don't know?

Shivani912: No, but yes there should be one. I think the problem is somewhere with ser/de, it somehow changes the sequence of validators in the set which then produces a wrong valset hash. I'm not sure how this is happening. But I'm sure this is not coming from Testgen at least. Currently, these tests were disabled on master because of this and re-sorting seemed to be a quick solution. I need to dig further to know where this problem is coming from.

melekes (approved):
👍

Shivani912:
Thanks for reviewing @melekes :) I'll wait for one more approval from @romac or @andrey-kuprianov before merging!

light-client/tests/model_based.rs
Shivani912: Issue for this: #687

andrey-kuprianov: Ok, I think I know what is happening here. When we transform tests from Apalache counterexamples, the validators are in no particular order. At the same time new deserialization requires them to be sorted according to the power, and then lexicographically. Therefore the validation fails. One more point for implementing #665.

andrey-kuprianov (approved):
Looks good to me! Thanks, @Shivani912 for fixing the tests!

We need to rethink the model-based testing for Tendermint-rs in general, besides fixing the current issues, and automate it further, to reduce such manual work as much as possible. Let's address this in the near future.

Shivani912:
> Looks good to me! Thanks, @Shivani912 for fixing the tests!
> 
> We need to rethink the model-based testing for Tendermint-rs in general, besides fixing the current issues, and automate it further, to reduce such manual work as much as possible. Let's address this in the near future.

Awesome, that's exactly where my thoughts were going!

PR #696, opened 2020-11-20
p2p: migrate secret_connection from tmkms

### LINKED ISSUE #691: p2p: migrate secret_connection to new p2p crate
As we're starting to move towards a functioning seed node, the first practical step is to move [secret_connection](https://github.com/iqlusioninc/tmkms/tree/develop/src/connection) back into this repo. 

### Acceptance Criteria

* [x] bootstraped `p2p` crate
* [x] migrated `secret_connection` module from tmkms
* [x] landed update in tmkms with the new dependency

/cc @tony-iqlusion 

***

### Rationale for a new crate

To enable networked processes which can communicate via the p2p layer with other nodes in a heterogeneous tendermint installation, we would like to provide a crate. Which can be dependent on in isolation and will provide utility for for currently specified connections and provides the structure for future implementations. Originally envisioned in #7, with the advent the timing to bring it back into this repository seems fitting. And we've come full circle since #27. :>

melekes:
Refs #691

NOTE: this is a copy of the `secret_connection` from tmkms with `eyre` errors and without `generate_key` function.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


p2p/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name       = "p2p"
```
thanethomson, line 2: Should we not maybe call this `tendermint-p2p`? That'd fit with the naming convention of the other crates. Plus there's already a [p2p crate](https://crates.io/crates/p2p).

`p2p` is fine though for the directory name, of course.

p2p/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+prost = "0.6"
+prost-amino = "0.6"
+prost-amino-derive = "0.6"
+prost-derive = "0.6"
+rand_core = { version = "0.5", features = ["std"] }
+sha2 = "0.9"
+subtle = "2"
+subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
+tendermint = { version = "=0.17.0-rc3", features = ["secp256k1"] }
+tendermint-proto = "=0.17.0-rc3"
```
xla, line 33: Dependencies on crate in the same repository should be local relative `../<name>` and ideally bundled together for easier visibility.

xla (changes requested):
Dope! Had a first pass mostly for the crate setup. Will look deeper into the code in a second pass.

p2p/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+merlin = "2"
+prost = "0.6"
+prost-amino = "0.6"
+prost-amino-derive = "0.6"
+prost-derive = "0.6"
+rand_core = { version = "0.5", features = ["std"] }
+sha2 = "0.9"
+subtle = "2"
+subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
+tendermint = { version = "=0.17.0-rc3", features = ["secp256k1"] }
```
tony-iqlusion, line 32: You probably don't need the `secp256k1` feature

p2p/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+hkdf = "0.10.0"
+merlin = "2"
+prost = "0.6"
+prost-amino = "0.6"
+prost-amino-derive = "0.6"
+prost-derive = "0.6"
+rand_core = { version = "0.5", features = ["std"] }
+sha2 = "0.9"
+subtle = "2"
+subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
```
tony-iqlusion, line 31: The `bech32-preview` feature is probably unnecessary

p2p/src/secret_connection.rs
```diff
@@ -0,0 +1,420 @@
...
+        protocol_version: Version,
+    ) -> Result<SecretConnection<IoHandler>> {
+        let local_pubkey = PublicKey::from(local_privkey);
+
+        // Generate ephemeral keys for perfect forward secrecy.
+        let (local_eph_pubkey, local_eph_privkey) = gen_eph_keys();
+
+        // Write local ephemeral pubkey and receive one too.
+        // NOTE: every 32-byte string is accepted as a Curve25519 public key
+        // (see DJB's Curve25519 paper: http://cr.yp.to/ecdh/curve25519-20060209.pdf)
```
tony-iqlusion, line 71: This note is probably superfluous as the code in `protocol.rs` rejects low order points

p2p/src/secret_connection.rs
```diff
@@ -0,0 +1,420 @@
...
+
+        // Reject all-zero outputs from X25519 (i.e. from low-order points)
+        //
+        // See the following for information on potential attacks this check
+        // aids in mitigating:
+        //
+        // - https://github.com/tendermint/kms/issues/142
+        // - https://eprint.iacr.org/2019/526.pdf
+        if shared_secret.as_bytes().ct_eq(&[0x00; 32]).unwrap_u8() == 1 {
+            return Err(Error::InvalidKey).wrap_err("low-order points found");
```
tony-iqlusion, line 88: If this happens, it might be worth noting in the error message this is a potential MitM attack

p2p/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[dependencies]
+chacha20poly1305 = "0.7"
+ed25519-dalek = "1"
+eyre = "0.6"
+hkdf = "0.10.0"
+merlin = "2"
+prost = "0.6"
+prost-amino = "0.6"
+prost-amino-derive = "0.6"
```
tony-iqlusion, line 26: Not necessarily required for this PR, but you might consider adding an `amino` cargo feature and making these dependencies `optional`

ebuchman: Should we not be able to remove these altogether?

ebuchman: Right, are you trying to maintain a backwards compatible KMS, so one version can run on both pre and post stargate? I actually don't see how that would be possible since tendermint-rs isn't backwards compatible? I too would probably prefer to drop all the legacy support if possible.

p2p/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+//! p2p
+//!
+//! The Tendermint P2P stack.
+
+// Tip: Deny warnings with `RUSTFLAGS="-D warnings"` environment variable in CI
```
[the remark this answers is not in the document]
melekes: probably not. copied it from `light-node` crate.

p2p/Cargo.toml
melekes: I actually removed legacy versions in the first commit, but decided to bring them back in https://github.com/informalsystems/tendermint-rs/pull/696/commits/24b16ba928a145dfbf3d15ec8c167ee21d3ce420 exactly because TMKMS breakage.

melekes:
Thanks for the speedy feedback everyone 🚀

p2p/Cargo.toml
melekes: Should I use multiple locations https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations so crate can be published on crates.io? otherwise it may not be possible: "Note: crates.io does not allow packages to be published with path dependencies (path dev-dependencies are ignored)."

p2p/Cargo.toml
tony-iqlusion: @ebuchman TMKMS now vendors the old privval Amino message types in addition to using `tendermint`/ `tendermint-proto` v0.17.0-rc3 and is able to sign on all Tendermint versions.

ebuchman: Wow nice work!

p2p/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+//! The Tendermint P2P stack.
+
+#![forbid(unsafe_code)]
+#![deny(
+    warnings,
+    missing_docs,
```
xla, line 6: While definitely sensible to have, during ongoing development these impede quite a lot. A better strategy is to have these not be denies in here, but enforcing them in CI.

p2p/src/error.rs
```diff
@@ -0,0 +1,20 @@
+//! Error types
+
+pub use eyre::{Report, Result};
```
xla, line 3: Why are these re-exported here?

p2p/src/secret_connection.rs
```diff
@@ -0,0 +1,418 @@
+//! `SecretConnection`: Transport layer encryption for Tendermint P2P connections.
+
+mod amino_types;
```
xla, line 3: For me it's important to have some structured for the header of the rust files. Usually I apply the following order: 

* imports
  * std
  * third-party
  * first-party
  * crate
  * super
  * mod
* mods
  * declarations
  * exports

p2p/Cargo.toml
xla: Good call actually, will tmkms depend on the crates.io version? @tony-iqlusion

p2p/src/error.rs
melekes: I guess so others could use `crate::error::Report` instead of `eyre::Report`, but we're not doing it => removed. Thanks

p2p/Cargo.toml
tony-iqlusion: You can include a path dependency so long as there's also a version number.

TMKMS will need a crates.io release, yes.

melekes:
Made `amino` optional here.

xla (approved):
👛 🎣 ⛓ ✴️

PR #703, opened 2020-11-25
Added tendermint docker config

### LINKED ISSUE #702: Build our own docker images for integration testing in CI
We're relying on "latest" images of beta software for tendermint and we might also want gaia nodes in our CI testing. Best to have our own images so we supply the exact version we want to use.

greg-szabo:
Closes #702 .

Configuration to build tendermint docker images with default config. A few of these were built already with different versions of tendermint and uploaded to DockerHub here: https://hub.docker.com/repository/docker/informaldev/tendermint#

The integration tests were updated to run against a tendermint image close to "0.34.0-rc4" that was used in a previous Stargate testnet.

When upgrading Tendermint to be compatible with 034.0-rc5 (or later), we should also update these images to do proper integration testing.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,12 +94,11 @@ jobs:
           command: test
           args: --all-features --no-fail-fast -p tendermint-testgen
 
-  # Todo: Update the docker image version as soon as Tendermint Core v0.34 is released.
   tendermint-integration:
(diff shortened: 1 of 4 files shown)


melekes (approved):
👍

Issue #705, opened 2020-11-26
Specify the proposer in the validator set of fetched light blocks


romac:
This is required for IBC on-chain validation: https://github.com/informalsystems/ibc-rs/blob/d732aefe59787a95ef6df20d00f5c5e4ebbbb8a9/relayer/src/chain/cosmos.rs#L416-L436

PR #706, opened 2020-11-26
Specify proposer when constructing validator set in light client

romac:
Close: #705

This PR also refactors the constructors of `validator::Set` and precomputes the total voting power in the set.

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


ancazamfir (approved):
Looks good. Also tried it with some changes in `ibc-rs` and all is fine. Thanks!

tendermint/src/validator.rs
```diff
@@ -51,14 +49,25 @@ impl From<Set> for RawValidatorSet {
 }
 
 impl Set {
-    /// constructor
+    /// Constructor
     pub fn new(
         validators: Vec<Info>,
         proposer: Option<Info>,
-        total_voting_power: vote::Power,
+        total_voting_power: Option<vote::Power>,
```
melekes, line 56: why would someone need to pass `total_voting_power` if it can be computed?

light-client/src/components/io.rs
```diff
@@ -147,7 +153,11 @@ mod prod {
             }
         }
 
-        fn fetch_validator_set(&self, height: AtHeight) -> Result<TMValidatorSet, IoError> {
+        fn fetch_validator_set(
+            &self,
+            height: AtHeight,
+            proposer_address: Option<TMAccountId>,
```
melekes, line 159: TMAccountId may not be the best name for validator ID. accounts usually refer to users accounts who have wallets & transact within a network

tendermint/src/validator.rs
romac: Yeah I wonder about this as well. Maybe @ebuchman knows why it was done like this?

light-client/src/components/io.rs
romac: That's the name of the type currently used for the address field in `validator::Info`: https://github.com/informalsystems/tendermint-rs/blob/c5fe2a88c80d9336bd14a9bcbf575e7a62f3a93b/tendermint/src/validator.rs#L145-L146

Perhaps we need a new type for this?

melekes: actually, validators also receive a reward. please ignore

tendermint/src/validator.rs
ebuchman: they do not I've been campaigning against total voting power and proposer fields in this type for a few weeks now! I suspect it was done without really thinking about it ...

thanethomson (approved):
LGTM 👍

PR #708, opened 2020-11-27
p2p: transport agnostic peer abstractions

### LINKED ISSUE #692: p2p: design transport agnostic peer abstraction
To enable flexibility with regards to transport implementations - concerning layer4, encryption, handshake - a proper abstraction should be designed. Aiming for clear boundaries into the core domain and enabling node implementations built on top  of it to be future proof should a migration to other transports take place. A comparable effort is under way in the Go implementation of Tendermint in [ADR 62](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-062-p2p-architecture.md). 

### Important consideration

* single connection, multiplexed streams
* attached metadata for
  * capability tracking
  * potential hole punching
  * potential QoS and rate limiting

### Acceptance criteria

* [ ] peer-reviewed ADR

xla:
[Rendered ADR 009](https://github.com/informalsystems/tendermint-rs/blob/xla/692-transport-agnostic-peer/docs/architecture/adr-009-transport-agnostic-peer-abstraction.md)

Closes #692


p2p/src/transport.rs
```diff
@@ -0,0 +1,22 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {}
+
+pub trait Connection: Clone {
+    type Stream;
+
+    fn open(&self) -> Result<Self::Stream, Error>;
+}
+
+pub trait Endpoint {
```
melekes, line 10: why the endpoint initiates the connection? shouldn't it represent a IP/port or other remote/local address?

p2p/src/transport.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    fn connect(&self) -> Result<Self::Connection, Error>;
+}
+
+pub trait Transport {
+    type Connection: Connection;
+    type Endpoint: Endpoint<Connection = <Self as Transport>::Connection>;
+    type Incoming: Iterator<Item = Result<<Self as Transport>::Connection, Error>>;
+
+    fn bind(&self) -> Result<(Self::Endpoint, Self::Incoming), Error>;
```
melekes, line 21: is bind like a constructor? where it gets endpoint and incoming connections?

p2p/src/transport.rs
```diff
@@ -0,0 +1,22 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {}
+
+pub trait Connection: Clone {
+    type Stream;
+
+    fn open(&self) -> Result<Self::Stream, Error>;
```
melekes, line 7: what about closing the connection?

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,90 @@
...
+    incoming: I,
+}
+impl<E, I> State for Running<E, I>
+where
+    E: Endpoint,
+    I: Iterator<Item = Result<E::Connection, transport::Error>>,
+{
+}
+
+struct Supervisor<T, St>
```
melekes, line 28: how does supervisor learns about disconnects?

melekes (commented):
good start 👍 left some initial comments. we can discuss them in a meeting too

p2p/src/transport.rs
xla: It's a constructor that returns a working endpoint and the stream of incoming connections. Inspired by the syscall it is to indicate that after that call connections can be created and accepted.

p2p/src/transport.rs
xla: It does have a listen_addr, it represents the local binding returned by bind, so a working socket if you will.

p2p/src/transport.rs
xla: added.

p2p/src/supervisor.rs
xla: Valid question, needs to be determined.

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum Error {}
+
+pub trait Connection: Clone {
+    type Stream;
+
+    fn advertised_addrs(&self) -> Vec<SocketAddr>;
+    fn close(&self) -> Result<(), Error>;
+    fn open(&self) -> Result<Self::Stream, Error>;
+    fn peer_id(&self) -> PeerId;
```
melekes, line 23: don't think underlying transport should be aware of the peer abstraction. what necessitated this?

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+    type Connection;
+
+    fn connect(&self) -> Result<Self::Connection, Error>;
+    fn listen_addrs(&self) -> SocketAddr;
+}
+
+pub trait Transport {
+    type Connection: Connection;
+    type Endpoint: Endpoint<Connection = <Self as Transport>::Connection>;
+    type Incoming: Iterator<Item = Result<<Self as Transport>::Connection, Error>>;
```
melekes, line 37: what about outgoing connections?

p2p/src/peer.rs
```diff
@@ -0,0 +1,59 @@
...
+            _state: PhantomData,
+        }
+    }
+}
+
+impl<Conn> Iterator for Peer<Conn, Connected>
+where
+    Conn: Connection,
+{
+    type Item = Result<Message, Error>;
```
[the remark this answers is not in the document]
xla: The concept of an `id` should not leave the domain of the networking. For current connections they are used to byte prefix messages. Iff a stream-based transport is used, the streams are "typed" as in in the upgrade to a certain stream will depend on the stream handshake negotiating the messages communicated over it.

In this case `Message` is well typed and the only other information is origin which we have in the form of the `PeerId`. So inputs into the rest of the system and its protocols will be something like `(PeerId, Message)`. I assume the confusion stems from `Message` being an empty enum at this point. What I imagine it to look like is to encompass all possible domain variants. Assuming the node is running the following protocols: `Consensus` and`, `Pex` the set of messages could look like this:

```
enum Consensus {
    PreVote(..),
    Vote(..),
}

enum Pex {
    ReceiveAddrs(..),
    SendAddrs(..),
}

enum Message {
    Consensus(Consensus),
    Pex(Pex),
}
```

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub addr: SocketAddr,
+    pub advertise_addrs: Vec<SocketAddr>,
+    pub peer_id: PeerId,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {}
+
+pub trait Connection: Clone {
+    type Stream;
```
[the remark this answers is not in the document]
xla: Good observation, we likely want these constraints and I amended in aedfa44

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    fn advertised_addrs(&self) -> Vec<SocketAddr>;
```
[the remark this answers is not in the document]
xla: This is in preparation for use-cases where we support either more elaborate heuristics on the address book and enable hole punching, which depends on the sent package to carry the right public address at point of termination. I put in here to hint at that, we can take it out if it is confusing.

p2p/src/transport.rs
xla: The `PeerId` is not part of the peer abstraction, it's the essential identity of a node. My standing assumption is that we will not be able to transform the handshake into a protocol and have to leave that up-to the transport implementations. So they will need to have access to it to exchange it. Furthermore we likely want to enable early termination based on duplicates in connected IPs/PeerIds and down the line based on deny lists in address book implementations.

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn close(&self) -> Result<(), Error>;
+    fn open(&self) -> Result<Self::Stream, Error>;
+    fn peer_id(&self) -> PeerId;
+    fn remote_addr(&self) -> SocketAddr;
```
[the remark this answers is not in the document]
xla: Good point, added in 3c4c3cf

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+pub trait Endpoint {
+    type Connection;
+
+    fn connect(&self) -> Result<Self::Connection, Error>;
+    fn listen_addrs(&self) -> SocketAddr;
```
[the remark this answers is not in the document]
xla: Imagine we want to bind ot IPv4 and IPv6 at the same time, that will likely require two distinct network stacks. We can always argue that the `Supervisor` could run a distinct transport for either and then we just need a single `listen_addr`.

p2p/src/transport.rs
xla: We should definitely distinguish the direction of the connection. Maybe that is better done in the `Connection` type.

p2p/src/transport.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+pub trait Transport {
+    type Connection: Connection;
+    type Endpoint: Endpoint<Connection = <Self as Transport>::Connection>;
+    type Incoming: Iterator<Item = Result<<Self as Transport>::Connection, Error>>;
+
+    fn bind(&self, bind_info: BindInfo) -> Result<(Self::Endpoint, Self::Incoming), Error>;
```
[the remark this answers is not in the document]
xla: What would you expect to be specific per transport in the info?

p2p/src/transport.rs
xla: Expressed in b5b0643

p2p/src/transport.rs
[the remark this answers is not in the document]
melekes: probably nothing. sorry, I was thinking where transport configuration will occur (e.g. https://github.com/cloudflare/quiche/blob/e3ce6eb5bd2347d6b46d69ce9786a6132028e140/examples/http3-server.rs#L90)

p2p/src/transport.rs
melekes: > So they will need to have access to it to exchange it.

secret connection operates on public keys, not peer ids. or peer ids == public keys? if so, that's fine

xla: As we established given a `PublicKey` a `PeerId` can be derived. Should there be a need to get hold of that it can be computed ont he spot. So this should indeed be `public_key`. Thanks @melekes for the clarification.

p2p/src/peer.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+}
+
+impl<Conn> Peer<Connected<Conn>>
+where
+    Conn: Connection,
+{
+    pub fn run(self, stream_ids: Vec<StreamId>) -> Result<Peer<Running<Conn>>> {
+        // FIXME(xla): No queue should be unbounded, backpressure should be finley tuned and/or
```
melekes, line 69: is there a way for `Peer` to signal `Supervisor` that it must stop reading bytes from `Connection` (more than the peer can handle at the moment)?

p2p/src/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub trait Connection: Send {
+    type Error: 'static + fmt::Display + std::error::Error + Send + Sync;
+    type Read: Read;
+    type Write: Write;
+
+    fn advertised_addrs(&self) -> Vec<SocketAddr>;
+    fn close(&self) -> Result<()>;
+    fn local_addr(&self) -> SocketAddr;
+    fn open_bidirectional(
```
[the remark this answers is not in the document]
xla: Directionality here expressed which way data can flow, in a bidirectional scenario both sides can send and receive. Where a unidirectional stream would be only for sending on on side and receiving on the other. This is largely decoupled from the OSI layers used, as it is a logical separation, independent of the use of TCP, UDP, etc.

AFAIU `open` seems to be the most common terminology for streams. A close can usually be expressed by either dropping the handles, or observe an `EOF`.

p2p/src/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn connect(&self, info: ConnectInfo) -> Result<Self::Connection>;
+    fn listen_addrs(&self) -> SocketAddr;
+}
+
+pub trait Transport {
+    type Connection: Connection;
+    type Endpoint: Endpoint<Connection = <Self as Transport>::Connection>;
+    type Incoming: Iterator<Item = Result<<Self as Transport>::Connection>> + Send;
+
+    fn bind(&self, bind_info: BindInfo) -> Result<(Self::Endpoint, Self::Incoming)>;
```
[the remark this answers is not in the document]
xla: There will definitely be a memory transport. If that will work with  dummy IPs I'm not sure yet. In the end it could be true that the bindinfo is trasnport specific. Will leave it as an open point.

p2p/src/peer.rs
xla: Could you explain how the peer knows it cant handle more? As in how would it determine that?

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,241 @@
...
+                    for (id, peer) in &state.upgraded {
+                        selector = selector.recv(&peer.state.receiver, move |res| {
+                            Input::Receive(*id, res.unwrap())
+                        });
+                    }
+
+                    selector.wait()
+                };
+
+                for event in state.transition(input) {
```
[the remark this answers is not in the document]
xla: Flume has a feature flag called `eventual-fairness`, which is enabled by default. From the docs for `wait`:

> If the eventual-fairness feature flag is enabled, this method is fair and will handle a random event of those that are ready

p2p/src/peer.rs
melekes: because each peer has a queue (which at the moment is unbounded, but I suppose will be of a fixed size)

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+
+After the successful bind the caller holds an `Endpoint` as well as a stream of
+incoming `Connection`s. Which is a standardised way to connect to new peers and
+react to newly connected ones respectively.
+
+``` rust
+pub trait Endpoint: Send {
+    type Connection;
+
+    fn connect(&self, info: ConnectInfo) -> Result<Self::Connection>;
```
melekes, line 50: What goes into `ConnectInfo`? Is it transport specific?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+    fn listen_addrs(&self) -> SocketAddr;
```
melekes, line 51:
```suggestion
    fn listen_addrs(&self) -> Vec<SocketAddr>;
```

or 

```
    fn listen_addr(&self) -> SocketAddr;
```

melekes, line 51:
```suggestion
    fn listen_addrs(&self) -> Vec<SocketAddr>;
```

or 

```
    fn listen_addr(&self) -> SocketAddr;
```

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+}
+```
+
+Centerpiece of the whole shebang is the `Connection`. It represents a connected
+peer and provides the primitives to get data and send data from a peer. It is
+designed with the outlook to support stream based trasnports down the road.
```
melekes, line 57:
```suggestion
designed with the outlook to support stream based transports down the road.
```

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+While being open to enable feature parity with current production installations
+based on tendermint-go's `MConn`.
+
+``` rust
+pub trait Connection: Send {
+    type Error: 'static + fmt::Display + std::error::Error + Send + Sync;
```
melekes, line 63: Is it necessary for the `Error` to have a static lifetime?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+``` rust
+impl<Conn> Peer<Connected<Conn>>
+where
+    Conn: Connection,
+{
+    pub fn run(self, stream_ids: Vec<StreamId>) -> Result<Peer<Running<Conn>>> {
+        // ...
+    }
+
+    fn stop(self) -> Result<Peer<Stopped>> {
```
melekes, line 100: why it's possible to stop `Connected` (not `Running`) peer?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+        // ...
+    }
+}
+
+impl<Conn> Peer<Running<Conn>>
+where
+    Conn: Connection,
+{
+    pub fn send(&self, message: message::Send) -> Result<()> {
+        // ...
```
melekes, line 110: do I understand correctly that routing (mapping message::Send type onto a certain stream) happens here?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
melekes, line 63: Is it necessary for the `Error` to have a static lifetime?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+
+* Significant set of abstractions need to be satisfied for a new transport
+  implementation
+* Non-stream based transports need to be fitted into this model
+
+### Neutral
+
+## Open Questions
+
+* Should serialization be transport specific to allow for optimisations on the
```
melekes, line 186: NOTE: tendermint-go has decided on unified serialization (protobuf) across all transports for the sake of simplicity.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+
+Proposed
+
+## Consequences
+
+### Positive
+
+* Unified way to bootstrap and integrate transports
+* Potential for connecting different wire transports in the same process
+* Rest of the domain is simply concerned with `PeerId`s as identity
```
melekes, line 174:
```suggestion
* Rest of the domain is simply concerned with `node::Id`s as identity
```

brapse:
@erikgrinaker brought up some interesting points in Slack which I would copy and past here with permission or perhaps he could chime in here directly.

p2p/src/peer.rs
xla: Good to raise this early. Intuitively this would necessitate to make backpressue a first-class concept, which protocols that rely on it have to handle. Particularly if a send to a peer fails it should generate a new event which is passed back to the rest of the system. Which in turn could be used to decide if certain/all messages are being held or dropped. A major challenge here is that the message that experiences backpressure might not be the offending one as the message streams to peers are heterogenous. It could be more specific as in which one of the streams inside of a peer is slow to push to. Open to suggestions of how to incorporate it in the rest of the system.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: Both `Connected` and `Running` have an associated `stop` method. :)

xla: I forgot to include it in the ADR, the code has it.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: I'd be surprise if we can get away without it as we likely we want to pass it around for tracing/introspection. We can omit it in the ADR.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: Actually it's in there. See above.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: That's exactly right, the message multiplexing over the streams is responsibility of the peer implementation. Maintaining this across transports seems an important guarantee to provide.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: Great, thanks for the context. If this is indeed the vision @brapse @greg-szabo @erikgrinaker? We should stick to the plan that connections provide byte streams and we maintain a single implementation of the serialisation.

Personally I would benefit from a brush up on how currently the byte layout looks on the wire with TCP connections. Are we still byte-prefixing to determine the `channel`?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
melekes: > Both Connected and Running have an associated stop method. :)

yeah, so my question is why it's possible to stop a non-running peer? it seems weird to me

xla: Ah right, I miss-read your question. From my perspective a `Connected` peer already holds onto some resources (i.e. the connection)  and maybe there is a valid use-case where it shouldn't be transitioned into running state but dropped before.
While we can rely on the `Drop` semantics alone I wanted a transition in the terminal `Stopped` state for both "starting" states.

erikgrinaker:
> @erikgrinaker brought up some interesting points in Slack which I would copy and past here with permission

Sure, feel free to include them here.

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,279 @@
...
+            let conn = upgrade_rx.recv().unwrap();
+            let peer = peer::Peer::from(conn);
+
+            upgraded_tx
+                .send(Input::Upgraded(peer.id, peer.run(vec![])))
+                .unwrap();
+        });
+
+        // MAIN
+        thread::spawn(move || {
```
thanethomson, line 135: I think I understand what you're achieving here with spawning the main supervisor loop as a separate thread and then returning the supervisor instance (i.e. so you can send it commands from the initiating thread), but how does the owner know if the supervisor's main thread has panicked and is no longer available? (e.g. there are a bunch of `unwrap`s in the loop)

thanethomson (commented):
Had a first pass at reviewing this, and I really like the simplicity of the abstractions! Makes a lot of sense :+1: 

What're the plans for documentation?

p2p/src/supervisor.rs
romac: Perhaps we could refactor this so that the `run` method returns a `Result` and leave it to the caller to spawn a thread for the supervisor itself, much like we do [in the light client](https://github.com/informalsystems/tendermint-rs/blob/3c7754b98e09d965973bf6580d48dcce9216e022/light-client/src/supervisor.rs#L330).

This would require splitting the `Supervisor` struct into an actual `Supervisor` with a `run` method, and a clonable `SupervisorHandle` which can then be use to send commands to the supervisor.

```rust
impl Supervisor {
  // option 1
  pub fn new() -> Self;
  pub fn handle(&self) -> SupervisorHandle;
  // option 2
  pub fn new() -> (Self, SupervisorHandle);

  // common
  pub fn run(self) -> Result<(), SupervisorError>;
}

#[derive(Clone)]
struct SupervisorHandle {
    command: Sender<Command>,
    events: Receiver<Event>,
}
```

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+        };
+
+        let id = match pk {
+            PublicKey::Ed25519(ed25519) => node::Id::from(ed25519),
+            _ => panic!(),
+        };
+
+        Peer {
+            id,
+
```
romac, line 61: nitpick: Extraneous whitespace.

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+{
+    fn from(connection: Direction<Conn>) -> Peer<Connected<Conn>> {
+        let pk = match &connection {
+            Direction::Incoming(conn) => conn.public_key(),
+            Direction::Outgoing(conn) => conn.public_key(),
+        };
+
+        let id = match pk {
+            PublicKey::Ed25519(ed25519) => node::Id::from(ed25519),
+            _ => panic!(),
```
romac, line 56: We should return an error and thus use `TryFrom` if the conversion can fail.

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+    error: Option<Error>,
+}
+impl State for Stopped {}
+
+pub struct Peer<St>
+where
+    St: private::Sealed,
+{
+    pub id: node::Id,
+
```
romac, line 40: nitpick: Extraneous whitespace.

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn stop(self) -> Result<Peer<Stopped>> {
+        match self.state.connection {
+            Direction::Incoming(conn) | Direction::Outgoing(conn) => conn.close()?,
+        }
+
+        Ok(Peer {
+            id: self.id,
+
```
romac, line 143: nitpick: Extraneous whitespace.

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    pub fn stop(self) -> Result<Peer<Stopped>> {
+        match self.state.connection {
+            Direction::Incoming(conn) | Direction::Outgoing(conn) => conn.close()?,
+        }
+
+        Ok(Peer {
+            id: self.id,
+
```
romac, line 165: nitpick: Extraneous whitespace.

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,279 @@
...
+        let (endpoint, mut incoming) = transport.bind(transport::BindInfo {
+            addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 12345),
+            advertise_addrs: vec![SocketAddr::new(
+                IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
+                12345,
+            )],
+            public_key: PublicKey::from_raw_ed25519(&[
+                215, 90, 152, 1, 130, 177, 10, 183, 213, 75, 254, 211, 201, 100, 7, 58, 14, 225,
+                114, 243, 218, 166, 35, 37, 175, 2, 26, 104, 247, 7, 81, 26,
+            ])
```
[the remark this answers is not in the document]
romac: Those could be given to the `new` method I describe in my [later comment](https://github.com/informalsystems/tendermint-rs/pull/708/files#r582755198).

p2p/src/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+    Pex,
+}
+
+pub enum Direction<Conn> {
+    Incoming(Conn),
+    Outgoing(Conn),
+}
+
+pub trait Connection: Send {
+    type Error: 'static + fmt::Display + std::error::Error + Send + Sync;
```
romac, line 28: The `fmt::Display` bound is redundant here as it's [already required](https://doc.rust-lang.org/stable/std/error/trait.Error.html) by the `std::error::Error` trait.

romac (changes requested):
Aside from the few comments inline, we should also fix the various clippy warnings and add some doc comments.

Otherwise, this all looks really good! Very nice fit for the typestate pattern 💯

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+impl<Conn> Peer<Running<Conn>>
+where
+    Conn: Connection,
+{
+    pub fn send(&self, message: message::Send) -> Result<()> {
+        // TODO(xla): Map message to stream id.
+        todo!()
```
romac, line 155: Reminder that this `todo!` should be addressed.

p2p/src/supervisor.rs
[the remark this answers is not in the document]
xla: > Are these parameters going to be loaded from somewhere/configurable?

It will become a parameter in the construction of the supervisor.

p2p/src/supervisor.rs
xla: The extra indirection of the handle does not add much value here as the returned `Supervisor` acts as such. The `uwrap`s will be refactored out and the public interface will indicate if the internal machinery dissolved/imploded.

Beyond that there doesn't seem to be some gain in having extra lifecycle control on the caller side, a transport is already passed in with the intent to be bound and managed. It is questionable what could happen between the construction and explicit call of `run` that wold be meaningful. In turn it could lead to unexpected behaviour, when the run is never spawned and the transport is bound and its resources are locked up.

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+
+pub struct Running<Conn> {
+    connection: Direction<Conn>,
+    pub receiver: Receiver<message::Receive>,
+    senders: HashMap<StreamId, Sender<Box<dyn message::Outgoing>>>,
+}
+impl<Conn> State for Running<Conn> {}
+
+pub struct Stopped {
+    error: Option<Error>,
```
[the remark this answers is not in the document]
xla: > Does this not actually represent two possible states: failed (with error), and stopped (without error)?

In both cases the peer is stopped, in my mental model it now can have a reason why it stopped. Alternatively encoding a reason that is non-optional could encompass all possible reasons (even non-unexpected ones) why the peer transitioned into the `Stopped` state.

> Might also be useful to expose the error publicly so it's accessible by the peer's owner.

Great point.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+After the successful bind the caller holds an `Endpoint` as well as a stream of
+incoming `Connection`s. Which is a standardised way to connect to new peers and
+react to newly connected ones respectively.
+
+``` rust
+pub trait Endpoint: Send {
+    type Connection;
+
+    fn connect(&self, info: ConnectInfo) -> Result<Self::Connection>;
+    fn listen_addrs(&self) -> Vec<SocketAddr>;
```
melekes, line 51: 1) can we add a comment here and for all other functions as well?

2) can we move this into `Incoming`? it makes more sense since the iterator (TcpListener.incoming() in case of MConnection) is being used there.

p2p/src/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::net::SocketAddr;
+
+use tendermint::public_key::PublicKey;
+
+use eyre::Result;
+
+pub struct BindInfo {
+    pub addr: SocketAddr,
+    pub advertise_addrs: Vec<SocketAddr>,
+    pub public_key: PublicKey,
```
melekes, line 12: I don't think public_key is enough. `SecretConnection` needs `private_key`

xla: Which type would you suggest to use here?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: > can we add a comment here and for all other functions as well?

Will be addressed together with all other clippy warnings, once it's final.

> can we move this into `Incoming`? it makes more sense since the iterator (TcpListener.incoming() in case of MConnection) is being used there.

Not quite getting the meaning here. `Incoming` is an iterator, not a trait that is implemented.

p2p/src/message.rs
```diff
@@ -0,0 +1,16 @@
+pub enum Receive {
```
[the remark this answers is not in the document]
xla: https://github.com/informalsystems/tendermint-rs/pull/708#discussion_r591355073

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+            };
+
+            {
+                let tx = receive_tx.clone();
+
+                thread::spawn(move || {
+                    loop {
+                        // Read bytes
+                        // Deserialize into typed message
+                        // send on receiver_send
```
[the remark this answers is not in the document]
xla: They are left as todos to resemble the order of steps, still need to be implemented, there as a reminder and for the reviewer to understand what is happening here.

p2p/src/peer.rs
```diff
@@ -0,0 +1,182 @@
...
+
+                    // The only error possible is a disconnection.
+                    if res.is_err() {
+                        break;
+                    }
+
+                    let msg = res.unwrap();
+
+                    // Serialise message
+                    // write bytes
```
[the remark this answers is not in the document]
xla: https://github.com/informalsystems/tendermint-rs/pull/708#discussion_r581891250

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
xla: I wouldn't assume so. It would need some `node::Id`, `ip:port`/`dns endpoint` afaict.

p2p/src/transport.rs
melekes:
```rust
    pub private_key: ed25519::Keypair,
```

?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
melekes: > Not quite getting the meaning here. Incoming is an iterator, not a trait that is implemented.

please ignore

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+    type Error: std::error::Error + Send + Sync + 'static;
+    type Read: Read;
+    type Write: Write;
+
+    fn advertised_addrs(&self) -> Vec<SocketAddr>;
+    fn close(&self) -> Result<()>;
+    fn local_addr(&self) -> SocketAddr;
+    fn open_bidirectional(
+        &self,
+        stream_id: &StreamId,
```
melekes, line 72: does this have to be a pointer?

xla: Not necessarily, we can likely have `Copy` on `StreamId`, it's just good practice if the caller doesn't consume it to just pass a reference. Does it give you trouble in the implementation?

melekes: I thought for simple types (int, uint, etc.) it is better to copy. It's certainly a case for Go https://twitter.com/akaliaev/status/1370274749545320448

p2p/src/peer.rs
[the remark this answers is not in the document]
thanethomson: When would this need to be implemented?

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,418 @@
...
+    /// (as far as applicable for the transport) so the caller can use the command input and
+    /// consume events.
+    ///
+    /// # Errors
+    ///
+    /// * If the bind of the transport fails
+    #[allow(clippy::too_many_lines)]
+    pub fn run<T>(transport: &T) -> Result<Self>
+    where
+        T: transport::Transport + Send + 'static,
```
melekes, line 120: why `T` (transport) has a static lifetime?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,194 @@
...
+    ) -> Result<(Self::Read, Self::Write), Self::Error>;
+    fn public_key(&self) -> PublicKey;
+    fn remote_addr(&self) -> SocketAddr;
```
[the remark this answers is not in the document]
xla: For the closed case we should be able to return the information from when the connection was running. Is there ever the case where we have a connection that is not open yet?

p2p/src/supervisor.rs
xla: Could you provide more context about what makes this lifetime so surprising?

romac: I am not fully sure what happens there but it seems that, while `transport` itself is not explicitly moved between threads, some values of type `<T as Transport>::Endpoint` and `<T as Transport>::Connection` are. Rust apparently requires a `'static` bound on `T` for the types above to be considered `'static` as well. Not fully clear to me why that is.

On the other hand, the `Send` bound does not seem like it's needed (ie. one can remove it and the code still compiles).

Another way of working around that, would be to remove the `'static` bound here, and add it to the associated types of `Transport`:

```rust
pub trait Transport {
    type Connection: Connection + 'static;
    type Endpoint: Endpoint<Connection = Self::Connection> + 'static;
    type Incoming: Iterator<Item = Result<Self::Connection>> + Send + 'static;
```

but I am not sure which approach is more modular.

---

Just in case, some background for why the `'static` bound is needed for moving values of a generic types between threads: https://users.rust-lang.org/t/why-does-thread-spawn-need-static-lifetime-for-generic-bounds/4541/2

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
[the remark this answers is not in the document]
melekes: Right, I was just looking at `TcpStream` `local_addr` and `peer_addr` methods that both return `Result<SocketAddr>` (rather than `SocketAddr`) even though it's also not possible to construct a non-connected stream. 

At the moment, this is how `local_addr` implemented:

```rust
    fn local_addr(&self) -> SocketAddr {
        self.stream
            .local_addr()
            .unwrap()
    }
```

if there's an error, the program will panic, which is not good.

p2p/src/supervisor.rs
xla: Besides @romac alternative another version would be:

``` rust
    pub fn run<T>(transport: &T, info: transport::BindInfo) -> Result<Self>
    where
        T: transport::Transport + Send,
        T::Connection: 'static,
        T::Endpoint: 'static,
        T::Incoming: 'static,
```

as I generally prefer lifetime bounds to the latest possible moment. Yet at that point it doesn't really make a difference from constraining `Transport` directly.

> yeah, so from my newbie perspective `'static` lifetime should be applied sparingly. Does the transport really have to live for the duration of the entire program? I guess it's fine as only a few transports will be created, but still if it's possible to avoid `'static` I'd remove it.

Glad you bring that up as this is a common misconception. Personally I greatly improved my intuition of lifetimes by reading the sections in the Rustonomicon: https://doc.rust-lang.org/nomicon/lifetimes.html and this great piece: https://github.com/pretzelhammer/rust-blog/blob/master/posts/common-rust-lifetime-misconceptions.md

In this case we could change the lifetime to be named, but that would will trickle through the entire assortment of types and will need to be carried almost everywhere. satisfied.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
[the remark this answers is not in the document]
xla: It doesn't need to be done that way, right? We can also memorise this information when we construct the connection and return the stored value whenever the method is called. Or am I missing an important detail here? In my understanding the address shouldn't change over the lifetime of the conn.

p2p/src/supervisor.rs
romac: > yeah, so from my newbie perspective `'static` lifetime should be applied sparingly. Does the transport really have to live for the duration of the entire program?

That's not what a `'static` lifetime _bound_ means. There is a difference between a `'static` _lifetime_ and a `'static` _lifetime bound_. The former indeed means that the reference is valid for the entire duration of the program, while the latter means that the type bounded by it does not contain any reference which has a lifetime less than `'static`.

For example, the following types satisfy the `'static` lifetime bound:

```rust
u32
struct Foo(u32);
String
enum { Bar(String), Baz(bool) }
struct Test { x: &'static Foo }
```

On the other hand, this type will not satisfy the `'static` bound unless `'a == 'static`:

```rust
struct Abc<'a> {
  x: &'a i32
}
```

eg.

```rust
fn foo<'a>(x: Abc<'a>) {
  // Abc<'a> does not satisfy the 'static bound because we don't know anything about 'a (it's universally quantified)
}

fn bar(x: Abc<'static>) {
  // Abc<'static> satisfies the 'static bound because the reference it contains is 'static
}
````

Hope that clears things up :)

Comment (romac) [reply p2p/src/supervisor.rs:119]: Thanks @xla for the link! This section in particular explains this stuff in a much better way than I did above: https://github.com/pretzelhammer/rust-blog/blob/master/posts/common-rust-lifetime-misconceptions.md#2-if-t-static-then-t-must-be-valid-for-the-entire-program

Comment (romac) [reply p2p/src/supervisor.rs:119]: > Besides @romac alternative another version would be:
> 
> ```rust
>     pub fn run<T>(transport: &T, info: transport::BindInfo) -> Result<Self>
>     where
>         T: transport::Transport + Send,
>         T::Connection: 'static,
>         T::Endpoint: 'static,
>         T::Incoming: 'static,
> ```
> 
> as I generally prefer lifetime bounds to the latest possible moment. Yet at that point it doesn't really make a difference from constraining `Transport` directly.

Agreed, that sounds like the best solution to me.

**Comment by thanethomson** [comment]
Had another pass and made some minor recommendations in #880.

When I run the following using the code from #880 (after fixing the initial clippy errors):

```bash
cargo clippy --all-features --all-targets -- -Dwarnings -Drust-2018-idioms
```

I get 166 errors and 1 warning. It seems as though this is largely due to unexported types. There are, however, 2 errors that stand out that say to me that there still needs to be a little more clarity in the API boundaries in the supervisor's types:

```
error: private type `supervisor::Input` in public interface (error E0446)
  --> p2p/src/supervisor/protocol.rs:19:5
   |
19 |     pub fn transition(&mut self, input: Input) -> Vec<Output> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D private-in-public` implied by `-D warnings`
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #34537 <https://github.com/rust-lang/rust/issues/34537>
```

and similarly for

```
error: private type `supervisor::Output` in public interface (error E0446)
  --> p2p/src/supervisor/protocol.rs:19:5
   |
19 |     pub fn transition(&mut self, input: Input) -> Vec<Output> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #34537 <https://github.com/rust-lang/rust/issues/34537>
```

Making these types public has a viral effect, necessitating a bunch of types (that look like they're meant to be internal ones to the crate) end up needing to be public too.

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,517 @@
...
+            let handle = thread::Builder::new()
+                .name(name.clone())
+                .spawn(move || Self::upgrade::<T>(connected, input_tx, peers, upgrade_rx))?;
+
+            handles.push((name, handle));
+        }
+
+        let mut protocol = Protocol::default();
+
+        loop {
```
thanethomson, line 241: Am I correct in saying that this is the core event loop of the `Supervisor`, and it handles message/event coordination across all worker threads and peers?

Also, similar to my earlier question, what happens if one of the worker threads die unexpectedly during this event loop's execution?

thanethomson (commented):
I've given the latest changes another pass, with specific emphasis on trying to improve my understanding of the `Supervisor` and its possible failure modes.

Cargo.toml
```diff
@@ -5,11 +5,12 @@ members = [
     "light-client",
     "light-client-js",
     "p2p",
+    "pbt-gen",
     "proto",
     "rpc",
     "tendermint",
-    "testgen",
-    "pbt-gen"
+    "test",
```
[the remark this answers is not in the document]
xla: > Is the idea here to have a dedicated crate for cross-crate integration testing?

That's one of the use-cases for this crate.

> It might be good to add a `README.md` to the `test` crate to explain what it's for.

Good shout.

test/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#[cfg(test)]
+mod pipe;
+
+#[cfg(test)]
+mod test;
```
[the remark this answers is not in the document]
xla: All modules outside of `test` are utilities that can be used when testing tendermint code, so possibly even external to this repo. Everything under `test` is for execution inside of this repo while avoiding bloat in other crates, esp. when it comes to cross-crate integration testing.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+strongly correspond to the acquisition and lifecycle management of network
+resources on the system.
+
+``` rust
+pub trait Transport {
+    type Connection: Connection;
+    type Endpoint: Endpoint<Connection = <Self as Transport>::Connection>;
+    type Incoming: Iterator<Item = Result<<Self as Transport>::Connection>> + Send;
+
+    fn bind(&self, bind_info: BindInfo) -> Result<(Self::Endpoint, Self::Incoming)>;
```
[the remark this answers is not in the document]
xla: Indeed shutdown is pointless as the transport itself is consumed. Shutdown semantics should happen when the endpoint is dropped. Updated the ADR and the code.

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+    fn shutdown(&self) -> Result<()>;
```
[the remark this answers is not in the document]
xla: https://github.com/informalsystems/tendermint-rs/pull/708#discussion_r648122718

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,517 @@
...
+type Connected<T> =
+    Arc<Mutex<HashMap<node::Id, transport::Direction<<T as transport::Transport>::Connection>>>>;
+type Peers<T> = Arc<
+    Mutex<HashMap<node::Id, peer::Peer<peer::Running<<T as transport::Transport>::Connection>>>>,
+>;
+
+#[allow(clippy::needless_pass_by_value)]
+impl Supervisor {
+    #[allow(clippy::too_many_lines)]
+    fn main<T>(
```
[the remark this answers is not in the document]
xla: Do you have ideas of how to refactor it meaningfully to reduce the body size?

docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+
+Centerpiece of the whole shebang is the `Connection`. It represents a connected
+peer and provides the primitives to get data and send data from a peer. It is
+designed with the outlook to support stream based transports down the road.
+While being open to enable feature parity with current production installations
+based on tendermint-go's `MConn`.
+
+``` rust
+pub trait StreamSend {
+    fn send(msg: Vec<u8>) -> Result<()>;
```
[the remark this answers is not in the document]
xla: Until we see strong reasons to go with `Bytes` the generic approach will suffice. Updated ADR and code.

p2p/src/supervisor.rs
```diff
@@ -0,0 +1,517 @@
...
+        T: transport::Transport + Send + 'static,
+    {
+        let (endpoint, incoming) = transport.bind(info)?;
+        let (command_tx, command_rx) = unbounded();
+        let (event_tx, event_rx) = unbounded();
+
+        thread::spawn(move || {
+            match Self::main::<T>(command_rx, event_tx.clone(), endpoint, incoming) {
+                Ok(handles) => {
+                    for (_name, handle) in handles {
```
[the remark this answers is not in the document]
xla: > Does that ordering matter?

Ordering does not matter, unless you are seeing a case where it would.

> What happens when one of the threads beyond the `supervisor-accept` (first) thread fail? e.g. if the `supervisor-message` thread fails, but the `supervisor-accept` thread continues running, won't this effectively leave the supervisor in a partially degraded/invalid state of operation, where it can accept incoming peer connections but can't handle messages from them?

All subroutines have two channels, one which they read from to initiate the next operation (including accept) and one to send the result of the operation. If one of them fails the supervisor run loop is terminated - which in turn means that all Senders are dropped and therefore the receive errors. Symmetrically if one of the subroutiens fails (the code is crafted in a way that all side-effects are accounted for so panics in subroutines should be eliminated) the Sender will error on the next operation. With this approach there will be convergence onto the actual state of affairs, there is room for a single in-flight operation to be underway before the machinery is dismantled, which in my view is a worthwhile trade-off as long as the entire system converges towards the true state. Super important to surface if we see scenarios where this can leave the p2p stack in a partially degraded state.

p2p/src/supervisor.rs
xla: > Am I correct in saying that this is the core event loop of the `Supervisor`, and it handles message/event coordination across all worker threads and peers (i.e. peer threads)?

Yes!



> Also, similar to my earlier question, what happens if one of the worker threads die unexpectedly during this event loop's execution?

See https://github.com/informalsystems/tendermint-rs/pull/708#discussion_r648137076

xla:
Superseded by #904

PR #718, opened 2020-12-01
Automatically de/serialize ABCI event attributes from/to base64

### LINKED ISSUE #717: abci Tag deserializer
`DeliverTx` and rpc block_result `Response` include events in the form `Vec<Event>`, where `Event` is defined [here](https://github.com/informalsystems/tendermint-rs/blob/25f7ba5e99726375140d0d6c84f067dd4d23b03d/tendermint/src/abci/responses.rs#L12=L31) like this:
```
/// Event
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct Event {
    /// Event type
    #[serde(rename = "type")]
    pub type_str: String,

    /// Attributes
    pub attributes: Vec<Tag>,
}
```
Where `Tag` is defined [here](https://github.com/informalsystems/tendermint-rs/blob/25f7ba5e99726375140d0d6c84f067dd4d23b03d/tendermint/src/abci/tag.rs#L9-L15)
The `Tag` fields are base64 encoded but we do not `decode` when deserializing them, so the the `Event` looks something like:

```Event {
    type_str: "send_packet",
    attributes: [
        Tag {
            key: Key(
                "cGFja2V0X3RpbWVvdXRfaGVpZ2h0",
            ),
            value: Value(
                "MC00ODQw",
            ),
        },
...
```

instead of:
```
Event {
    type_str: "send_packet",
    attributes: [
        Tag {
            key: Key(
                "packet_timeout_height",
            ),
            value: Value(
                "0-4840",
            ),
        },
...
```
In IBC we need to extract some info from these tags (in particular the ones in `DeliverTx`) so we currently have to do something like:
```
let key = String::from_utf8(base64::decode(tag.key.to_string().as_bytes())?)?;
```

I think this should be done in `tendermint-rs` instead.

thanethomson:
Closes #717 

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,16 @@
 - `[rpc]` The `WebSocketClient` now adds support for all remaining RPC requests
   by way of implementing the `Client` trait ([#646])
 - `[rpc]` Support for the `tx_search` RPC endpoint has been added ([#701])
+- `[rpc]` Responses that include events now automatically have their tag
+  key/value pairs decoded from base64, where previously tag key/value pairs
+  were Base64-encoded ([#717])
 - `[tendermint]` (Since v0.17.0-rc3) Bech32 encoding fix ([#690])
 
 [#425]: https://github.com/informalsystems/tendermint-rs/issues/425
 [#646]: https://github.com/informalsystems/tendermint-rs/pull/646
 [#690]: https://github.com/informalsystems/tendermint-rs/issues/690
 [#701]: https://github.com/informalsystems/tendermint-rs/pull/701
+[#717]: https://github.com/informalsystems/tendermint-rs/issues/717
 
 ## v0.17.0-rc3
 
--- a/proto/src/serializers/bytes.rs
+++ b/proto/src/serializers/bytes.rs
@@ -42,6 +42,16 @@ pub mod base64string {
         base64::decode(&string).map_err(serde::de::Error::custom)
     }
 
+    /// Deserialize base64string into String
+    pub fn deserialize_to_string<'de, D>(deserializer: D) -> Result<String, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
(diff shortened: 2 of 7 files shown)


greg-szabo (commented):
The code is technically correct and I like that the serialization is done using serde attributes (instead of custom derives) but I have some reservations about where the serialization module is located.

While trying to separate serialization from domain types, there were a few places where serialization code was gathered:

The `tendermint_proto::serializers` module contains most of the "raw" serializer/deserializer code in a reusable format. Things like serializing from a string that is supposed to be an integer, etc.
The `tendermint::serializers` module contains serializers that have specific validation requirements on the Tendermint domain-type level: for example a `Hash` is just a hexstring (serializer implemented in `tendermint_proto`) but it also has a length requirement that is validated in the domain type, hence the serializer needs to call that validation (in the form of a `TryFrom` trait in this case, which is implemented in `tendermint`).

Technically the `base64string` module would belong into `tendermint_proto` since it doesn't have any domain-type specific restrictions. I understand that this sounds a bit weird, since nothing in `tendermint-proto` is actually using the module.

So I leave it to you to decide if it's worth putting it there or keep it in `tendermint`, but I would really like to move it into one of the `serializers` modules, both for reusability and to keep the idea of a slow but constant separation of serialization from domain types.

thanethomson:
> So I leave it to you to decide if it's worth putting it there or keep it in tendermint, but I would really like to move it into one of the serializers modules, both for reusability and to keep the idea of a slow but constant separation of serialization from domain types.

Cool, I've moved just the deserializer into the `tendermint-proto` crate for now, since the serializer works for `String`s. If we need a completely separate serialize/deserialize method then I can abstract it out into a separate module in the `bytes.rs` file, but it doesn't seem necessary right now.

melekes (approved):
👍

Issue #661, opened 2020-11-06
Domain type constructors and getters


greg-szabo:
With the introduction of domain type restrictions, we need all domain types to be able to construct valid structures using a `new` method and we need the domain type fields hidden, so an invalid struct cannot be created.
We also need getters for the fields that were hidden.

Example implementation in `SignedHeader`.

thanethomson:
One of the challenges with the approach of providing a constructor that validates its input is that our more complex domain types rely on other domain types (e.g. the `Block` domain type needs a `Header`, `transaction::Data`, `evidence::Data` and an `Option<Commit>` to construct). One could simply continue to use the current constructor, which looks as follows:

```rust
impl Block {
    pub fn new(
        header: Header,
        data: transaction::Data,
        evidence: evidence::Data,
        last_commit: Option<Commit>,
    ) -> Result<Self, Error> {
        // ... validation logic ...
        Ok(Block {
            header,
            data,
            evidence,
            last_commit,
        })
    }
}
```

The problem with this is that you need to do all the `TryInto` conversions for all of these related types, which includes error handling/mapping, in the outer context (e.g. within the `TryFrom<RawBlock> for Block` implementation). If you have `TryFrom` implementations for multiple different source/serialization types (e.g. Protobuf and JSON), then you need to handle these field conversion failures in each and every `TryFrom` implementation.

Compare this to an approach where we require dynamic input parameters:

```rust
impl Block {
    pub fn new(
        header: impl TryInto<Header>,
        data: impl TryInto<transaction::Data>,
        evidence: impl TryInto<evidence::Data>,
        last_commit: Option<impl TryInto<Commit>>,
    ) -> Result<Self, Error> {
        let header = header.try_into()?;
        let data = data.try_into()?;
        // ... etc.
        // ... validation logic ...
        Ok(Block {
            header,
            data,
            evidence,
            last_commit,
        })
    }
}
```

In such a case, the conversion attempts for all sub-types takes place within the constructor.

It can also improve developer ergonomics, because you can supply any type for a constructor parameter whose `TryInto` implementation facilitates conversion.

Are there other such approaches that could facilitate better developer ergonomics, and cut down on the required amount of code to achieve what we're looking to achieve here? Or better yet, are there perhaps libraries that can help to facilitate this with a minimal amount of coding from our side?

greg-szabo:
[...]

PR #719, opened 2020-12-02
rpc: Add support for consensus_state endpoint

### LINKED ISSUE #674: Support consensus state in tendermint rpc
I would like to be able to write consensus monitoring systems in rust.

The consensus_state endpoint can be very helpful for doing that.

This can be used to badger validators who aren't live yet.

thanethomson:
Closes #674 

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/rpc/src/client.rs
+++ b/rpc/src/client.rs
@@ -118,7 +118,7 @@ pub trait Client {
         self.perform(commit::Request::new(height.into())).await
     }
 
-    /// `/consensus_state`: get current consensus state (UNSTABLE)
+    /// `/consensus_state`: get current consensus state
     async fn consensus_state(&self) -> Result<consensus_state::Response> {
         self.perform(consensus_state::Request::new()).await
     }
--- a/rpc/src/client.rs
+++ b/rpc/src/client.rs
@@ -118,6 +118,11 @@ pub trait Client {
         self.perform(commit::Request::new(height.into())).await
     }
 
+    /// `/consensus_state`: get current consensus state (UNSTABLE)
+    async fn consensus_state(&self) -> Result<consensus_state::Response> {
+        self.perform(consensus_state::Request::new()).await
+    }
+
     /// `/validators`: get validators a given height.
     async fn validators<H>(&self, height: H) -> Result<validators::Response>
     where
--- a/rpc/src/endpoint/consensus_state.rs
+++ b/rpc/src/endpoint/consensus_state.rs
@@ -1,10 +1,16 @@
 //! `/consensus_state` endpoint JSON-RPC wrapper
 
-use crate::Method;
+use crate::{Error, Method};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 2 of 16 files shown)


rpc/src/client.rs
```diff
@@ -118,6 +118,11 @@ pub trait Client {
         self.perform(commit::Request::new(height.into())).await
     }
 
+    /// `/consensus_state`: get current consensus state (UNSTABLE)
```
greg-szabo, line 121: What is unstable here? The endpoint definition?

rpc/src/endpoint/consensus_state.rs
```diff
@@ -0,0 +1,387 @@
...
+pub struct HeightRoundStep {
+    /// Current block height
+    pub height: Height,
+    /// Current consensus round
+    pub round: Round,
+    /// Current consensus step
+    pub step: i8,
+}
+
+impl Serialize for HeightRoundStep {
```
greg-szabo, line 80: I know we deprioritized separating JSON serialization in the code, but would it make sense to add one-off JSON structs when we implement new domain types?

This seems like a prime example where we can create a HeightRoundStep JSON type that has i64/i64/i8 fields and then add a `#[serde(try_from="json-type", into="json-type")]` to the top of this struct.

I'm asking because tracking custom serialization implementations can become tricky and this implementation has public fields (described as a Type 0 struct in issue #661 ) but domain knowledge restrictions in the serialization (which I consider no-no and should be either in a TryFrom trait or a constructor as described by Type 1 and Type 2 structs in the same issue). I would guess that this doesn't cause issues in regular cases, because most of the fields are well guarded (Height and Round are domain types), but I can for example create a HeightRoundStep with step == -2 which I think is invalid.

So, even if we keep the custom serialization, it seems that a constructor might be warranted.

melekes (approved):
👍

rpc/src/client.rs
thanethomson: I don't know exactly what's unstable about this - that's just what it says in the Go code for this endpoint 😁

rpc/src/endpoint/consensus_state.rs
thanethomson: We could separate the JSON struct from the domain type for `HeightRoundStep`, but where would we put it?

rpc/src/client.rs
ebuchman: LOL this is how I used to mark APIs in Tendermint Go that weren't going to be covered by SemVer because they might change ... not sure it should be carried forward

thanethomson: Cool, so I'll take out the "unstable" notice here 😁

greg-szabo (approved):

rpc/src/endpoint/consensus_state.rs
thanethomson: Chatted to @greg-szabo about this, and we'll follow up here in a separate issue (#749).

PR #720, opened 2020-12-02
Re-enable fuzzing

### LINKED ISSUE #710: Border Control Testing to re-enable trivial fuzzing
**Context:**
"Fuzzing" was introduced to MBT so that we can test two things at data structure level:
1. messing with data fields like a hash or a signature must fail block verification,
2. incoming data structures need to be **correctly constructed** 
(for e.g. a `SignedHeader` cannot have a `Header` and a `Commit` pointing to different heights)

**Problem:**
For point 2, we have tests that does this kind of fuzzing but these scenarios cannot be tested with the current test driver because it primarily focuses on verification testing. And so, the current fuzzing tests have been disabled because some of them fail to be deserialized into domain types. 

**Solution:**
We need to test exactly that and this will require a different kind of testing. We call it "Border Control" (thanks @andrey-kuprianov for naming!) because we want to test if a fuzzed data is considered meaningful or discarded by the LightClient.

Shivani912:
closes #710 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/light-client/tests/model_based.rs
+++ b/light-client/tests/model_based.rs
@@ -16,6 +16,7 @@ use tendermint_testgen::{
 };
 use serde_json::Error;
 use serde::de::DeserializeOwned;
+use crate::LiteVerdict::Invalid;
 
 fn testgen_to_anon(tm_lb: TMLightBlock) -> AnonLightBlock {
     AnonLightBlock {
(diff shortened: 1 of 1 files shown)


andrey-kuprianov (approved):
Great work, @Shivani912! 

I also very much like your idea of generalizing to add `LiteVerdict::ParseError` :+1:

Shivani912:
Thanks @andrey-kuprianov ! 

> I also very much like your idea of generalizing to add LiteVerdict::ParseError 

Awesome, I think this will be helpful when we implement Testgen fuzzer :)

PR #724, opened 2020-12-06
Bump `k256` crate dependency to v0.6

tony-iqlusion:
Release notes: https://github.com/RustCrypto/elliptic-curves/pull/253

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -57,7 +57,7 @@ tendermint-proto = { version = "0.17.0-rc3", path = "../proto" }
 toml = { version = "0.5" }
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
 
-k256 = { version = "0.5", optional = true, features = ["ecdsa"] }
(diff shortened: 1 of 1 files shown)


tony-iqlusion:
Note: this release adds a `k256::PublicKey` type which ensures public keys are valid secp256k1 curve points:

https://docs.rs/elliptic-curve/0.7.1/elliptic_curve/public_key/struct.PublicKey.html

It might make sense to switch to it, however that'd also require changing the current `tendermint::PublicKey::as_bytes()` method (which borrows) to a `::to_bytes()` method which allocates / returns something like `Vec<u8>`, as the `k256::PublicKey` type does not retain the encoded form.

xla:
> It might make sense to switch to it, however that'd also require changing the current `tendermint::PublicKey::as_bytes()` method (which borrows) to a `::to_bytes()` method which allocates / returns something like `Vec<u8>`, as the `k256::PublicKey` type does not retain the encoded form.

@tony-iqlusion Do you have any intuition what the cost would be for this change?

xla (approved):
LGTM

👑 🍊 🚆 ♒️

PR #730, opened 2020-12-10
Adds testgen validator set

Shivani912:
This PR is step 1 of #665
* [X] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/testgen/src/validator_set.rs
+++ b/testgen/src/validator_set.rs
@@ -1,0 +1,78 @@
+use gumdrop::Options;
+use serde::{Deserialize, Serialize};
+use simple_error::*;
+use crate::{
+    validator::{
+        Validator as Validator,
+        generate_validators,
+    },
+    Generator,
+    helpers::*
+};
+use tendermint::{validator};
+
+#[derive(Debug, Options, Serialize, Deserialize, Clone)]
+pub struct ValidatorSet {
+    #[options(
+        parse(try_from_str = "parse_as::<Vec<Validator>>"),
+        help = "validators (required)"
+    )]
+    validators: Option<Vec<Validator>>,
+}
+
+impl ValidatorSet {
+    pub fn new(ids: Vec<&str>) -> Self {
+        let validators = ids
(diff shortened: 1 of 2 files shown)


testgen/src/validator_set.rs
```diff
@@ -0,0 +1,88 @@
...
+use simple_error::*;
+use tendermint::validator;
+
+#[derive(Debug, Options, Serialize, Deserialize, Clone)]
+pub struct ValidatorSet {
+    #[options(
+        parse(try_from_str = "parse_as::<Vec<Validator>>"),
+        help = "validators (required)"
+    )]
+    validators: Option<Vec<Validator>>,
```
melekes, line 17: could a validator set ever be empty?

Shivani912: Ideally, not. But in testgen we want to allow such situations to be able to test them.

andrey-kuprianov: yes, as this is primarily for testing, we do want to test situations when the validator set coming from the wire is empty.

andrey-kuprianov (changes requested):
Thanks for implementing this, @Shivani912! 
As the first draft it looks ok, but it is not at all integrated with the rest of the datastructures. Here are some points:

* everywhere where `Vec<Validator>` is used, e.g. in the `Header`, `ValidatorSet` should be used instead.
 * `generate_validators` should disappear from `validator`, and become just part of `ValidatorSet::generate()`
 * `ValidatorSet` should have the options to set `total_voting_power` and `proposer`. If they are not set, they should be calculated automatically.

These are the main points for now. There are two options on how to proceed:
 1. Either you can make these changes now;
 2. Or create a follow-up issue for that (or comment in the original issue #665), and now focus on writing the unit tests for the LightClient.

andrey-kuprianov (approved):
After a discussion, we've decided to merge this PR now, and address the comments above at a later time point. @Shivani912 could you please put those into #665? Thanks!

PR #733, opened 2020-12-10
Added version information to protobuf structs

greg-szabo:
A minor enhancement that allows the `tendermint-proto` crate users to check what the source of the protobuf structs are.

`tendermint_proto::meta::REPOSITORY` and `tendermint_proto::meta::COMMITISH` are autogenerated constants when creating the structs.

Edit: it also fixes a random issue with `git2::Oid::from_str()`. It didn't surface before and it might be a bug in the patch version of git2 but this way it's slightly more robust.

Edit2: This is in preparation for the final Go v0.34.0-compatible Tendermint-rs v0.17.0. There will be another PR where we update the version number and fix the breaking changes.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


melekes (approved):
👍

PR #740, opened 2020-12-14
Remove `total_voting_power` parameter from `validator::Set::new`

### LINKED ISSUE #739: Remove `total_voting_power` parameter from `validator::Set::new`
as discussed in https://github.com/informalsystems/tendermint-rs/pull/706/files#r542483341

romac:
Close #739 

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


tendermint/src/validator.rs
```diff
@@ -30,11 +30,12 @@ impl TryFrom<RawValidatorSet> for Set {
...
             .map(TryInto::try_into)
             .collect();
-        Ok(Self::new(
-            unsorted_validators_result?,
-            value.proposer.map(TryInto::try_into).transpose()?,
-            Some(value.total_voting_power.try_into()?),
-        ))
+
+        Ok(Self {
+            validators: unsorted_validators_result?,
```
thanethomson, line 35: This no longer sorts the validator set (which is done in `Set::new()`) - is that intentional?

romac: Oh good catch! Will fix this right now

thanethomson (approved):
LGTM 👍

PR #746, opened 2020-12-15
testgen: Compute `last_block_id` hash when generating a light chain

### LINKED ISSUE #744: testgen: LightChain does not generate blocks correctly
The `LightChain` helper in `test gen` does not properly generate light blocks:

a) when invoked with `LightChain::default_with_length(N)` it generates a chain of length `N - 1` instead of `N`
b) while the first block has correct height 1, the subsequent blocks all have height 2

### LINKED ISSUE #745: testgen: Compute `last_block_id` property when generating a chain of light blocks
I would like to use `testgen` to generate a sequence of light blocks in order to test my implementation of backward verification.

Since backward verification requires comparing a header's hash against the `last_block_id` hash stored in the next block, we need to compute this property when generating a chain of light blocks in the `LightChain` helper.

romac:
Close #744
Close #745

* [x] Wrote tests
* [x] Updated CHANGELOG.md


Shivani912 (approved):
Thank you for doing this @romac ! ❤️

Shivani912 (approved):
Approved 👍

PR #747, opened 2020-12-16
kvstore test with built-in docker invoke

greg-szabo:
Please compare to #748 

This is example code how we could expand our testing framework with Rust-native test prerequisites built into `build.rs`.

Prerequisite:
* docker daemon installed and running

Benefits:
* No additional dependencies
* Runs natively with `cargo test` if the endpoint is already running
* Runs natively with `cargo test --features docker` if the endpoint has to be set up
* Nice message during build notifying what docker image is going to be run

Disadvantages:
* The shiplift library used to manage docker is out-of-date and had to be pointed to master.
* The docker image cannot be shut down automatically at the end of the test run. `docker stop kvstore-test` has to be run or Tendermint will keep running in the background.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


greg-szabo:
We are dropping this idea in favour of #748 .

PR #752, opened 2020-12-17
Bump `k256` crate dependency to v0.7

tony-iqlusion:
This is a minor bugfix release that required some breaking changes in the public API.

None of the bugs impact `tendermint-rs`, but it would be nice to ensure it's using the latest version before the final v0.17 release.

Release notes: https://github.com/RustCrypto/elliptic-curves/pull/261


thanethomson (approved):
Awesome, thanks @tony-iqlusion!

PR #771, opened 2021-01-05
Update prost to 0.7

jkilpatr:
Prost has a new major release with significant improvements

https://github.com/danburkert/prost/releases/tag/v0.7.0

This patch only updates two dependencies Prost and bytes since Prost 0.7 updated it's own bytes dependency to 1.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### IMPROVEMENTS:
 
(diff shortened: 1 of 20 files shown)


thanethomson:
Thanks for this @jkilpatr, but we had to do a bigger upgrade in #783.

Prost is now at 0.7 though in tendermint-rs on `master`.

Issue #774, opened 2021-01-07
`Time` values are not always formatted correctly, causing the light client to sometimes return malformed light blocks


romac:
@ancazamfir recently discovered that the Rust relayer would sometimes send malformed light blocks to the chain, eg. in a  ClientUpdate message. The chain would then fail with the following error:

```
header failed basic validation:
  commit signs block A195DA45ACDC71387E33B7FAF7756425483FAB5DA21D8B0D02A4D5A5772A95F8
  header is block E1372844E223FDBB7209B08AA1231ED9C9D85AFBCF72FF9278ABE02A12B549EC
```

It appears that light blocks sometimes get corrupted when stored and retrieved from the LightStore, specifically the `signed_header.header.time` field:

```diff
         chain_id: chain::Id(ibc-0),
         height: block::Height(3962),
         time: Time(
-            2021-01-06T21:50:58.067816Z,
+            2021-01-06T21:50:58.678160Z,
         ),
         last_block_id: Some(
             Id {
```

Because this field is part of the hash of the header, the chain would then compute a different hash than the commit's last_block_id hash, yielded the error above. The problem stems from the custom  `serde` serializer for `Time` values, which is used by `serde_cbor` to encode the light blocks as a byte array before storing them in the `sled`-backed light store.

I have tracked this to the [custom RFC3339 formatter](https://github.com/informalsystems/tendermint-rs/blob/e65337a4ec107274bc43834c1d4f1106815cd3ad/proto/src/serializers/timestamp.rs#L55-L72) which we use to format `tendermint::Time` via `tendermint_proto::serializers::timestamp::Timestamp`.

This function incorrectly trims the leading zeros of the nanoseconds digits, which would break the `Time` -> String -> `Time` roundtrip.

For example, the date "2021-01-06T21:50:58.067816Z" is serialized as "2021-01-06T21:50:58.67816Z", which then decodes to "2021-01-06T21:50:58.678160Z", shifting the leading zero to the end.

This explains why this error pops up non-deterministically, as it requires a light block whose time would feature one or more zeros in its subseconds digits for the formatting to break.

romac:
This glitch was not caught by the Rust light client because light blocks are verified _before_ they are stored in the light store (which is when they are serialized and the glitch occurs). On the other hand, the light client always fetches a block from the light store before returning it upon successful verification.

PR #775, opened 2021-01-07
Fix formatting of tendermint::Time values

romac:
Close: #774 

The `Time` datatype needs to be formatted the same way as
Go's `time.RFC3339Nano` format, ie. a RFC3339 date-time
with left-padded nanoseconds digits without trailing zeros
and no trailing dot.

This is taken care of by the `to_rfc339_nanos` function in
the `tendermint_proto::serializers::timestamp` module.

Prior to this commit, this function would incorrectly
trim the leading zeros of the nanoseconds digits, which
would break the `Time` -> String -> `Time` roundtrip.

For example, the date "2021-01-07T17:52:04.034475Z" was serialized
as "2021-01-07T17:52:04.34475Z", which then decodes to
"2021-01-07T17:52:04.344750Z", shifting the leading zero to the end.

This commit fixes `to_rfc339_nanos` function by using `chrono`'s
formatting facilities to preserve the precision of the nanoseconds,
and then manually trim the trailing zeros and dot.

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@
 [#758]: https://github.com/informalsystems/tendermint-rs/pull/758
 [cargo make]: https://github.com/sagiegurari/cargo-make
 
+### BUG FIXES:
+
+- `[tendermint]` `Time` values were not always formatted properly,
+  causing the light client to sometimes return malformed light blocks. ([#774])
+
+[#774]: https://github.com/informalsystems/tendermint-rs/issue/774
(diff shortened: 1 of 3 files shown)


ancazamfir (approved):
LGTM, thanks Romain for getting to the root cause of this!

thanethomson (approved):
Awesome, thanks for catching this Romain!

Issue #764, opened 2020-12-28
Upgrade prost and bytes dependency to 0.7 / 1.0


greg-szabo:
bytes 1.0 was released and is used in prost 0.7.

Both bytes and prost should be upgraded.

Todo: add impact of upgrade here.

PR #783, opened 2021-01-18
Update Tokio to 1.0, Hyper to 0.14, Prost to 0.7 and Bytes to 1.0

romac:
Close: #764

This PR updates the following dependencies:

  - `tokio` to `1.0`
  - `hyper` to `0.14`
  - `prost` to `0.7`
  - `bytes` to `1.0`
  - `async-tungstenite` to `0.12`

---

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


rpc/src/client/transport/websocket.rs
```diff
@@ -301,35 +301,35 @@ impl WebSocketClientDriver {
     pub async fn run(mut self) -> Result<()> {
         let mut ping_interval =
             tokio::time::interval_at(Instant::now().add(PING_INTERVAL), PING_INTERVAL);
-        let mut recv_timeout = tokio::time::delay_for(PING_INTERVAL);
+
         loop {
+            let next_msg = tokio::time::timeout(RECV_TIMEOUT, self.stream.next());
```
romac, line 306: @thanethomson This uses `RECV_TIMEOUT` as the timeout for the next message, instead of `PING_INTERVAL`. Can you please confirm that this is correct?

romac: @thanethomson I've also had to implement the timeout in a different way, because [`Sleep::reset`](https://docs.rs/tokio/1.0.2/tokio/time/struct.Sleep.html#method.reset) now requires the self-type `Pin<&mut Sleep>` instead of just `&mut Sleep`, and thus takes ownership of the pinned `Sleep` value, which makes it impossible to `.await` it later in the loop.

You can see the error for yourself if you replace the whole `run` method with

```rust
    /// Executes the WebSocket driver, which manages the underlying WebSocket
    /// transport.
    pub async fn run(mut self) -> Result<()> {
        let mut ping_interval =
            tokio::time::interval_at(Instant::now().add(PING_INTERVAL), PING_INTERVAL);

        let recv_timeout = tokio::time::sleep(RECV_TIMEOUT);
        tokio::pin!(recv_timeout);

        loop {
            tokio::select! {
                Some(res) = self.stream.next() => match res {
                    Ok(msg) => {
                        // Reset the receive timeout every time we successfully
                        // receive a message from the remote endpoint.
                        recv_timeout.reset(Instant::now().add(RECV_TIMEOUT));
                        self.handle_incoming_msg(msg).await?
                    },
                    Err(e) => return Err(
                        Error::websocket_error(
                            format!("failed to read from WebSocket connection: {}", e),
                        ),
                    ),
                },
                Some(cmd) = self.cmd_rx.recv() => match cmd {
                    DriverCommand::Subscribe(subs_cmd) => self.subscribe(subs_cmd).await?,
                    DriverCommand::Unsubscribe(unsubs_cmd) => self.unsubscribe(unsubs_cmd).await?,
                    DriverCommand::SimpleRequest(req_cmd) => self.simple_request(req_cmd).await?,
                    DriverCommand::Terminate => return self.close().await,
                },
                _ = ping_interval.tick() => self.ping().await?,
                _ = &mut recv_timeout => {
                    return Err(Error::websocket_error(format!(
                        "reading from WebSocket connection timed out after {} seconds",
                        RECV_TIMEOUT.as_secs()
                    )));
                }
            }
        }
    }
```

```
error[E0382]: borrow of moved value: `recv_timeout`
   --> rpc/src/client/transport/websocket.rs:330:21
    |
306 |         tokio::pin!(recv_timeout);
    |         -------------------------- move occurs because `recv_timeout` has type `std::pin::Pin<&mut tokio::time::Sleep>`, which does not implement the `Copy` trait
...
314 |                         recv_timeout.reset(Instant::now().add(RECV_TIMEOUT));
    |                         ------------ value moved here, in previous iteration of loop
...
330 |                 _ = &mut recv_timeout => {
    |                     ^^^^^^^^^^^^^^^^^ value borrowed here after move

error[E0382]: borrow of moved value: `recv_timeout`
   --> rpc/src/client/transport/websocket.rs:330:21
    |
306 |         tokio::pin!(recv_timeout);
    |         -------------------------- move occurs because `recv_timeout` has type `std::pin::Pin<&mut tokio::time::Sleep>`, which does not implement the `Copy` trait
...
314 |                         recv_timeout.reset(Instant::now().add(RECV_TIMEOUT));
    |                         ------------ value moved here, in previous iteration of loop
...
330 |                 _ = &mut recv_timeout => {
    |                     ^^^^^^^^^^^^^^^^^ value borrowed here after move
```

Please let me know what you think of the new solution with `timeout` combinator :)

rpc/src/client/transport/websocket.rs
```diff
@@ -586,7 +586,7 @@ mod test {
             loop {
                 tokio::select! {
                     Some(ev) = self.event_rx.recv() => self.publish_event(ev),
-                    Some(res) = self.listener.next() => self.handle_incoming(res.unwrap()).await,
+                    Ok((stream, _)) = self.listener.accept() => self.handle_incoming(stream).await,
```
romac, line 589: @thanethomson This is imho cleaner as it avoids an `unwrap` but will silently discard errors when accepting a new connection whereas the previous solution would panic. Does this work for you?

rpc/src/client/transport/websocket.rs
```diff
@@ -678,8 +678,8 @@ mod test {
         async fn run(mut self) -> Result<()> {
             loop {
                 tokio::select! {
-                    Some(res) = self.conn.next() => {
-                        if let Some(ret) = self.handle_incoming_msg(res.unwrap()).await {
+                    Some(Ok(msg)) = self.conn.next() => {
+                        if let Some(ret) = self.handle_incoming_msg(msg).await {
```
romac, line 682: @thanethomson Same here

tendermint/src/time.rs
```diff
@@ -33,11 +33,7 @@ impl TryFrom<Timestamp> for Time {
             nanos: value.nanos,
         };
 
-        Ok(SystemTime::try_from(prost_value)
-            .map_err(|e| {
-                Kind::OutOfRange.context(format!("time before EPOCH by {} seconds", e.as_secs()))
-            })?
-            .into())
+        Ok(SystemTime::from(prost_value).into())
```
romac, line 36: The conversion from a prost `Timestamp` to `SystemTime` is now infallible because `prost_types::Timestamp` now provides a `Into<SystemTime>`.

rpc/src/client/transport/websocket.rs
romac: @thanethomson What do you think about introducing an enum to help with the readability of the client loop:

```rust
    pub async fn run(mut self) -> Result<()> {
        let mut ping_interval =
            tokio::time::interval_at(Instant::now().add(PING_INTERVAL), PING_INTERVAL);

        enum NextMsg {
            Ok(Message),
            StreamEnded,
            Timeout,
            WebSocketError(async_tungstenite::tungstenite::Error),
        }

        loop {
            let next_msg =
                tokio::time::timeout(RECV_TIMEOUT, self.stream.next()).map(|res| match res {
                    Ok(Some(Ok(msg))) => NextMsg::Ok(msg),
                    Ok(Some(Err(e))) => NextMsg::WebSocketError(e),
                    Ok(None) => NextMsg::StreamEnded,
                    Err(_) => NextMsg::Timeout,
                });

            tokio::select! {
                res = next_msg => match res {
                    NextMsg::Ok(msg) => {
                        self.handle_incoming_msg(msg).await?
                    },
                    NextMsg::StreamEnded => {
                        // Websocket stream is over, let's continue in case
                        // we still receive commands via the `cmd_rx` channel.
                        continue;
                    },
                    NextMsg::Timeout => return Err(Error::websocket_error(format!(
                        "reading from WebSocket connection timed out after {} seconds",
                        RECV_TIMEOUT.as_secs()
                    ))),
                    NextMsg::WebSocketError(e) => return Err(Error::websocket_error(
                        format!("failed to read from WebSocket connection: {}", e),
                    )),
                },
                Some(cmd) = self.cmd_rx.recv() => match cmd {
                    DriverCommand::Subscribe(subs_cmd) => self.subscribe(subs_cmd).await?,
                    DriverCommand::Unsubscribe(unsubs_cmd) => self.unsubscribe(unsubs_cmd).await?,
                    DriverCommand::SimpleRequest(req_cmd) => self.simple_request(req_cmd).await?,
                    DriverCommand::Terminate => return self.close().await,
                },
                _ = ping_interval.tick() => self.ping().await?,
            }
        }
    }
```

thanethomson: @romac you're right, the `recv_timeout` should've been reset with `RECV_TIMEOUT` and not `PING_INTERVAL`.

But with regard to the timeout combinator, correct me if I'm wrong but it looks like it changes the behaviour of the loop. The expected behaviour is that the `recv_timeout` must only be reset when we receive a message.

The new behaviour seems to reset the timeout on each iteration of the loop, which would also then happen when we (1) receive any command, and (2) send out a ping.

rpc/src/client/transport/websocket.rs
thanethomson: In testing code I'd generally prefer `.unwrap()` (personal preference). Absolutely nothing should go wrong here, and the test should break clearly and definitively if something fails.

tendermint/src/time.rs
thanethomson: Nice!

rpc/src/client/transport/websocket.rs
romac: > But with regard to the timeout combinator, correct me if I'm wrong but it looks like it changes the behaviour of the loop. The expected behaviour is that the recv_timeout must only be reset when we receive a message.

Yeah absolutely, I understood the original semantics but somehow didn't realize that this would change the behavior.

Thankfully, I managed to restore the original behavior in 477881026205da4d8911c5e59a9bccd9e5e55f15. They key was to use [`Pin::as_mut`](https://doc.rust-lang.org/stable/std/pin/struct.Pin.html#method.as_mut) to reborrow the pinned `&mut Sleep` in order to call `reset` on it.

rpc/src/client/transport/websocket.rs
romac: Makes sense! Fixed in f8a37a67034122dd8e11e522716b10b33c219b34.

rpc/src/client/transport/websocket.rs
romac: Restored panicking behavior in f84e6aa7bb5fdd79e2b5a3caf65088c033150d17

thanethomson (approved):
Exciting times! 🕺🎉

PR #794, opened 2021-01-28
Add minimal blocking ABCI crate

### LINKED ISSUE #29: Absorb rust-abci

thanethomson:
Closes #29

In order to reduce the amount of code for review, I've extracted portions of #750 and expanded on them, but purely for the blocking case (using the Rust standard library's networking functionality). I've now included the full range of ABCI requests/responses, but purely using their Protobuf representations (also to reduce the sheer quantity of code). Now that the investigative work in #750 has been done, adding `async` compatibility should be a relatively simple (but probably tedious) task - TBD in separate PRs.

There's a minimal `kvstore-rs` application that can be built into a binary, and the basic functionality works with a fresh Tendermint node. See the ABCI crate README for details:

:book: [README rendered](https://github.com/informalsystems/tendermint-rs/blob/thane/minimal-blocking-abci/abci/README.md)

* [x] Wrote tests
* [x] Updated CHANGELOG.md


abci/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+mod server;
+
+// Common exports
+pub use application::Application;
+#[cfg(feature = "client")]
+pub use client::{Client, ClientBuilder};
+pub use result::{Error, Result};
+pub use server::{Server, ServerBuilder};
+
+// Example applications
```
thanethomson, line 17: Do we need to expose these apps?

thanethomson: Looking at this again now, the integration tests need these apps to be exposed, and they might actually be useful as components of test suites (e.g. a full node will need to import an application like the kvstore for in-process testing).

abci/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[features]
+client = []
+echo-app = []
+kvstore-app = []
+binary = [ "simple_logger", "structopt" ]
+
+[dependencies]
+bytes = "1.0"
+eyre = "0.6"
+log = "0.4"
```
romac, line 27: What's the reasoning for using `log` instead of `tracing`, like we do in `tendermint-rpc` or `ibc-rs`?

abci/src/application/kvstore.rs
```diff
@@ -0,0 +1,273 @@
...
+        result_tx: Sender<Option<String>>,
+    },
+    /// Commit the current state of the application, which involves recomputing
+    /// the application's hash.
+    Commit { result_tx: Sender<(i64, Vec<u8>)> },
+}
+
+fn channel_send<T>(tx: &Sender<T>, value: T) -> Result<()> {
+    tx.send(value)
+        .map_err(|e| Error::ChannelSend(e.to_string()))
```
romac, line 268: Can we hold on to the original error here rather than its string representation?

abci/src/application/kvstore.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+fn channel_recv<T>(rx: &Receiver<T>) -> Result<T> {
+    rx.recv().map_err(|e| Error::ChannelRecv(e.to_string()))
```
romac, line 272: Same here

abci/src/codec.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            // Try to decode an incoming message from our buffer first
+            match decode_length_delimited::<I>(&mut self.read_buf) {
+                Ok(opt) => {
+                    if let Some(incoming) = opt {
+                        return Some(Ok(incoming));
+                    }
+                }
+                Err(e) => return Some(Err(e)),
```
romac, line 76: imho I find the following version of the code a bit easier to understand, but that's completely subjective so feel free to ignore.

```suggestion
                Ok(Some(incoming)) => return Some(Ok(incoming)),
                Err(e) => return Some(Err(e)),
                _ => (), // not enough data to decode a message, let's continue.
```

abci/src/codec.rs
```diff
@@ -0,0 +1,174 @@
...
+    let encoded_len = match decode_varint(&mut tmp) {
+        Ok(len) => len,
+        Err(e) => {
+            return if src_len <= MAX_VARINT_LENGTH {
+                // We've potentially only received a partial length delimiter
+                Ok(None)
+            } else {
+                Err(e)
+            };
+        }
```
romac, line 148: The following reads a bit better to me, but again feel free to ignore.

```suggestion
        // We've potentially only received a partial length delimiter
        Err(e) if src_len <= MAX_VARINT_LENGTH => return Ok(None),
        Err(e) => return Err(e),
```

romac (commented):
Great stuff, very clean and minimal, <3 the test suite :)

Left a few minor comments but otherwise it looks really good to me!

abci/Cargo.toml
thanethomson: I could switch over to `tracing` - are we going to be sticking with `tracing` then for `tendermint-rpc` and `ibc-rs` for the foreseeable future?

abci/src/application/kvstore.rs
thanethomson: I might be able to do that with `eyre` - I'll try it out :+1: Originally I had trouble with this because [SendError](https://doc.rust-lang.org/std/sync/mpsc/struct.SendError.html) is generic over `T`, which would imply that the error needs to be generic over `T` too. But if the error can be wrapped as a `Box<dyn std::error::Error>` this shouldn't be a problem, I think.

abci/src/codec.rs
thanethomson: Applied :+1:

abci/src/application/kvstore.rs
romac: Oh I see. Since it's just an example app it does not matter too much, just a nice-to-have (probably not worth it if we have to pull in `eyre`).

abci/Cargo.toml
romac: Yeah, we intend to stick with `tracing` in `ibc-rs`, as it's strictly more powerful than the other crates we've looked at (even if we do not use it to its full potential at the moment).

abci/src/application/kvstore.rs
thanethomson: Here's the conundrum with eyre, as per https://docs.rs/eyre/0.6.5/eyre/struct.Report.html:

> `Report` requires that the error is `Send`, `Sync`, and `'static`.

This would imply that `T` must be `Send`, `Sync` and `'static` too. In this case I'm trying to send a variant of the `Command` enum:

```rust
enum Command {
    /// Get the height of the last commit.
    GetInfo { result_tx: Sender<(i64, Vec<u8>)> },
    /// Get the key associated with `key`.
    Get {
        key: String,
        result_tx: Sender<(i64, Option<String>)>,
    },
    /// Set the value of `key` to to `value`.
    Set {
        key: String,
        value: String,
        result_tx: Sender<Option<String>>,
    },
    /// Commit the current state of the application, which involves recomputing
    /// the application's hash.
    Commit { result_tx: Sender<(i64, Vec<u8>)> },
}
```

`std::sync::mpsc::Sender` is `!Sync`, and I ideally need to be able to send it across the channel in order to get a dedicated response asynchronously.

Perhaps this is a hint that a better overall design is possible? I suppose falling back to `Arc<Mutex<...>>` instead of the handle/driver architecture would solve this. Or simply "flattening" `std::sync::mpsc::SendError<T>`, like I'm doing in the original code here. I can't think why the `SendError` would need to be generic over `T` in the first place?

abci/Cargo.toml
thanethomson: Refactored! :+1:

abci/src/application/kvstore.rs
romac: I think the current solution and architecture are fine.

Just wondering, have you tried just casting `SendError` to `Box<dyn std::error::Error>` and wrapping that one in `ChannelSend` (without using `eyre` at all)?

---

Regarding `SendError` being generic over `T`: that's because it's defined as

```rust
pub struct SendError<T>(pub T);
```

which allows you to get back the value that failed to be sent.

abci/src/codec.rs
```diff
@@ -0,0 +1,166 @@
...
+pub type ServerCodec<S> = Codec<S, Request, Response>;
+
+#[cfg(feature = "client")]
+/// The client sends outgoing requests, and receives incoming responses.
+pub type ClientCodec<S> = Codec<S, Response, Request>;
+
+/// Allows for iteration over `S` to produce instances of `I`, as well as
+/// sending instances of `O`.
+///
+/// `S` must implement [`std::io::Read`] and [`std::io::Write`].
```
xla, line 28: Instead of documenting it, `S` could be constrained here.

romac: The pattern of constraining struct type params when not absolutely necessary* is discouraged in the Rust ecosystem (most notably in the standard library): https://github.com/rust-lang/api-guidelines/issues/6

As such, the current implementation looks just fine to me.

\* A struct should only have trait bounds on its type params if it needs to refer to a type member defined in a trait, eg.

```rust
trait Foo {
  type Bar;
}

struct Toto<F: Foo> {
  tata: F::Bar
}
```

xla: Thanks for the pointers in the guidelines. It is sensible to not constraint the params here, but I'd drop the doc hint for it then.

abci/src/application.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// Provide information about the ABCI application.
+    fn info(&self, _request: RequestInfo) -> ResponseInfo {
+        Default::default()
+    }
+
+    /// Called once upon genesis.
+    fn init_chain(&self, _request: RequestInitChain) -> ResponseInitChain {
+        Default::default()
```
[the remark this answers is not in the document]
thanethomson: I was modelling it on what the `BaseApplication` does in the Go version: https://github.com/tendermint/tendermint/blob/master/abci/types/application.go#L34

abci/src/codec.rs
thanethomson: Cool, I've removed the comment :+1:

tools/docker/tendermint-0.34.0/.gitignore
```diff
@@ -0,0 +1 @@
+tendermint
```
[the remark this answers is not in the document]
thanethomson: Done

abci/src/application/kvstore.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// Constructor.
+    pub fn new() -> (Self, KeyValueStoreDriver) {
+        let (cmd_tx, cmd_rx) = channel();
+        (Self { cmd_tx }, KeyValueStoreDriver::new(cmd_rx))
+    }
+
+    /// Attempt to retrieve the value associated with the given key.
+    pub fn get<K: AsRef<str>>(&self, key: K) -> Result<(i64, Option<String>)> {
+        let (result_tx, result_rx) = channel();
+        channel_send(
```
melekes, line 33: Could you explain why you've chosen this concurrency model (communication via channels) over something like a mutex & hashmap? _As someone who's learning Rust it will be useful to me._

abci/src/codec.rs
```diff
@@ -0,0 +1,164 @@
...
+        // data to try to decode the result.
+        src.advance(delim_len + (encoded_len as usize));
+
+        let mut result_bytes = BytesMut::from(tmp.split_to(encoded_len as usize).as_ref());
+        Ok(Some(M::decode(&mut result_bytes)?))
+    }
+}
+
+// encode_varint and decode_varint will be removed once
+// https://github.com/tendermint/tendermint/issues/5783 lands in Tendermint.
```
melekes, line 156: note the issue is closed

melekes (dismissed):
👍

abci/src/application/kvstore.rs
thanethomson: This may be a personal preference/readability thing :grin: I'd say there are pros and cons to each approach. I've always tended to prefer an actor-style model for managing state in concurrent systems (explicit-lock-free state encapsulation with message passing and pattern matching on messages), which I've tried to implement here in a lightweight way. It's easier for me to reason about such state machines in general ([like this](https://github.com/informalsystems/tendermint-rs/pull/794/files#diff-d3dc4894c066ca996f470f741a50cd983e04000bd8dce1960ce0adccce6dad72R211)).

When it comes to contention, I'd imagine using mutexes would push bottlenecks down the stack towards the networking layer. With the actor-based model it'd be able to leverage memory (in the unbounded channel) to avoid pushing the bottleneck towards the networking layer. On more constrained devices, this channel-based approach wouldn't provide any benefit for the additional complexity cost, so mutexes would probably be the way to go.

Practically both models would probably work the same for this specific application, although I'm not sure which one would provide better performance (I'd have to benchmark to find out). I'd imagine that the mutex-based approach would probably provide better performance, speed-wise, as long as the app can keep up with incoming requests.

abci/src/codec.rs
thanethomson: Isn't it only going to be released in Tendermint v0.35? All our tests are against v0.34.x right now.

abci/src/application/kvstore.rs
thanethomson: > Just wondering, have you tried just casting SendError to Box<dyn std::error::Error> and wrapping that one in ChannelSend (without using eyre at all)?

`T` would still need to implement `Sync`, afaik :disappointed:

abci/src/codec.rs
melekes: yeah, because it's breaking

brapse (approved):
Good stuff & discussion 👍

tomtau:
I don't have admin rights on https://github.com/tendermint/rust-abci but after the ABCI crate here gets published, perhaps it'll be good to archive the old repo and point to here?
@xla and @liamsi should have rights on https://crates.io/crates/abci

abci/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "tendermint-abci"
```
tomtau, line 2:
```suggestion
name = "abci"
```

it can be called just this and published as a new breaking version of https://crates.io/crates/abci -- @xla @liamsi can help with that

PR #798, opened 2021-02-01
Fix panic in evidence serialization

### LINKED ISSUE #782: Panic in evidence serializers
The [serializers for `tendermint::types::{Evidence, evidence::Sum}`](https://github.com/informalsystems/tendermint-rs/blob/4cbf727/proto/src/serializers/evidence.rs) are incomplete and may [cause a panic in the relayer](https://github.com/informalsystems/ibc-rs/pull/516#issuecomment-758745231).

The panic happens in this `From` instance, which should likely be converted to a `TryFrom` instance:

https://github.com/informalsystems/tendermint-rs/blob/4cbf72738906facfa320745d2f1df223f707ee7e/proto/src/serializers/evidence.rs#L30-L38

Additionally, these `From` instances are currently `unimplemented!`:

https://github.com/informalsystems/tendermint-rs/blob/4cbf72738906facfa320745d2f1df223f707ee7e/proto/src/serializers/evidence.rs#L40-L50

The comment seems to imply they won't ever be called, so it shouldn't be a problem, but perhaps it would worth looking into removing them altogether while we're at it.

thanethomson:
Closes #782 

* [x] Wrote tests
* [x] Updated CHANGELOG.md


romac (commented):
Seems like a good fix!

But I wonder if we actually need this whole custom serializer at all?

Disclaimer: It's 2am here and I am likely missing something, so please bear with me :)

It seems to me that this custom `EvidenceVariant` is defined the very same way as its counterpart in the generated code.

Custom `EvidenceVariant`:

```rust
 #[derive(Clone, PartialEq, ::serde::Deserialize, ::serde::Serialize)]
 #[serde(tag = "type", content = "value")]
 pub enum EvidenceVariant {
     #[serde(rename = "tendermint/DuplicateVoteEvidence")]
     DuplicateVoteEvidence(crate::tendermint::types::DuplicateVoteEvidence),
     #[serde(rename = "tendermint/LightClientAttackEvidence")]
    LightClientAttackEvidence(crate::tendermint::types::LightClientAttackEvidence),
}
```

Generated code in `proto/src/prost/tendermint.types.rs`

```rust
    #[derive(Clone, PartialEq, ::prost::Oneof, ::serde::Deserialize, ::serde::Serialize)]
    #[serde(tag = "type", content = "value")]
    pub enum Sum {
        #[prost(message, tag = "1")]
        #[serde(rename = "tendermint/DuplicateVoteEvidence")]
        DuplicateVoteEvidence(super::DuplicateVoteEvidence),
        #[prost(message, tag = "2")]
        #[serde(rename = "tendermint/LightClientAttackEvidence")]
        LightClientAttackEvidence(super::LightClientAttackEvidence),
    }
```

Could we instead simply remove the `#[serde(from = "crate::serializers::evidence::EvidenceVariant", into = "crate::serializers::evidence::EvidenceVariant")]` annotation from the generated `Evidence` type and get the very same behavior wrt. to serialization without having to introduce these weird `From` instances?

---

Okay, apparently it's not as simple as I though since my branch where I implemented the changes suggested above fails to deserialize the evidence: https://github.com/informalsystems/tendermint-rs/runs/1811124652

It's still not clear to me why this fails, so I would be grateful if someone can enlighten me :)

romac:
Oh I think I got it: the RPC endpoint actually returns the `EvidenceVariant` directly rather than wrapped in an `Evidence` with a `sum` field. Nevermind then, we likely can't do much different than what we are doing here.

romac:
I still wonder if there's not another way to work around this discrepancy between the Protobuf definition and the JSON format returned via RPC.

Would it help if we made Evidence into a DomainType, like we do for other datatypes?

thanethomson:
Yeah the whole reason for this custom serializer is because of how Protobuf wraps its `enum` types as structs. We generally use the `enum` type directly (and it's far more ergonomic to do so in Rust), whereas the generated Protobuf/prost types optionally wrap the `enum` in a struct.

Technically, we should never really encounter an instance where we receive evidence, but the `sum` field is `None`, because I'd imagine we'd rather receive an empty `EvidenceList`. I'm still not sure why this is happening on the IBC side.

> Would it help if we made Evidence into a DomainType, like we do for other datatypes?

We do actually have a domain type for it in the `tendermint` crate :grin:  This struct in the `tendermint-proto` crate just performs the JSON de/serialization duties on behalf of the Protobuf struct, because we wanted to keep the JSON serialization out of the `tendermint` crate.

I have to say, it looks like the dream of a single set of data structures is quite far off. Part of this has to do with the annoying way in which prost generates `enum` types. Right now we generally have 2 sets of structs (prost and domain types), but in some cases it's unavoidable to also have separate JSON types (a third set).

If we were separating things "properly", we'd have one set of dedicated Protobuf types, one set of JSON types, and one set of domain types. But the sheer amount of manual code and maintenance required for that doesn't really seem justified, especially since so much of it's boilerplate.

---

On a separate note, it looks like the `kvstore-test` is failing consistently in this build for an unrelated serialization problem: https://github.com/informalsystems/tendermint-rs/pull/798/checks?check_run_id=1809593214 (the consensus state deserialization's failing, but it's not clear why). What makes this more frustrating is how these tests pass just fine on my local machine :neutral_face:

romac (approved):

PR #802, opened 2021-02-05
p2p: apply state machine pattern when building SecretConnection

### LINKED ISSUE #735: p2p: rewrite SecretConnection handshake using the state machine pattern
Rewrite the SecretConnection handshake using the learnings from https://github.com/informalsystems/tendermint-rs/pull/708. The state machine pattern hopefully will allow us: a) to test each step (~state) separately b) enforce the rules (i.e. one can't go to creating challenge (state 2) before receiving the remote ephemeral key (state 1).

Resources:

- https://hoverbear.org/blog/rust-state-machine-pattern/
- https://deislabs.io/posts/a-fistful-of-states/

melekes:
Split SecretConnection::new into 3 states:

- new (_no separate state really_)
- awaiting ephemeral key (`AwaitingEphKey`)
- awaiting auth signature (`AwaitingAuthSig`)

First, you need to use Handshake::new to create a Handshake state
machine (STM), which is immediately put into AwaitingEphKey state.
When you obtain a remote ephemeral key, you're supposed to call
`got_key` fn to transition STM to the next state - `AwaitingAuthSig`.

NOTE: Handshake does not do IO and does not care how exchange is done.

Once you have the remote auth signature, you're supposed to call
`got_signature` fn, which returns the final piece - remote pubkey.

The construction of `SecretConnection` is now complete.

Closes #735

* [ ] ~~Wrote tests~~ https://github.com/informalsystems/tendermint-rs/pull/776 
* [ ] ~~Updated CHANGELOG.md~~

### DIFF
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -42,6 +42,13 @@ pub const DATA_MAX_SIZE: usize = 1024;
 const DATA_LEN_SIZE: usize = 4;
 const TOTAL_FRAME_SIZE: usize = DATA_MAX_SIZE + DATA_LEN_SIZE;
 
+/// Handshake is a process of establishing the SecretConnection between two peers.
+/// Specification: https://github.com/tendermint/spec/blob/master/spec/p2p/peer.md#authenticated-encryption-handshake
+struct Handshake<S> {
+    protocol_version: Version,
+    state: S,
+}
+
 /// Handshake states
 
 /// AwaitingEphKey means we're waiting for the remote ephemeral pubkey.
@@ -59,14 +66,8 @@ struct AwaitingAuthSig {
     local_signature: ed25519::Signature,
 }
 
-/// Handshake is a process of establishing the SecretConnection between two peers.
-/// Specification: https://github.com/tendermint/spec/blob/master/spec/p2p/peer.md#authenticated-encryption-handshake
-struct Handshake<S> {
-    protocol_version: Version,
-    state: S,
-}
-
 impl Handshake<AwaitingEphKey> {
+    /// Initiate a handshake.
     pub fn new(
         local_privkey: ed25519::Keypair,
         protocol_version: Version,
@@ -93,7 +94,11 @@ impl Handshake<AwaitingEphKey> {
         &mut self,
         remote_eph_pubkey: EphemeralPublic,
     ) -> Result<Handshake<AwaitingAuthSig>> {
-        let local_eph_privkey = self.state.local_eph_privkey.take().unwrap();
+        let local_eph_privkey = self
+            .state
+            .local_eph_privkey
+            .take()
+            .wrap_err("forgot to call Handshake::new?")?;
         let local_eph_pubkey = EphemeralPublic::from(&local_eph_privkey);
 
         // Compute common shared secret.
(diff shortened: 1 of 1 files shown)


p2p/src/secret_connection.rs
```diff
@@ -148,19 +168,79 @@ impl<IoHandler: Read + Write + Send + Sync> SecretConnection<IoHandler> {
...
+        // Compute a local signature (also recv_cipher & send_cipher)
+        let mut h = h.got_key(remote_eph_pubkey)?;
+
+        let mut sc = SecretConnection {
+            io_handler,
+            protocol_version,
+            recv_buffer: vec![],
+            recv_nonce: Nonce::default(),
+            send_nonce: Nonce::default(),
+            recv_cipher: h.state.recv_cipher.clone(),
```
melekes, line 227: note: I had to clone ciphers. Otherwise, I'd have to move them out of the Handshake struct.

p2p/src/secret_connection.rs
```diff
@@ -42,38 +42,59 @@ pub const DATA_MAX_SIZE: usize = 1024;
...
-impl<IoHandler: Read + Write + Send + Sync> SecretConnection<IoHandler> {
-    /// Returns authenticated remote pubkey
-    pub fn remote_pubkey(&self) -> PublicKey {
-        self.remote_pubkey.expect("remote_pubkey uninitialized")
-    }
+/// Handshake is a process of establishing the SecretConnection between two peers.
+/// Specification: https://github.com/tendermint/spec/blob/master/spec/p2p/peer.md#authenticated-encryption-handshake
+struct Handshake<S> {
+    protocol_version: Version,
+    state: S,
```
xla, line 66: One significant change we should land is to have `S` be constrained. Otherwise at the moment a `Handshake` can be constructed with any `S`. Commonly the `State` trait is sealed away in a private module, together with the impls to make it impossible for it to be constructed from the outside with anything but a sanctioned list of states. For reference:

https://github.com/informalsystems/tendermint-rs/blob/6cfeeff8ff58d2cf58f5dc40a60dc356db9504c2/p2p/src/peer.rs#L171-L182

xla (changes requested):

p2p/src/secret_connection.rs
melekes: > from the outside with anything but a sanctioned list of states

but Handshake is not exposed. How one can construct it from the outside?

p2p/src/secret_connection.rs
melekes: It turns out it's not so easy. In addition to the ciphers, there is also a nonce, which is incremented every time we send a packet with encrypted data. So I would need to extract 4 methods (encrypt, decrypt, read, write; read/write also depend on buffers), which doesn't seem like such a good idea to me.

xla: Sensible.

p2p/src/secret_connection.rs
xla: The example from above would only be meaningful, if Handshake would be allowed to be publicly constructed, so we can disregard.

xla (approved):
The nightly issue, is unrelated to this PR, should be safe to ignore.

🕖 📤 🆙 🌎

PR #671, opened 2020-11-12
Expose light client verifier to JavaScript 

romac:
See: #463

Basic prototype that exposes the verifier to JS via wasm-pack/wasm-bindgen.

Rust wrapper: https://github.com/informalsystems/tendermint-rs/blob/romac/wasm-verifier/wasm-client/src/lib.rs
Usage from JS: https://github.com/informalsystems/tendermint-rs/blob/romac/wasm-verifier/wasm-client/app/index.js

To make this work, one currently has to:

1. Run `wasm-pack build` from within the `wasm-client` folder
2. Move the `app` directory and run `npm start`
3. Open `http://localhost:8080/` and look at the console
4. It should display:

```
{
  "Ok": {
    "Invalid": {
      "NotWithinTrustPeriod": {
        "expires_at": "1970-01-15T00:00:03Z",
        "now": "2020-10-21T12:40:04.160328400Z"
      }
    }
  }
}
```

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

PR #812, opened 2021-02-13
Add tendermint-light-client-js crate

### LINKED ISSUE #463: light-client: wasm compilation
_Edited due to summarize the discussion bellow_

## Context

The light client core verification algorithm can verify an application state using only a sequence of headers. This has uses in the `light-node` where it can perform IO and fetch headers from full nodes. Another use for the light client verification is within the context of IBC where a full node will verify data from foreign chains before including it in local transactions. In this way the light client verification works as a pure function without IO but instead relies on the relayer to feed it data.

## Goal

We want to be able to compile our light-client verification to WASM such that it can be used by other chains to verify foreign data. This requires a build target that is independent from WASM incompatible libraries such as net2/socket2 or other IO operations.

## Status

- [x] The `tendermint` crate successfully builds for both `wasm32-unknown-unknown` and `wasm32-wasi` targets (#553)
- [x] The `light-client` crate successfully builds for both `wasm32-unknown-unknown` and `wasm32-wasi` targets (#553)
- [x] The `rpc` crate successfully builds for both `wasm32-unknown-unknown` and `wasm32-wasi` targets if its `client` feature is disabled.
- [ ] Implement a WASM-compatible RPC client in the `rpc` crate using the [`fetch` API exposed by `web-sys`](https://rustwasm.github.io/docs/wasm-bindgen/examples/fetch.html)
- [ ] npm package which exports the light client API to JavaScript via [`wasm-pack`](https://github.com/rustwasm/wasm-pack)

thanethomson:
Closes #463 

The vast majority of this PR was taken from #671 with very few changes (thanks @romac for breaking ground there!). I basically added a UI to allow for quick/easy tweaking of the inputs ~~, and will be looking at defining a custom data structure to provide a result back to JavaScript as opposed to serializing Rust's `Result` type (TBD).~~

**EDIT**: I'll leave the `verify` function parameters and return types as-is for now. We can always look at finding better ways of expressing the types in later versions.

:book: [Example README](https://github.com/informalsystems/tendermint-rs/blob/thane/463-wasm-light-client/light-client-js/examples/verifier-web/README.md)

* [x] Wrote tests
* [x] Updated CHANGELOG.md


webmaster128:
> to provide a result back to JavaScript as opposed to serializing Rust's `Result` type

Result can be used with wasm-bindgen. See https://rustwasm.github.io/docs/wasm-bindgen/reference/types/result.html. I never used it though.

thanethomson:
> Result can be used with wasm-bindgen. See https://rustwasm.github.io/docs/wasm-bindgen/reference/types/result.html. I never used it though.

I'm quite happy to leave it as-is, if that works for you @webmaster128. The result comes through as per the example in the README. It's just a little weird with all the different naming conventions, but it works :grin:

romac (commented):
Looks all good to me! Great job @thanethomson!

The one things we may still want to doo, is add a unit test which calls the exported `verify` function with some `JsValue`s.

thanethomson:
> The one things we may still want to doo, is add a unit test which calls the exported verify function with some JsValues.

Cool! @romac I've now added two basic tests of the `verify` method which are now tested in CI too against Chrome and Firefox.

light-client/src/components/verifier.rs
```diff
@@ -15,7 +15,7 @@ use serde::{Deserialize, Serialize};
 
 /// Represents the result of the verification performed by the
 /// verifier component.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
romac, line 18: Can we also add `Eq` here while we're at it?

romac (changes requested):
Great stuff! Could you also please fix the warnings in `light-client-js/src/utils.rs` with an `#[allow(dead_code)]` on `set_panic_hook`?

light-client/src/components/verifier.rs
thanethomson: Done!

thanethomson:
> Great stuff! Could you also please fix the warnings in `light-client-js/src/utils.rs` with an `#[allow(dead_code)]` on `set_panic_hook`?

Done :grin:

romac (approved):

PR #815, opened 2021-02-15
Property based tests of Time

shonfeder:
This is an initial effort towards adding an extensible foundation for PBT, as per #303.

The main intents of this contribution are:

- Evaluate whether PBT will add value to our test harness.
- Begin to establish a library of combinators for generating elementary arbitrary values (it is no accident that we begin with a combinators for time).

## Summary

- Add generators for arbitrary RFC3339 timestamps (as strings) and `DateTime<Utc>` structs (including within a specified range). These will be generally useful in PBT tests, any time we need to construct objects that contain time structs or timestamps. In a followup, I expect these strategies will be moved into their own library that can be used in different tests, which will considerably reduce the the amount of code in the test here. So the amount of code added to support this test is not particular to this alone, but is an investment in PBT more generally (I also expect generation of times will be one of the more involved constructions).
- Fixed the conversion `From` for `Time` to `SystemTime` to prevent panics on older dates. See  f19a6b1 for details.
 
 ---

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/tendermint/src/time.rs
+++ b/tendermint/src/time.rs
@@ -185,7 +185,7 @@ mod tests {
             // map the JSON `encoded_time` to back to the inital `time`.
             let encoded_time = serde_json::to_value(&time).unwrap();
             let decoded_time = serde_json::from_value(encoded_time.clone()).unwrap();
-            assert_eq!(time, decoded_time);
+            prop_assert_eq!(time, decoded_time);
         }
     }
 
--- a/tendermint/src/time.rs
+++ b/tendermint/src/time.rs
@@ -321,7 +321,7 @@ mod tests {
             // map the JSON `encoded_time` to back to the inital `time`.
             let time: Time = datetime.into();
             let json_encoded_time = serde_json::to_value(&time).unwrap();
-            let decoded_time: Time = serde_json::from_value(json_encoded_time.clone()).unwrap();
+            let decoded_time: Time = serde_json::from_value(json_encoded_time).unwrap();
             prop_assert_eq!(time, decoded_time);
         }
 
@@ -338,7 +338,7 @@ mod tests {
             // range. Tho we do incidentally test the inversion as well.
             let time: Time = stamp.parse().unwrap();
             let json_encoded_time = serde_json::to_value(&time).unwrap();
-            let decoded_time: Time = serde_json::from_value(json_encoded_time.clone()).unwrap();
+            let decoded_time: Time = serde_json::from_value(json_encoded_time).unwrap();
             prop_assert_eq!(time, decoded_time);
         }
     }
--- a/tendermint/src/time.rs
+++ b/tendermint/src/time.rs
@@ -41,7 +41,7 @@ impl From<Time> for Timestamp {
     fn from(value: Time) -> Self {
         // prost_types::Timestamp has a SystemTime converter but
         // tendermint_proto::Timestamp can be JSON-encoded
-        let prost_value = prost_types::Timestamp::from(value.to_system_time().unwrap());
+        let prost_value = prost_types::Timestamp::from(SystemTime::from(value));
         Timestamp {
             seconds: prost_value.seconds,
             nanos: prost_value.nanos,
@@ -77,12 +77,6 @@ impl Time {
     /// Return an RFC 3339 and ISO 8601 date and time string with 6 subseconds digits and Z.
     pub fn to_rfc3339(&self) -> String {
         timestamp::to_rfc3339_nanos(&self.0)
-    }
-
-    /// Convert [`Time`] to [`SystemTime`]
-    pub fn to_system_time(&self) -> Result<SystemTime, Error> {
-        let duration_since_epoch = self.duration_since(Self::unix_epoch())?;
-        Ok(UNIX_EPOCH + duration_since_epoch)
     }
 }
 
@@ -120,7 +114,7 @@ impl From<SystemTime> for Time {
 
 impl From<Time> for SystemTime {
     fn from(t: Time) -> SystemTime {
-        t.to_system_time().unwrap()
+        t.0.into()
     }
 }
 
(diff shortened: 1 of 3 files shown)


shonfeder:
We're currently failing due to testing times beginning from 0, but I see our library is currently set up to panic on any dates prior to the epoch. Is this desired behavior, indicating I should reduce the range of times generated?

thanethomson:
>  the `to_std` can't cope with negative values (so any time before the epoch)

This makes sense to me, and should be considered expected behaviour, I think. If we look at where all of the tendermint-rs code's being applied, nobody should be running it with a clock set to any time pre-epoch.

> `can_parse_rfc_3339_timestamps` is failing on the timestamp `0000-02-30T00:00:00-00:00`

Yeah I was wondering about this. Any valid RFC3339-compatible parser will fail on that timestamp because there'll never be 30 days in February, right? Same goes for other months that only have 30 days and not 31.

Is it possible to encode that in the testing? I suppose you'd then have to cater for leap years, where February has 29 days?

shonfeder:
Thanks, Thane! I can indeed make a modification to only construct time stamps for valid times. But I wanted to be sure that failure on syntactically correct timestamps that are for invalid times is desired. I'll make those constraints on the inputs! Thanks :pray:

shonfeder:
Just one thought re: 

> nobody should be running it with a clock set to any time pre-epoch.

I wanted to check on this *not* because I thought anyone would be legitimately running a clock before the epoch, but because I wanted to be sure this isn't a possible attack vector. Presumably, a malicious actor could send a message with an old timestamp. Is it a problem for a node using this if their code blows up with a panic in that case? What would be the outcome from a user perspective in that event? Or, do we have some other safeguards in place that will prevent an old (but valid) timestamp ever making it into deserialization?

More to the point: why do we panic here instead of surfacing an error and responding appropriately?

shonfeder:
The failure is from the docs job, and doesn't appear to be related to my changes:

```
error[E0658]: linking to associated items of raw pointers is experimental
Error:    |
   = note: see issue #80896 <https://github.com/rust-lang/rust/issues/80896> for more information
   = help: add `#![feature(intra_doc_pointers)]` to the crate attributes to enable
   = note: rustdoc does not allow disambiguating between `*const` and `*mut`, and pointers are unstable until it does

Error: aborting due to previous error
For more information about this error, try `rustc --explain E0658`.
error: could not document `pin-utils`
Error: could not document `pin-utils`
Caused by:
  process didn't exit successfully: `rustdoc --edition=2018 --crate-type lib --crate-name pin_utils /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/pin-utils-0.1.0/src/lib.rs --cap-lints allow -o /home/runner/work/tendermint-rs/tendermint-rs/target/doc --error-format=json --json=diagnostic-rendered-ansi -L dependency=/home/runner/work/tendermint-rs/tendermint-rs/target/debug/deps --crate-version 0.1.0` (exit code: 1)
warning: build failed, waiting for other jobs to finish...
Warning: error: build failed
Error: The process '/usr/share/rust/.cargo/bin/cargo' failed with exit code 101
```

tendermint/src/time.rs
```diff
@@ -171,14 +299,47 @@ mod tests {
...
+            ]
+        ) {
+            // ser/de of rfc3339 timestamps is safe if it never panics.
+            // This differes from the the inverse test in that we are testing on
+            // arbitrarily generated textual timestamps, rather than times in a
+            // range. Tho we do incidentally test the inversion as well.
+            let time: Time = stamp.parse().unwrap();
+            let json_encoded_time = serde_json::to_value(&time).unwrap();
+            let decoded_time: Time = serde_json::from_value(json_encoded_time).unwrap();
+            prop_assert_eq!(time, decoded_time);
```
shonfeder, line 342: This test ends up duplicating elements of the previous two, but it is ultimately testing a different property than the other two, as explained in the comment.

I'm not sure how I feel about this overlap. I could remove the `can_parse` test, since that is fully subsumed by this property, but I think we should maintain two separate tests, one to check for arbitrarily generated `DateTime` structs (as we do for `serde_from_value`) and one to check for arbitrarily generated time stamps (as we do here).

I"m open to other suggestions tho!

tendermint/src/time.rs
```diff
@@ -150,11 +144,145 @@ pub trait ParseTimestamp {
...
+    // timestamps. See https://github.com/chronotope/chrono/issues/537
+    fn min_time() -> DateTime<Utc> {
+        Utc.timestamp(-9999999999, 0)
+    }
+
+    fn max_time() -> DateTime<Utc> {
+        Utc.timestamp(99999999999, 0)
+    }
+
+    fn num_days_in_month(year: i32, month: u32) -> u32 {
```
thanethomson, line 169: What does this method add beyond something trivial like:

```rust
fn num_days_in_month(year: i32, month: u32) -> u32 {
    match month {
        2 => if year % 4 == 0 { 29 } else { 28 },
        1 | 3 | 5 | 7 | 8 | 10 | 12 => 31,
        4 | 6 | 9 | 11 => 30,
    }
}
```

shonfeder: It just offloads the calculation of dates to the datetime library we've imported for that very purpose :) 

We can compute it ourselves, but, assuming we are using the Gregorian calendar, it's slightly more involved than checking for divisibility by 4. See https://docs.microsoft.com/en-us/office/troubleshoot/excel/determine-a-leap-year#how-to-determine-whether-a-year-is-a-leap-year 

I don't feel super strongly that we must use chrono here. But as a matter of principle, if I'm using an external library to deal with some tricky domain (and time is tricky), I prefer to consistently delegate to the library wherever possible and not substitute in my own home-rolled algorithms. I find the latter is often a source of subtle bugs, since I usually don't haven't deeply internalized all the aspects of the domain (which is why I'm using the external lib in the first place :)).

thanethomson: Makes sense!

thanethomson (approved):
LGTM! :+1:

PR #820, opened 2021-02-23
Add RPC support for secure connections and HTTP proxies

### LINKED ISSUE #819: Support for https proxies in Tendermint RPC
During the stargate launch the earlier available RPC node was the cosmostation's node. I couldn't use it because it was behind an HTTPs proxy.

Tendermint RPC uses the Tendermint Address type which only takes TCP as a network protocol. As result the information to perform an HTTPS handshake on connection isn't passed to hyper when constructing a http connection.

It would be nice to be able to support https proxies.

I suggest we extend the Address type to natively support http and https.

thanethomson:
Closes #819

This PR is quite a large one unfortunately, because there's no easy way to test proxied connections or secure connections (or proxied secure connections :grin:). I therefore had to:

1. Add TLS support to the RPC clients (e.g. for when the RPC endpoint's behind a reverse proxy).
2. Add HTTP/S proxy support.
3. Introduce a CLI called `tendermint-rpc`, which provides a simple interface for interacting with RPC clients over `http://`, `https://`, `ws://` and `wss://`, as well as HTTP proxies, for testing purposes. This necessitated the development of a parser for subscription queries, because I wanted to test subscriptions from the command line too.

The documentation for the CLI is available here:

:book: [Rendered tendermint-rpc CLI README](https://github.com/informalsystems/tendermint-rs/blob/thane/819-rpc-tls/rpc/README.md#cli)

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


rpc/src/client/transport/http.rs
```diff
@@ -32,14 +33,50 @@ use std::io::Read;
...
-    port: u16,
+    inner: sealed::HttpClient,
+}
+
+impl HttpClient {
+    /// Construct a new Tendermint RPC HTTP/S client connecting to the given
+    /// URL.
+    pub fn new<U>(url: U) -> Result<Self>
+    where
+        U: TryInto<HttpClientUrl, Error = Error>,
```
romac, line 46: Nevermind, I just saw the two calls to `try_into` on lines 51 and 53, meaning that even if the given value is of type `HttpClientUrl`, the conversion into a `hyper::Uri` can still fail.

---

I think using `TryInto` here instead of `Into` or just the actual type is a bit anti-modular.

With `TryInto`, if I already hold a value of type `HttpClientUrl`, I will still be forced to deal with the potential error returned by `new` because of the inner call to `try_into` even if it cannot fail.

While I agree that it's a nice way to keep the calling code short, I think it should be the responsibility of the caller to pass in a valid `HttpClientUrl` (or perhaps a type which implements an infallible `Into<HttpClientUrl>`.

By changing the bound to `Into` (or by just using `HttpClientUrl` since we don't have any `From` instance for this type at the moment), we would go from:

```rust
let literal_case = HttpClient::new("https://localhost:9091")?;

let url: HttpClientUrl = ???;
let known_good_value = HttpClient::new(known_good_value)?;
```

to

```rust
let literal_case = HttpClient::new("https://localhost:9091".parse()?); // moved the error handling one level up

let url: HttpClientUrl = ???;
let known_good_value = HttpClient::new(known_good_value); // no need to deal with error here
```

romac (approved):
💯

tony-iqlusion:
Note: tested and working on our end for an HTTP CONNECT egress proxy. Thank you! 🎉

PR #825, opened 2021-03-03
p2p: MConnection transport

### LINKED ISSUE #814: p2p: implement MConnection transport
Implement the transport, which mimics the [tendermint-go's MConnection](https://github.com/tendermint/spec/blob/master/spec/p2p/connection.md#p2p-multiplex-connection) - i.e. multiplexes messages (from different reactors) over a single TCP stream.

melekes:
Closes #814 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


p2p/src/transports/mconnection/mconnection.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    fn remote_addr(&self) -> SocketAddr {
+        self.stream.peer_addr().unwrap()
+    }
+
+    fn open_bidirectional(
+        &self,
+        _stream_id: &StreamId,
+    ) -> Result<(&Self::Read, &Self::Write), Self::Error> {
+        Ok((&self.secret_connection, &self.secret_connection))
```
melekes, line 104: wonder if I should use smth like Rc or Arc to share secret_connection between multiple callers...

melekes:
NOTE: this is based on @xla PR https://github.com/informalsystems/tendermint-rs/pull/708, so we should probably merge it before this lands.

p2p/src/transports/mconnection.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    fn remote_addr(&self) -> SocketAddr {
+        self.stream.peer_addr().unwrap()
+    }
+
+    fn open_bidirectional(
+        &self,
+        _stream_id: &StreamId,
+    ) -> Result<(&Self::Read, &Self::Write), Self::Error> {
+        Ok((&self.secret_connection, &self.secret_connection))
```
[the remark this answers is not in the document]
xla: Thinking more about it, the implementation of `MConnection` should take care of the keep-alive semantics and unpack the `PacketMessage` into prefixed streams through `open-bidirectional. Specifically this packet [`Sum`](Thinking more about it, the implementation of `MConnection` should take care of the keep-alive semantics and unpack the `PacketMessage` into prefixed streams through `open-bidi) should be handled in here and the `PacketMsg` case should be unpacked and send to the receivers of a specific stream.

p2p/src/transport.rs
```diff
@@ -1,18 +1,26 @@
...
 use eyre::Result;
 
+use super::secret_connection;
+
 pub struct BindInfo {
     pub addr: SocketAddr,
-    pub advertise_addrs: Vec<SocketAddr>,
-    pub public_key: PublicKey,
+    pub private_key: ed25519::Keypair,
+    pub protocol_version: secret_connection::Version,
```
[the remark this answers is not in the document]
melekes: Good point! Let's leave it in doubt for now. I think at some point we'll just drop amino support and remove amino related code.

p2p/src/transports/mconnection.rs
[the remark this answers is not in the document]
melekes: > This is not quite what we want. Opening a bidirectional stream means that all bytes read and wrote from and to it will be only for messages of that stream id. In the case of MConn that is symmetric with ChanneldID.

Right, so we'll need to wrap `&self.secret_connection` into something that takes `streamId` into account (i.e. unpacks the message, checks if streamId == thisStreamId and, depending on the result, proxies the message or drops it).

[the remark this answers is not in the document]
melekes: > PacketMessage

where I can find it? or it does not exist yet?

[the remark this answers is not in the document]
xla: All the p2p messages are in here https://docs.rs/tendermint-proto/0.18.1/tendermint_proto/p2p/packet/enum.Sum.html

The one which needs to be used for the multiplexing is https://docs.rs/tendermint-proto/0.18.1/tendermint_proto/p2p/struct.PacketMsg.html it has the channel id and the bytes that need to be pass around.

p2p/src/transport/mconnection.rs
```diff
@@ -0,0 +1,384 @@
...
+        // The socket will be closed when the MConnectionTransport is dropped.
+        Ok(())
+    }
+}
+
+impl Read for VirtualStream {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        loop {
+            let mut tmp_buf = Vec::new();
+            // FIXME: we should probably read in loop until decode_length_delimited succeds
```
melekes, line 266: question to all of the team I guess: 

why we're using prost lib and not https://github.com/stepancheg/rust-protobuf, where decode_length_delimited_from_reader exists and we don't need to implement it ourselves?

xla: @greg-szabo @thanethomson @tarcieri

liamsi: I can only comment on this from a historical perspective: Prost was originally chosen (back in the KMS) as it was much easier to modify and use to quickly hack together an amino-compatible implementation in Rust for the few messages that the KMS needed to understand. Also, it seemed like the most mature and maintained choice back then.
That said, we kinda sticked to prost even here because it was still nice that you were not forced to use protobuf files and generated (difficult to read) rust code derived from these proto-files. Instead you could just use the derive marco and you wouldn't really need to care about protobuf defineitons. I'm not sure these reasons are still valid (probably not). If there are pros than simply the method `decode_length_delimited_from_reader` then it might be a good time to reconsider choices. if the only reason is the reader method, why not simply implement once and upstream it to prost. But I guess this is very orthogonal to what this PR is trying to achieve and probably deserves its own issue.

thanethomson: Yeah it might be worthwhile to revisit this again sometime soon. Does `rust-protobuf` allow for customization of `serde` options at the struct and field-level? Right now we make extensive use of this in our [proto-compiler](https://github.com/informalsystems/tendermint-rs/tree/v0.18.1/tools/proto-compiler) tool.

p2p/src/lib.rs
```diff
@@ -30,4 +30,4 @@ pub mod secret_connection;
 mod message;
 mod peer;
 mod supervisor;
-mod transport;
+pub mod transport;
```
melekes, line 33: note: this is needed for doc test below to succeed

tendermint/src/node/id.rs
```diff
@@ -104,7 +104,7 @@ impl TryFrom<PublicKey> for Id {
     fn try_from(pk: PublicKey) -> Result<Self, Self::Error> {
         match pk {
             PublicKey::Ed25519(ed25519) => Ok(Id::from(ed25519)),
-            _ => Err(Kind::UnsupportedKeyType.into()),
+            // _ => Err(Kind::UnsupportedKeyType.into()),
```
melekes, line 107: this was giving me unreachable code error

xla: This happens when compiled without all features enabled, otherwise it should be correct when the tendermint crate is compiled with its features as one of the variants is guarded:

https://github.com/informalsystems/tendermint-rs/blob/2609b60755114c73ab6b5d4b169463e1de95b243/tendermint/src/public_key.rs#L49-L57

melekes: I would expect this to be specified in the readme or `Makefile`. @thanethomson should I update the readme or create a Makefile target?

p2p/src/transport/mconnection.rs
```diff
@@ -0,0 +1,520 @@
...
+        virtual_streams: Arc<RwLock<HashMap<StreamId, Sender<PacketMsg>>>>,
+    ) {
+        loop {
+            // If any stream is trying to send a message, go for it.
+            if let Ok(msg) = rx.try_recv() {
+                let mut buf = Vec::new();
+                msg.encode_length_delimited(&mut buf)
+                    .expect("encode to always succeed");
+                if let Err(e) = secret_connection.write_all(&buf) {
+                    println!("can't write to connection: {}. exiting...", e);
```
melekes, line 153: @thanethomson @xla what is the convention around logging inside the "command loop" running in a separate thread? how do I inform the caller about any errors? Should I return a channel and write any errors to it?

xla: Is this a terminal condition for the entire transport?

melekes: No, just for that particular connection.

xla: This should be surfaced through one of the public methods of `Connection` so whenever the owner tries to interact with it, by opening a stream or reading from a stream.

xla:
@melekes Checking in on the status here, is this good for another review?

melekes:
> @melekes Checking in on the status here, is this good for another review?

Yep

xla (changes requested):
This is well advanced and in many ways we could just merge it in and then iterate on it. If that is the preferred way forward to avoid further drag we should do that.

Two observations:

I think my naive view on that the streams should just implement `Read` and `Write` was ill-informed. It muddies the interface of the transport as ultimately my current believe is that the transport should be in charge of the framing and therefore should emit entire `Vec<u8>`. While that might not be the optimal solution if down the line we have to deal with large streams of data for a message, we should look at that problem when it gets to that and maybe look to other solutions than piggybacking it on the message abstraction. A reasonable new interpretation of the `Stream` could look like this:

``` rust
pub trait StreamSend {
    fn send(msg: Vec<u8>) -> Result<()>;
}

pub trait Connection: Send {
    type Error: std::error::Error + Send + Sync + 'static;
    type StreamRecv: Iterator<Item = Result<Vec<u8>>> + Send;
    type StreamSend: StreamSend;

    fn advertised_addrs(&self) -> Vec<SocketAddr>;
    fn close(&self) -> Result<()>;
    fn local_addr(&self) -> SocketAddr;
    fn open_bidirectional(
        &self,
        stream_id: StreamId,
    ) -> Result<(Self::StreamRecv, Self::StreamSend), Self::Error>;
    fn public_key(&self) -> PublicKey;
    fn remote_addr(&self) -> SocketAddr;
}
```

Secondly in the current form a lot of the current error cases/termination reasons are swallowed and only surfaced via `println`s. IMHO we should aim from the start at very honest code and make the effort to get these cases dealt with. The reason I'm of that believe is, that my assumption is it will positively inform the design when we have to deal with all scenarios. Likely the complexity will increase because of it, but the foundational nature of the code asks for it.

p2p/src/transport/mconnection.rs
```diff
@@ -0,0 +1,520 @@
...
+                // if we want to support old chains that are using Amino encoding
+                // if not, we should just hardcode the value
+                protocol_version: Version::V0_34,
+            },
+        ))
+    }
+
+    /// Noop.
+    fn shutdown(&self) -> Result<()> {
+        // The socket will be closed when the MConnectionTransport is dropped.
```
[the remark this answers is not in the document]
melekes: > The packets in tendermint p2p have an EOF under which condition should those be sent?

It's used when data size is bigger than the allowed limit (per one package). we then must split the data into N pieces and only send `eof: true` when data is exhausted.

p2p/src/transport/mconnection.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    /// Advances the iterator and returns the next `MConnection`.
+    fn next(&mut self) -> Option<Result<MConnection>> {
+        let public_key = self.private_key.public;
+
+        match self
+            .tcp_listener
+            .incoming()
+            .next()
+            .expect("Incoming to always return Some") // it's safe to unwrap here because Incoming never returns None
```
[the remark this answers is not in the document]
melekes: will add a link to std docs

melekes:
>  think my naive view on that the streams should just implement Read and Write was ill-informed. It muddies the interface of the transport as ultimately my current believe is that the transport should be in charge of the framing and therefore should emit entire Vec<u8>.

Agree with the reasoning here. One thing is maybe change `Vec<u8>` to `&[u8]`.

> Secondly in the current form a lot of the current error cases/termination reasons are swallowed and only surfaced via printlns.

I am not sure this is correct. printlns are there for additional info. if the loop terminates, all streams will start returning errors to the client.

tony-iqlusion:
> What are the advantages of using the &[u8] over a Vec<u8>?

Another alternative to consider here is [`Bytes`](https://docs.rs/bytes/), which provides reference counted buffers. It's pretty commonly used in the Tokio/`async` ecosystem, but could work here as well.

Issue #822, opened 2021-02-25
Explore interaction of model-based and property-based testing


andrey-kuprianov:
Following the efforts to apply PBT in #709 (see [backward.rs](https://github.com/informalsystems/tendermint-rs/blob/v0.18.1/light-client/tests/backward.rs)), #792, #815, as well as discussion within the MBT team, we need to understand how to properly integrate MBT and PBT.

The high-level idea is not to reinvent the wheel by writing our own fuzzers, and instead to leverage the PBT infrastructure as done by `proptest`. In that, the test cases that span over several computation steps will be generated by MBT, and then mutated by PB strategies tendermint-rs datastructures. The main complication here is how to correctly express what changes should happen to the expected test outcome when the intermediate computation steps are mutated.

As the first step for the exploration we have decided to replace one of the manual fuzzing steps as done in [model_based.rs](https://github.com/informalsystems/tendermint-rs/blob/e6a31686300915175be2279a95636869eda31312/light-client/tests/model_based.rs#L611-L631) with the `proptest`-based one, in particular to replace fuzzing of Time.

For successful completion of this issue:
- [ ] write a PB-test that accepts the MB-test case, and fuzzes time fields in various test case blocks
- [ ] fuzzing should be done using combinators over the `proptest` strategies developed in #815
- [ ] the PB-test should explore different options wrt. how the fuzzed time relates to other times in the test case:
   - [ ] to the trusting period
   - [ ] to the time of neighbouring blocks
 Depending on that the expected verdict should be modified accordingly.
- [ ] (optional) explore possible new macros to concisely express such MBT-specific PB-tests/strategies.

Issue #821, opened 2021-02-23
Add property based tests for the VotingPower calculator


shonfeder:
Followup to #333 

In the exploratory work of #815, our focused application of PBT surfaced an edge case that could produce fatal errors. Consequently, in discussion with @thanethomson, we decided it was worth investing in further application of PBT, both to improve the depth of tests and to continue building up a library of combinators to expedite writing PBT of future development.

The proposed target for the next step of this work is the [`VotingPowerCalculator`](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/operations/voting_power.rs#L105). 

This work item decomposes into two principle parts:

1. Defining combinators for generating arbitrary inputs to `voting_power_in`.
2. Identifying the essential properties of this computation that should be checked, and writing the tests.

**What's the definition of "done" for this issue?**

1. There are combinators for generating arbitrary values of the following types:
    -   [ ] **SignedHeader:** /src/block/signed_header.rs
        -   [ ] **Header:** /src/block/header.rs
            -   [x]  **Version:** /src/block/header.rs
            -   [x] **Height:** /src/block/height.rs
            -   [x] **Time:** /src/time.rs
            -   [ ] **Hash:** /src/hash.rs
            -   [ ] **AppHash:** /src/hash.rs
        -   [ ] **Commit:** /src/block/commit.rs
    -   [ ] **ValidatorSet:** /src/validator.rs
        -   [ ] **Info:** /src/validator.rs
            -   [ ] **ID:** /src/account.rs
            -   [ ] **PublicKey:** /src/public_key.rs
            -   [ ] **Power:** /src/vote/power.rs
            -   [ ] **ProposerPriority:** /src/validator.rs
        -   [ ] **TrustThresholdFraction:** /src/trust_threshold.rs
2. There are property based tests for essential properties of the calculation written, and these are tagged with the logical unit they are derived from.

shonfeder:
Yesterday we talked about where these generators might live, given that:

- they will likely need to be used by tests that outside of the tendermint crate
- we don't want to bloat tendermint with code that is only used in tests.

@romac advised that the best course of action would be to put the generators in a model guarded by a disjunctive feature that's satisfied either by `test` or `proprest` (or whatever other name we'd like here) using https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute. E.g., 

```rust
#[cfg(any(test, proptest)]
```

[...]

PR #829, opened 2021-03-17
Move time proptest generators into their own crate

### LINKED ISSUE #828: Create a crate for PBT combinators
We'll want to share the PBT generation combinators between tests (see #822 and #821).

The obvious (and [preferred?](https://stackoverflow.com/a/44545091/1187277)) way to share code between tests is by creating a new crate for the shared code, which is then [added to the dev dependencies](https://stackoverflow.com/a/33025972/1187277), but not published.

As an aside, some of the generators may actually end up being generally useful, in which case we might consider publishing the crate.

## Acceptance Criteria

- [ ] A new crate is added for proptest generators
- [ ] The proptest generators added in #815 are moved out of the test module where they were first introduced and into a module int he new crate.

shonfeder:
Closes #828

This allows us to use the timestamp and DateTime generators in any other
crate in our workspace. It should help advance #822 and #821.

-------

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/pbt-gen/Cargo.toml
+++ b/pbt-gen/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "pbt-gen"
+name = "tendermint-pbt-gen"
 version = "0.1.0"
 authors = ["Shon Feder <shon@informal.systems>"]
 edition = "2018"
@@ -10,7 +10,13 @@ description = """
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+
+default = ["time"]
+
+time = ["chrono"]
+
 [dependencies]
 
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4", features = ["serde"], optional = true}
 proptest = "0.10.1"
--- a/pbt-gen/Cargo.toml
+++ b/pbt-gen/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "pbt-gen"
+version = "0.1.0"
+authors = ["Shon Feder <shon.feder@gmail.com>"]
+edition = "2018"
+description = """
+            An internal crate provoding proptest generators used accross our
+            crates and not depending on any code internal to those crates.
+            """
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+
+chrono = { version = "0.4", features = ["serde"] }
+proptest = "0.10.1"
--- a/pbt-gen/Cargo.toml
+++ b/pbt-gen/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
 name = "pbt-gen"
 version = "0.1.0"
-authors = ["Shon Feder <shon.feder@gmail.com>"]
+authors = ["Shon Feder <shon@informal.systems>"]
 edition = "2018"
 description = """
-            An internal crate provoding proptest generators used accross our
+            An internal crate providing proptest generators used across our
             crates and not depending on any code internal to those crates.
             """
 
--- a/pbt-gen/src/time.rs
+++ b/pbt-gen/src/time.rs
@@ -1,5 +1,5 @@
-//! [proptest](https://github.com/AltSysrq/proptest) generators for time-like
-//! objects.
+//! Provides [proptest](https://github.com/AltSysrq/proptest) generators for
+//! time-like objects.
 
 use std::convert::TryInto;
 
--- a/pbt-gen/src/time.rs
+++ b/pbt-gen/src/time.rs
@@ -76,7 +76,7 @@ prop_compose! {
     ///          Utc.ymd(2006, 2, 24).and_hms(0,0,0), // DOD
     ///        ),
     ///     ) {
-    ///       prop_assert!(time_in_butlers_lifespan == time_in_butlers_lifespan)
+    ///       prop_assert!(time_in_luxemburgs_lifespan != time_in_butlers_lifespan)
     ///     }
     /// }
     /// ```
--- a/pbt-gen/src/time.rs
+++ b/pbt-gen/src/time.rs
@@ -1,0 +1,182 @@
+//! [proptest](https://github.com/AltSysrq/proptest) generators for time-like
+//! objects.
+
+use std::convert::TryInto;
+
+use chrono::{DateTime, NaiveDate, TimeZone, Timelike, Utc};
+use proptest::prelude::*;
+
+/// Any higher, and we're at seconds
+pub const MAX_NANO_SECS: u32 = 999_999_999u32;
+
+/// The most distant time in the past for which chrono produces correct
+/// times from [Utc.timestamp](chrono::Utc.timestamp).
+///
+/// See https://github.com/chronotope/chrono/issues/537.
+///
+/// ```
+/// use pbt_gen;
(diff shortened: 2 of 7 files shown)


shonfeder:
Tagging in @Shivani912 for review since I think this might help with code reuse in ongoing work?

pbt-gen/src/time.rs
```diff
@@ -0,0 +1,182 @@
...
+    ///        time_in_luxemburgs_lifespan in pbt_gen::time::arb_datetime_in_range(
+    ///          Utc.ymd(1871, 3, 5).and_hms(0,0,0), // DOB
+    ///          Utc.ymd(1919, 1, 15).and_hms(0,0,0), // DOD
+    ///        ),
+    ///        time_in_butlers_lifespan in pbt_gen::time::arb_datetime_in_range(
+    ///          Utc.ymd(1947, 6, 22).and_hms(0,0,0), // DOB
+    ///          Utc.ymd(2006, 2, 24).and_hms(0,0,0), // DOD
+    ///        ),
+    ///     ) {
+    ///       prop_assert!(time_in_butlers_lifespan == time_in_butlers_lifespan)
```
Shivani912, line 79: isn't this expected to be not equal?
```suggestion
    ///       prop_assert!(time_in_butlers_lifespan != time_in_butlers_lifespan)
```

Shivani912 (changes requested):
Thanks @shonfeder ! It's great we now have a crate for this. I was indeed looking forward to it :)) 

I have one small comment, otherwise looks good to me!

pbt-gen/src/time.rs
shonfeder: Indeed! I messed this up. I also put the same lifespan twice, I now see. Thanks for catching my mistake!

Shivani912: Ouch, I didn't see that 😄

Shivani912 (dismissed):

pbt-gen/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "pbt-gen"
+version = "0.1.0"
+authors = ["Shon Feder <shon.feder@gmail.com>"]
```
thanethomson, line 4: Would you prefer to use your Gmail address over your Informal address here?

pbt-gen/src/time.rs
```diff
@@ -0,0 +1,182 @@
...
+use chrono::{DateTime, NaiveDate, TimeZone, Timelike, Utc};
+use proptest::prelude::*;
+
+/// Any higher, and we're at seconds
+pub const MAX_NANO_SECS: u32 = 999_999_999u32;
+
+/// The most distant time in the past for which chrono produces correct
+/// times from [Utc.timestamp](chrono::Utc.timestamp).
+///
+/// See https://github.com/chronotope/chrono/issues/537.
```
thanethomson, line 15:
```suggestion
/// See <https://github.com/chronotope/chrono/issues/537>.
```

pbt-gen/src/time.rs
```diff
@@ -0,0 +1,182 @@
...
+/// assert_eq!(pbt_gen::time::min_time().to_string(), "1653-02-10 06:13:21 UTC".to_string());
+/// ```
+pub fn min_time() -> DateTime<Utc> {
+    Utc.timestamp(-9999999999, 0)
+}
+
+/// The most distant time in the future for which chrono produces correct
+/// times from [Utc.timestamp](chrono::Utc.timestamp).
+///
+/// See https://github.com/chronotope/chrono/issues/537.
```
thanethomson, line 29:
```suggestion
/// See <https://github.com/chronotope/chrono/issues/537>.
```

pbt-gen/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "pbt-gen"
```
thanethomson, line 2: We typically prefix all crates' names within this repository with `tendermint-`, like:

```suggestion
name = "tendermint-pbt-gen"
```

thanethomson (commented):
Left a couple of minor comments. I think having this as a separate crate is a great idea! :+1:

pbt-gen/Cargo.toml
shonfeder: I thought we might omit the long prefix since it's just for internal use, but happy to follow the convention if that matters here.

It will make it awkward to call functions using the current convention in the examples (where the short crate name is used as an accessor, like `pbt_gen::time::arb_datetime()`. Would you recommend I rename the `time` module to something less general, or do you think we'll end up just importing the particular functions directly?

pbt-gen/Cargo.toml
shonfeder: Ah no, that was an oversight due to the template generation. Thanks.

pbt-gen/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+authors = ["Shon Feder <shon.feder@gmail.com>"]
+edition = "2018"
+description = """
+            An internal crate provoding proptest generators used accross our
```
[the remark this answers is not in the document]
shonfeder: Thanks!

pbt-gen/Cargo.toml
romac: I think both names are fine as long as it does not depend on anything from `tendermint`, in case I would argue for the `tendermint-` prefix.

Regarding the issue of short crate name for easy invocations, you can rename a crate when you specify it as a dependency in `Cargo.toml`. The following will add the `tendermint-pbt-gen` crate as a dependency and let you use it as `pbt_gen` in the code.

```toml
[dependencies]
pbt-gen = { package = "tendermint-pbt-gen", version = "0.1" }
```

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml

pbt-gen/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+description = """
+            An internal crate providing proptest generators used across our
+            crates and not depending on any code internal to those crates.
+            """
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+
+chrono = { version = "0.4", features = ["serde"] }
```
romac, line 15: Can we feature-guard the `chrono` dev and associated proptest strategies? If this crate grows or end up being published, it would be nice to only pull in whatever strategies (and their dependencies) are actually needed by the downstream project.

romac: I guess at the moment there is only a single set of strategies, so it does not matter much. But it could still be feature-guarded and enabled by default.

pbt-gen/Cargo.toml
shonfeder: I think this cannot depend on anything from `tendermint`, because `tendermint` depends on it (as a dev dependency), and, iiuc, crate dependencies cannot be circular?

Nice to learn about dep renaming! Thanks :) I think it may be clearer to just rename the crate on `use` tho. I'll try that and see how that sits with folks.

pbt-gen/Cargo.toml
shonfeder: Done!

thanethomson (approved):
LGTM! :+1:

PR #837, opened 2021-03-25
Fix consensus state vote summary deserialization

### LINKED ISSUE #836: Intermittent failure in CI when deserializing consensus state vote summaries
The most recent example of this is here: https://github.com/informalsystems/tendermint-rs/pull/834/checks?check_run_id=2192937025

I've managed to figure out that sometimes the vote type is encoded as a number (`1` or `2`), and sometimes it's encoded as a string (`SIGNED_MSG_TYPE_PREVOTE`), and the code currently only caters for the former.

**Steps to reproduce**

Try deserializing the string causing the CI failure:

```
Vote{0:2DA21E474F57 384/00/SIGNED_MSG_TYPE_PREVOTE(Prevote) 8FA9FD23F590 2987C33E8F87 @ 2021-03-25T12:12:03.693870115Z}
```

When we can parse these kinds of vote summaries.

thanethomson:
Closes #836 

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/rpc/src/endpoint/consensus_state.rs
+++ b/rpc/src/endpoint/consensus_state.rs
@@ -2,7 +2,6 @@
 
 use crate::{Error, Method};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use std::convert::TryFrom;
 use std::fmt;
 use std::str::FromStr;
 use subtle_encoding::hex;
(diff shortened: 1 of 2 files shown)


rpc/src/endpoint/consensus_state.rs
```diff
@@ -384,4 +389,11 @@ mod test {
         };
         assert_eq!(summary.to_string(), "Vote{0:000001E443FD 1262197/00/1(Prevote) 634ADAF1F402 7BB974E1BA40 @ 2019-08-01T11:52:35.513572509Z}");
     }
+
+    #[test]
+    fn deserialization_regressions() {
+        for s in DESERIALIZATION_REGRESSIONS {
+            let _ = VoteSummary::from_str(s).unwrap();
```
romac, line 396: Can we test here that the parse `VoteSummary` contains the expected data?

romac (changes requested):
Good stuff :)

rpc/src/endpoint/consensus_state.rs
thanethomson: Sure! :+1:

shonfeder (approved):
Looks like the requested change her was addressed, so I'll provide the green check :)

romac (approved):
Looks good!

We might want to revisit this code altogether at some point, perhaps to rewrite it to use parser combinators or something like that as the format is a bit too involved for manual parsing. But in the meantime, with the new tests, we have some confidence that this works as expected and can refactor later.

thanethomson:
> We might want to revisit this code altogether at some point, perhaps to rewrite it to use parser combinators or something like that as the format is a bit too involved for manual parsing. But in the meantime, with the new tests, we have some confidence that this works as expected and can refactor later.

Ideally we'd avoid this kind of custom serialization format altogether from the server side :grin: It should be something that's encoded properly in JSON/Protobuf, and not a custom string format.

romac:
> Ideally we'd avoid this kind of custom serialization format altogether from the server side 😁 It should be something that's encoded properly in JSON/Protobuf, and not a custom string format.

💯

Issue #252, opened 2020-05-04
Research and implement tarpaulin instead of grcov


greg-szabo:
Not a done deal, but according to Tony, tarpaulin has less false negatives.
https://github.com/xd009642/tarpaulin

Liamsi:
related: https://github.com/informalsystems/tendermint-rs/issues/250

thanethomson:
I'm assuming this will be superseded by #811?

thanethomson:
Since our code coverage seems broken (and seems like it has been broken for some time), we should consider fixing it or integrating tarpaulin-based coverage calculation sooner rather than later.

@romac are you planning on taking #811 further? What are the pros/cons at this point of going with `cargo-llvm-cov`? I see it's still labelled as being "experimental" right now?

romac:
> are you planning on taking #811 further?

I wish I could but I don't have the bandwidth at the moment.

> What are the pros/cons at this point of going with `cargo-llvm-cov`?

Pros:
- the coverage should be much more precise
Cons:
- at the moment I was not able to get codecov.io to understand it's output on anything other than toy crates
- `-Zinstrument-coverage` is still pretty new so there might be issues

> I see it's still labelled as being "experimental" right now?

The wrapper itself is marked as experimental, but it basically only wraps `cargo build` (adding nightly-only options), `llvm-profdata`, and `llvm-cov`.

romac:
By the way, we use tarpaulin on ibc-rs but have seen no real improvements (eg. the coverage still jumps around by 30% sometimes, etc) hence why I was looking into source-based code coverage.

thanethomson:
Currently superseded by #840.

We can reopen this issue again in future if we want to reconsider using tarpaulin.

PR #840, opened 2021-03-25
Source-based coverage

vitorenesduarte:
With this PR, we start tracking coverage using LLVM's source-based coverage: https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html

- pin nightly version to avoid spurious errors (it's unlikely we need to bump this often, unless we need new compiler features)
- pin grcov version to avoid https://github.com/mozilla/grcov/issues/555
- set `LLVM_PROFILE_FILE="%m.profraw"` to avoid overwrites of coverage data (which occur when several test binaries are run for the same crate): https://github.com/rust-lang/rust/issues/79899
  - maybe the 30% coverage decreases (https://github.com/informalsystems/tendermint-rs/pull/839#issuecomment-807115109 and https://github.com/informalsystems/tendermint-rs/issues/252#issuecomment-805623519) are due to something like this

With this, coverage is at 67.7%, which seems to be an all-time high 😄 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/.github/actions-rs/grcov.yml
+++ b/.github/actions-rs/grcov.yml
@@ -1,6 +1,0 @@
-branch: true
-ignore-not-existing: true
-llvm: true
-output-type: lcov
-output-file: ./lcov.info
-prefix-dir: /home/user/build/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,14 +153,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
+          profile: minimal
           toolchain: nightly
           override: true
-      - uses: actions-rs/cargo@v1
-        with:
-          command: clean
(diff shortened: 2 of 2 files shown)


thanethomson (approved):
Let's give this a try :+1: I'm looking forward to the results. Thanks @vitorenesduarte!

romac:
Awesome, thanks a lot @vitorenesduarte !

marco-c:
@vitorenesduarte could you try the latest version of grcov, with LLVM_PROFILE_FILE set to `%p-%m.profraw` instead of `%m.profraw`?

vitorenesduarte:
> @vitorenesduarte could you try the latest version of grcov, with LLVM_PROFILE_FILE set to `%p-%m.profraw` instead of `%m.profraw`?

@marco-c trying it here: https://github.com/informalsystems/tendermint-rs/pull/864

PR #842, opened 2021-03-25
Add TLS support for the Light Client

### LINKED ISSUE #835: TLS support for Light Client
At the moment, the Light Client makes use of [`tendermint::net::Address`](https://github.com/informalsystems/tendermint-rs/blob/630ca026009dd9be335656dac2df29b45d46715e/tendermint/src/net.rs#L24) to keep track of RPC addresses, but this doesn't allow for the full range of RPC functionality introduced in #820 (e.g. TLS and proxy support).

We should update the Light Client to use the new URL type, since it's not really peer addresses the Light Client's interested in, but their RPCs' URLs.

Some preliminary work was done in [this commit](https://github.com/informalsystems/tendermint-rs/pull/834/commits/14ad69f02839dabd7c8937d2ba885f0aa2783f94), but I had to roll it back because it broke the WASM build.

Once we can successfully run the Light Node against nodes with secure endpoints (i.e. HTTPS).

thanethomson:
Closes #835.

It additionally introduces a little more info-level logging in the Light Node.

I've successfully manually tested the Light Node against two different Cosmos Hub validators, both accessed via HTTPS.

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


light-node/src/config.rs
```diff
@@ -41,7 +41,8 @@ pub struct LightNodeConfig {
 pub struct LightClientConfig {
     /// Address of the Tendermint fullnode to connect to and
     /// fetch LightBlock data from.
-    pub address: tendermint::net::Address,
+    #[serde(with = "tendermint_proto::serializers::from_str")]
```
romac, line 44: Can we rather have a custom `Serialize/Deserialize` instance for `tendermint_rpc::Url`, so that one does not have to provide such an attribute everywhere we want to store it within a serializable struct?

romac (changes requested):
Nice! Just left a single request.

light-node/src/config.rs
thanethomson: Makes sense, will do! :+1:

romac (approved):

PR #844, opened 2021-03-26
Remove kvstore-integration-latest test from CI

### LINKED ISSUE #843: kvstore-integration test hanging in CI
The vast majority of CI runs this past while have resulted in the `kvstore-integration-latest` test hanging, and it's unclear as to why unless you watch the run take place.

It looks like there's some kind of serialization incompatibility with the latest versions of Tendermint, and this causes one of the subscription tests to hang indefinitely. A reasonable solution here would be to impose a time limit on the subscription test's execution.

Here's some of the test output:

```
...
running 13 tests
test rpc::abci_query ... FAILED
test rpc::abci_info ... FAILED
test rpc::block ... FAILED
test rpc::block_results ... FAILED
test rpc::commit ... FAILED
test rpc::blockchain ... FAILED
test rpc::consensus_state ... ok
test rpc::genesis ... ok
test rpc::health ... ok
test rpc::net_info ... ok
test rpc::status_integration ... FAILED
test rpc::transaction_subscription ... FAILED
test rpc::subscription_interface ... test rpc::subscription_interface has been running for over 60 seconds
```

**Steps to reproduce**

Submit a PR and then watch the `kvstore-integration` test's execution.

When this test either passes or fails within a reasonable period of time (1-5mins). We can look at fixing the serialization compatibilities in a future PR targeting `tendermint/tendermint:latest`.

thanethomson:
Closes #843 

The test's still most likely going to be valuable, but just not on each PR necessarily.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -125,28 +125,6 @@ jobs:
         env:
           RUST_LOG: debug
 
-  kvstore-integration-latest:
-    runs-on: ubuntu-latest
-    services:
-      tendermint:
-        image: tendermint/tendermint:latest
-        ports:
-          - 26656:26656
-          - 26657:26657
-          - 26660:26660
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          override: true
-      - uses: actions-rs/cargo@v1
-        # Don't fail CI due to integration failures on unstable/unreleased versions
-        continue-on-error: true
-        with:
-          command: test-all-features
-          args: --manifest-path tools/kvstore-test/Cargo.toml
-
   nightly-coverage:
     runs-on: ubuntu-latest
     steps:


xla (approved):
📴 🎴 🙁 🕔

.github/workflows/test.yml
```diff
@@ -125,28 +125,6 @@ jobs:
         env:
           RUST_LOG: debug
 
-  kvstore-integration-latest:
-    runs-on: ubuntu-latest
-    services:
-      tendermint:
-        image: tendermint/tendermint:latest
```
marbar3778, line 132: Could you not add --mode validator as an arg? this will make it work again.

thanethomson: We could. But @cmwaters mentioned you'd be changing this again sometime soon - would that break it again?

thanethomson: Up to now it's been a nice-to-have to test against `tendermint/tendermint:latest` in every PR (we don't fail CI if the test fails). We do, however, still want to continue testing against the latest stable releases (the `kvstore-integration-stable` test).

I was thinking we could move this `-latest` test out of this part of the CI to rather make it a daily [scheduled job](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events)?

marbar3778: It will not break, the only change that is being proposed is to default to a validator instead of a fullnode

thanethomson: @marbar3778 do you perhaps know how to add that parameter to a _service_ for a GitHub Action? Right now the only way I can find to override the default `start` command (replacing it with `start --mode validator`) would be to build a separate Tendermint image or come up with a complicated Docker Compose-based solution.

Just overriding the entrypoint won't work, unfortunately, since the `CMD` is defined in the Tendermint Dockerfile.

thanethomson:
I think this is safe to merge, with #845 as a follow-up to keep the test.

PR #853, opened 2021-03-31
p2p: use async bipipe

### LINKED ISSUE #841: Non-deterministic hanging test in tendermint_p2p::secret_connection
I was running some tests today (on `master`) and found one of the P2P tests hanging indefinitely: https://github.com/informalsystems/tendermint-rs/blob/4b0318c4c7e2150565c5ded593280df4b9ac7ad9/p2p/src/secret_connection.rs#L539

Not sure if this is caused by a problem in the `pipe` crate or by the non-deterministic nature of who reads/writes first from/to the bidirectional buffer when spawning the write and read threads in the test.

cc @melekes @xla 

**Steps to reproduce**

It's not guaranteed to reproduce the issue, but if I run the following around 5-10 times the test usually hangs at least once:

```
cargo test -p tendermint-p2p --all-features
```

When we can reliably execute the test without it hanging.

melekes:
Closes #841

A copy of the `pipe` crate (https://github.com/arcnmx/pipe-rs) is added to the `tendermint-p2p` crate. I had to add a new method - `async_bipipe_buffered` to make the handshake tests always pass. `async_bipipe_buffered` method uses the `unbounded` channel (rather than the blocking one) for communication - hence the prefix. Note I also removed the methods and structures from `pipe` clone that we're not using at the moment.

* [x] Wrote tests
* [ ] ~~Updated CHANGELOG.md~~

### DIFF
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -19,6 +19,7 @@ description = """
 chacha20poly1305 = "0.7"
 ed25519-dalek = "1"
 eyre = "0.6"
+flume = "0.10"
 hkdf = "0.10.0"
 merlin = "2"
 prost = "0.7"
@@ -39,7 +40,6 @@ prost-amino = { version = "0.6", optional = true }
 prost-amino-derive = { version = "0.6", optional = true }
 
 [dev-dependencies]
-crossbeam-channel = "^0.5.0"
 readwrite = "^0.1.1"
 
 [features]
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -39,7 +39,8 @@ prost-amino = { version = "0.6", optional = true }
 prost-amino-derive = { version = "0.6", optional = true }
 
 [dev-dependencies]
-pipe = { version = "0", features = ["bidirectional"], git = "https://github.com/melekes/pipe-rs" }
+crossbeam-channel = "^0.5.0"
+readwrite = "^0.1.1"
 
 [features]
 amino = ["prost-amino", "prost-amino-derive"]
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -39,7 +39,7 @@ prost-amino = { version = "0.6", optional = true }
 prost-amino-derive = { version = "0.6", optional = true }
 
 [dev-dependencies]
-pipe = { version = "0", features = ["bidirectional"] }
+pipe = { version = "0", features = ["bidirectional"], git = "https://github.com/melekes/pipe-rs" }
 
 [features]
 amino = ["prost-amino", "prost-amino-derive"]
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -1,6 +1,6 @@
 //! The Tendermint P2P stack.
 
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
 #![deny(
     trivial_casts,
     trivial_numeric_casts,
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -1,6 +1,6 @@
 //! The Tendermint P2P stack.
 
-#![deny(unsafe_code)]
+#![forbid(unsafe_code)]
 #![deny(
     trivial_casts,
     trivial_numeric_casts,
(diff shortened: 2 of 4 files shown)


melekes:
I will see if this could be upstreamed (I doubt that, but I will try nonetheless)

xla:
> I will see if this could be upstreamed (I doubt that, but I will try nonetheless) [arcnmx/pipe-rs#10](https://github.com/arcnmx/pipe-rs/issues/10)

If we have doubts that it will be upstream-ed should we just maintain a pipe implementation in the p2p package for our testing purposes?

thanethomson:
> If we have doubts that it will be upstream-ed should we just maintain a pipe implementation in the p2p package for our testing purposes?

Yeah this makes sense to me :+1:

melekes:
> > I will see if this could be upstreamed (I doubt that, but I will try nonetheless) [arcnmx/pipe-rs#10](https://github.com/arcnmx/pipe-rs/issues/10)
> 
> If we have doubts that it will be upstream-ed should we just maintain a pipe implementation in the p2p package for our testing purposes?

done 👍

p2p/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! The Tendermint P2P stack.
 
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
```
xla, line 3: Why changing this?

p2p/Cargo.toml
```diff
@@ -39,7 +39,8 @@ prost-amino = { version = "0.6", optional = true }
 prost-amino-derive = { version = "0.6", optional = true }
 
 [dev-dependencies]
-pipe = { version = "0", features = ["bidirectional"] }
+crossbeam-channel = "^0.5.0"
```
xla, line 42: Can we converge on one channel implementation?

melekes: while it should be possible to rewrite `pipe` impl to use `flume` instead of `crossbeam-channel`, `flume` does not support (nor does it plan to) broadcast channels https://github.com/zesterer/flume/issues/68#issuecomment-761032975. I'm using broadcast channels in https://github.com/informalsystems/tendermint-rs/pull/825/files#diff-6adb8bb700e22bde6a4c4e334b9e1a36fafae1026f988f9d52836690ed97cf68R88 to propagate data read from socket to virtual streams.

p2p/src/lib.rs
thanethomson: Is there any way to avoid `unsafe` code? We're trying to have no `unsafe` code in the whole tendermint-rs repo.

xla: The only usage of `unsafe` in the pipe implementation is:

https://github.com/informalsystems/tendermint-rs/blob/9b7fac4e4fd720e90b76a2c7a72ebfa9ac629c9b/p2p/src/secret_connection/pipe.rs#L159-L172

AFAIU it's solely to eliminate that branch during compilation, I reckon we can refactor that away.

p2p/Cargo.toml
xla: Broadcast semantics shouldn't be too hard to build, given that we will hold onto the set of all virtual streams.

p2p/src/lib.rs
thanethomson: Why not just `.unwrap()`?

p2p/src/secret_connection/pipe.rs
```diff
@@ -0,0 +1,188 @@
...
+
+fn epipe() -> io::Error {
+    io::Error::new(io::ErrorKind::BrokenPipe, "pipe reader has been dropped")
+}
+
+impl PipeBufWriter {
+    #[inline]
+    /// Gets a reference to the underlying `Sender`
+    pub fn sender(&self) -> &Sender<Vec<u8>> {
+        self.sender.as_ref().unwrap()
```
xla, line 82: Let's use `expect` here to cntextualise the panic.

xla (dismissed):
🛋 🐻 🅰️ 🙌

xla (approved):
🌭 🎃 🍘 🗽

PR #855, opened 2021-04-06
rpc: add `Client::wait_until_healthy` method

tony-iqlusion:
Adds a method which repeatedly polls the `/health` endpoint until it returns a successful response or a specified timeout elapses.

This is useful when writing tests that need to wait for a node to boot up (e.g. inside of a docker container), e.g.:

https://github.com/cosmos/cosmos-rust/pull/71/files#diff-0e31f23fc223d1d543169a682e0f1093c1cc53c6a312febce7d5d1854919fafaR84-R96

### DIFF
--- a/rpc/src/client.rs
+++ b/rpc/src/client.rs
@@ -14,14 +14,17 @@ pub use transport::websocket::{WebSocketClient, WebSocketClientDriver, WebSocket
 
 use crate::endpoint::validators::DEFAULT_VALIDATORS_PER_PAGE;
 use crate::endpoint::*;
+use crate::error::Error;
 use crate::paging::Paging;
 use crate::query::Query;
 use crate::{Order, Result, SimpleRequest};
 use async_trait::async_trait;
+use std::time::Duration;
 use tendermint::abci::{self, Transaction};
 use tendermint::block::Height;
 use tendermint::evidence::Evidence;
 use tendermint::Genesis;
+use tokio::time;
 
 /// Provides lightweight access to the Tendermint RPC. It gives access to all
 /// endpoints with the exception of the event subscription-related ones.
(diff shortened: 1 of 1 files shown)


rpc/src/client.rs
```diff
@@ -221,6 +224,31 @@ pub trait Client {
...
+    /// the given `timeout` has elapsed.
+    async fn wait_until_healthy<T>(&self, timeout: T) -> Result<()>
+    where
+        T: Into<Duration> + Send,
+    {
+        let timeout = timeout.into();
+        let poll_interval = Duration::from_millis(200);
+        let mut attempts_remaining = timeout.as_millis() / poll_interval.as_millis();
+
+        while let Err(_) = self.health().await {
```
thanethomson, line 237:
```suggestion
        while self.health().await.is_err() {
```

Quick fix for the clippy suggestion.

thanethomson (commented):
This is a great idea! Have you tested it locally on your side?

tony-iqlusion:
I updated the PR linked in the description to use code that is now nearly identical to this and it's working as expected locally:

https://github.com/cosmos/cosmos-rust/pull/71/files#diff-0e31f23fc223d1d543169a682e0f1093c1cc53c6a312febce7d5d1854919fafaR97-R117

thanethomson (commented):
Awesome. I'd like to include this in the v0.19.0 release. Could you please add this to the `CHANGELOG` under the `IMPROVEMENTS` section?

thanethomson:
Actually don't worry about it - I'll add something tomorrow when I finalize the release. It'll be easier that way so we don't have merge conflicts in the changelog.

tony-iqlusion:
I noticed one small problem with this trying to use it.

After updating to `tendermint-rpc` v0.19, I was still using `tcp://` URLs, and even though those were failing to parse, this code waited the full timeout and just said it timed out, making it a bit hard to debug.

Perhaps it should look for certain kinds of errors, i.e. ones relating to the remote node being down, and bubble all others up.

Otherwise I switched my code to use it and it's working great!

thanethomson:
> After updating to tendermint-rpc v0.19, I was still using tcp:// URLs, and even though those were failing to parse, this code waited the full timeout and just said it timed out, making it a bit hard to debug.

Yes, I think I've just found a small bug in the `rpc::Url` -> `uri::Uri` conversion that would cause this to happen when specifying a `tcp://` address. I'll open an issue :+1:

Issue #856, opened 2021-04-06
Reorganize `tendermint::abci` to clean up module organization and conform to Rust RFC356


hdevalence:
Currently, the ABCI types in `tendermint_proto` have the structural content filled in, but aren't nicely organized yet:

<img height="500" alt="Screen Shot 2021-04-06 at 12 40 27 PM" src="https://user-images.githubusercontent.com/44879/113773111-68c96c80-96da-11eb-951c-b7a76139d195.png">
<img width="550" alt="Screen Shot 2021-04-06 at 12 40 35 PM" src="https://user-images.githubusercontent.com/44879/113773115-6bc45d00-96da-11eb-840b-1a3573217216.png">

Rust's [RFC356](https://rust-lang.github.io/rfcs/0356-no-module-prefixes.html) recommends that type names should avoid having common prefixes, and instead use module paths to factor a name into components, for instance having `abci::request::{Commit, CheckTx}` instead of `abci::{RequestCommit, RequestCheckTx}`.  This has the advantage that the user of the name can decide which parts of the name are required, writing any of

* `abci::request::Commit`,
* `request::Commit`,
* `Commit`,

depending on how much context is assumed at the use site. It also makes navigating the related types in the docs more clear, since the types are organized into categories.

**What's the definition of "done" for this issue?**

*To tidy the module organization:*

- [ ] rename and move `RequestFoo` to `request::Foo`;
- [ ] move the `Request` enum to `request` and `pub use` it in the `abci` module;
- [ ] rename and move `ResponseFoo` to `response::Foo`;
- [ ] move the `Response` enum to `response` and `pub use` it in the `abci` module;
- [ ] eliminate the `response_apply_snapshot_chunk` and `response_offer_snapshot` modules, by either:
  - **option A**: rename `response_apply_snapshot_chunk::Result` to `response::ApplySnapshotChunkResult` (least impact on code that currently uses the structure, it's a clean rename);
  - **option B**: change the data modeling of `ApplySnapshotChunk` and `OfferSnapshot` to hold enums internally (need to make non-mechanical changes to using code, but potentially cleaner in the long run);
  
I would pick option A above.

*To fill in the documentation:*

- [ ] Document all requests and responses and all fields, copying descriptions from [the Methods and Types spec page](https://docs.tendermint.com/master/spec/abci/abci.html).

*To make it easier to model the four ABCI connections:*

Create per-connection enums:

[...]

Issue #669, opened 2020-11-11
Retiring `anomaly`


tarcieri:
FYI, I'm planning on deprecating the [`anomaly`](https://github.com/iqlusioninc/crates/tree/develop/anomaly) error handling crate, which tendermint-rs presently uses.

There have been a number of error handling crates that have surpassed it coming out of the [error handling working group](https://blog.rust-lang.org/inside-rust/2020/09/18/error-handling-wg-announcement.html).

I would suggest [`eyre`](https://github.com/yaahc/eyre) as a replacement.

romac:
Great news! Will `abscissa` move to `eyre` as well? If so, any ETA on that?

tarcieri:
@romac I believe @yaahc has a branch which moves Abscissa to (color-)`eyre` but I'm not sure about the status of that

tarcieri:
FYI, here's an example of moving from `anomaly` to `eyre`: https://github.com/iqlusioninc/crates/pull/555

romac:
@tarcieri Thanks for the pointers!

PS: By the way, by "great news" I didn't mean to imply that `anomaly` was a bad library, but rather that since the ecosystem seems to have moved away from it in favor of `eyre` (something I've been considering doing for our libs as well), I was happy to see that you were now recommending it as well over `anomaly`.

tarcieri:
No worries, I’m happy there are better libraries emerging in this space as well

ebuchman:
Re #160 (closed for this issue) we should still write a small ADR for error handling in the repo. The comment https://github.com/informalsystems/tendermint-rs/issues/150#issuecomment-586148337 provides a great starting point for an ADR. Also relevant https://github.com/informalsystems/tendermint-rs/pull/158#issuecomment-588290556

tarcieri:
This issue was mostly an FYI. It seems like progress is better tracked in #160.

DaviRain-Su:
Can we  use `anyhow` to handle error?

tony-iqlusion:
Some other downstream projects have switched to `eyre`, which is a fork of `anyhow`. `eyre` supports adding arbitrary context objects to errors.

[...]

PR #862, opened 2021-04-14
abci: Add domain types

hdevalence:
Incomplete implementation of #856

Still to do:

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


hdevalence:
I regenerated the protos against tendermint `master` to avoid having to do extra work later to align with changes that already exist upstream now.  The tests seem to still pass, but if the idea is that `tendermint-rs` is tracking a specific release I guess this PR will have to wait.

I commented out some doctests that failed locally prior to making any changes -- not sure whether that's a local machine problem, or if the CI isn't running the doctests.

The conversions between domain types and protos have a lot of `&'static str` errors.  I'm not sure what the crate error modeling strategy is like, so that could be changed in this PR or in a later one.

I still need to fill in the request types and arrange those nicely, and touch up the docs.

There's existing stub code in the `abci` module that seems to be used for handling some RPC calls, and it should be replaced by complete modeling of the ABCI types.

There are a bunch of `*Params` structures that aren't properly part of ABCI, but come from other proto definitions.  Presumably this means they shouldn't live in the `abci` module, but somewhere else. I'm not sure where.  Hopefully I didn't duplicate anything, but I didn't see any existing ones.

There are no tests for the conversions.  Ideally, it would be possible to use `proptest` to generate randomized messages and check the serialization round-trips, but I'm not sure if that infrastructure already exists. If it doesn't, I would not like to implement a bunch of one-off custom serialization tests, since I think that doing that has relatively low assurance vs effort (compared to, e.g., just using the code to talk to an ABCI client).

tendermint/src/abci/kind.rs
```diff
@@ -0,0 +1,23 @@
+/// A category of ABCI method.
+///
+/// ABCI methods are split into four categories. Tendermint opens one ABCI
+/// connection for each category and refers to these categories as *connections*,
+/// but nothing actually restricts an ABCI connection from calling methods in
+/// multiple categories.
+///
+/// This enum breaks out the `Flush` method as a distinct category, since it is
+/// used to control the execution of other methods.
+pub enum MethodKind {
```
thanethomson, line 10: While I know it's correct to have some kind of indication as to which requests apply to which connection from a modelling perspective, is it actually useful to an ABCI developer? Do you have any particular use cases in mind?

tendermint/src/abci/response.rs
```diff
@@ -0,0 +1,1027 @@
...
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#info)
+#[derive(Clone, PartialEq, Eq, Debug, Default)]
+pub struct Info {
+    /// Some arbitrary information.
+    pub data: String,
+    /// The application software semantic version.
+    pub version: String,
+    /// The application protocol version.
+    pub app_version: u64,
+    /// The latest block for which the app has called [`Commit`](super::request::Commit).
```
thanethomson, line 58: For me, `cargo doc` complains that there's "no item named `Commit` in module `request`"

tendermint/src/abci/response.rs
```diff
@@ -0,0 +1,1027 @@
...
+    pub last_block_height: i64,
+    /// The latest result of [`Commit`](super::request::Commit).
```
thanethomson, line 60: Same as for line 58

tendermint/src/abci/response.rs
```diff
@@ -0,0 +1,1027 @@
...
+///
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#checktx)
+#[derive(Clone, PartialEq, Eq, Debug, Default)]
+pub struct CheckTx {
+    /// The response code.
+    ///
+    /// Transactions where `code != 0` will be rejected; these transactions will
+    /// not be broadcast to other nodes or included in a proposal block.
+    /// Tendermint attributes no other value to the response code.
+    pub code: u32,
```
thanethomson, line 137: There's a struct called `Code` in `tendermint/src/abci/code.rs` that we could use here.

tendermint/src/abci/response.rs
```diff
@@ -0,0 +1,1027 @@
...
+///
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#delivertx)
+#[derive(Clone, PartialEq, Eq, Debug, Default)]
+pub struct DeliverTx {
+    /// The response code.
+    ///
+    /// This code should be `0` only if the transaction is fully valid. However,
+    /// invalid transactions included in a block will still be executed against
+    /// the application state.
+    pub code: u32,
```
thanethomson, line 169: Same as for line 137.

tendermint/src/abci/response.rs
```diff
@@ -0,0 +1,1027 @@
...
+            Response::OfferSnapshot(x) => Ok(Self::OfferSnapshot(x)),
+            Response::LoadSnapshotChunk(x) => Ok(Self::LoadSnapshotChunk(x)),
+            Response::ApplySnapshotChunk(x) => Ok(Self::ApplySnapshotChunk(x)),
+            _ => Err("wrong request type"),
+        }
+    }
+}
+
+// =============================================================================
+// Protobuf conversions
```
thanethomson, line 565: Our current convention is to organize all code relating to a specific type together with that type (e.g. all code relating to the `Exception` struct should be located directly underneath the struct, as opposed to anywhere else in the module).

tendermint/src/abci/types.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// Used to inform Tendermint of changes to the validator set.
+///
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#validatorupdate)
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct ValidatorUpdate {
+    /// The validator's public key.
+    pub pub_key: PublicKey,
+    /// The validator's voting power.
+    pub power: i64,
```
thanethomson, line 36: Could the `Power` struct in `tendermint/src/vote/power.rs` be useful here?

thanethomson (commented):
Thanks so much for all this work! :tada: I've done a first pass here and had a few comments/questions, but it's already looking like it's in really good shape.

tendermint/src/abci/kind.rs
hdevalence: Yep, in https://github.com/hdevalence/tower-abci I use this to split different requests into different queues that can either go to different backing services (e.g., info requests are processed by a replica of the state), or to the same service with different priorities, etc.

tendermint/src/abci/response.rs
hdevalence: Yes, but I didn't use it because I'm not sure that type is well-modeled: it's possible to create an error `Code` with value `0` that will be treated as a success by Tendermint, which doesn't seem right.  Maybe using a `NonZeroU32` would work, but I didn't want to get into it yet.

tendermint/src/abci/response.rs
hdevalence: I think this kind of functional grouping is clearer -- the protobuf conversion code is pretty repetitive boilerplate, so it's nice to be able to just look at all of the structure definitions together, rather than alternating between boilerplate -- but I'm happy to change it for consistency if you'd prefer.

tendermint/src/abci/types.rs
hdevalence: Probably!

tendermint/src/abci/response.rs
thanethomson: Yes, our current convention is actually to split out each struct into its own file and co-locate all `impl`s along with that struct.

tendermint/src/abci/response.rs
thanethomson: Yes, you're right - we could leave this as a `u32` for now then :+1:

cwgoes:
Bump. What's the status here / is there anything we can help with? (ref https://github.com/hdevalence/tendermint-rs/pull/1)

thanethomson:
> Bump. What's the status here / is there anything we can help with? (ref [hdevalence#1](https://github.com/hdevalence/tendermint-rs/pull/1))

Also happy to help on this!

hdevalence:
Hey, sorry for the delay -- I wasn't able to pick it up again in June after the initial review because of some other work, but I've rebased it on top of the current state of the `master` branch to get ready to drive it to completion.

tendermint/src/abci/response.rs
hdevalence: Great, I'll split out the `Request` and `Response` variant types.

tendermint/src/abci/response.rs
hdevalence: Fixed

tendermint/src/abci/response.rs
hdevalence: Split out the types into individual files, each with all the proto impls.

tendermint/src/abci/response.rs
hdevalence: Fixed

hdevalence:
Update:

- I regenerated the protos against the current upstream `master` rather than the `master` as of mid-April when I opened the PR, and fixed any inconsistencies the upstream proto changes introduced relative to the rest of `tendermint-rs`.
- I decoupled the original ABCI-for-RPC types from the new ABCI-for-ABCI types by copying them to the RPC crate. While this means the RPC crate now has duplicate definitions of ABCI types, it means that that code can be refactored with the ABCI code in chunks, rather than getting it all done in this PR (which is big enough as is, and has been unmerged for long enough to start risking problems -- sorry about that!)

Remaining work:
- [ ] Another pass over the ABCI types to catch types that could be refined (e.g., sticking in `Power`, `AppHash`, etc
- [ ] Aligning the original ABCI types with the new ones (without respect to the RPC code). For instance, using the response code structure, etc.

I'll post an update when I've finished those two and this is ready for re-review.

hdevalence:
I've rebased this branch on top of current `master` to be able to get it to completion.

Currently, it's broken, because the conversions between the domain types and the proto types used `&'static str` errors, but it's no longer possible to use a `&'static str` as a source error.

Before trying to correct this, I'd like to make sure I understand how the error modeling is supposed to work. I looked at the `flex-error` docs, but it seems to be an abstraction over other error crates, so I'm not sure what the intended goal is or how to use it.  The change to `flex-error` was in PR #923, which references discussion about an error overhaul in #669, but that issue just discusses some possibilities, and doesn't have any discussion motivating the creation of `flex-error`.

thanethomson:
> Before trying to correct this, I'd like to make sure I understand how the error modeling is supposed to work. I looked at the `flex-error` docs, but it seems to be an abstraction over other error crates, so I'm not sure what the intended goal is or how to use it. The change to `flex-error` was in PR #923, which references discussion about an error overhaul in #669, but that issue just discusses some possibilities, and doesn't have any discussion motivating the creation of `flex-error`.

Sure I can give some context there! [flex-error](https://github.com/informalsystems/flex-error/) does, as you say, abstract over other error handling crates, and our main goals there are:

1. Harmonizing the error handling approach across both tendermint-rs and ibc-rs
2. `no_std` support for as many crates in tendermint-rs and ibc-rs as possible/necessary

hdevalence:
> > Before trying to correct this, I'd like to make sure I understand how the error modeling is supposed to work. I looked at the `flex-error` docs, but it seems to be an abstraction over other error crates, so I'm not sure what the intended goal is or how to use it. The change to `flex-error` was in PR #923, which references discussion about an error overhaul in #669, but that issue just discusses some possibilities, and doesn't have any discussion motivating the creation of `flex-error`.
> 
> Sure I can give some context there! [flex-error](https://github.com/informalsystems/flex-error/) does, as you say, abstract over other error handling crates, and our main goals there are:
> 
>     1. Harmonizing the error handling approach across both tendermint-rs and ibc-rs
> 
>     2. `no_std` support for as many crates in tendermint-rs and ibc-rs as possible/necessary

Okay, great, is there any guidance on how I should use it and create error variants? As a workaround, 1d5fce2 adds a catch-all conversion for `&'static str`, but this isn't ideal.

Otherwise, I've finished rebasing this PR onto current master, and modulo the issues in the commits marked DROP/FIXME, it is probably finished enough that it makes sense to consider merging; further work like refining all of the types to align with the other domain types is probably something that could be done incrementally.

hdevalence:
FYI, you can view a rendered version of what the new code looks like here: https://rustdoc.penumbra.zone/main/tendermint/abci/index.html (at least for now)

hdevalence:
Bumping this -- it would be great to get this over the finish line, since it's sort of unwieldy to rebase :)

tendermint/src/block.rs
```diff
@@ -35,18 +35,18 @@ use tendermint_proto::Protobuf;
...
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[non_exhaustive]
+#[serde(try_from = "RawBlock", into = "RawBlock")]
 pub struct Block {
     /// Block header
     pub header: Header,
 
     /// Transaction data
-    pub data: transaction::Data,
+    pub data: Vec<Vec<u8>>,
```
thanethomson, line 44: What's the reasoning behind the change from `Data` (which is `Vec<u8>`) to `Vec<Vec<u8>>`?

tendermint/src/abci/types.rs
```diff
@@ -0,0 +1,306 @@
...
+pub struct Evidence {
+    /// The kind of evidence.
+    ///
+    /// Note: this field is called `type` in the protobuf, but we call it `kind`
+    /// to avoid the Rust keyword.
+    pub kind: EvidenceKind,
+    /// The offending validator.
+    pub validator: Validator,
+    /// The height when the offense occurred.
+    pub height: i64,
```
thanethomson, line 83: Can this height validly be 0 or negative?

If not, I'd recommend using the [`Height`](https://github.com/informalsystems/tendermint-rs/blob/5bf6bae177cebb32bd9926f300d6e42edc23041c/tendermint/src/block/height.rs#L16) domain type.

(Applies to all other height fields)

tendermint/src/abci/types.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// The corresponding time when the offense occurred.
+    pub time: DateTime<Utc>,
+    /// Total voting power of the validator set at `height`.
+    ///
+    /// This is included in case the ABCI application does not store historical
+    /// validators, cf.
+    /// [#4581](https://github.com/tendermint/tendermint/issues/4581)
+    pub total_voting_power: i64,
```
thanethomson, line 91: Same as the height-related comment, but using the [`Power`](https://github.com/informalsystems/tendermint-rs/blob/5bf6bae177cebb32bd9926f300d6e42edc23041c/tendermint/src/vote/power.rs#L13) domain type.

tendermint/src/abci/types.rs
```diff
@@ -0,0 +1,306 @@
...
+/// Information on the last block commit.
+///
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#lastcommitinfo)
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct LastCommitInfo {
+    /// The commit round.
+    ///
+    /// Reflects the total number of rounds it took to come to consensus for the
+    /// current block.
+    pub round: i32,
```
thanethomson, line 103: See [`Round`](https://github.com/informalsystems/tendermint-rs/blob/5bf6bae177cebb32bd9926f300d6e42edc23041c/tendermint/src/block/round.rs#L13).

thanethomson (commented):
I've given this a first pass and have added some comments/questions. Looking really good so far, thanks for all this work @hdevalence! :tada: 

Please feel free to add issues to capture follow-up work that should happen after this PR's merged.

Also, I see that there are some minor-looking CI errors atm, e.g. https://github.com/informalsystems/tendermint-rs/runs/4045520320?check_suite_focus=true

rpc/src/abci.rs
```diff
@@ -0,0 +1,34 @@
+//! Old ABCI structures, formerly defined in `tendermint::abci`.
```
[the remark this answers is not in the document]
hdevalence: Yep, that's the eventual goal:

> I decoupled the original ABCI-for-RPC types from the new ABCI-for-ABCI types by copying them to the RPC crate. While this means the RPC crate now has duplicate definitions of ABCI types, it means that that code can be refactored with the ABCI code in chunks, rather than getting it all done in this PR.

tendermint/src/abci/request.rs
```diff
@@ -0,0 +1,306 @@
...
+mod echo;
+mod end_block;
+mod info;
+mod init_chain;
+mod load_snapshot_chunk;
+mod offer_snapshot;
+mod query;
+
+pub use apply_snapshot_chunk::ApplySnapshotChunk;
+pub use begin_block::BeginBlock;
```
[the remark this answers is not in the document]
hdevalence: I don't, but they're not changed as of 0.35.

proto/src/prost/tendermint.crypto.rs
```diff
@@ -71,3 +50,26 @@ pub struct ProofOps {
...
+    #[derive(Clone, PartialEq, ::prost::Oneof)]
+    pub enum Sum {
+        #[prost(bytes, tag="1")]
+        #[serde(rename = "tendermint/PubKeyEd25519", with = "crate::serializers::bytes::base64string")]
+        Ed25519(::prost::alloc::vec::Vec<u8>),
+        #[prost(bytes, tag="2")]
+        #[serde(rename = "tendermint/PubKeySecp256k1", with = "crate::serializers::bytes::base64string")]
+        Secp256k1(::prost::alloc::vec::Vec<u8>),
+        #[prost(bytes, tag="3")]
+        Sr25519(::prost::alloc::vec::Vec<u8>),
```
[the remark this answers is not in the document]
hdevalence: Sure, I added just enough to fix the compiler errors, happy to pull in some commits if you'd like to flesh it out more.

tendermint/src/block.rs
hdevalence: `Data` isn't a `Vec<u8>`, it holds an `Option<Vec<Transaction>`, where `Transaction` is an opaque wrapper around `Vec<u8>` that blocks using the normal `Vec` methods.  The transaction data is a `Vec<Vec<u8>>` because it's a list of byte sequences, and IMO it's better to use ordinary Rust types that provide convenient methods, unless there is actually some invariant to maintain.

tendermint/src/abci/request.rs
```diff
@@ -0,0 +1,306 @@
...
+            _ => Err("wrong request type"),
+        }
+    }
+}
+
+// =============================================================================
+// Protobuf conversions
+// =============================================================================
+
+// XXX(hdevalence): these all use &'static str for now, this should be fixed
```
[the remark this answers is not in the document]
hdevalence: Sure, or I could fix the error types -- per my comment above, I don't know how `flex-error` is supposed to be used.

tendermint/src/abci/request.rs
```diff
@@ -0,0 +1,306 @@
...
+pub use info::Info;
+pub use init_chain::InitChain;
+pub use load_snapshot_chunk::LoadSnapshotChunk;
+pub use offer_snapshot::OfferSnapshot;
+pub use query::Query;
+
+/// All possible ABCI requests.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum Request {
+    #[doc = include_str!("doc/request-echo.md")]
```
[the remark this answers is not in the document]
hdevalence: It looks like the CI failures are because the CI is hardcoding a version of nightly 6 months older than the stable version it uses. I think the fix would be to either remove the nightly-dependent CI check, or to make it always use the latest nightly.

tendermint/src/block.rs
thanethomson: Sorry, I was looking at the wrong `Data` struct for some reason ([this one](https://github.com/informalsystems/tendermint-rs/blob/60e1a0a35db90de3de75bfe11e9a2445a7e22431/tendermint/src/abci/data.rs#L10)).

This is something that I've been wondering about for some time now more generally: whether our `Height`/`Round`/etc. types actually add much value beyond just using native Rust types like `u64`/`u32`/etc.

I personally like the simplicity of using Rust native types as far as possible.

tendermint/src/abci/request.rs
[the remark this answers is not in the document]
thanethomson: I think the best place to start with `flex-error` is probably to just understand what its `define_error` macro does: https://docs.rs/flex-error/0.4.4/flex_error/macro.define_error.html - and then there are quite a few variations of examples in the `tendermint` crate showing how it's used.

Its syntax for defining errors was a bit odd for me at first, given that it's not idiomatic Rust, but it is concise. It automatically generates convenient error constructor methods (e.g. the [`InvalidKey` entry](https://github.com/informalsystems/tendermint-rs/blob/60e1a0a35db90de3de75bfe11e9a2445a7e22431/tendermint/src/error.rs#L16) automatically generates an [`invalid_key` constructor](https://github.com/informalsystems/tendermint-rs/blob/5bf6bae177cebb32bd9926f300d6e42edc23041c/tendermint/src/public_key.rs#L65)), which is nice.

I'd imagine that, when adding functionality to the `tendermint` crate, you could just add a variant to the [`Error`](https://github.com/informalsystems/tendermint-rs/blob/60e1a0a35db90de3de75bfe11e9a2445a7e22431/tendermint/src/error.rs#L12) type for each kind of error you'd like to produce, and then use its automatically generated constructor method to construct it.

tendermint/src/block.rs
tarcieri: `Height` used to enforce an invariant (NonZeroU64) but it was changed to allow 0 too at which point having a newtype for it provided little value

thanethomson:
@hdevalence, when you get a chance, could you please update this PR to resolve the conflicts and the CI errors? We've updated the Rust nightly version used for coverage checks in #1017.

hdevalence:
Picking this up again now.

tendermint/src/abci/params.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::prelude::*;
+
+use chrono::Duration;
+use core::convert::{TryFrom, TryInto};
+
+/// All consensus-relevant parameters that can be adjusted by the ABCI app.
+///
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#consensusparams)
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct ConsensusParams {
```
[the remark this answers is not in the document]
hdevalence: Merged structs in https://github.com/informalsystems/tendermint-rs/pull/862/commits/259ddd9d648721accdc98bc76e5b67dee610bfa4

proto/src/prost/tendermint.crypto.rs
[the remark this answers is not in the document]
hdevalence: Fixed in  https://github.com/informalsystems/tendermint-rs/pull/862/commits/1acc5b3c61a540d608dbb7ff66a75a63f43e22ef

tendermint/src/block.rs
hdevalence: Having newtypes is really useful for enforcing semantic consistency (am I accidentally using a round in place of a height?), but I think the important thing is that if there aren't any additional invariants on the underlying type, the newtype should have public inner fields, so that people can use all of the existing API.

tendermint/src/abci/types.rs
hdevalence: Fixed in 243452df

tendermint/src/abci/types.rs
hdevalence: Fixed in 243452df

tendermint/src/abci/types.rs
hdevalence: Fixed in fde60809

tendermint/src/error.rs
```diff
@@ -193,5 +193,22 @@ define_error! {
 
         TrustThresholdTooSmall
             |_| { "trust threshold too small (must be >= 1/3)" },
+
+        Other
+            { msg: String }
+            | e | { format_args!("error: {}", e.msg) },
```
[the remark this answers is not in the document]
hdevalence: I just cleaned up all the errors instead in 0dcdc6f

tendermint/src/abci/request.rs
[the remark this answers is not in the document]
hdevalence: Fixed in 0dcdc6f

hdevalence:
@thanethomson I've rebased this again and deduplicated types per your suggestions (thanks for pointing those out!).  I also aligned all the errors, and then rebased again to eliminate the placeholder commits labeled "drop me".

So, this should be good to go.

thanethomson:
Superseded by (and included in) #1022.

hdevalence:
🎊 Thanks so much for getting this over the finish line!!

PR #863, opened 2021-04-16
rpc: add support for `/tx` endpoint

tony-iqlusion:
Support for finding a transaction directly by its hash.

In theory this endpoint should have read-your-writes (RYW) consistency whereas `/tx_search` does not.

See: https://github.com/tendermint/tendermint/issues/6359


thanethomson (approved):
LGTM! :+1:

PR #866, opened 2021-04-20
Verify `PeerList` with Stainless

yannbolliger:
This is a first trial to introduce `cargo stainless` and let it run on the `PeerList`.

**Problems that came up:**

- [x] No support for constants epfl-lara/rust-stainless#73
- [x] No type aliases https://github.com/epfl-lara/rust-stainless/issues/141
- [x] ~~`contracts` crate has a `==>` macro feature for implies in specs.~~ (won't fix)
- [x] Stainless collections don't have the same names as Rust's std ones. Examples: Stainless Map has `contains` instead of `contains_key`, the Set has `add` instead of `insert`. https://github.com/epfl-lara/rust-stainless/issues/144
- [x] No mutable references/borrowing, this was easily circumvented by locally mutating and consuming the `self`: 858da861b11a3a5ff1f0c8fa6e26775cad0ed1b3. https://github.com/epfl-lara/rust-stainless/issues/145
- [x] External files with modules of the same crate work. But, the `mod xx; use xx::*;` and the corresponding `use super::*;` fail as unsupported trees. https://github.com/epfl-lara/rust-stainless/issues/146
- [x] Sometimes, Stainless complains that it can't extract the tree kind lifetime. Could these just be ignored? Example: `pub fn first(&self) -> Option<&T> { ... }`. https://github.com/epfl-lara/rust-stainless/issues/147
- [x] No higher-order functions for things like `iter().all(...)`. This was easily circumvented by creating some new set methods.
- [x] The trait bounds that are now required on `stainless::{Set, Map}`, namely `T:  Eq + Hash + Clone`, force users that also implement something generic (for example the `ListSet`) to put the same trait bounds on their implementation. These bounds then create evidence arguments everywhere that are for types that are not extracted. Hence, Stainless fails with `missing dependencies`. **Unfortunately, this makes the new runtime implementations for `stainless::{Set, Map}` quite unusable.**
    - [x] One idea was to conditionally (on `--cfg stainless`) compile with the trait bounds for the runtime version and without them for `cargo stainless`. This fails due to the one-crate-only compilation model. The `stainless::{Set, Map}` are outside of the current crate. _Could a prelude be the solution to this?_
    - [x] Another idea would be to ignore/erase all the three traits `Eq + Hash + Clone` and their corresponding evidence args.
    - [ ] The final idea is https://github.com/epfl-lara/rust-stainless/issues/143

**Findings (Tendermint-related)**

- [x] The postconditions on the two methods `replace_faulty_witness`, `replace_faulty_primary` only hold if the precondition includes/validates the invariant. Should this be added / was that implicitly assumed?

---

*  Referenced an issue explaining the need for the change
*  Updated all relevant documentation in docs
* Updated all code comments where relevant
*  Wrote tests
*  Updated CHANGELOG.md


romac:
> The postconditions on the two methods replace_faulty_witness, replace_faulty_primary only hold if the precondition includes/validates the invariant. Should this be added / was that implicitly assumed?

Good catch, feel free to open a PR to fix this, and make sure to mention there how you figured this out :)

thanethomson:
Is this PR still relevant in any way?

PR #872, opened 2021-04-30
Bump `k256` crate dependency to v0.8

tarcieri:
Bumps `k256` to the latest release.

It seems like some refactoring of the public keys might be in order at this point, but I decided to start with a simple version bump.

I could also potentially add NIST P-256 support with the `p256` crate too.

### DIFF
--- a/tendermint/Cargo.toml
+++ b/tendermint/Cargo.toml
@@ -39,8 +39,6 @@ bytes = "1.0"
 chrono = { version = "0.4", features = ["serde"] }
 ed25519 = "1"
 ed25519-dalek = { version = "1", features = ["serde"] }
-# TODO(thane): Remove once https://github.com/myrrlyn/funty/issues/3 is fixed
-funty = "=1.1.0"
 futures = "0.3"
 num-traits = "0.2"
 once_cell = "1.3"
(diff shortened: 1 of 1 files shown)


xla (approved):
📓 🖖 🎭 🙏

tarcieri:
> What you got in mind for this? Worth capturing in a follow-up issue?

Yeah, probably.

I think it'd be good to switch to using the [`k256::PublicKey`](https://docs.rs/k256/0.8.0/k256/type.PublicKey.html) or [`k256::ecdsa::VerifyingKey`](https://docs.rs/k256/0.8.0/k256/ecdsa/struct.VerifyingKey.html) types as the [`tendermint::public_key::Secp256k1`](https://docs.rs/tendermint/0.19.0/tendermint/public_key/type.Secp256k1.html) key type.

Those types ensure that the public key represents a valid elliptic curve point. It's presently using an `EncodedPoint` type which represents a SEC1 serialization of an elliptic curve point that isn't guaranteed to be valid.

One drawback of this approach is that [`tendermint::PublicKey::as_bytes()`](https://docs.rs/tendermint/0.19.0/tendermint/public_key/enum.PublicKey.html#method.as_bytes) couldn't exist in its current form, since the `k256` types represent a decoded curve point and require an extra step to re-serialize them.

PR #874, opened 2021-05-04
Update the TLA specs used in MBT with the current Apalache type annotations

shonfeder:
In order to make use of the latest Apalache for MBT, we need to update the typing to use the current annotation practice.

Discussed with @andrey-kuprianov and @konnov via Slack.

### DIFF
--- a/light-client/tests/support/model_based/MC4_4_faulty.tla
+++ b/light-client/tests/support/model_based/MC4_4_faulty.tla
@@ -6,10 +6,40 @@ TARGET_HEIGHT == 4
 TRUSTING_PERIOD == 1400 \* two weeks, one day is 100 time units :-)
 IS_PRIMARY_CORRECT == FALSE
 
+\* @typeAlias: BLOCKHEADER = [height: Int, time: Int, lastCommit: Set(Str), VS: Set(Str), NextVS: Set(Str)];
+\* @typeAlias: BLOCK = [header: BLOCKHEADER, Commits: Set(Str)];
+\* @typeAlias: BLOCKSTATUS = Int -> Str;
+\* @typeAlias: BLOCKS = Int -> BLOCK;
+MC4TypeAliases == TRUE
+
 VARIABLES
-  state, nextHeight, fetchedLightBlocks, lightBlockStatus, latestVerified,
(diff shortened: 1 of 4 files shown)


light-client/tests/support/model_based/MC4_4_faulty.tla
```diff
@@ -6,10 +6,40 @@ TARGET_HEIGHT == 4
 TRUSTING_PERIOD == 1400 \* two weeks, one day is 100 time units :-)
 IS_PRIMARY_CORRECT == FALSE
 
+\* @typeAlias: BLOCKHEADER = [height: Int, time: Int, lastCommit: Set(Str), VS: Set(Str), NextVS: Set(Str)];
+\* @typeAlias: BLOCK = [header: BLOCKHEADER, Commits: Set(Str)];
+\* @typeAlias: BLOCKSTATUS = Int -> Str;
+\* @typeAlias: BLOCKS = Int -> BLOCK;
+MC4TypeAliases == TRUE
+
 VARIABLES
```
shonfeder, line 15: @konnov is there a reasonable way to avoid having to repeat these annotations here?

I tired extending the Lightclient003 module, but that caused operator def conflicts.

konnov: I think the hack by Ilina was to refactor these repetitive definitions in a separate module like `TypeDefs` and extend that one.

konnov (approved):
Looks great! Thanks for updating the annotations!

light-client/tests/support/model_based/MC4_4_faulty.tla
shonfeder: Ah! Clever. We should do that, but I'll wait until we've consolidated the specs before a more invasive refactoring. Thanks! :)

andrey-kuprianov (approved):
Great work, Shon! 

Many thanks for doing that -- now with an example transformation it's much easier to understand what needs to be done.

I do like that in many places annotations of empty sets are gone:) I don't like that so many additional annotations are needed when instantiating other modules... But I guess we will address this in the future:)

Thanks again!

shonfeder:
Thanks for the reviews!

PR #881, opened 2021-05-17
Remove `tendermint-light-node` crate

### LINKED ISSUE #879: Remove/archive the Light Node
As per our discussion in the tendermint-rs call yesterday, if nobody's using it directly we should strongly consider removing the `light-node` crate from the repository.

The reasons for this are:

1. The code to facilitate a Light Node is generally used as part of another higher-level library or application (e.g. [ibc-rs](https://github.com/informalsystems/ibc-rs)), and isn't generally directly instantiated by itself (except perhaps for testing purposes).
2. We should eventually consider positioning the tendermint-rs repo as a framework/collection of libraries, as opposed to mixing in binaries. Binaries should possibly be separated out at a later date as issues relating to them will often be operational and tangential to framework-level concerns.
3. We want to minimize the amount of code we need to maintain in this repo.

When the `light-node` crate's been deleted from this repo.

thanethomson:
Closes #879 

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


melekes (approved):
👍

Issue #824, opened 2021-03-03
Status of no_std compatibility


romac:
This issue tracks the current status of `no_std` compatibility for the `tendermint` and `ibc` crates.

# Dependency on `std`

## `tendermint`

Pervasive

## `ibc`

Pervasive

# Dependencies

### no_std by default
- [async-trait](https://lib.rs/async-trait)
- [num-derive](https://lib.rs/num-derive)
- [zeroize](https://lib.rs/zeroize)
- [serde-repr](https://lib.rs/serde-repr)

### no_std via feature
- [anomaly](https://lib.rs/anomaly)
- [prost/prost-type](https://lib.rs/prost/prost-type)
- [chrono](https://lib.rs/chrono)
- [bytes](https://lib.rs/bytes)
- [ed25519](https://lib.rs/ed25519)
- [ed25519-dalek](https://lib.rs/ed25519-dalek)
- [futures](https://lib.rs/futures)
- [num-traits](https://lib.rs/num-traits)
- [once-cell](https://lib.rs/once-cell)
- [sha2](https://lib.rs/sha2)
- [signature](https://lib.rs/signature)
- [subtle](https://lib.rs/subtle)
- [subtle-encoding](https://lib.rs/subtle-encoding)
- [serde](https://lib.rs/serde)
- [serde-json](https://lib.rs/serde-json)
- [serde-bytes](https://lib.rs/serde-bytes)

### incompatible
- [thiserror](https://lib.rs/thiserror) (need to use https://github.com/dtolnay/thiserror/pull/64#issuecomment-735805334)
- [toml](https://lib.rs/toml)  (only used for parsing Tendermint config, could be feature-guarded for std)

### no_std by default
- [dyn_clonable](https://lib.rs/dyn_clonable)

### no_std via feature
- [regex](https://lib.rs/regex)
- [tracing](https://lib.rs/tracing) (unsure but could switch to tracing-core)

### std-only
- [bech32](https://lib.rs/bech32)

andynog:
the `ibc-rs` dependency on `bech32` will be removed when some refactoring happens as part of [ibc-rs issue #635](https://github.com/informalsystems/ibc-rs/issues/635)

zmanian:
We have alloc only bech32 implementation that can be used in subtle-encoding 

https://crates.io/crates/subtle-encoding

DaviRain-Su:
[...]

PR #890, opened 2021-06-02
enable tendermint no_std and error handle by flex-error

DaviRain-Su:
https://github.com/informalsystems/tendermint-rs/issues/824 ,Support to Tendermint no_std features.
For error handle use `anyhow`, `flex-eror` and use `no-std-compat` to handle no_std features primitives.

`flex-error` crates: https://github.com/informalsystems/flex-error/
An experimental PR for ibc-rs is available at informalsystems/ibc-rs https://github.com/informalsystems/ibc-rs/pull/988.

This is a follow-up to support tendermint no_std features https://github.com/informalsystems/tendermint-rs/pull/889.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


proto/Cargo.toml
```diff
@@ -20,7 +20,8 @@ all-features = true
 prost = "0.7"
 prost-types = "0.7"
 bytes = "1.0"
-anomaly = "0.2"
+anyhow = {version = "1.0", default-features = false, git = "https://github.com/DaviRain-Su/anyhow.git", branch = "open-stderror"}
```
soareschen, line 23: The version of `anyhow` here gives a compile error:

```
error[E0433]: failed to resolve: use of undeclared crate or module `anyhow`
    |
272 | pub type BoxError = alloc::Box<dyn anyhow::StdError + Send + Sync>;
    |                                    ^^^^^^ use of undeclared crate or module `anyhow`

```

proto/src/lib.rs
```diff
@@ -20,7 +20,17 @@ mod tendermint;
...
-use anomaly::BoxError;
+
+#[cfg(not(feature = "std"))]
+extern crate alloc;
+
+#[cfg(not(feature = "std"))]
+pub use alloc::boxed::Box;
+
+#[cfg(feature = "std")]
+pub use std::boxed::Box;
```
soareschen, line 31: I think we can probably don't use feature flags here and use `alloc` crate directly. When compiled with `std` it should automatically link with the one provided by `std`.

tendermint/src/error.rs
```diff
@@ -1,206 +1,213 @@
...
-    Parse,
+    Length
+    [DisplayError<Error>]
+    |_| { format_args!("length error") },
 
-    /// Network protocol-related errors
-    #[error("protocol error")]
-    Protocol,
+    Parse
+    [DisplayError<Error>]
```
soareschen, line 28: We only need to wrap upstream errors if they are coming from code not belong to the us. If you just want to attach additional details in string, you can instead define suberrors such as `Parse { reason: String }` and omit the error source.

Better yet, we should split out the different suberrors such as:

```rust
ParseUrl
  { url: String }
  [ DisplayError<url::ParseError> ]
  | e | { format_args!("error parsing url {}", e.url) }
InvalidHashLength
  { length: usize }
  | e | { format_args!("hash has invalid length {}", e.length) }
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,213 @@
...
-    /// Proposer not found in validator set
-    #[error("proposer with address '{}' not found in validator set", _0)]
-    ProposerNotFound(account::Id),
-}
+    InvalidHashSize
+    [DisplayError<Error>]
+    |_| { format_args!("invalid hash: expected hash size to be 32 bytes") },
+
+    NoTimestamp
+    [DisplayError<Error>]
```
soareschen, line 68: There is no need for any error source here, since there is no additional information needed.

tendermint/src/vote.rs
```diff
@@ -65,7 +68,7 @@ impl TryFrom<RawVote> for Vote {
 
     fn try_from(value: RawVote) -> Result<Self, Self::Error> {
         if value.timestamp.is_none() {
-            return Err(NoTimestamp.into());
+            return Err(error::no_timestamp_error(anyhow::anyhow!("no timestamp error")));
```
soareschen, line 71: This can be simplified into just

```rust
return Err(error::no_timestamp_error());
```

Behind the scene `flex-error` automatically calls `anyhow` to capture the stack trace already.

soareschen (changes requested):
I went through the changes a little and given some suggestions on how you can better define the errors.

proto/Cargo.toml
DaviRain-Su: I have reset this change.

proto/src/lib.rs
DaviRain-Su: I have reset this change.

DaviRain-Su:
I made some fixes and upgraded flex-error to the latest

tendermint/src/error.rs
DaviRain-Su: I have reset this change.

tendermint/src/vote.rs
DaviRain-Su: I have reset this change.

p2p/Cargo.toml
```diff
@@ -30,7 +31,6 @@ subtle-encoding = { version = "0.5" }
 thiserror = "1"
 x25519-dalek = "1.1"
 zeroize = "1"
-
```
xla, line 33: This line deletion is not necessary.

p2p/Cargo.toml
```diff
@@ -19,6 +19,7 @@ description = """
 chacha20poly1305 = "0.7"
 ed25519-dalek = "1"
 eyre = "0.6"
+anyhow = "1.0"
```
xla, line 22: Dependencies are ordered alphanumerically.

.gitignore
```diff
@@ -23,3 +23,6 @@ light-client-js/pkg/
 light-client-js/examples/verifier-web/node_modules/
 light-client-js/examples/verifier-web/dist/
 light-client-js/examples/verifier-web/package-lock.json
+
+# clion
+.idea
```
xla, line 28: This exclude is not meaningful outside of specific personal setups. Why do you want to see it included?

xla (changes requested):

p2p/Cargo.toml
xla: Why the introduction of the extra dependency?

.gitignore
romac: Yes, this should probably go in `.git/info/exclude`, cf. http://git-scm.com/docs/gitignore.

p2p/Cargo.toml
romac: Can you also explain why the `p2p` crate now needs to depend on `anyhow`?

DaviRain-Su: Oh, sorry, I forget delete this crate. The current is can delete.

.gitignore
DaviRain-Su: > This should probably go in `.git/info/exclude` (cf. http://git-scm.com/docs/gitignore).

Oh, Thanks for you, This is help me.

DaviRain-Su:
The above issue has been fixed.

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        |_| { format_args!("missing evidence field") },
+
+        MissingTimestamp
+        |_| { format_args!("missing timestamp field") },
+
+        InvalidBlock
+        [DisplayError<Error>]
+        |_| { format_args!("invalid block") },
+
+        InvalidFristBlock
```
soareschen, line 113: There is a typo and this don't seem to be used

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        |_| { format_args!("missing data field") },
+
+        MissingEvidence
+        |_| { format_args!("missing evidence field") },
+
+        MissingTimestamp
+        |_| { format_args!("missing timestamp field") },
+
+        InvalidBlock
+        [DisplayError<Error>]
```
soareschen, line 110: The use of this do not require any source, so this can be

```rust
InvalidBlock { reason: String }
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        |_| { format_args!("invalid block") },
+
+        InvalidFristBlock
+        [DisplayError<Error>]
+        |_| { format_args!("invalid first block") },
+
+        MissingVersion
+        |_| { format_args!("missing version") },
+
+        InvalidHeader
```
soareschen, line 120: This is not used.

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        [DisplayError<Error>]
+        |_| { format_args!("invalid header") },
+
+        InvalidFirstHeader
```
soareschen, line 124: This error do not have a source and is used in one place. So it can be

```rust
InvalidFirstHeader
  |_| { format_args!("last_block_id is not null on first height") }
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        [DisplayError<Error>]
+        |_| { format_args!("invalid first header") },
+
+        InvalidSignature
```
soareschen, line 128:
```rust
InvalidSignature { reason: String }
```

tendermint/src/block/commit_sig.rs
```diff
@@ -74,47 +77,61 @@ impl TryFrom<RawCommitSig> for CommitSig {
...
+                return Err(error::invalid_validator_address_error());
             }
             return Ok(CommitSig::BlockIdFlagCommit {
                 validator_address: value.validator_address.try_into()?,
-                timestamp: value.timestamp.ok_or(Kind::NoTimestamp)?.try_into()?,
+                timestamp: value
+                    .timestamp
+                    .ok_or(error::no_timestamp_error())?
+                    .try_into()
+                    .map_err(|e: std::convert::Infallible| {
```
soareschen, line 107: `Infallible` can never happen. So you can eliminate it with just `match e {}`

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+
+        ProposerNotFound
+        {account: account::Id}
+        |e| { format_args!("proposer with address '{0}' no found in validator set", e.account) },
+
+        InFallible
+        [DisplayError<Error>]
+        |_| { format_args!("infallible") },
+
+        ChronoParse
```
soareschen, line 174:
```rust
ChronoParse
  [ DisplayError<chrono::ParseError> ]
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        [DisplayError<Error>]
+        |_| { format_args!("chrono parse error") },
+
+        SubtleEncoding
```
soareschen, line 178:
```rust
SubtleEncoding
  [ DisplayError<subtle_encoding::Error> ]
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        [DisplayError<Error>]
+        |_| { format_args!("subtle encoding error") },
+
+        SerdeJson
+        [DisplayError<Error>]
```
soareschen, line 183:
```rust
SerdeJson
   [ DisplayError<serde_json::Error> ]
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
+        |_| { format_args!("serde json error") },
 
-impl Kind {
-    /// Add additional context.
-    pub fn context(self, source: impl Into<BoxError>) -> Context<Kind> {
-        Context::new(self, Some(source.into()))
+        Toml
```
soareschen, line 186:
```rust
Toml
  [ DisplayError<toml::de::Error> ]
```

tendermint/src/error.rs
```diff
@@ -1,206 +1,190 @@
...
-    #[error("proposer with address '{}' not found in validator set", _0)]
-    ProposerNotFound(account::Id),
-}
+        InvalidHashSize
+        |_| { format_args!("invalid hash: expected hash size to be 32 bytes") },
+
+        NoTimestamp
+        |_| { format_args!("no timestamp") },
+
+        InvalidTimestamp
```
soareschen, line 71:
```rust
NonZeroTimestamp
  | _ | { "absent commitsig has non-zero timestamp" }
```

soareschen (changes requested):

tendermint/src/error.rs
DaviRain-Su: Oh, It seems to be.

tendermint/src/error.rs
DaviRain-Su: enn.

tendermint/src/error.rs
DaviRain-Su: en

DaviRain-Su:
@soareschen  Have modified all the problems you mentioned.

tendermint/src/block/commit_sig.rs
soareschen: This is still not fixed. The `Infallible` variant can also be removed.

soareschen (changes requested):

tendermint/src/block/commit_sig.rs
DaviRain-Su: You'll report an error if you don't clearly identify it.

tendermint/src/block/commit_sig.rs
```diff
@@ -74,47 +77,53 @@ impl TryFrom<RawCommitSig> for CommitSig {
...
             }
             return Ok(CommitSig::BlockIdFlagCommit {
                 validator_address: value.validator_address.try_into()?,
-                timestamp: value.timestamp.ok_or(Kind::NoTimestamp)?.try_into()?,
+                timestamp: value
+                    .timestamp
+                    .ok_or(error::no_timestamp_error())?
+                    .try_into()
+                    .map_err(|e | {
+                        error::in_fallible_error(e)
```
soareschen, line 102: See the documentation for `Infallible` here: https://doc.rust-lang.org/std/convert/enum.Infallible.html

Basically `Infallible` is an empty enum, which means there is no way you can ever construct a valid value for `Infallible`. So if we have a function that is given an `Infallible` value, we know that can never happen, so that part of the code will never actually get executed. As a result, we can discard an infallible value `e: Infallible` using `match e {}`.

Strictly speaking, we should never need to use `Infallible` in this case. Because a `TryFrom` with `Infallible` error is just equivalent to `Try`. So the type `Timestamp` in this case should just implement `Try` instead of `TryFrom`:


```diff
--- a/tendermint/src/time.rs
+++ b/tendermint/src/time.rs
- impl TryFrom<Timestamp> for Time {
-     type Error = Infallible;
- 
-     fn try_from(value: Timestamp) -> Result<Self, Self::Error> {
-         // prost_types::Timestamp has a SystemTime converter but
-         // tendermint_proto::Timestamp can be JSON-encoded
-         let prost_value = prost_types::Timestamp {
-             seconds: value.seconds,
-             nanos: value.nanos,
-         };
- 
-         Ok(SystemTime::from(prost_value).into())
-     }
- }
+ impl From<Timestamp> for Time {
+     fn from(value: Timestamp) -> Self {
+         // prost_types::Timestamp has a SystemTime converter but
+         // tendermint_proto::Timestamp can be JSON-encoded
+         let prost_value = prost_types::Timestamp {
+             seconds: value.seconds,
+             nanos: value.nanos,
+         };
+ 
+         SystemTime::from(prost_value).into()
+     }
+ }
```

Then here we can just rewrite it as

```rust
timestamp: value
    .timestamp
    .ok_or(error::no_timestamp_error())?
    .into()
```

See the commit https://github.com/informalsystems/ibc-rs/pull/988/commits/6b397f183c7ceb1d3bbf6f396b4f208590dbbf11 for similar work done at https://github.com/informalsystems/ibc-rs/pull/988.

DaviRain-Su: I understand thank you.

DaviRain-Su:
I made some fixes.

soareschen:
@DaviRain-Su I notice that right now in master, the `tendermint`, `tendermint-rpc`, and `tendermint-light-client-js` crates are already compilable in wasm using the `cargo build-wasm-tendermint` and `build-wasm-light-client` commands. Does that means that the crates already work in WebAssembly and Substrate despite not yet supporting `no_std`?

@thanethomson do you have any context on how the Wasm builds of Tendermint works?

DaviRain-Su:
> @DaviRain-Su I notice that right now in master, the `tendermint`, `tendermint-rpc`, and `tendermint-light-client-js` crates are already compilable in wasm using the `cargo build-wasm-tendermint` and `build-wasm-light-client` commands. Does that means that the crates already work in WebAssembly and Substrate despite not yet supporting `no_std`?
> 
> @thanethomson do you have any context on how the Wasm builds of Tendermint works?

I think it might not be supported, I found that some of the dependency crates don't have default-feature = false turned on.

DaviRain-Su:
@soareschen When I use `cargo check -p tendermint --target wasm32-unknown-unknown` to compile tendermint it gives the following error.
`
error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /Users/davirain/.cargo/registry/src/rsproxy.cn-8f6827c7555bfaf8/getrandom-0.2.2/src/lib.rs:213:9
    |
213 | /         compile_error!("target is not supported, for more information see: \
214 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |_________________________________________________________________________^
`

soareschen:
@DaviRain-Su did you tried building with the current branch or on master? I tried checking and building `tendermint` with the `wasm32-unknown-unknown` target on master and it works.

DaviRain-Su:
> @DaviRain-Su did you tried building with the current branch or on master? I tried checking and building `tendermint` with the `wasm32-unknown-unknown` target on master and it works.

Yes, the current branch and the master branch are the same, and also encountered the same problem as in compiling `ibc-proto`.

soareschen:
Closing this as the related changes have been merged through other PRs or addressed in other issues.

PR #891, opened 2021-06-02
Switch to `unclog` to parallelize changelog updates

thanethomson:
As per out-of-band discussions, we noticed that we often end up in a situation where we're working on several PRs simultaneously, each of which modifies the `CHANGELOG.md` file but makes changes to distinct parts of the code base. This inevitably results in annoying merge conflicts purely related to the changelog.

The simplest solution seems to be to disaggregate changelog entries and build `CHANGELOG.md` from those entries as we perform a release.

For this, I built [unclog](https://github.com/informalsystems/unclog), which is a small CLI tool written in Rust, to help manage these disaggregated changelog entries.

* [ ] Wrote tests

### DIFF
--- a/.changelog/README.md
+++ b/.changelog/README.md
@@ -1,0 +1,13 @@
+# CHANGELOG
+
+This folder contains raw changelog data that can be built using [`unclog`].
+
+```bash
+# Install unclog (if you haven't already)
+cargo install unclog
+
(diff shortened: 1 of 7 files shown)


xla (approved):
⬆️ 🏥 💂 🌊

Issue #892, opened 2021-06-02
Pre-v0.20.0 release tracking ticket for: `k256` v0.9 update, #873, upstreaming cosmos-sdk-rs's `AccountId`


tony-iqlusion:
This is a tracking issue for some changes I'd like to get into tendermint-rs v0.20, some of which I've discussed with @thanethomson:

- [x] #894: `k256` v0.9
- [ ] #873: refactor public keys (PR: #900)
- [ ] #895: upstream `AccountId` from cosmos-sdk-rs

## #894: `k256` v0.9

I plan on releasing a new version of `k256` (secp256k1 library) in the next few days. Hopefully I can get an upgrade in before the next release.

The biggest reason why you'd want this is it eliminates `bitvec` as a mandatory dependency, and `bitvec` has a number of transitive dependencies, so it would help cutting down on the overall dependency count.

For us at iqlusion, we're writing a key manager for Gravity Bridge (based on Signatory) which uses tendermint-rs and would like to be able to take advantage of the latest features in this crate.

## #873: refactor public keys (PR: #900)

Right now the `tendermint::public_key::Secp256k1` type performs no validation of public keys. As documented in #873, ensuring keys are valid is a beneficial invariant to maintain.

As it were,  `tendermint::public_key::Ed25519` used to behave this way. When it was switched to validate public keys, some issues with the test vectors in use were discovered. I think it would be good to ensure keys are validated (i.e. checked to decode to curve points which are valid solutions to the curve equation) to prevent similar bugs.

## #895: upstream `AccountId` from cosmos-sdk-rs

This one seems somewhat debatable, but as part of working on cosmos-sdk-rs I wrote a type for representing Bech32 `AccountId`s:

https://github.com/cosmos/cosmos-rust/blob/5999ee12f658195ebc57e84adaff4bff9c425553/cosmos-sdk-rs/src/base.rs

This is effectively a newtype whose contents are always valid Bech32 and as such also carries along the "human readable prefix" . The current `tendermint::account::Id` (which I think I'm largely guilty for) does not.

thanethomson:
With regard to the `AccountId` type, what value does this add at the `tendermint-rs` level? It's not something that Tendermint itself really understands, is it?

tony-iqlusion:
FYI, PRs are now open for all of these. #894 was the only one I really cared strongly about and it's landed.

tony-iqlusion:
Will go ahead and close this issue as the relevant PRs are all open

[...]

Issue #816, opened 2021-02-16
funty 1.2.0 breaks tendermint-rs build


thanethomson:
As per https://github.com/myrrlyn/funty/issues/3

**Steps to reproduce**

Build `master` with funty v1.2.0 and you should get such compiler errors:

```
error[E0034]: multiple applicable items in scope
   --> /home/thane/.cargo/registry/src/github.com-1ecc6299db9ec823/bitvec-0.18.4/src/field.rs:307:25
    |
307 |         check("load", len, M::BITS);
    |                               ^^^^ multiple `BITS` found
    |
note: candidate #1 is defined in the trait `BitMemory`
   --> /home/thane/.cargo/registry/src/github.com-1ecc6299db9ec823/bitvec-0.18.4/src/mem.rs:26:2
    |
26  |     const BITS: u8 = mem::size_of::<Self>() as u8 * 8;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: candidate #2 is defined in the trait `IsNumber`
   --> /home/thane/.cargo/registry/src/github.com-1ecc6299db9ec823/funty-1.2.0/src/lib.rs:144:2
    |
144 |     const BITS: u32;
    |     ^^^^^^^^^^^^^^^^
help: disambiguate the associated constant for candidate #1
    |
307 |         check("load", len, BitMemory::BITS);
    |                            ^^^^^^^^^^^^^^^
help: disambiguate the associated constant for candidate #2
    |
307 |         check("load", len, IsNumber::BITS);
    |                            ^^^^^^^^^^^^^^
```

**What's the definition of "done" for this issue?**

When `master` builds again without breaking due to these errors.

PR #894, opened 2021-06-08
Bump `k256` crate dependency to v0.9; MSRV 1.51+

tony-iqlusion:
As discussed in #892.

This release makes `bitvec` an optional dependency (gated under the `bits` feature). Since tendermint-rs doesn't need it at all, it's disabled as per default. This eliminates some transitive dependencies, which I know have been a problem in the past (#816).

The internal implementation is now using const generics, so this crate has a MSRV of 1.51+.


xla (approved):
◀️ 🌠 🕸 🏩

PR #895, opened 2021-06-09
tendermint: add `account::Address` (Bech32) type

tony-iqlusion:
See #892.

Adds a newtype for a Bech32-encoded string representation of a public key fingerprint, a.k.a. an account id/address.

Adapted from the `cosmos_sdk::AccountId` type:

https://docs.rs/cosmos_sdk/0.1.1/cosmos_sdk/struct.AccountId.html

This type is effectively a string serialization of the existing `account::Id` type, which is extracted into a module and slightly modified in this PR as well:

- Visibility changed to: `Id(pub [u8; Self::LENGTH])`
- `TryFrom<&[u8]>` impl added
- `Id::new` changed to be fallible and accept a byte slice. Previous usages that accepted a fixed-size array can now use `Id(arr)` instead.

The rationale for changing the visibility of the `Id` type is because, at least as it works at present, it does not enforce any kind of invariant.

However, the newly added `Address` type does: it ensures it contains a valid Bech32-encoded string.


tony-iqlusion:
Closing per discussion in #892

PR #904, opened 2021-06-13
p2p: Propose transport agnostic abstractions

xla:
Describes broadly a new set of abstractions to support a transport
agnostic p2p stack. Bundled together are the types and traits in
`p2p::transport`. Other parts like the `Supervisor` will land in
follow-up change-sets.

Closes #692

***
[Rendered ADR-009](https://github.com/informalsystems/tendermint-rs/blob/xla/692-p2p-transport/docs/architecture/adr-009-transport-agnostic-peer-abstraction.md)
Supersedes #708

PR #905, opened 2021-06-13
p2p: Implement Supervisor

xla:
*Work In Progress*

This is a placeholder PR to hold all the outstanding changes for the p2p work, will be broken apart further.

Stacked on top of #904


thanethomson:
I've been looking through the abstractions again in the `p2p/src/transport.rs` module, and wanted to ask if there's a particular reason why "read/write" and "send/receive" naming conventions are mixed in the P2P architecture?

For example, the [`Connection` trait](https://github.com/informalsystems/tendermint-rs/blob/d8e18c647cd8695d16610c4292b15ec6d1b45fbc/p2p/src/transport.rs#L71) has `StreamRead` and `StreamSend` associated types (instead of `StreamRead`/`StreamWrite` or `StreamSend`/`StreamReceive`).

xla:
> For example, the [`Connection` trait](https://github.com/informalsystems/tendermint-rs/blob/d8e18c647cd8695d16610c4292b15ec6d1b45fbc/p2p/src/transport.rs#L71) has `StreamRead` and `StreamSend` associated types (instead of `StreamRead`/`StreamWrite` or `StreamSend`/`StreamReceive`).

In order to distinguish physical connection semantics I'm keen on using send/receive for streams. The inconsistency you are pointing out is definitely an oversight and needs to be addressed.

Issue #907, opened 2021-06-16
net::Address Display omits peer_id


tzemanovic:
The parser `FromStr` for `net::Address` parses out a peer_id from a TCP address, if any, but the `Display` impl omits it.

thanethomson:
Closed by #908

PR #908, opened 2021-06-16
Tendermint config handle optional values + fix `net::Address` Display

### LINKED ISSUE #906: Tendermint config optional values parsing
The Tendermint config parser from toml uses a custom serde deserializing function for optional config values `deserialize_optional_value`, which requires an empty string when there's no value. If the optional field is missing from the config altogether, it fails with e.g.:

"missing field `tls_cert_file` for key `rpc` at line 96 column 1"

This could be relaxed as the Tendermint node does accept missing optional fields.

Furthermore, the serialization for the config omits the fields with optional values when there's no value, so it cannot parse back what it has written.

tzemanovic:
closes #906, #907

* [x] Wrote tests

### DIFF
--- a/.changelog/unreleased/bug-fixes/908-config-opt-values.md
+++ b/.changelog/unreleased/bug-fixes/908-config-opt-values.md
@@ -1,0 +1,1 @@
+- `[tendermint]` Better handling of optional values in TendermintConfig ([#908](https://github.com/informalsystems/tendermint-rs/issues/908))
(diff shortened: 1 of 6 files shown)


thanethomson (approved):
Thanks for this!

For some reason the doc test (which runs on Rust nightly) keeps failing here. I ran the doc tests locally on my machine for this PR, using nightly, and it compiles just fine, so I'm not sure what the problem is.

PR #914, opened 2021-06-22
Unpin Tokio version and fix docs test in CI

### LINKED ISSUE #911: Unpin Tokio version once nightly docs compile again
Following from #910, where I pinned the Tokio version to 1.6.2 to deal with https://github.com/tokio-rs/tokio/issues/3880, we should unpin it and rather use a newer version once that's fixed.

thanethomson:
Closes #911 

Seems like we were using the wrong environment variable in CI: https://github.com/tokio-rs/tokio/issues/3880#issuecomment-866056546

* [ ] Wrote tests

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,7 +56,7 @@ jobs:
           override: true
       - uses: actions-rs/cargo@v1
         env:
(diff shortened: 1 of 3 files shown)


xla (approved):
♠️ 🎤 ⌨ ↗️

PR #917, opened 2021-06-23
update Tendermint genesis

### LINKED ISSUE #916: Tendermint genesis is outdated
The genesis module doesn't seem to be compatible with Tendermint v0.34.x:
- missing `initial_height` field
- missing `consensus.block.time_iota_ms`
- incompatible type for `app_hash`
- `validators.voting_power` is not accepted by Tendermint Go, the default should be `validators.power`
- missing optional field `validators.name`

tzemanovic:
closes #916 

* [x] Wrote tests

### DIFF
--- a/tendermint/src/genesis.rs
+++ b/tendermint/src/genesis.rs
@@ -25,7 +25,8 @@ pub struct Genesis<AppState = serde_json::Value> {
     pub validators: Vec<validator::Info>,
 
     /// App hash
-    pub app_hash: String,
+    #[serde(with = "serializers::bytes::hexstring")]
+    pub app_hash: Vec<u8>,
 
     /// App state
     #[serde(default)]
--- a/tendermint/src/genesis.rs
+++ b/tendermint/src/genesis.rs
@@ -2,11 +2,7 @@
 
 use crate::serializers;
 use crate::{chain, consensus, validator, Time};
-use chrono::DateTime;
-use serde::de::Error;
-use serde::{Deserialize, Deserializer, Serialize};
-use std::convert::TryFrom;
-use tendermint_proto::google::protobuf::Timestamp;
+use serde::{Deserialize, Serialize};
 
 /// Genesis data
 #[derive(Clone, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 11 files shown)


tendermint/src/genesis.rs
```diff
@@ -24,25 +25,9 @@ pub struct Genesis<AppState = serde_json::Value> {
     pub validators: Vec<validator::Info>,
 
     /// App hash
-    #[serde(skip_serializing_if = "Vec::is_empty", with = "serde_bytes")]
-    pub app_hash: Vec<u8>,
+    pub app_hash: String,
```
thanethomson, line 28: What's your reasoning behind this change?

thanethomson (commented):
Just one question from my side. Otherwise looks good! :pray:

tendermint/src/genesis.rs
tzemanovic: In our app, we're updating the genesis.json generated from Tendermint Go. It sets `"app_hash": ""`, which is parsed as an empty `Vec`, but when we serialize it back to json, it omits the field completely. Then on subsequent run, the parser fails due to `Error("missing field `app_hash`", ...)`

thanethomson: And what if you changed it to:

```suggestion
    #[serde(with = "serde_bytes")]
    pub app_hash: Vec<u8>,
```

In the Go version of Tendermint it's modeled as `HexBytes`, which is an array of bytes, and I think it should be modeled similarly here.

tzemanovic: unfortunately, this writes back `"app_hash": []` into JSON, which is not accepted :/

tzemanovic: perhaps this could use a custom serialize/deserialize with hex encoding/decoding

tzemanovic: `#[serde(with="serializers::bytes::hexstring")]` does it, I'll fix it up

thanethomson (approved):
LGTM! :+1:

PR #920, opened 2021-06-24
Rebuild Protobuf data structures for Tendermint v0.34.9

### LINKED ISSUE #871: prost 0.7 uses core types instead of std
Prost 0.7 moved all internally used types (Vec, String, etc) generated from the core library, instead of std.

This is cool for our no-std approach but will it have any impact on downstream projects?

Get an answer if we're ok going ahead with using `core` types in `tendermint-proto`.
We should rebuild the proto library if this is ok.

thanethomson:
Closes #871 

Additionally just updates `tendermint-rs` to ensure compatibility with the latest released version of Gaia. I've given the (automatically produced) changes a once-over and it doesn't look like there are any meaningful changes.

If so, then this PR effectively just reduces the noise for any future rebuilds of the serialization types to make those changes easier to see.

* [ ] Wrote tests


xla (approved):
⛺️ 😝 📒 🏪

PR #921, opened 2021-06-25
Add equality/ordering/hash trait bounds to RPC URL types

### LINKED ISSUE #919: Add `PartialEq, Eq, PartialOrd, Ord, Hash` bounds to `tendermint_rpc::Url`
On order to implement runtime configuration reload in Hermes, we need to be able to compare two configurations for equality.
We are currently not able to do that, because our configuration struct contains a few fields of type `tendermint_rpc::Url`, which does not implement the `Eq` trait.

It would be great if we could derive this trait for `tendermint_rpc::Url`, and add other standard trait bounds while we're at it. The full list I am thinking of is: `PartialEq, Eq, PartialOrd, Ord, Hash`.

`tendermint_rpc::Url` implements `PartialEq, Eq, PartialOrd, Ord, Hash`.

thanethomson:
Closes #919 

* [ ] Wrote tests


romac (approved):
💙

Issue #859, opened 2021-04-07
Reintroduce query parsing for `tendermint-rpc`


thanethomson:
#858 removed RPC query parsing to address #857. We should reintroduce it, but possibly look at replacing `rust-peg` with a different parser (e.g. `nom` or `LALRPOP`).

**What's the definition of "done" for this issue?**

Once we have reintroduced:
- [ ] RPC query parsing from arbitrary strings.
- [ ] Subscription functionality in the [`tendermint-rpc` CLI](https://github.com/informalsystems/tendermint-rs/tree/master/rpc/src/client/bin)

PR #929, opened 2021-07-16
Propagate JSON-RPC errors through the Rust subscription

romac:
Close: #932

Emit JSON-RPC errors sent through the WS subscription through the `Subscription` channel, so that the Hermes event monitor can detect those (eg. when the WS subscription is closed by the Tendermint node) and attempt to resubscribe.

* [ ] Wrote tests


rpc/src/client/transport/router.rs
```diff
@@ -17,36 +19,54 @@ pub struct SubscriptionRouter {
     // their result channels. Used for publishing events relating to a specific
     // query.
     subscriptions: HashMap<String, HashMap<String, SubscriptionTx>>,
+
+    // A map of subscription ids to their queries
+    sub_to_query: HashMap<String, String>,
 }
 
 impl SubscriptionRouter {
+    pub fn publish_error(&mut self, id: &str, err: Error) -> PublishResult {
```
romac, line 28: Not sure why Clippy is complaining as this method is used in `rpc/src/client/transport/websocket.rs` at line 688.

rpc/src/client/transport/websocket.rs
```diff
@@ -650,38 +652,69 @@ impl WebSocketClientDriver {
...
+            self.respond_to_pending_command(pending_cmd, msg).await?;
         };
+
         // We ignore incoming messages whose ID we don't recognize (could be
         // relating to a fire-and-forget unsubscribe request - see the
         // publish_event() method below).
         Ok(())
     }
 
+    async fn publish_error(&mut self, id: &str, err: Error) {
```
romac, line 687: There's a bit of duplicated code between `publish_error` and `publish_event`, we should probably clean this up

rpc/src/response.rs
```diff
@@ -51,6 +51,12 @@ where
         &self.id
     }
 
+    /// Convert this wrapper into the underlying error, if any
+    #[allow(dead_code)]
```
romac, line 55:
```suggestion
```

rpc/src/client/transport/router.rs
```diff
@@ -17,36 +19,54 @@ pub struct SubscriptionRouter {
     // their result channels. Used for publishing events relating to a specific
     // query.
     subscriptions: HashMap<String, HashMap<String, SubscriptionTx>>,
```
romac, line 21: We should probably use type aliases or even new types here, eg.

```suggestion
    subscriptions: HashMap<SubscriptionQuery, HashMap<SubscriptionId, SubscriptionTx>>,
```

rpc/src/client/transport/router.rs
```diff
@@ -17,36 +19,54 @@ pub struct SubscriptionRouter {
...
+
+    // A map of subscription ids to their queries
+    sub_to_query: HashMap<String, String>,
```
romac, line 24: Same here

```suggestion
    sub_to_query: HashMap<SubscriptionId, SubscriptionQuery>,
```

thanethomson, line 24: Is this an optimization to allow for reverse lookup of queries from their subscription IDs?

rpc/src/client/transport/router.rs
```diff
@@ -98,7 +123,7 @@ impl Default for SubscriptionRouter {
 pub enum PublishResult {
     Success,
     NoSubscribers,
-    AllDisconnected,
+    AllDisconnected(String),
```
thanethomson, line 126:
```suggestion
    // All subscriptions for the given query have disconnected.
    AllDisconnected(String),
```

rpc/src/client/transport/router.rs
thanethomson: Might be due to the features enabled in CI when running the clippy check.

rpc/src/client/transport/router.rs
thanethomson: Makes sense. I do have an issue open for reintroducing query parsing into the codebase (#859) that'll provide the `Query` type.

rpc/src/client/transport/router.rs
romac: Yes, maybe I should i prove the naming and the comment if that wasn't so clear.

thanethomson: It's more the optimization I was wondering about. I generally prefer to have a "single source of truth" when it comes to the data being managed, because data duplication results in additional complexity. In this case, what's the computational cost of doing a query lookup from an ID from the `subscriptions` data structure?

We could do something like:

```rust
/// Returns the query string associated with the subscription with the given ID.
pub fn get_query_by_id(&self, id: &str) -> Option<String> {
    self.subscriptions.iter().find_map(|(query, subs)| {
        if subs.contains_key(id) {
            Some(query.clone())
        } else {
            None
        }
    })
}
```

romac: I share your concern about the single source of truth but I am not a big fan of iterating over the map to find the matching entry. How about we can extract this into a generic data structure with proper invariants enforced by the `contracts` create, to guarantee that the data does not go out of sync? Maybe we can even prove it correct with the Rust Stainless front end ;)

thanethomson: Just so I understand, are there any drawbacks of iterating through the map apart from the additional computational overhead?

For me, the optimized version provides `O(1)` lookup time at the cost of additional data structure management complexity (the two data structures become implicitly intertwined without any compile-time or runtime guarantees as to their coherence, and potentially complicating future changes). It introduces this data structure complexity for an edge case that is, at least partially, a bug in Tendermint.

From what I can see, the unoptimized, but simpler version provides `O(N)` lookup time (since each ID lookup is `O(1)`), where `N` = no. of unique queries for which there are subscriptions. How many subscription queries do we anticipate an RPC client will be creating simultaneously? 10's, 100's, 1000's?

romac: Good point, we will probably only ever have a handful of queries per RPC client (currently 2), so yeah I agree with you that this optimization is perhaps unnecessary 👍

.changelog/unreleased/bug-fixes/930-newblock-event-parse.md
```diff
@@ -0,0 +1,2 @@
+- Fix bug where `NewBlock` events emitted by Tendermint could not be parsed because of a missing field ([#930](https://github.com/informalsystems/tendermint-rs/issues/930))
```
thanethomson, line 1:
```suggestion
- `[tendermint-rpc]` Fix bug where `NewBlock` events emitted by Tendermint could not be parsed because of a missing field ([#930](https://github.com/informalsystems/tendermint-rs/issues/930))
```

thanethomson (commented):
Final minor nit, then I'd say we're good to go here!

I think there should be some follow-up work, which I can look into: to avoid all the `allow(dead_code)` directives, we need to fix the imports according to the features enabled. We should probably find a way to avoid importing the subscription router unless a subscription-enabled transport is required (e.g. the `websocket` feature).

thanethomson (approved):

PR #934, opened 2021-07-19
Ensure correct `TrustThresholdFraction` construction

### LINKED ISSUE #924: Trust threshold validation
The [TrustThresholdFraction](https://github.com/informalsystems/tendermint-rs/blob/master/tendermint/src/trust_threshold.rs#L25) as currently defined can be instantiated with an invalid `numerator`/`denominator` combination (see [ibc-rs/#1149](https://github.com/informalsystems/ibc-rs/issues/1149) and [ibc-rs/!1157](https://github.com/informalsystems/ibc-rs/pull/1157)), so we could enforce validation by making it's internal fields non-public and providing a constructor that always validates. 

A `TrustThresholdFraction` should only be constructible iff all below conditions are satisfied ->
* denominator != 0
* numerator * 3 >= denominator
* numerator < denominator

thanethomson:
Closes #924 

* [x] Wrote tests

### DIFF
--- a/tendermint/src/trust_threshold.rs
+++ b/tendermint/src/trust_threshold.rs
@@ -2,8 +2,10 @@
 
 use std::fmt::{self, Debug, Display};
 
-use crate::serializers;
+use crate::error::Error;
+use crate::{serializers, Kind};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::convert::TryFrom;
 
 /// TrustThreshold defines how much of the total voting power of a known
 /// and trusted validator set is sufficient for a commit to be
@@ -22,16 +24,22 @@ pub trait TrustThreshold: Copy + Clone + Debug + Serialize + DeserializeOwned {
 /// Some clients might require more than +1/3 and can implement their own
 /// [`TrustThreshold`] which can be passed into all relevant methods.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(
+    try_from = "RawTrustThresholdFraction",
+    into = "RawTrustThresholdFraction"
+)]
 pub struct TrustThresholdFraction {
-    /// Numerator of the trust threshold fraction
-    #[serde(with = "serializers::from_str")]
-    pub numerator: u64,
-    /// Numerator of the trust threshold fraction
-    #[serde(with = "serializers::from_str")]
-    pub denominator: u64,
+    numerator: u64,
+    denominator: u64,
 }
 
 impl TrustThresholdFraction {
+    /// Constant for a trust threshold of 1/3.
+    pub const ONE_THIRD: Self = Self {
+        numerator: 1,
+        denominator: 3,
+    };
+
     /// Constant for a trust threshold of 2/3.
     pub const TWO_THIRDS: Self = Self {
         numerator: 2,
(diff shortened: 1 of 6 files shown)


tendermint/src/trust_threshold.rs
```diff
@@ -42,15 +50,47 @@ impl TrustThresholdFraction {
...
-        if numerator <= denominator && denominator > 0 && 3 * numerator >= denominator {
-            Some(Self {
-                numerator,
-                denominator,
-            })
-        } else {
-            None
+    /// In any other case we return an error.
+    pub fn new(numerator: u64, denominator: u64) -> Result<Self, Error> {
+        if numerator > denominator {
```
hu55a1n1, line 55:
```suggestion
        if numerator >= denominator {
```
I believe that the `=1` case is ambiguous here because a `1/3` threshold would require `+1/3` of the total voting power and correspondingly a threshold `=1` would require `>1` of the total voting power. What do you think?

tendermint/src/trust_threshold.rs
```diff
@@ -73,3 +113,74 @@ impl Display for TrustThresholdFraction {
...
+#[derive(Serialize, Deserialize)]
+pub struct RawTrustThresholdFraction {
+    #[serde(with = "serializers::from_str")]
+    numerator: u64,
+    #[serde(with = "serializers::from_str")]
+    denominator: u64,
+}
+
+#[cfg(test)]
+mod test {
```
hu55a1n1, line 128: Thanks @thanethomson! Beautifully done! 💯👌

tendermint/src/trust_threshold.rs
```diff
@@ -22,16 +24,22 @@ pub trait TrustThreshold: Copy + Clone + Debug + Serialize + DeserializeOwned {
 /// Some clients might require more than +1/3 and can implement their own
 /// [`TrustThreshold`] which can be passed into all relevant methods.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(
+    try_from = "RawTrustThresholdFraction",
+    into = "RawTrustThresholdFraction"
+)]
 pub struct TrustThresholdFraction {
-    /// Numerator of the trust threshold fraction
-    #[serde(with = "serializers::from_str")]
```
romac, line 27: Do we know why we are serializing the numerator and denominator as strings?

tools/kvstore-test/tests/light-client.rs
```diff
@@ -74,23 +74,18 @@ fn make_supervisor() -> Supervisor {
     let node_address: rpc::Url = "http://127.0.0.1:26657".parse().unwrap();
 
     let options = light_client::Options {
-        trust_threshold: TrustThreshold {
-            numerator: 1,
-            denominator: 3,
-        },
+        trust_threshold: TrustThreshold::new(1, 3).unwrap(),
```
hu55a1n1, line 77: Just a minor suggesiton - if you like we could use `Default::default()` here. Also, the `impl default for TrustThresholdFraction` could reuse the `ONE_THIRD` const.

tendermint/src/trust_threshold.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/issues/1149

tendermint/src/trust_threshold.rs
thanethomson: They're `u64`, which can only fully be represented in JSON by a string afaik.

tendermint/src/trust_threshold.rs
thanethomson: Done

tendermint/src/trust_threshold.rs
thanethomson: Cool! Wasn't sure if proptest was a bit overkill for this, but I think it's still worthwhile to have more of this kind of testing in the codebase.

tools/kvstore-test/tests/light-client.rs
thanethomson: Done

hu55a1n1 (approved):

PR #935, opened 2021-07-20
Release v0.21.0

thanethomson:
:book: [CHANGELOG rendered](https://github.com/informalsystems/tendermint-rs/blob/release/v0.21.0/CHANGELOG.md#v0210)

The audit check will continue to fail due to #925 

* [ ] Wrote tests


CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
...
+
+### IMPROVEMENTS
+
+- `[tendermint-proto]` Regenerate the Rust equivalents of the Tendermint
+  Protobuf structures for Tendermint v0.34.9
+  ([#871](https://github.com/informalsystems/tendermint-rs/issues/871))
+- `[tendermint-rpc]` Add `PartialEq`, `Eq`, `PartialOrd`, `Ord` and `Hash` trait
+  bounds to the RPC URL types
+  ([#919](https://github.com/informalsystems/tendermint-rs/issues/919))
+- `[tendermint-rpc]` Propagate JSON-RPC errors through the Rust subscription ([#932](https://github.com/informalsystems/tendermint-rs/issues/932))
```
romac, line 43: I guess this is technically a breaking change since it changes the signature of the subscription API, ie. it provides a `Stream` of `Result<Event, Error>` instead of a stream of `Event`s.

thanethomson: Good catch. I'll update the CHANGELOG accordingly :+1:

romac: My bad, I should have filed the entry properly in the first place.

romac (approved):
Pre-approving already :)

CHANGELOG.md
thanethomson: I see that you only modified the `SubscriptionRouter`'s `publish` method's API though, which is only effectively used internally by the WebSocket transport. None of the public-facing transport or subscription API was changed, was it?

romac: Right, my bad!

xla (approved):
🚁 🍤 💽 💁

The change to Informal Systems as the author for some of the crates is sub-optimal I think.

thanethomson:
> The change to Informal Systems as the author for some of the crates is sub-optimal I think.

How so? :slightly_smiling_face: I'd imagine it'd just be easier to have a single point of contact for Informal Systems. All other authors' details are kept as-is.

PR #936, opened 2021-07-22
Fix Warning in Cargo Build

JonathanLorimer:
I believe that the "blessed" way to add profile settings in Cargo workspaces is with package overrides now
- [Cargo Documentation](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides)
- [Stack Overflow Thread](https://stackoverflow.com/questions/45794917/specific-profiles-for-workspace-members)

I was getting warnings when building which suggested that the profile setting in `light-client-js` wasn't doing anything because profiles need to be set on the project manifest.

* [ ] Wrote tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,7 @@ members = [
 exclude = [
     "proto-compiler"
 ]
+
+[profile.release.package.tendermint-light-client-js]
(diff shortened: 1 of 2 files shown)


thanethomson (approved):
Awesome, thanks @JonathanLorimer! :pray:

PR #938, opened 2021-07-23
Implement full-duplex secret connection

thanethomson:
Work towards #814 

Something like this is necessary for the MConnection to be able to safely read from and write to the underlying stream from multiple threads (e.g. one thread for reading, one thread for writing).

* [x] Wrote tests

### DIFF
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -227,7 +227,7 @@ impl Handshake<AwaitingAuthSig> {
 /// not both simultaneously).
 ///
 /// If, however, the underlying I/O handler class implements
-/// [`crate::transport::TryClone`], then you can use
+/// [`tendermint_std_ext::TryClone`], then you can use
 /// [`SecretConnection::try_split`] to split the `SecretConnection` into its
 /// sending and receiving halves. Each of these halves can then be used in a
 /// separate thread to facilitate full-duplex communication.
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -221,6 +221,12 @@ impl Handshake<AwaitingAuthSig> {
 
 /// Encrypted connection between peers in a Tendermint network.
 ///
+/// ## Connection integrity and failures
+///
+/// Due to the underlying encryption mechanism (currently [RFC 8439]), when a
+/// read or write failure occurs, it is necessary to disconnect from the remote
+/// peer and attempt to reconnect.
+///
 /// ## Half- and full-duplex connections
 /// By default, a `SecretConnection` facilitates half-duplex operations (i.e.
 /// one can either read from the connection or write to it at a given time, but
@@ -232,9 +238,11 @@ impl Handshake<AwaitingAuthSig> {
 /// sending and receiving halves. Each of these halves can then be used in a
 /// separate thread to facilitate full-duplex communication.
 ///
-/// ## Contract
+/// ## Contracts
 ///
 /// When reading data, data smaller than [`DATA_MAX_SIZE`] is read atomically.
+///
+/// [RFC 8439]: https://www.rfc-editor.org/rfc/rfc8439.html
 pub struct SecretConnection<IoHandler> {
     io_handler: IoHandler,
     protocol_version: Version,
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -469,8 +469,8 @@ fn encrypt(
     send_nonce: &Nonce,
     sealed_frame: &mut [u8; TAG_SIZE + TOTAL_FRAME_SIZE],
 ) -> Result<()> {
-    debug_assert!(!chunk.is_empty(), "chunk is empty");
-    debug_assert!(
+    assert!(!chunk.is_empty(), "chunk is empty");
+    assert!(
         chunk.len() <= TOTAL_FRAME_SIZE - DATA_LEN_SIZE,
         "chunk is too big: {}! max: {}",
         chunk.len(),
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -351,7 +351,7 @@ where
     /// ## Errors
     /// Fails when the `try_clone` operation for the underlying I/O handler
     /// fails.
-    pub fn try_split(self) -> Result<(Sender<IoHandler>, Receiver<IoHandler>)> {
+    pub fn split(self) -> Result<(Sender<IoHandler>, Receiver<IoHandler>)> {
         let remote_pubkey = self.remote_pubkey.expect("remote_pubkey to be initialized");
         Ok((
             Sender {
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -1,12 +1,12 @@
 //! `SecretConnection`: Transport layer encryption for Tendermint P2P connections.
 
-use std::{
-    cmp,
-    convert::{TryFrom, TryInto},
-    io::{self, Read, Write},
-    marker::{Send, Sync},
-    slice,
-};
+use std::cmp;
+use std::convert::{TryFrom, TryInto};
+use std::io::{self, Read, Write};
+use std::marker::{Send, Sync};
+use std::slice;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
 
 use chacha20poly1305::{
     aead::{generic_array::GenericArray, AeadInPlace, NewAead},
@@ -219,6 +219,28 @@ impl Handshake<AwaitingAuthSig> {
     }
 }
 
+// Macro usage allows us to avoid unnecessarily cloning the Arc<AtomicBool>
+// that indicates whether we need to terminate the connection.
+//
+// Limitation: this only checks once prior to the execution of an I/O operation
+// whether we need to terminate. This should be sufficient for our purposes
+// though.
+macro_rules! checked_io {
+    ($term:expr, $f:expr) => {{
+        if $term.load(Ordering::SeqCst) {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "secret connection was terminated elsewhere by previous error",
+            ));
+        }
+        let result = { $f };
+        if result.is_err() {
+            $term.store(true, Ordering::SeqCst);
+        }
+        result
+    }};
+}
+
 /// Encrypted connection between peers in a Tendermint network.
 ///
 /// ## Connection integrity and failures
@@ -249,6 +271,7 @@ pub struct SecretConnection<IoHandler> {
     remote_pubkey: Option<PublicKey>,
     send_state: SendState,
     recv_state: ReceiveState,
+    terminate: Arc<AtomicBool>,
 }
 
 impl<IoHandler: Read + Write + Send + Sync> SecretConnection<IoHandler> {
@@ -293,6 +316,7 @@ impl<IoHandler: Read + Write + Send + Sync> SecretConnection<IoHandler> {
                 recv_nonce: Nonce::default(),
                 recv_buffer: vec![],
             },
+            terminate: Arc::new(AtomicBool::new(false)),
         };
 
         // Share each other's pubkey & challenge signature.
@@ -334,11 +358,13 @@ where
                 io_handler: self.io_handler.try_clone()?,
                 remote_pubkey,
                 send_state: self.send_state,
+                terminate: self.terminate.clone(),
             },
             Receiver {
                 io_handler: self.io_handler,
                 remote_pubkey,
                 recv_state: self.recv_state,
+                terminate: self.terminate,
             },
         ))
     }
@@ -346,17 +372,23 @@ where
 
 impl<IoHandler: Read> Read for SecretConnection<IoHandler> {
     fn read(&mut self, data: &mut [u8]) -> io::Result<usize> {
-        read_and_decrypt(&mut self.io_handler, &mut self.recv_state, data)
+        checked_io!(
+            self.terminate,
+            read_and_decrypt(&mut self.io_handler, &mut self.recv_state, data)
+        )
     }
 }
 
 impl<IoHandler: Write> Write for SecretConnection<IoHandler> {
     fn write(&mut self, data: &[u8]) -> io::Result<usize> {
-        encrypt_and_write(&mut self.io_handler, &mut self.send_state, data)
+        checked_io!(
+            self.terminate,
+            encrypt_and_write(&mut self.io_handler, &mut self.send_state, data)
+        )
     }
 
     fn flush(&mut self) -> io::Result<()> {
-        self.io_handler.flush()
+        checked_io!(self.terminate, self.io_handler.flush())
     }
 }
 
@@ -378,6 +410,7 @@ pub struct Sender<IoHandler> {
     io_handler: IoHandler,
     remote_pubkey: PublicKey,
     send_state: SendState,
+    terminate: Arc<AtomicBool>,
 }
 
 impl<IoHandler> Sender<IoHandler> {
@@ -389,11 +422,14 @@ impl<IoHandler> Sender<IoHandler> {
 
 impl<IoHandler: Write> Write for Sender<IoHandler> {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        encrypt_and_write(&mut self.io_handler, &mut self.send_state, buf)
+        checked_io!(
+            self.terminate,
+            encrypt_and_write(&mut self.io_handler, &mut self.send_state, buf)
+        )
     }
 
     fn flush(&mut self) -> io::Result<()> {
-        self.io_handler.flush()
+        checked_io!(self.terminate, self.io_handler.flush())
     }
 }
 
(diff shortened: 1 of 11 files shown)


p2p/src/secret_connection.rs
```diff
@@ -218,16 +220,49 @@ impl Handshake<AwaitingAuthSig> {
...
-    send_nonce: Nonce,
     recv_cipher: ChaCha20Poly1305,
     send_cipher: ChaCha20Poly1305,
     remote_pubkey: Option<PublicKey>,
-    recv_buffer: Vec<u8>,
+    // The portion of the connection state that is mutated when sending.
+    send_state: Arc<Mutex<SecretConnectionSendState>>,
+    // The portion of the connection state that is mutated when receiving.
+    recv_state: Arc<Mutex<SecretConnectionRecvState>>,
 }
```
tony-iqlusion, line 266: Why not just `Arc<Mutex<SecretConnection<_>>`?

thanethomson: We could, but then it wouldn't be full-duplex, would it? (i.e. reading would block writing, and vice-versa).

I have two main assumptions here:

1. That `std::net::TcpStream` is full-duplex (I'm not 100% sure if it is though).
2. That the MConnection, which will make use of the `SecretConnection`, will need full-duplex communication (e.g. while we're receiving large chunks of data from a peer, we may need to write out to that peer simultaneously).

tony-iqlusion: Aah, okay, well in that case why not have a type for a half-duplex socket which only has a key for sending in one particular direction?

thanethomson: So do you mean to have a `SecretConnectionSender` and a `SecretConnectionReceiver`?

thanethomson: Part of the benefit I see of the approach I've implemented here is that it can be used in a single-threaded way, as is used in the KMS, which shouldn't change the existing communication pattern. But if the underlying `IoHandler` implements `TryClone` then you can have full-duplex communication in a multi-threaded way.

tony-iqlusion: You could still compose the full duplex version from the half-duplex versions. It'd require always cloning, but I don't think that should be a problem.

thanethomson: Hmmm, yes, if we go that route we'd probably need to redo the tests for the `SecretConnection`, because they currently depend on this [pipe](https://github.com/arcnmx/pipe-rs) lib whose endpoints can't be cloned.

tarcieri: Perhaps you could use a Unix domain socket instead?

tony-iqlusion: Perhaps you could use a Unix domain socket instead? (although I suppose that would preclude testing on Windows, at least those tests anyway)

thanethomson: Not sure why TCP-based tests weren't used in the first place, but they seem to work just fine for our purposes (and act as decent integration tests too), so we may as well just use TCP-based tests right?

tony-iqlusion: Yeah TCP works just fine and is portable, the only tricky part is locating an unused port and handling the error case if something is bound to the test port

thanethomson: Afaik Rust handles that automatically when you `bind` to `127.0.0.1:0`, doesn't it?

p2p/src/secret_connection.rs
```diff
@@ -452,6 +496,23 @@ where
     }
 }
 
+impl<IoHandler: TryClone> TryClone for SecretConnection<IoHandler> {
```
soareschen, line 499: What is the advantage of using this trait as compared to holding an `Arc<IoHandler>` and `Clone`?

p2p/src/secret_connection.rs
```diff
@@ -431,11 +475,11 @@ where
                 data_copy = &[0_u8; 0];
             }
             let sealed_frame = &mut [0_u8; TAG_SIZE + TOTAL_FRAME_SIZE];
-            let res = self.encrypt(chunk, sealed_frame);
+            let res = self.encrypt(chunk, &send_state.send_nonce, sealed_frame);
             if let Err(err) = res {
                 return Err(io::Error::new(io::ErrorKind::Other, err.to_string()));
             }
-            self.send_nonce.increment();
+            send_state.send_nonce.increment();
```
soareschen, line 482: Should we have higher level abstraction to guarantee that the nonce is always incremented after being acquired? Or should it never be incremented at all if there is error? Would the state be inconsistent if the line above or below returns error prematurely?

p2p/src/secret_connection.rs
```diff
@@ -431,11 +475,11 @@ where
                 data_copy = &[0_u8; 0];
             }
             let sealed_frame = &mut [0_u8; TAG_SIZE + TOTAL_FRAME_SIZE];
-            let res = self.encrypt(chunk, sealed_frame);
+            let res = self.encrypt(chunk, &send_state.send_nonce, sealed_frame);
```
soareschen, line 478: We can do

```rust
self.encrypt(chunk, &send_state.send_nonce, sealed_frame)
  .map_err(|err| io::Error::new(io::ErrorKind::Other, err.to_string()))?;
```

With the `?` short circuiting the loop and returns the error result.

p2p/src/secret_connection.rs
```diff
@@ -357,18 +392,23 @@ where
 {
     // CONTRACT: data smaller than DATA_MAX_SIZE is read atomically.
     fn read(&mut self, data: &mut [u8]) -> io::Result<usize> {
-        if !self.recv_buffer.is_empty() {
-            let n = cmp::min(data.len(), self.recv_buffer.len());
-            data.copy_from_slice(&self.recv_buffer[..n]);
+        let mut recv_state = self
+            .recv_state
+            .lock()
+            .expect("failed to obtain lock on secret connection recv state");
```
soareschen, line 398: Should we return error result here instead of panicking? Same for the other uses of `.expect()`.

p2p/src/transport.rs
```diff
@@ -140,6 +140,32 @@ where
...
+    fn try_clone(&self) -> Result<Self, Self::Error>;
+}
+
+impl TryClone for TcpStream {
+    type Error = std::io::Error;
+
+    fn try_clone(&self) -> Result<Self, Self::Error> {
+        // Uses the TcpStream struct's method. See
+        // https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name
+        self.try_clone()
```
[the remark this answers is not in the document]
thanethomson: It wasn't obvious to me whether the `TcpStream`'s `try_clone` method was going to be called here or the `TryClone` trait's `try_clone`, so I thought I'd leave a comment here for posterity to explain it.

p2p/src/secret_connection.rs
thanethomson: See https://github.com/informalsystems/tendermint-rs/pull/938#discussion_r675794117

p2p/src/secret_connection.rs
thanethomson: How would we (and should we be able to) recover from such an error?

My understanding is that, when [`Mutex::lock`](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock) fails, it's due to a panic while the lock is held by another thread.

p2p/src/transport.rs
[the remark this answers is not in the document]
thanethomson: Oh, also, I tried explicitly calling `TcpStream::try_clone(self)` here and clippy complains that the better approach is to use `Self::try_clone(self)`, which, to me, isn't nearly as clear as `TcpStream::try_clone(self)`.

thanethomson:
I've converted this back to a draft PR while I rework things a bit as per Tony's suggestions.

p2p/src/secret_connection.rs
tarcieri: Preventing nonce reuse is the most important thing as it fails catastrophically with ChaCha20Poly1305, but also if it's ever incremented without being used it will be out-of-sync with the peer and abort the connection that way.

So it's best to have a sort of simple state machine that aborts the whole connection on errors, and if that can wipe the nonce somehow that's best, but so long as you can prevent repeat nonce reuse that's probably the most important thing.

tony-iqlusion: Preventing nonce reuse is the most important thing as it fails catastrophically with ChaCha20Poly1305, but also if it's ever incremented without being used it will be out-of-sync with the peer and abort the connection that way.

So it's best to have a sort of simple state machine that aborts the whole connection on errors, and if that can wipe the nonce somehow that's best, but so long as you can prevent repeat nonce reuse that's probably the most important thing.

p2p/src/secret_connection.rs
soareschen: Sounds reasonable. Perhaps we can add a comment here that we are panicking because a poisoned mutex is unrecoverable. This is mainly to distinguish it from the other uses of `.expect` in the same file, which IMO should be refactored to return `Result::Err` instead.

tony-iqlusion (dismissed):
Looks good now!

xla (commented):
Very sensible, most left some clarifying questions. Are the concerns addressed wrt nonce handling in https://github.com/informalsystems/tendermint-rs/pull/938#discussion_r676424442?

p2p/src/transport.rs
```diff
@@ -140,6 +140,32 @@ where
...
     /// # Errors
     ///
-    /// * If resource allocation fails for lack of priviliges or being not available.
+    /// * If resource allocation fails for lack of privileges or being not available.
     fn bind(self, bind_info: BindInfo<A>) -> Result<(Self::Endpoint, Self::Incoming)>;
 }
+
+/// Types that can be cloned where success is not guaranteed can implement this
+/// trait.
+pub trait TryClone: Sized {
```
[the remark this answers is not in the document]
thanethomson: So concretely, do you mean creating a new crate in the tendermint-rs repo called `std-ext`? Or perhaps `tendermint-std-ext` to avoid potential crate naming clashes in future?

p2p/src/secret_connection.rs
```diff
@@ -509,3 +460,176 @@ pub fn sort32(first: [u8; 32], second: [u8; 32]) -> ([u8; 32], [u8; 32]) {
...
+/// Encrypt AEAD authenticated data
+#[allow(clippy::cast_possible_truncation)]
+fn encrypt(
+    chunk: &[u8],
+    send_cipher: &ChaCha20Poly1305,
+    send_nonce: &Nonce,
+    sealed_frame: &mut [u8; TAG_SIZE + TOTAL_FRAME_SIZE],
+) -> Result<()> {
+    debug_assert!(!chunk.is_empty(), "chunk is empty");
+    debug_assert!(
```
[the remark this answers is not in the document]
thanethomson: Good question. @tony-iqlusion?

p2p/src/secret_connection.rs
```diff
@@ -285,170 +297,109 @@ impl<IoHandler: Read + Write + Send + Sync> SecretConnection<IoHandler> {
...
-        if chunk_length as usize > DATA_MAX_SIZE {
-            return Err(io::Error::new(
-                io::ErrorKind::Other,
-                format!("chunk is too big: {}! max: {}", chunk_length, DATA_MAX_SIZE),
-            ));
-        }
+/// The sending end of a [`SecretConnection`].
+pub struct Sender<IoHandler> {
+    io_handler: IoHandler,
+    remote_pubkey: Option<PublicKey>,
```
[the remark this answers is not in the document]
thanethomson: Right - that should probably not be optional.

p2p/src/secret_connection.rs
```diff
@@ -509,3 +460,176 @@ pub fn sort32(first: [u8; 32], second: [u8; 32]) -> ([u8; 32], [u8; 32]) {
...
+            b"",
+            in_out,
+            tag.into(),
+        )
+        .map_err(|_| Error::Crypto)?;
+
+    Ok(in_out.len())
+}
+
+// CONTRACT: data smaller than DATA_MAX_SIZE is read atomically.
```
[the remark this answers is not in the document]
thanethomson: I was transferring the comments from the previous iterations of the code. It still seems like something that's worth communicating to users, so perhaps I'll put it into the `SecretConnection`'s API docs.

thanethomson:
> Are the concerns addressed wrt nonce handling in [#938 (comment)](https://github.com/informalsystems/tendermint-rs/pull/938#discussion_r676424442)?

I'm not at all familiar with the low-level details of ChaCha20Poly1305.

@tony-iqlusion, does this mean that, if there's any failure to read from/write to a `SecretConnection`, the whole connection needs to be dropped and re-established to be able to continue communicating? If so that's something I should probably add to the documentation.

p2p/src/secret_connection.rs
thanethomson: Fortunately in the new iteration there's no need for mutexes any more :grin:

p2p/src/transport.rs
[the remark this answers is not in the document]
xla: Yeah that was my meaning indeed, sorry for leaving it up for interpretation.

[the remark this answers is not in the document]
thanethomson: Cool, I can kick off a new crate :+1: Any particular naming preferences? I feel like `tendermint-std-ext` is probably a good bet.

tony-iqlusion:
@thanethomson other than a timeout (which you could potentially retry), yes, any other errors should abort the connection, or at least that's how something like the Noise Protocol state machine works

xla:
> @tony-iqlusion, does this mean that, if there's any failure to read from/write to a `SecretConnection`, the whole connection needs to be dropped and re-established to be able to continue communicating? If so that's something I should probably add to the documentation.

Can we implement a switch with kills both hands if either of them fail?

thanethomson:
> Can we implement a switch with kills both hands if either of them fail?

Sure, I could look into that.

tony-iqlusion:
Shouldn't be too tricky with something like an `AtomicBool`

thanethomson:
The latest commits build incrementally on previous ones, so review shouldn't be too complex.

One question though for those who've done/seen this sort of performance benchmarking in Rust before: what's the potential performance hit of loading an `AtomicBool`'s value before each time we read from/write to the underlying TCP stream?

tony-iqlusion:
> what's the potential performance hit of loading an AtomicBool's value before each time we read from/write to the underlying TCP stream?

Vanishingly small compared to the system call overhead of an I/O operation. On x86(-64) in particular, the architecture provides a strongly consistent memory model, which means `AtomicBool` is effectively zero overhead versus an unsynchronized load/store regardless of what consistency model you choose.

Things are a bit more complicated on e.g. ARM, where it will require flushing (potentially multicore) CPU pipelines, but again, minuscule compare to the overhead of a system call.

thanethomson:
Sounds good, thanks for the insight @tony-iqlusion! :pray:

p2p/src/secret_connection.rs
```diff
@@ -285,170 +334,126 @@ impl<IoHandler: Read + Write + Send + Sync> SecretConnection<IoHandler> {
...
+    /// [`tendermint_std_ext::TryClone`], this attempts to split such a
+    /// connection into its sending and receiving halves.
+    ///
+    /// This facilitates full-duplex communications when each half is used in
+    /// a separate thread.
+    ///
+    /// ## Errors
+    /// Fails when the `try_clone` operation for the underlying I/O handler
+    /// fails.
+    pub fn try_split(self) -> Result<(Sender<IoHandler>, Receiver<IoHandler>)> {
```
tony-iqlusion, line 354: Personally I'd just call this `split` unless there's already an analogous `split` method (i.e. `try_*` is helpful for disambiguating a fallible alternative to an existing method)

xla (commented):
With https://github.com/informalsystems/tendermint-rs/pull/905#issuecomment-887541465 in mind should the internals and the handles acquired from a `split` be named `Reader` and `Writer` respectively? Esp. as they implement `Read` and `Write` and receiver and sender can be reserved for abstractions which offer full messages as they input/output.

thanethomson:
> With [#905 (comment)](https://github.com/informalsystems/tendermint-rs/pull/905#issuecomment-887541465) in mind should the internals and the handles acquired from a `split` be named `Reader` and `Writer` respectively? Esp. as they implement `Read` and `Write` and receiver and sender can be reserved for abstractions which offer full messages as they input/output.

I've been trying to figure out which set of idioms (either "send/receive" or "read/write") should apply where, and it seems to me like "read/write" seems more general than "send/receive" (e.g. the latter doesn't make too much sense when interacting with the filesystem).

Networked entities seem to send/receive more than read/write. This is probably due to the fact that low-level socket functions like [`send`](https://www.man7.org/linux/man-pages/man2/send.2.html) employ "send/receive" semantics on networking connections. As such, it probably makes more sense to keep the `SecretConnection`'s "send/receive" semantics. It's already pretty pervasive throughout the `SecretConnection`-related code.

But either way, as long as it's consistent, I'd be fine with it. My issue in https://github.com/informalsystems/tendermint-rs/pull/905#issuecomment-887541465 was that it was inconsistent.

xla (approved):
🏌 😜 🏷 🏉

PR #941, opened 2021-08-05
Fix recent clippy errors on `master`

thanethomson:
Some new clippy errors have appeared recently on `master` (most likely due to upgrading to Rust 1.54). This PR aims to fix them.

See https://github.com/informalsystems/tendermint-rs/runs/3242281271

* [ ] Wrote tests

### DIFF
--- a/light-client/src/store/memory.rs
+++ b/light-client/src/store/memory.rs
@@ -81,6 +81,7 @@ impl LightStore for MemoryStore {
             .map(|(_, e)| e.light_block.clone())
     }
 
+    #[allow(clippy::needless_collect)]
     fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock>> {
         let light_blocks: Vec<_> = self
             .store
--- a/light-client/src/components/verifier.rs
+++ b/light-client/src/components/verifier.rs
@@ -113,8 +113,8 @@ impl Verifier for ProdVerifier {
             &*self.voting_power_calculator,
             &*self.commit_validator,
             &*self.hasher,
-            &trusted,
-            &untrusted,
+            trusted,
+            untrusted,
             options,
             now,
         )
--- a/light-client/src/evidence.rs
+++ b/light-client/src/evidence.rs
@@ -70,7 +70,7 @@ mod prod {
         #[pre(self.peer_map.contains_key(&peer))]
         fn rpc_client_for(&self, peer: PeerId) -> Result<rpc::HttpClient, IoError> {
             let peer_addr = self.peer_map.get(&peer).unwrap().to_owned();
-            Ok(rpc::HttpClient::new(peer_addr).map_err(IoError::from)?)
+            rpc::HttpClient::new(peer_addr).map_err(IoError::from)
         }
     }
 }
--- a/light-client/src/macros.rs
+++ b/light-client/src/macros.rs
@@ -4,7 +4,7 @@
 #[macro_export]
 macro_rules! bail {
     ($kind:expr) => {
-        return Err($kind.into());
+        return Err($kind.into())
     };
 }
 
--- a/light-client/src/peer_list.rs
+++ b/light-client/src/peer_list.rs
@@ -115,7 +115,7 @@ impl<T> PeerList<T> {
     /// - The given peer id must not be the primary peer id.
     /// - The given peer must be in the witness list
     #[pre(faulty_witness != self.primary && self.witnesses.contains(&faulty_witness))]
-    #[post(Self::invariant(&self))]
+    #[post(Self::invariant(self))]
     pub fn replace_faulty_witness(&mut self, faulty_witness: PeerId) -> Option<PeerId> {
         let mut result = None;
 
@@ -137,7 +137,7 @@ impl<T> PeerList<T> {
     ///
     /// ## Errors
     /// - If there are no witness left, returns `ErrorKind::NoWitnessLeft`.
-    #[post(ret.is_ok() ==> Self::invariant(&self))]
+    #[post(ret.is_ok() ==> Self::invariant(self))]
     pub fn replace_faulty_primary(
         &mut self,
         primary_error: Option<Error>,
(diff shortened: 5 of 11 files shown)


light-client/src/store/memory.rs
```diff
@@ -81,6 +81,7 @@ impl LightStore for MemoryStore {
             .map(|(_, e)| e.light_block.clone())
     }
 
+    #[allow(clippy::needless_collect)]
```
romac, line 84: Why can't we just return the boxes iterator directly here? I assume you've tried that already, so just curious what's the issue here.

romac (approved):
Quick question, per-approving already

light-client/src/store/memory.rs
thanethomson: Yeah it's a good question. I refactored the code to the following:

```rust
    fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock>> {
        let light_blocks = self
            .store
            .iter()
            .filter(|(_, e)| e.status == status)
            .map(|(_, e)| e.light_block.clone());

        Box::new(light_blocks)
    }
```

Is that what you mean?

If so, this is the error I get from the compiler:

```
error[E0759]: `self` has an anonymous lifetime `'_` but it needs to satisfy a `'static` lifetime requirement
  --> light-client/src/store/memory.rs:87:14
   |
84 |       fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock>> {
   |              ----- this data with an anonymous lifetime `'_`...
85 |           let light_blocks = self
   |  ____________________________-
86 | |             .store
   | |__________________- ...is captured here...
87 |               .iter()
   |                ^^^^
...
91 |           Box::new(light_blocks)
   |           ---------------------- ...and is required to live as long as `'static` here
   |
help: to declare that the trait object captures data from argument `self`, you can add an explicit `'_` lifetime bound
   |
84 |     fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock> + '_> {
   |                                                                          ^^^^

```

I'm not totally comfortable just yet with the way that these lifetime bounds work (due to lack of experience with using them). Does declaring `that the trait object captures data from argument 'self'` make sense to you? What does this mean, practically?

thanethomson: And if I apply the suggestion, we'll need to modify the trait signature:

```
error: `impl` item signature doesn't match `trait` item signature
  --> light-client/src/store/memory.rs:84:5
   |
84 |     fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock> + '_> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ found `fn(&MemoryStore, types::Status) -> Box<dyn Iterator<Item = types::LightBlock>>`
   | 
  ::: light-client/src/store.rs:48:5
   |
48 |     fn all(&self, status: Status) -> Box<dyn Iterator<Item = LightBlock>>;
   |     ---------------------------------------------------------------------- expected `fn(&MemoryStore, types::Status) -> Box<(dyn Iterator<Item = types::LightBlock> + 'static)>`
   |
   = note: expected `fn(&MemoryStore, types::Status) -> Box<(dyn Iterator<Item = types::LightBlock> + 'static)>`
              found `fn(&MemoryStore, types::Status) -> Box<dyn Iterator<Item = types::LightBlock>>`
   = help: the lifetime requirements from the `impl` do not correspond to the requirements in the `trait`
   = help: verify the lifetime relationships in the `trait` and `impl` between the `self` argument, the other inputs and its output
```

tony-iqlusion: @thanethomson heh, that's a really "fun" syntax it's nearly impossible to discover without assistance from rustc...

That said, why not write it with `impl Trait` instead of `Box` + `dyn`?

```rust
    fn all(&self, status: Status) -> impl Iterator<Item = LightBlock> + '_ {
       self
            .store
            .iter()
            .filter(|(_, e)| e.status == status)
            .map(|(_, e)| e.light_block.clone())
    }
```

romac: @tarcieri Because `all` is a trait method and `impl Trait` return types in trait methods are not supported (yet?)

tony-iqlusion: Aah right, would probably require GATs

thanethomson: So after refactoring the trait to include the lifetime bound, I get:

```
error[E0373]: closure may outlive the current function, but it borrows `status`, which is owned by the current function
  --> light-client/src/store/memory.rs:88:21
   |
88 |             .filter(|(_, e)| e.status == status)
   |                     ^^^^^^^^             ------ `status` is borrowed here
   |                     |
   |                     may outlive borrowed value `status`
   |
note: closure is returned here
  --> light-client/src/store/memory.rs:91:9
   |
91 |         Box::new(light_blocks)
   |         ^^^^^^^^^^^^^^^^^^^^^^
help: to force the closure to take ownership of `status` (and any other referenced variables), use the `move` keyword
   |
88 |             .filter(move |(_, e)| e.status == status)
   |                     ^^^^^^^^^^^^^

```

But the `Status` type implements `Copy`, so why would I need to `move` it into the closure? Is it because the copy would only happen just before the filter closure's executed?

romac: Sounds like closures borrow the variables they close over by default, even if they are `Copy`. Moving it into the closure will just copy it into it and should just work(tm)

romac: I wonder if we should just return a `Vec` here rather than an iterator. This code is very suboptimal already (and would deserve a rewrite) so I don't think we mind the extra allocations.

thanethomson: > I wonder if we should just return a `Vec` here rather than an iterator.

That probably makes sense, given that we're cloning light blocks anyways. So should we perhaps look at changing the method signature to:

```rust
    /// Get an iterator of all light blocks with the given status.
    fn all(&self, status: Status) -> Vec<LightBlock>;
```

thanethomson: For now though, perhaps we should just leave the `#[allow(clippy::needless_collect)]`? We can fix this in a follow-up PR, as it touches various aspects of the codebase and is a breaking API change.

thanethomson: I've opened #943 to follow up on this.

PR #951, opened 2021-08-23
[WIP] Light client wasm browser support

### LINKED ISSUE #809: Network I/O support for WASM build of Light Client
Follows from #463.

In order to achieve a fully functional, WASM-compatible Light Client (and supervisor), [tendermint-rpc](https://github.com/informalsystems/tendermint-rs/tree/master/rpc) should compile to WASM and be usable from JavaScript, most likely making use of the [fetch API](https://rustwasm.github.io/docs/wasm-bindgen/examples/fetch.html).

* The `tendermint-rpc` crate must compile to WASM with network I/O support (feature-guarded).
* We must provide a working JavaScript example that:
  1. Loads a Light Client WASM binary,
  2. Instantiates a Supervisor, and
  3. Executes one of the verification algorithms against multiple Tendermint full nodes.

shravanshetty1:
closes #809 

Summary:
* Io, EvidenceReporter and ForkDetector are now async traits. Conditionally compiling this would be tedious.
* Removed Send + Sync assertion from above traits since wasm futures can never return Send+Sync since they use Rc (unless I am wrong). This means supervisor cannot be shared between threads, currently using tokio to run supervisor in the main thread as an async task.
This can probably be conditionally compiled, when not in wasm Send+Sync will be enabled.
* Disabled a few contracts to allow setting a peer as both a primary and witness, this is useful for testing.

TODO:
* Port tendermint-rpc from hyper to reqwest for wasm compatibility
* Conditionally compile as many features as possible
* Make tests use tokio to avoid panic

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ Cargo.lock
 
 # Proptest regressions dumps
 **/*.proptest-regressions
+.idea
 
 # Light Client WASM
 light-client-js/pkg/
--- a/light-client/Cargo.toml
+++ b/light-client/Cargo.toml
@@ -51,7 +51,7 @@ serde_cbor = "0.11.1"
 serde_derive = "1.0.106"
 sled = { version = "0.34.3", optional = true }
 static_assertions = "1.1.0"
-tokio = { version = "1.0", features = ["rt"], optional = true }
+tokio = { version = "1.0", features = ["rt-multi-thread"], optional = true }
 flex-error = { version = "0.4.1", default-features = false }
 async-trait = "0.1.51"
 async-recursion = "0.3.2"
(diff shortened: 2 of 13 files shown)


shravanshetty1:
@thanethomson Does this make sense? If there is a better approach please let me know so I dont want to waste time on this.

shravanshetty1:
> @shravanshetty1 thanks for this! Yes, it's something we've been thinking of.
> 
> For context, we would need to coordinate this work with the [ibc-rs](https://github.com/informalsystems/ibc-rs) team, because the relayer (Hermes) relies on the Light Client code, so I can't guarantee we'd be able to incorporate this code in the next month or two, given our current roadmap, priorities and resources. I'll chat to the team and get back to you here.
> 
> Secondly, I've modified #809 to try to break it up a little, so implementing WASM support doesn't end up being one massive PR (such PRs usually take a long time to review and are difficult to incorporate given other changes we have planned for the Light Client; see #956, for example).

I just modified the current client in a hacky way, to make it compatible with my application. I probably dont have the time to invest into making this production ready which will take alot more work.

Feel free to recommend anyone to use my fork for browser compatibility if they dont mind a hacky solution till a production ready solution is available, will be closing this PR, thanks!

PR #960, opened 2021-08-27
Refactor Light Client verification predicates interface for use from IBC

### LINKED ISSUE #956: Refactor Light Client interface to better support IBC-rs and ICS07
As discussed in today's tendermint-rs call, right now there's some ongoing work in ibc-rs (see https://github.com/informalsystems/ibc-rs/pull/1252) that needs to make use of some of the code from the `tendermint-light-client` crate, but the interfaces currently exposed don't really work well for use from ibc-rs.

For example, the [VerificationPredicates trait](https://github.com/informalsystems/tendermint-rs/blob/bbad6faf622fe3b812075969e436e5a52353be2f/light-client/src/predicates.rs#L26) makes use of `LightBlock`s as parameters, but ibc-rs doesn't make use of the same `LightBlock` constructs.

When the Light Client API's been refactored in such a way that we can use its code from https://github.com/informalsystems/ibc-rs/pull/1252

thanethomson:
Closes #956 

The main aim of this PR is to remove references to the `LightBlock` type from the verification predicates, as the `LightBlock` type isn't used in https://github.com/informalsystems/ibc-rs/pull/1252

Tomorrow I'll try working on refactoring that PR to make use of this code to see if it will work.

* [ ] Wrote tests

### DIFF
--- a/light-client/src/components/verifier.rs
+++ b/light-client/src/components/verifier.rs
@@ -2,7 +2,7 @@
 
 use crate::operations::voting_power::VotingPowerTally;
 use crate::predicates as preds;
-use crate::types::VerificationState;
+use crate::types::VerifyParams;
 use crate::{
     errors::ErrorExt,
     light_client::Options,
@@ -56,8 +56,8 @@ pub trait Verifier: Send + Sync {
     /// Perform the verification.
     fn verify(
         &self,
-        untrusted: VerificationState<'_>,
-        trusted: VerificationState<'_>,
+        untrusted: VerifyParams<'_>,
+        trusted: VerifyParams<'_>,
         options: &Options,
         now: Time,
     ) -> Verdict;
@@ -138,8 +138,8 @@ where
     ///   the trusted block.
     fn verify(
         &self,
-        untrusted: VerificationState<'_>,
-        trusted: VerificationState<'_>,
+        untrusted: VerifyParams<'_>,
+        trusted: VerifyParams<'_>,
         options: &Options,
         now: Time,
     ) -> Verdict {
--- a/light-client/src/components/verifier.rs
+++ b/light-client/src/components/verifier.rs
@@ -2,6 +2,7 @@
 
 use crate::operations::voting_power::VotingPowerTally;
 use crate::predicates as preds;
+use crate::types::LightBlockState;
 use crate::{
     errors::ErrorExt,
     light_client::Options,
@@ -9,7 +10,7 @@ use crate::{
         CommitValidator, Hasher, ProdCommitValidator, ProdHasher, ProdVotingPowerCalculator,
         VotingPowerCalculator,
     },
-    types::{LightBlock, Time},
+    types::Time,
 };
 use preds::{
     errors::{VerificationError, VerificationErrorDetail},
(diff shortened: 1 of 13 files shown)


thanethomson:
The rustdoc error seems to only pop up with Rust nightly 1.57.0, whereas it compiles just fine with 1.56.0.

It's a problem in `prost-derive`'s documentation:

```
error[E0034]: multiple applicable items in scope
Error:    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/prost-derive-0.7.0/src/lib.rs:109:14
    |
109 |             .intersperse(quote!(|));
    |              ^^^^^^^^^^^ multiple `intersperse` found
    |
    = note: candidate #1 is defined in an impl of the trait `Iterator` for the type `Map<I, F>`
    = note: candidate #2 is defined in an impl of the trait `Itertools` for the type `T`
help: disambiguate the associated function for candidate #1
    |
105 ~         let tags = Iterator::intersperse(field
106 +             .tags()
107 +             .into_iter()
108 +             .map(|tag| quote!(#tag)), {
109 +         let mut _s = $crate::__private::TokenStream::new();
110 +         $crate::quote_each_token!(_s $($tt)*);
  ...
help: disambiguate the associated function for candidate #2
    |
105 ~         let tags = Itertools::intersperse(field
106 +             .tags()
107 +             .into_iter()
108 +             .map(|tag| quote!(#tag)), {
109 +         let mut _s = $crate::__private::TokenStream::new();
110 +         $crate::quote_each_token!(_s $($tt)*);
  ...
```

light-client/src/components/verifier.rs
```diff
@@ -55,75 +56,184 @@ pub trait Verifier: Send + Sync {
...
+        ));
+
+        // Ensure the header validator hashes match the given validators
+        verdict!(self.predicates.validator_sets_match(
+            untrusted.validators,
+            untrusted.signed_header.header.validators_hash,
+            &self.hasher,
+        ));
+
+        // TODO(thane): Is this check necessary for IBC?
```
adizere, line 171: @josef-widder We may need your help here.

It seems the IBC handler (ICS 07) has no access to the set of `next_validators` for the untrusted header. For this reason, this field is an `Option` here. When the IBC handler calls into `verify` it does so with this field to `None` (cf. https://github.com/informalsystems/ibc-rs/pull/1321/files#diff-9a833365bf13b4ff1ce2cc757bf18ebabea582b2b97774040e0b7be03fcacbf4R93-R95). This check will effectively be disabled in the context of the IBC handler. This is probably a problem and we need to sort it out.

light-client/src/components/verifier.rs
```diff
@@ -55,75 +56,184 @@ pub trait Verifier: Send + Sync {
...
+            untrusted.signed_header,
+            untrusted.validators,
+            &self.voting_power_calculator,
+        ));
+
+        Verdict::Success
     }
 }
+
+/// The default production implementation of the [`PredicateVerifier`].
```
xla, line 237: This implies that there is a common understanding what a "production" environment constitutes. More qualification would aid in contextualised this.

xla (commented):
Structurally the refactoring makes sense to me, can't comment much on the internals of the verification.

light-client/src/types.rs
```diff
@@ -117,6 +139,27 @@ impl LightBlock {
     pub fn height(&self) -> Height {
         self.signed_header.header.height
     }
+
+    /// Obtain the verification parameters for the light block when using it as
+    /// trusted state.
+    pub fn as_trusted_state(&self) -> TrustedBlockState<'_> {
```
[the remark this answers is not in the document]
thanethomson: Not with the way the `TrustedBlockState` and `UntrustedBlockState` structs are defined, since they take references to (potentially large) data structures specifically to avoid having to clone those data structures.

light-client/src/types.rs
```diff
@@ -117,6 +139,27 @@ impl LightBlock {
...
+            header_time: self.signed_header.header.time,
+            height: self.signed_header.header.height,
+            next_validators: &self.next_validators,
+            next_validators_hash: self.signed_header.header.next_validators_hash,
+        }
+    }
+
+    /// Obtain the verification parameters for the light block when using it as
+    /// untrusted state.
+    pub fn as_untrusted_state(&self) -> UntrustedBlockState<'_> {
```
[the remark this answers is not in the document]
thanethomson: See above

light-client/src/types.rs
```diff
@@ -77,6 +77,28 @@ impl Status {
...
+pub struct UntrustedBlockState<'a> {
+    pub signed_header: &'a SignedHeader,
+    pub validators: &'a ValidatorSet,
+    pub next_validators: Option<&'a ValidatorSet>,
+}
+
+impl<'a> UntrustedBlockState<'a> {
+    /// Convenience method to expose the height of the associated header.
+    pub fn height(&self) -> Height {
+        self.signed_header.header.height
```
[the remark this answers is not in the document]
thanethomson: Convenience :slightly_smiling_face:

light-client/src/components/verifier.rs
```diff
@@ -55,75 +56,184 @@ pub trait Verifier: Send + Sync {
...
-/// - A commit validator
-/// - A header hasher
-///
-/// For regular use, one can construct a standard implementation with `ProdVerifier::default()`.
-pub struct ProdVerifier {
-    predicates: Box<dyn VerificationPredicates>,
-    voting_power_calculator: Box<dyn VotingPowerCalculator>,
-    commit_validator: Box<dyn CommitValidator>,
-    hasher: Box<dyn Hasher>,
+macro_rules! verdict {
```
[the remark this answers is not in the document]
thanethomson: The `PredicateVerifier::verify` method returns a `Verdict`, not a `Result`, so it wouldn't be compatible with those macros.

light-client/src/components/verifier.rs
thanethomson: Yes this is something we've been talking/thinking about for a year now :smile: We're definitely open to suggestions on the naming front.

The original purpose, afaik, was to disambiguate mock versions from non-mock versions of the types, but I think @romac can elaborate on this better than I can?

light-client/src/components/verifier.rs
adizere: FYI this concern need not block the PR. (Josef will only return from holidays after Sept 27). Maybe @ancazamfir could help us sort out the problem.

xla (approved):
LGTM

🛀 😷 🌽 💐

light-client/src/components/verifier.rs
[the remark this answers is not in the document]
romac: Note that we'll eventually be able to just use `?` once https://doc.rust-lang.org/stable/std/ops/trait.Try.html stabilizes.

light-client/src/components/verifier.rs
romac: Yeah I am not super happy with the current terminology (that I came up with). I guess we need to differentiate between:
a) the trait (previously `Predicates`, now `Verifier`)
b) the "production" implementation (previously `ProdPredicates`, now `ProdVerifier`)
c) the potential mock implementations of the trait (no use for that yet)

I agree that the `Prod` prefix is subpar, maybe `Default` or `Standard` would be a better fit?

light-client/src/types.rs
```diff
@@ -77,6 +77,28 @@ impl Status {
     }
 }
 
+/// Trusted block parameters needed for light client verification.
+pub struct TrustedBlockState<'a> {
```
romac, line 81: 😍

romac (approved):
Amazing job, @thanethomson! Very much an improvement over the previous API 💯

PR #965, opened 2021-09-02
Refactor: makes the light-client crate async

### LINKED ISSUE #953: `async`-ify the light client
As per a recent discussion on the weekly ibc-rs call, the ibc-rs team potentially wants to move towards `async`-ifying the I/O components of the project. As such it would make sense to `async`-ify the `tendermint-light-client`.

This would also help towards #809, since right now the Light Client exposes a non-`async` API (which, under the hood, wraps an `async` interface). Usage at a higher level complicates the architecture, since it'll most likely be used from an `async` outer wrapper. The chain of use would then look like `async -> sync -> async`, which doesn't make any sense.

When the `tendermint-light-client` crate exposes an `async` API for all methods requiring I/O instead of its current non-`async` one.

celsobonutti:
Closes #953 

The main goal of this PR is to make the `tendermint-light-client` crate asynchronous. 

Summary: 
* Io, EvidenceReporter and ForkDetector are now async_trait;
* Tests updated accordingly

- [x] Referenced an issue explaining the need for the change
- [x] Updated all relevant documentation in docs (I think only the types were changed, couldn't find any relevant docs)
- [x] Updated all code comments where relevant (same as above)
- [x] Wrote tests (I've updated the existing tests to use the async API)
- [x] Updated CHANGELOG.md


celsobonutti:
Hey @thanethomson, do you have any idea why fmt and docs are breaking? `kvstore-integration` test also seem to be passing, but failing due to an error in to cargo

thanethomson:
> Hey @thanethomson, do you have any idea why fmt and docs are breaking? `kvstore-integration` test also seem to be passing, but failing due to an error in to cargo

Seems like it's just this file: https://github.com/informalsystems/tendermint-rs/pull/965/checks?check_run_id=3538829694#step:4:37

celsobonutti:
Thanks a lot! 😊

thanethomson:
Just to keep you posted about this @celsobonutti, we're only going to be able to finalize and merge this once ibc-rs starts migrating to `async`, which might be some time from now.

celsobonutti:
Hey @thanethomson, thanks for the info! I've also been unable to work on this lately due to work, but I hope I'll have time to do fix the remaining test soon 😊

xla (changes requested):
Dropping the implementation of the `block_on` in this crate and just replacing it with `await` calls introduces a breaking change as before there was a timeout on the io operations.

xla:
@celsobonutti Thanks for the work so far, this PR is now superseded by #995. Would be great to get your input over there.

celsobonutti:
hey @xla, thanks for taking over with this! Unfortunately, I won't have much time to pay attention to this issue currently.

PR #973, opened 2021-09-15
Disable Light Client model-based tests by default

### LINKED ISSUE #968: Model-based tests take a long time
**Steps to reproduce**

Running `cargo test`, I get
```
Running tests/model_based.rs (target/debug/deps/model_based-0199679b3102f8af)

running 1 test
test run_model_based_single_step_tests has been running for over 60 seconds
test run_model_based_single_step_tests ... ok
```

after several minutes.  This increases the cycle time of working on the codebase.  It would be better if the model-based tests were off by default, and ran in CI, to keep the test workflow quick-by-default.

Default testing workflow is faster.

thanethomson:
Closes #968 

* [ ] Wrote tests

### DIFF
--- a/.changelog/config.toml
+++ b/.changelog/config.toml
@@ -1,0 +1,1 @@
+project_url = 'https://github.com/informalsystems/tendermint-rs'
--- a/.changelog/unreleased/improvements/968-mbt.md
+++ b/.changelog/unreleased/improvements/968-mbt.md
@@ -1,0 +1,3 @@
+- `[tendermint-light-client]` Model-based tests are now disabled
+  by default and can be enabled through the `mbt` feature
+  ([#968](https://github.com/informalsystems/tendermint-rs/issues/968))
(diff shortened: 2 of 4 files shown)


.changelog/config.toml
```diff
@@ -0,0 +1 @@
+project_url = 'https://github.com/informalsystems/tendermint-rs'
```
xla, line 1: This change seems unrelated to the premise of this PR, could you contextualise what constitutes this change?

xla (approved):
Besides the one comment, this is looking good.

🏢 ☺️ 🈴 🐝

.changelog/config.toml
thanethomson: It's a prerequisite for being able to automatically generate changelog entries from the CLI using unclog, as per https://github.com/informalsystems/unclog#adding-a-new-unreleased-entry

I could've probably made the change in a separate PR, but it's so small, and I had to do it anyways, that I thought I'd just add it in here.

Issue #974, opened 2021-09-16
rpc: Allow configuring the underlying `tungstenite` WebSocket connection


romac:
Osmos nodes regularly [send messages over WebSocket which are 250MB in size](https://github.com/informalsystems/ibc-rs/issues/1299), which is more than the default [`max_message_size` setting ](https://docs.rs/tungstenite/0.12.0/tungstenite/protocol/struct.WebSocketConfig.html#structfield.max_message_size) of the underlying WebSocket connection. Because of this, such messages are currently rejected by the underlying library and trigger an error within Hermes, which then also completely misses the message.

While we will eventually move away from the current event-based model in Hermes, and will therefore not need to use the WebSocket client anymore, we are not there yet and would like to see if we can figure out a way to receive such messages (what this will do to the performance of the relayer remains to be seen).

The easiest way to experiment with this would be to expose the underlying `tungstenite::protocol::WebSocketConfig` in `tendermint_rpc`, either directly or by using a facade that can convert into that.

**What's the definition of "done" for this issue?**

The `WebSocketClient` takes an optional `tungstenite::protocol::WebSocketConfig` parameter which will be use to configure the underlying WebSocket connection.

PR #975, opened 2021-09-16
Add `WebSocketClient::new_with_config` to specify the WebSocket connection settings

romac:
Closes: #974

This PR directly re-exports the `tungstenite::protocol::WebSocketConfig`. We could also define our own config type and convert it into the former if that's deemed preferable.

* [ ] Wrote tests


thanethomson (commented):
Looks great! Just a changelog entry and these minor tweaks and I think we're good to go here.

thanethomson (approved):

PR #978, opened 2021-09-16
Prepare for Prost v0.8+ upgrade

thanethomson:
Closes #925 

Right now I'm depending on the `prost*` crates at a specific commit after the attributes issue has been merged. This PR will be ready for review once Prost cuts a new release.

This PR helps to ensure compatibility before we upgrade.

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Added entry in `.changelog/`

PR #979, opened 2021-09-16
p2p: replace `prost-amino` with `prost`

tony-iqlusion:
The only message that `prost-amino` was still used for was `AuthSigMessage`, and the only way an amino-encoded version of that message differs from a proto is the amino type prefix on the `pub_key` field, i.e. `tendermint/PubKeyEd25519`.

This commit adds a test vector for a serialized amino message, and then changes `AuthSigMessage` to use `prost-derive`:

- An `AuthSigMessage::new` function takes care of adding the Amino prefix to the public key
- A `TryFrom` impl for `proto::p2p::AuthSigMessage` takes care of verifying and removing it

With that, `prost-amino` is no longer necessary and this commit therefore removes the last dependency on it.


p2p/Cargo.toml
```diff
@@ -19,13 +19,11 @@ description = """
     The Tendermint P2P stack in Rust.
     """
 
-[lib]
-test = false
```
tony-iqlusion, line 23: Not sure why this was here, but I added a test so I had to remove this to make the test work

xla: The reason it is declared is that all tests are in a separate crate over at https://github.com/informalsystems/tendermint-rs/tree/master/test

xla (commented):
I for one welcome the abolishment of amino! Clippy seems to complain in a couple of places and see my comment about tests, other than that this change looks good to me.

xla:
Could you also add a changelog entry.

p2p/Cargo.toml
tony-iqlusion: Aah, ok.

My test was mostly useful for ensuring message equivalence during development. Since I'm fairly confident of that at this point (and hopefully all of this can be removed soon) I can remove it and restore these lines.

xla: Sounds good!

thanethomson: Or keep the test and add it to the `test` crate?

tony-iqlusion: @thanethomson the API the test is written against isn't public.

I went ahead and removed it for now.

tony-iqlusion:
Changelog entry added, and looks like all of the tests are passing now except a docs failure happening internally within `prost-derive` due to what appears to be overlapping impls with `std`.

Not sure what to do about that but I would hope #978 can now address it.

p2p/Cargo.toml
```diff
@@ -53,5 +53,4 @@ tendermint-proto = { path = "../proto", version = "0.21.0" }
 tendermint-std-ext = { path = "../std-ext", version = "0.21.0" }
 
 # optional dependencies
-prost-amino = { version = "0.6", optional = true }
-prost-amino-derive = { version = "0.6", optional = true }
```
thanethomson, line 57: Amazing :grin:

thanethomson (approved):

thanethomson:
Awesome thanks @tony-iqlusion! Yes, #978 should address that. Once I've merged this I'll pull these changes into #978 to make 100% sure.

PR #980, opened 2021-09-17
Use `core` and `alloc` crates for `no_std` compatibility

soareschen:
Part of informalsystems/ibc-rs#1158

This follows similar work in informalsystems/ibc-rs#1156 to move closer to supporting no_std in tendermint-rs.

* [ ] Wrote tests


proto/src/lib.rs
```diff
@@ -107,9 +117,15 @@ pub mod serializers;
 /// // We expect a validation error here
 /// assert!(MyDomainType::decode(invalid_raw_bytes.as_ref()).is_err());
 /// ```
-pub trait Protobuf<T: Message + From<Self> + Default>
+pub trait Protobuf<T>
```
soareschen, line 120: I have modified the `Protobuf` trait such that it allows the message type to implement `TryFrom` instead of `From`. This allows safe parsing of the `TimeStamp` type, to prevent undefined behavior caused by integer overflow in DOS attack.

tendermint/src/time.rs
```diff
@@ -16,33 +18,35 @@ use crate::error::Error;
...
-        // tendermint_proto::Timestamp can be JSON-encoded
-        let prost_value = prost_types::Timestamp {
-            seconds: value.seconds,
-            nanos: value.nanos,
-        };
+impl TryFrom<Timestamp> for Time {
+    type Error = Error;
 
-        SystemTime::from(prost_value).into()
+    fn try_from(value: Timestamp) -> Result<Self, Error> {
```
soareschen, line 29: I implement `TryFrom` instead of `From` to allow safe conversion of the raw protobuf `Timestamp` to the `Time` type. This is necessary as one is using `i32` while the other use `u32`. Although the overflow is unlikely to occur, for security purpose it is still possible that an attacker may try to provide overflow time values to cause undefined behavior on the relayer or the chain. So it is better to return an error in case that happens.

As a result of this change, a number of `From` instances for various data structures that use `Time` have to be updated to implement `TryFrom` instead.

tendermint/src/time.rs
```diff
@@ -16,33 +18,35 @@ use crate::error::Error;
...
+
+        Ok(Time(Utc.timestamp(value.seconds, nanos)))
     }
 }
 
-impl From<Time> for Timestamp {
-    fn from(value: Time) -> Self {
-        // prost_types::Timestamp has a SystemTime converter but
-        // tendermint_proto::Timestamp can be JSON-encoded
-        let prost_value = prost_types::Timestamp::from(SystemTime::from(value));
```
soareschen, line 41: I remove the use of `SystemTime` here, as this is the only place where the `tendermint` crate is still using `std`.

tendermint/src/block.rs
```diff
@@ -84,14 +85,27 @@ impl TryFrom<RawBlock> for Block {
...
-            evidence: Some(value.evidence.into()),
-            last_commit: value.last_commit.map(Into::into),
-        }
+            evidence: Some(value.evidence.try_into()?),
+            last_commit: value.last_commit.map(|v| v.try_into()).transpose()?,
+        })
+    }
+}
+
+impl Serialize for Block {
```
soareschen, line 101: By changing from `TryFrom` to `Try`, it is not possible to auto derive the `Serialize` trait through type attributes. (https://github.com/serde-rs/serde/issues/1524) So I just implement the trait manually.

I'm not sure why there is a need to auto derive the `Serialize` trait other than for use in testing. In any case, I see that there are many custom `impl` blocks for `Serialize` in `tendermint`, so it should be fine to add a few more.

light-client/tests/model_based.rs
```diff
@@ -125,8 +125,8 @@ mod mbt {
     impl SingleStepTestFuzzer for HeaderVersionFuzzer {
         // TODO: rehash the header and re-compute commit with it
         // TODO: Unlike in tendermint-go, we don't assert for a particular version in rust
-        // TODO: Either add this check in verification or remove this test because otherwise there's no
-        // point of it
+        // TODO: Either add this check in verification or remove this test because otherwise there's
+        // no point of it
```
soareschen, line 129: It seems like there are some changes the Rust nightly that cause comments to be formatted slightly differently by `cargo fmt`.

tendermint/src/time.rs
```diff
@@ -106,33 +114,35 @@ impl From<Time> for DateTime<Utc> {
...
-        t.0.into()
-    }
-}
-
 impl Add<Duration> for Time {
-    type Output = Self;
+    type Output = Result<Self, Error>;
 
     fn add(self, rhs: Duration) -> Self::Output {
-        let st: SystemTime = self.into();
```
soareschen, line 125: I remove the use of `SystemTime` for addition/substration here to support no_std. To perform safe addition in `chrono`, we have to take care of integer overflows by returning error. This is important as the overflow result may cause computation such as trusted period to return incorrect result.

proto/src/lib.rs
thanethomson: Originally the idea here was that we wanted conversions from `SerializationType -> DomainType` to be fallible to allow for validation, and then from `DomainType -> SerializationType` to be infallible because `DomainType` should be a subset of `SerializationType`.

Changing this would imply that we could possibly have `DomainType`s that possibly do not have valid serialized representations, which implies to me that those `DomainType`s are incorrectly modeled, does it not?

So then the practical consequences of this architectural change would be:
1. All `From<DomainType> for SerializationType` impls now need to become `TryFrom<DomainType> for SerializationType`, touching all domain types
2. We need custom serde `Serialize` trait impls, as per https://github.com/informalsystems/tendermint-rs/pull/980/files#r717530415

This seems like a lot of work/code just to be able to cater for the `DomainTimestamp -> SerializedTimestamp` conversion, does it not? Is there no way to just implement a better `DomainTimestamp` whose conversion to `SerializedTimestamp` is infallible (as it should actually be)?

soareschen:
> Originally the idea here was that we wanted conversions from `SerializationType -> DomainType` to be fallible to allow for validation, and then from `DomainType -> SerializationType` to be infallible because `DomainType` should be a subset of `SerializationType`.
> Changing this would imply that we could possibly have `DomainType`s that possibly do not have valid serialized representations, which implies to me that those `DomainType`s are incorrectly modeled, does it not?

Just because the type signature of a function does not return an error, does not mean that there isn't actually an error. The purpose of defining type signatures is to reflect the actual behavior of the implementation as accurate as possible. In the case of the timestamp conversion, defining a `From` instance for it is simply ignoring the fact that there is a chance for that to fail. This kind of error-ignorance is especially common in languages like Go, where even when the code fails, a "valid" value is still always returned and the caller can pretend that the error never happens.

If we really want to ignore the failure and define a `From` instance for `Timestamp` using the underlying `chrono::DateTime`, there are a few choices:

1. Raise panic when conversion between unsigned and signed integers fail, allowing the relayer to potentially be DOS'ed with attacker sending an overflow timestamp.
2. Allow silent integer overflow using `as` coercion, allowing an attacker to send an overly large timestamp to go back in time and cause potential security vulnerabilities.
3. Default to 0 or max seconds/nanoseconds value when conversion fails.
4. Use an alternative library other than `chrono` so that it can process the seconds and nanoseconds in the same precision as the protobuf fields. (I am actually tempted on this, but couldn't find better alternatives)

I originally chose option 3 as the solution. But given that there aren't too many code changes, I do think that it is worth handling the error properly. I also had similar issue in `ibc-rs`, where I couldn't use the `Protobuf` trait because of the lack of `From` trait, and had to copy-paste the `decode` function with minor modification to workaround that.

Btw note that all types implementing `From` also automatically implement `TryFrom`, with the error type being `Infallible`.

thanethomson:
> Raise panic when conversion between unsigned and signed integers fail, allowing the relayer to potentially be DOS'ed with attacker sending an overflow timestamp.

Is this a real possibility? How would an attacker trigger an incorrect conversion from `Time -> Timestamp` (which is really a conversion from `chrono::DateTime<Utc> -> Timestamp`) from the outside?

The only way I could see that being a possibility is if we somehow constructed an incorrect `chrono::DateTime<Utc>` in the first place, right?

Here's my reasoning:
- [`chrono::DateTime::timestamp_subsec_nanos()`](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#method.timestamp_subsec_nanos) returns a `u32` (max value 2^32 > 4x10^9)
- The Protobuf `Timestamp::nanos` field is `i32` (max value 2^31 - 1, which is > 2x10^9)
- There are only 1x10^9 nanoseconds in a second, so any nanosecond value > 1x10^9 should be invalid

Hence if we want to prevent invalid nanosecond values, we should check that the `timestamp_subsec_nanos()` value of the constructed `chrono::DateTime<Utc>` is <= 10^9 when we're deserializing it, correct? Then there's never any chance of overflow and a `panic` would be justified if we did encounter a nanosecond overflow when we try to convert from `chrono::DateTime<Utc>` to Protobuf `Timestamp`.

xla (changes requested):
There is a bit of a pattern here where change-sets are proposed which often sneak in changes not directly pertaining to the set goal of the change-set. If we want to change the semantics of the Prototype trait this should be reasoned about and proposed in isolation instead of in the context of switching to `core` and `alloc`. It would be warranted to preface such a change with a small ADR so the design and rational can be discussed instead of it being intermingled with the review of the implementation.

soareschen:
> If we do somehow end up with a `chrono::DateTime<Utc>::timestamp_subsec_nanos()` value > `999_999_999` (as per the docs in the case of a [leap second](https://en.wikipedia.org/wiki/Leap_second)), should we not rather increment the seconds and use the nanoseconds modulo `1000_000_000`? How are leap seconds usually handled?

Yeah, in that particular case it shouldn't happen unless there is a bug in `chrono` itself. I suppose in case there is a conversion error, we could use `0` as the default subsecond nanos.

For the case of `TryFrom<Timestamp>`, we would still have to do safe conversion, since the protobuf can be supplied by an attacker. Btw I also just notice that we are supposed to use `Utc.timestamp_opt` instead of `Utc.timestamp` so that `chrono` does not panic in the case of an overflow input.

I'll check whether it is fine to have `TryFrom<Timestamp>` but `From<Time>`. The two asymmetric requirements still always confuse me. 

> There is a bit of a pattern here where change-sets are proposed which often sneak in changes not directly pertaining to the set goal of the change-set. If we want to change the semantics of the Prototype trait this should be reasoned about and proposed in isolation instead of in the context of switching to `core` and `alloc`. It would be warranted to preface such a change with a small ADR so the design and rational can be discussed instead of it being intermingled with the review of the implementation.

Yeah the changes get a bit large as I tried to get rid of the use of `std::time::SystemTime` ASAP. I will break this into two separate PRs instead.

soareschen:
Closing this PR in favor of #993 and #994.

It seems like we do able to avoid modifying the `Protobuf` trait if only `TryFrom<Timestamp>` is implemented. Thanks @thanethomson for pointing that out!

PR #982, opened 2021-09-17
Direct conversion between tendermint::Time and protobuf TimeStamp

soareschen:
This PR enables direct conversion between `tendermint::Time` and the protobuf `TimeStamp` type without having to go through `std::time::SystemTime`. This is needed so that we can remove the dependency to `std` for informalsystems/ibc-rs#1158.

I also make public the underlying `chrono::DateTime` field, as there is no reason to encapsulate `tendermint::Time` being a newtype wrapper to `chrono::DateTime`.

* [ ] Wrote tests

### DIFF
--- a/tendermint/src/time.rs
+++ b/tendermint/src/time.rs
@@ -1,6 +1,6 @@
 //! Timestamps used by Tendermint blockchains
 
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, TimeZone, Utc};
 use serde::{Deserialize, Serialize};
 
 use std::fmt;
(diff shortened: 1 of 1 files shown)


thanethomson:
There seem to be a bunch of [test failures](https://github.com/informalsystems/tendermint-rs/pull/982/checks?check_run_id=3631423388) here at the moment that seem directly related to the new time code.

There are also now errors in the [light client integration test](https://github.com/informalsystems/tendermint-rs/pull/982/checks?check_run_id=3631423577#step:5:374).

PR #986, opened 2021-09-20
Move out `tendermint::config` to `tendermint-config` crate

### LINKED ISSUE #983: Move `config` module out from `tendermint` into its own crate
Part of informalsystems/ibc-rs#1158.

With #980 there are still a few remaining use of `std` in the `tendermint` crate, which is mostly consist of filesystem and network operations to retrieve the config files. By moving `tendermint::config`, `tendermint::net` and their submodules to a separate crate, we are one step closer to no_std support for the `tendermint` crate.

soareschen:
Fixes #983.

* [ ] Wrote tests


xla (commented):
The metadata in the new cargo toml should be catered to the actual crate instead of a verbatim copy of the tendermint cargo manifest.

soareschen:
> The metadata in the new cargo toml should be catered to the actual crate instead of a verbatim copy of the tendermint cargo manifest.

Yeah I missed updating that after copying the boilerplate. Fixed it now.

soareschen:
Rehashing my note in #983:

I have taken a look at the use of `net::Address`, and it does not seem like the actual details are used anywhere. In #986 I moved out `tendermint::net` to `tendermint_config::net`, and changed the field in `tendermint::node::OtherInfo::rpc_address` to just `String`. If anyone needs to do the actual URL parsing, they can do it at the call site instead.

The main difference I see of using `String` instead of `Address` for `rpc_address` is the way it is deserialized. With the `Deserialize` implementation provided by `Address`, an invalid URL would result in an error. However I also do not see how this can be solved using feature flag. In the absense of the `url` crate, should the deserialization just succeed? That wouldn't be good since the crate would effectively be behaving differently with no invariant in place. Otherwise if it is possible to implement the parsing without using `url`, then it is better off to just not `url` in both cases in the first place.

Note that the other alternative is to just wait for servo/rust-url#717 to get merged and released.

light-client/src/components/verifier.rs
```diff
@@ -128,14 +128,12 @@ where
...
-    ///   validators
+    /// - Ensure the header next validator hashes match the given next validators
     /// - Additional implementation specific validation via `commit_validator`
     /// - Check that the untrusted block is more recent than the trusted state
-    /// - If the untrusted block is the very next block after the trusted block,
-    ///   check that their (next) validator sets hashes match.
-    /// - Otherwise, ensure that the untrusted block has a greater height than
-    ///   the trusted block.
+    /// - If the untrusted block is the very next block after the trusted block, check that their
+    ///   (next) validator sets hashes match.
```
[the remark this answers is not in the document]
soareschen: Not sure about that. This is somehow modified by `cargo fmt` even though no code changes are made.

light-client/src/components/verifier.rs
```diff
@@ -128,14 +128,12 @@ where
     ///
     /// - Ensure the latest trusted header hasn't expired
     /// - Ensure the header validator hashes match the given validators
-    /// - Ensure the header next validator hashes match the given next
-    ///   validators
+    /// - Ensure the header next validator hashes match the given next validators
```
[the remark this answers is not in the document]
soareschen: Not sure about that. This is somehow modified by `cargo fmt` even though no code changes are made.

config/src/net.rs
```diff
@@ -83,7 +96,8 @@ impl FromStr for Address {
         match url.scheme() {
             "tcp" => Ok(Self::Tcp {
                 peer_id: if !url.username().is_empty() {
-                    Some(url.username().parse()?)
+                    let username = url.username().parse().map_err(Error::tendermint)?;
```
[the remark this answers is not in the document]
soareschen: A new `Error` type is defined for `tendermint-config`. The error constructor `Error::tendermint` wraps a `tendermint::Error` into a `tendermint_config::Error`.

light-client/src/components/verifier.rs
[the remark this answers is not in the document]
seanchen1991: Oh, I meant the content of the actual comment, not the formatting that got changed (though to be fair, I probably wouldn't have noticed this comment if it hadn't gotten re-formatted) 🙂

tendermint/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 //! and remote procedure calls (JSON-RPC).
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
soareschen: Stale changes. Removed.

light-client/src/components/verifier.rs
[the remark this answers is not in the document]
soareschen: Yeah, I think you can ask in #tendermint on Slack to clarify.

config/Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[package]
+name       = "tendermint-config"
+version    = "0.21.0" # Also update `html_root_url` in lib.rs and
```
thanethomson, line 3:
```suggestion
version    = "0.22.0" # Also update `html_root_url` in lib.rs and
```

config/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+    tendermint-config provides functions for loading and validating Tendermint
+    configuration files. It is moved out of the tendermint crate for no_std support
+    in the tendermint crate.
+    """
+
+authors = [
+    "Informal Systems <hello@informal.systems>",
+    "Ismail Khoffi <Ismail.Khoffi@gmail.com>",
+    "ValarDragon <dojha12@gmail.com>",
+    "Tony Arcieri <tony@iqlusion.io>",
```
thanethomson, line 23:
```suggestion
```

rpc/Cargo.toml
```diff
@@ -74,6 +74,7 @@ serde = { version = "1", features = [ "derive" ] }
 serde_bytes = "0.11"
 serde_json = "1"
 tendermint = { version = "0.21.0", path = "../tendermint" }
+tendermint-config = { version = "0.21.0", path = "../config" }
```
thanethomson, line 77:
```suggestion
tendermint-config = { version = "0.22.0", path = "../config" }
```

config/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
+//! Tests for parsing configuration files.
+//!
+//! Test config files are located in the `tests/support/config` subdirectory.
+
+mod files {
```
thanethomson, line 5: Is this file not the same as `config/tests/config.rs`?

thanethomson: Also, what's the benefit of wrapping these tests in the `files` mod? Could we not just do away with the outer `mod files` wrapper?

soareschen: > Is this file not the same as `config/tests/config.rs`?

Oh yeah, that was probably a mistake during the file moving. I have removed the duplicate.


> Also, what's the benefit of wrapping these tests in the `files` mod? Could we not just do away with the outer `mod files` wrapper?

Yes we can. The `mod files` was there in the original code. I have moved them out now. One disadvantage is that git now recognize the changes as a new file rather than a moved file.

thanethomson: > Yes we can. The mod files was there in the original code. I have moved them out now. One disadvantage is that git now recognize the changes as a new file rather than a moved file.

That's fine, thanks!

abci/src/lib.rs
```diff
@@ -9,7 +9,7 @@
 //! have them interact. In practice, the client interaction will be performed
 //! by a full Tendermint node.
 //!
-//! ```rust
+//! ```ignore
```
thanethomson, line 12: Was this doctest failing?

soareschen: Yup. `cargo test` on the project root fails because of this and a few other doctests.

thanethomson: I've submitted #990 to fix these problems.

xla: It would be preferrable to not just ignore parts that break. The least would have been a tracking issue, a fix like @thanethomson provided is obviously preferrable.

soareschen: Nice! I will revert this change then.

tendermint/src/node/info.rs
```diff
@@ -82,7 +76,7 @@ pub struct OtherInfo {
     pub tx_index: TxIndexStatus,
 
     /// RPC address
-    pub rpc_address: net::Address,
+    pub rpc_address: String,
```
soareschen, line 79: One unresolved issue is whether using `String` instead of `net::Address` for `rpc_address` will break any functionality. The main difference is in the implementations of `Deserialize` and `Display`. Any idea of how this field is used in practice?

xla: @thanethomson Might have some input here.

thanethomson: No idea here actually.

PR #990, opened 2021-09-27
Fix `cargo test` and add check to CI

### LINKED ISSUE #989: `cargo test` fails with default features
Follows from https://github.com/informalsystems/tendermint-rs/pull/986/files#r716715195

There are a few misconfigured features when running without `--all-features`.

**Steps to reproduce**

Run `cargo test` on `master`.

1. `cargo test` shouldn't fail.
2. We should run `cargo test` in CI to test the default feature configuration.

thanethomson:
Closes #989 

* [ ] Wrote tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,18 @@ jobs:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
     if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
 
+  default-features:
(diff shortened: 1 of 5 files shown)


thanethomson:
> This still seems to fail on the docs in CI, looks like an import issue.

It's still the same old Prost v0.7.0 issue that's been plaguing us for some time now. See https://github.com/informalsystems/tendermint-rs/pull/990/checks?check_run_id=3722709931#step:4:598

Apparently they might be cutting a new Prost release this week sometime :crossed_fingers:

soareschen (approved):

PR #993, opened 2021-09-29
Use `core` and `alloc` crates for `no_std` compatibility (Take 2)

soareschen:
Part of informalsystems/ibc-rs#1158

This follows similar work in informalsystems/ibc-rs#1156 to move closer to supporting no_std in tendermint-rs.

This is a succession of #980 with the changes to replace the use of `std::time::SystemTime` being moved to another PR.

* [ ] Wrote tests


proto/src/serializers.rs
```diff
@@ -32,7 +32,7 @@
 //! Available serializers:
 //! i64                  <-> string:               #[serde(with="serializers::from_str")]
 //! u64                  <-> string:               #[serde(with="serializers::from_str")]
-//! std::time::Duration  <-> nanoseconds as string #[serde(with="serializers::time_duration")]
+//! core::time::Duration  <-> nanoseconds as string #[serde(with="serializers::time_duration")]
```
xla, line 35: Looks like this shifted the formatting and the entire second column should get an extra space.

xla (commented):
Quite some work, impressive. Would it be possible to consolidate the multiple preludes into a shared crate to ease with maintenance down the line?

soareschen:
> Would it be possible to consolidate the multiple preludes into a shared crate to ease with maintenance down the line?

I am a bit reluctant to expose the internal prelude as public interface. It will make it more difficult for us to remove it in the future. 

It is also not clear that if we do want to expose it, which crate should it be in. The lowest common denominator of all tendermint-rs and ibc-rs crates is probably `tendermint-proto`. But it also doesn't feel right to put it there. I also don't think it is worth a hassle to publish it as a separate crate.

In any case, I think this is just a temporary measure with minimal content that we can manage with. Once the `alloc::prelude` crate is graduated from nightly at https://github.com/rust-lang/rust/issues/58935, or in case the Rust team decide to go with different approaches, it is much easier for us to remove/alter the `prelude.rs` in the individual crates than having to do a major version release with deprecation and other hassles.

PR #994, opened 2021-09-29
Use chrono::DateTime instead of std::time::SystemTime

soareschen:
Part of informalsystems/ibc-rs#1158. This is a succession of #980

This PR removes the use of `std::time::SystemTime` and replace them with `chrono::DateTime` so that the `tendermint` crate can be no_std compliant.

* [ ] Wrote tests


xla (approved):
🕕 🌁 🏍 🚍

PR #997, opened 2021-09-29
ci: Build and check tools

### LINKED ISSUE #996: ci: tools workspace is not validated
Except for one set of tests the tools workspace is not considered during ci.

When changes which break any crate in the tools workspace are not accepted do to failing checks in the github workflows.

xla:
So far only the kvstore tests ran as part of the Github workfows. This
would leave opportunity for changes to introduce breakage to the builds
of the tools. In this change the same build and clippy stages are
introduced for the tools workspace that currently run for the top-level
one.

Closes #996 

* [ ] ~~Wrote tests~~
* [ ] ~~Added entry in `.changelog/`~~


.github/workflows/rust.yml
```diff
@@ -46,6 +46,21 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: clippy
+          override: true
+      - uses: actions-rs/clippy-check@v1
+        with:
+          name: clippy-tools-results
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --manifest-path tools/kvstore-test/Cargo.toml --all-features --all-targets -- -Dwarnings -Drust-2018-idioms
```
greg-szabo, line 62: I wonder if this should also be made into a Cargo alias so it's easy to run locally too.

greg-szabo (approved):

.github/workflows/rust.yml
xla: Yeah had the same intuition. There is a follow-up PR with a lot of improvements to the consistency of those commands.

thanethomson (approved):
Looks good! I assume resolving #1000 will fix the failing tools test.

PR #1005, opened 2021-10-06
meta: Adopt forked prost crates

### LINKED ISSUE #966: Current nightly (2021-09-07) breaks build
**Steps to reproduce**

xla:
Addresses the lack of a release of the upstream crates for a multitude
of reasons by adopting forked crates on crates.io based off of latest
master. Ideally this is a stop-gap solution until prost itself is
updated properly.

Fixes #966

* [ ] ~~Wrote tests~~


greg-szabo (approved):
I wonder if releasing a public crate is overkill and we could've maybe gotten away with copying the code into this repo (or maybe as a sub-module). But I'm not the one to argue, let's make this happen.

thanethomson:
> I wonder if releasing a public crate is overkill and we could've maybe gotten away with copying the code into this repo (or maybe as a sub-module). But I'm not the one to argue, let's make this happen.

Adding it as a submodule to tendermint-rs would've probably been far more work than just forking the existing project and republishing the crates under different names. It would've been far easier if we could point our crates to the official Prost repository at a specific commit, but then we wouldn't be able to publish our own crates.

PR #1017, opened 2021-11-10
ci: Update coverage check

thanethomson:
Follows from https://github.com/informalsystems/tendermint-rs/pull/862#discussion_r739678015

We could just go back to using `nightly`, but then we open ourselves up to arbitrary CI breakage.

* [ ] Wrote tests


greg-szabo (approved):
I agree. It's not optimal to need to update the nightly version regularly, but it gives the necessary stability for tests.

Issue #1023, opened 2021-11-21
Deprecation warnings with ed25519 1.3.0


mzabaluev:
tendermint uses API from `ed25519` that has become deprecated as of version 1.3.0.

**Steps to reproduce**

1. Change the directory to a fresh working copy of tendermint-rs.
2. Run `cargo update`.
3. Try to build.

**What's the definition of "done" for this issue?**

No deprecation warnings are observed.
The `tendermint` API should probably deprecate its standalone `ED25519_SIGNATURE_LENGTH` const, as the associated const on the signature struct type is now the bees' knees.

PR #1024, opened 2021-11-22
Update to deprecations in ed25519 1.3

mzabaluev:
Fixes: #1023

As of release 1.3.0, `ed25519` deprecates the standalone `SIGNATURE_LENGTH` constant in favor of the `Signature::BYTE_SIZE` associated item. Also, the `Signature::new` constructor is deprecated in favor of `Signature::from_bytes`.
Update the code to not trigger the deprecation lints.

In turn, deprecate what was formerly the reexport of that constant in the tendermint API.

* [ ] ~Updated all relevant documentation in docs~
* [x] ~Wrote tests~ - pure refactoring, exercised by existing tests.

### DIFF
--- a/tendermint/src/vote.rs
+++ b/tendermint/src/vote.rs
@@ -15,8 +15,6 @@ use core::fmt;
 use core::str::FromStr;
 
 use bytes::BufMut;
-use ed25519::Signature as Ed25519Signature;
-use ed25519::SIGNATURE_LENGTH as ED25519_SIGNATURE_LENGTH;
 use serde::{Deserialize, Serialize};
 
 use tendermint_proto::types::Vote as RawVote;
@@ -27,6 +25,7 @@ use crate::consensus::State;
 use crate::error::Error;
 use crate::hash;
 use crate::prelude::*;
+use crate::signature::Ed25519Signature;
 use crate::{account, block, Signature, Time};
 
 /// Votes are signed messages from validators for a particular block which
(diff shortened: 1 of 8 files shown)


romac (approved):
Sweet, thanks @mzabaluev! Will leave it to @thanethomson to take another look and do the merge.

tendermint/src/vote.rs
```diff
@@ -159,6 +158,7 @@ impl Vote {
 }
 
 /// Default trait. Used in tests.
+// FIXME: Does it need to be in public crate API? If not, replace with a helper fn in crate::test?
```
mzabaluev, line 161: To expand on this: it's probably not useful for non-test API users to have a `Default` impl for `Vote`, and there is potential for inadvertent misuse that is hard to audit for.

For test users, a dedicated helper fn would serve just as well, unless there are some commonly used cases that are more ergonomic with a `Default` impl. The dedicated function will be easy to find in code.

romac: Agreed, a dedicated helper fn sounds like the way to go. As far as I can tell this `Default` instance is only used in this very test (both directly via `Vote::default()` and indirectly via `..Default::default`): https://github.com/informalsystems/tendermint-rs/blob/9690eca1b97685f699da58c1d9f3f8a85d5c33a9/tendermint/src/vote/sign_vote.rs#L277

romac: Tracked in https://github.com/informalsystems/tendermint-rs/issues/1025

karim-agha:
Can you please merge this? It's blocking my build.

Issue #1012, opened 2021-10-25
Impact assessment of removing `chrono` dependency


thanethomson:
Following from https://github.com/informalsystems/tendermint-rs/issues/1007#issuecomment-951043387, it could possibly make sense to remove our dependency on `chrono` altogether in tendermint-rs.

The primary outcome of this issue should be a clear indication as to whether this is possible and what the impact of this will be.

mzabaluev:
It should be possible to migrate from `chrono` to `time` 0.3 unless you use some exotic chrono APIs, outside of basic datetime manipulation and serde. I've done this recently on a small code base in https://github.com/input-output-hk/jortestkit/pull/96

thanethomson:
Yeah that's what I'm hoping. It's not super high priority right now, but I was thinking that, if I had some time to kill in the next month or so, I'd do a little scouting through the codebase and write up an [ADR](https://github.com/informalsystems/tendermint-rs/tree/master/docs/architecture) to quantify the impact.

@mzabaluev you're welcome to take this on if you have the time to spare.

mzabaluev:
@thanethomson Will do, also as a practice in writing ADRs.

On the initial examination, exposing the chrono type wrapped by `Time` may not be such a good idea.

As a side observation, the serde impls for both `Time` and `Timestamp` only use the string format. This is going to be wasteful and embarrassing when somebody uses a non-human-readable format serializer on structures containing these types (outside of protobuf which does not use serde). Perhaps the serde impls should use field-wise serialization of the `Timestamp` struct with any binary serializers.

soareschen:
For your refactoring using `time`, please also refrain from using the std features by adding `default-features = false` to the `time` crate for the pure part of tendermint crates that are used by `ibc`. If the current time is needed, they should be passed as arguments by the impure parts of the tendermint crates.

PR #910, opened 2021-06-21
Fix latest clippy assertion failures

thanethomson:
Closes #909 

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [x] Added entry in `.changelog/`

Issue #865, opened 2021-04-19
What's the purpose of `tendermint::Time`?


hdevalence:
The `tendermint` crate has a `Time(DateTime<Utc>)` structure that wraps the `chrono::DateTime<Utc>` type and implements a bunch of custom methods for it.

Currently in #862, I [translated](https://github.com/informalsystems/tendermint-rs/pull/862/commits/eeafb5212d67c92e1bf40949e99d8b558d6b1c6b) the Protobuf `Timestamp` type directly to a `DateTime<Utc>`, which I think is a more ergonomic type for downstream users and would prefer to use, unless there is a reason not to use it.

I wasn't sure from reading the documentation what the purpose of the custom time type was, but I didn't see an explanation of how it differs from a `DateTime<Utc>`.  Because `Time` has `impl From<DateTime<Utc>> for Time`, the `tendermint` crate's public API is still locked to `chrono` `0.4`, and because `From` impls are infallible, there's no way for the `Time` type to enforce any additional invariants on a `DateTime<Utc>`.  So I'm not sure that there's really a meaningful encapsulation benefit.

Another possible motivation could be for use with JSON serialization -- from looking at the commit history, it seemed like there was a bug involved in the custom serialization (#775).  It's easy to add custom serialization to a newtype wrapper compared to a third party type.  But if this was the motivation, an alternative that doesn't involve creating and maintaining a new type could be to define JSON serialization on the `tendermint_proto` structs, rather than on all of the domain types, and use the `From`/`TryFrom` impls to apply that serialization to the domain types.

greg-szabo:
All right, so if I understand correctly, you're advocating for getting rid of tendermint::Time in lieu of tendermint_proto::Timestamp since there's no real added value.

So far, looking at the code I found that
* tendermint::Time in Go is also based on google::protobuf::Timestamp (https://github.com/tendermint/spec/blob/d46cd7f573a2c6a2399fcab2cde981330aa63f37/spec/core/data_structures.md#time)
* JSON serialization has to be moved into tendermint_proto, but we already control Timestamp there so it's not a big deal to apply it.
* Or maybe it is, we need to check: there is a JSON serialization already applied to tendermint_proto::google::protobuf::Timestamp for RFC3339 conversion. I have to check if it's the same thing implemented in tendermint and why was it implemented twice.

[...]

PR #1030, opened 2021-11-26
Replace chrono with time 0.3

mzabaluev:
Fixes: #1012, #1007, #1008

Important or breaking changes:

* Removed `chrono` from dependencies and public API, replaced by `time`.
* The inner value of the `Time` tuple struct is made private.
* Renamed `Time::as_rfc3339` to `to_rfc3339` to follow Rust naming conventions (somehow clippy got it wrong at the time of #910, but it no longer does).
* Instead of `From<chrono::DateTime<Utc>>`, `Time` now implements `TryFrom<time::OffsetDateTime>`. The reason is that `OffsetDateTime` supports a wider range of dates than is supported by RFC 3339, and `tendermint::Time` needs its string conversion to be standard-compliant and infallible.
* Added `Time` methods `checked_add`/`checked_sub`, to follow the pattern set by `SystemTime` and `time::OffsetDateTime`. In a later change, these could replace use of the `Add`/`Sub` operator impls, whose `Result` output types make for poor ergonomics in operator overloading.
* `rpc` uses `time` types `OffsetDateTime` and `Date` in query operands instead of their `chrono` counterparts.

* [x] Wrote tests


thanethomson:
I'd recommend targeting the `v0.23.x` branch with this PR for now, since I assume you want this change to make its way into ibc-rs soon?

I was planning on merging #1022 ASAP, then we'll need to backport these changes to `master`.

mzabaluev:
> I assume you want this change to make its way into ibc-rs soon?

No, we only really need it to fix the "broken window" of audit failures. Also, it's a breaking change, so it should properly be included starting from 0.24.x (assuming this project follows cargo's stricter semver rules).

thanethomson:
We're actually going to allow for breaking changes to the v0.23.x series from one "patch" version to the next. I should add a note that people need to pin a specific version of tendermint-rs in their dependencies.

This is one of the side effects of the [versioning strategy](https://github.com/informalsystems/tendermint-rs/discussions/1018) we're going with.

Since we're also pre-1.0, I think there's still some leniency here.

So I'd still recommend merging this into `v0.23.x` and then I can backport the changes once #1022's been merged. I really want to get that PR out of the way without more conflicts, as the work's been ongoing for 7 months now and it's time to push it over the finish line :sweat_smile:

hdevalence:
Sorry for missing this until now, but is the only motivation for removing `chrono` to avoid the `localtime_r` "issue"? As a user, I'd rather keep the `chrono` crate, since it has a much friendlier API.

mzabaluev:
> avoid the `localtime_r` "issue"?

Unfortunately, it creates a broken window syndrome in cargo-audit CI checks for anyone who uses the `tendermint` crate.
And it is a soundness issue that I'd rather not have lingering in a public-facing library stack.

> As a user, I'd rather keep the chrono crate, since it has a much friendlier API.

`time` 0.3 is arguably simpler, and the API details are mostly down to aesthetics. I've come to like the value initializer macros and `format_description!`.

Hiding the inner value of `Time` would help sidestep the debate here; we could feature-gate conversion impls from/to both `chrono` and `time` types so you could use whichever you prefer. However, none of them on their own can ensure sanity and stability of RFC 3339 serialization.

thanethomson:
> As a user, I'd rather keep the chrono crate, since it has a much friendlier API.

The bigger problem here is that it appears as though `chrono` is not as well-maintained as we need it to be, and doesn't look like that's going to change any time soon (e.g. its last official release was over a year ago).

mzabaluev:
There is some activity on fixing the issue in e.g. https://github.com/chronotope/chrono/pull/578. But there is no ETA on that yet.

tony-iqlusion:
> The reason is that OffsetDateTime supports a wider range of dates than is supported by RFC 3339, and tendermint::Time needs its string conversion to be standard-compliant and infallible.

Does Tendermint not require Z-normalized RFC3339? If not, it probably should.

mzabaluev:
> Does Tendermint not require Z-normalized RFC3339?

I believe `Time` can be parsed from any RFC3339 compliant string, but its conversion to `String` and, consequently, `Serialize` normalize to the `Z` timezone offset suffix.

The problem is, RFC 3339 requires the year to be CE (i.e. positive) in four digits, and both `chrono::DateTime<Utc>` and its counterpart from `time` allow dates outside of that range.

tony-iqlusion:
I asked about mandatory Z-normalization on this issue: https://github.com/tendermint/tendermint/issues/6233#issuecomment-981853880

Regardless I think you could use `PrimitiveDateTime` instead of `OffsetDateTime` as the inner time type for the `tendermint::Time` newtype, and then handle conversion to UTC at the time the date is parsed. That way non-UTC dates become unrepresentable, eliminating any potential time zone-related bugs.

mzabaluev:
> Regardless I think you could use `PrimitiveDateTime` instead of `OffsetDateTime` as the inner time type for the `tendermint::Time` newtype

:heavy_check_mark: This saves some struct space, too.

hdevalence:
> > avoid the `localtime_r` "issue"?
> 
> Unfortunately, it creates a broken window syndrome in cargo-audit CI checks for anyone who uses the `tendermint` crate.

This seems like a tooling issue with `cargo-audit`; if library authors feel the need to make their API worse to avoid a broken-window syndrome, it's unclear that the advisories have positive utility.

> And it is a soundness issue that I'd rather not have lingering in a public-facing library stack.
> 
> > As a user, I'd rather keep the chrono crate, since it has a much friendlier API.
> 
> `time` 0.3 is arguably simpler, and the API details are mostly down to aesthetics. I've come to like the value initializer macros and `format_description!`.

I don't agree that `time` API is simpler. It has no way to encode the timezone in the date-time type (with the exception of the oddly-named `OffsetDateTime`, which hardcodes UTC), so the user has to keep track on their own, and cannot rely on the type system for assurance.  This pushes the complexity onto the user, which is worse.

I also don't agree that there is a real soundness issue here. Yes, the OS can cause a segfault [if you use it wrong](https://twitter.com/richfelker/status/1450300830381445121?s=21). But there are plenty of other ways that I can cause similar "soundness" issues from safe Rust -- for instance, opening up `/proc/mem` and writing into it. Rust is great. Unfortunately, the rest of computing and our current operating systems are built with sharp edges, and denialism about this imposes massive collateral damage while providing very little benefit to assurance.

(For instance, I am far more concerned about bugs that arise from timezone confusion than I am about the possibility of a crash from someone improperly editing the environment).

> Hiding the inner value of `Time` would help sidestep the debate here; we could feature-gate conversion impls from/to both `chrono` and `time` types so you could use whichever you prefer. However, none of them on their own can ensure sanity and stability of RFC 3339 serialization.

This option puts extra configuration burden on the library and forces every user to jump through hoops to get a workable API, which seems like a movement in the opposite direction from the simplification I suggested in #865.

tony-iqlusion:
> Sorry for missing this until now, but is the only motivation for removing chrono to avoid the localtime_r "issue"?

`time` is also much more actively maintained than `chrono`, whose development moves at a glacial pace.

> It has no way to encode the timezone in the date-time type

To my knowledge `tendermint::Time` is exclusively used to represent BFT time, which by definition is always UTC. If that's the case, being able to carry along an additional time zone other than UTC is a misfeature.

tony-iqlusion:
Another problem with `chrono` in this particular application I recently came across was this issue: https://github.com/tendermint/tendermint/issues/6233

> The standard date/time lib for Rust, chrono, doesn't support this flavor of RFC3339 timestamps. Instead it defaults to including trailing zeros up to the precision of the internal time representation.

Not sure if `time` fares any better here offhand.

proto/src/chrono.rs
```diff
@@ -1,53 +0,0 @@
-use core::convert::TryInto;
-
-use chrono::{DateTime, Duration, TimeZone, Utc};
```
mzabaluev, line 3: The whole module is removed, rather than being replaced with `time` equivalents, because of the problems detailed below. Conversions from/to `timestamp::Time` can serve as domain-validating, fallible alternatives, until more direct convenience methods are reintroduced here if needed.

proto/src/chrono.rs
```diff
@@ -1,53 +0,0 @@
...
-
-use crate::google::protobuf as pb;
-
-impl From<DateTime<Utc>> for pb::Timestamp {
-    fn from(dt: DateTime<Utc>) -> pb::Timestamp {
-        pb::Timestamp {
-            seconds: dt.timestamp(),
-            // This can exceed 1_000_000_000 in the case of a leap second, but
```
mzabaluev, line 11: This was against the wire protocol [specification](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp).

proto/src/chrono.rs
```diff
@@ -1,53 +0,0 @@
...
-                .timestamp_subsec_nanos()
-                .try_into()
-                .expect("timestamp_subsec_nanos bigger than i32::MAX"),
-        }
-    }
-}
-
-impl From<pb::Timestamp> for DateTime<Utc> {
-    fn from(ts: pb::Timestamp) -> DateTime<Utc> {
-        Utc.timestamp(ts.seconds, ts.nanos as u32)
```
mzabaluev, line 23: This panicked if members of `ts` have invalid values (subject to `chrono`'s interpretation which does not readily conform to the protobuf specification).

As the most likely origin of `pb::Timestamp` is by having been parsed from an incoming gRPC message by prost without any validation as to its contents, this is an easy DoS bomb under any application developers who would make use of this conversion. Please consider using fallible conversions when dealing with untrusted values originating from the network.

proto/src/chrono.rs
```diff
@@ -1,53 +0,0 @@
...
-        // chrono's Duration stores the fractional part as `nanos: i32`
-        // internally but doesn't provide a way to access it, only a way to get
-        // the *total* number of nanoseconds. so we have to do this cool and fun
-        // hoop-jumping maneuver
-        let seconds = d.num_seconds();
-        let nanos = (d - Duration::seconds(seconds))
-            .num_nanoseconds()
-            .expect("we computed the fractional part, so there's no overflow")
-            .try_into()
-            .expect("the fractional part fits in i32");
```
mzabaluev, line 42: This could produce protobuf messages that violate the [specification](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp).

The whole approach of having infallible conversions that can only either panic or produce off-spec values for the wire protocol in case the destination's validity requirements are not met, needs to be reconsidered.

proto/src/chrono.rs
```diff
@@ -1,53 +0,0 @@
-use core::convert::TryInto;
-
-use chrono::{DateTime, Duration, TimeZone, Utc};
-
-use crate::google::protobuf as pb;
-
-impl From<DateTime<Utc>> for pb::Timestamp {
-    fn from(dt: DateTime<Utc>) -> pb::Timestamp {
-        pb::Timestamp {
-            seconds: dt.timestamp(),
```
mzabaluev, line 10: The timestamp's range as represented by `chrono::DateTime` can exceed the range valid for the protobuf `Timestamp` messages.

proto/src/chrono.rs
```diff
@@ -1,53 +0,0 @@
...
-
-        pb::Duration { seconds, nanos }
-    }
-}
-
-impl From<pb::Duration> for Duration {
-    fn from(d: pb::Duration) -> Duration {
-        // there's no constructor that supplies both at once
-        Duration::seconds(d.seconds) + Duration::nanoseconds(d.nanos as i64)
-    }
```
mzabaluev, line 52: This is actually possible to replace with an infallible `time` equivalent (and `time` _does_ have a one-stop shop constructor for that!), but it felt odd to leave a lone conversion standing when others were sent back to the drawing board.

We should consider introducing a newtype for durations in the tendermint API, with similar range constraint enforcement as maintained in this PR for `Time`.

mzabaluev:
> being able to carry along an additional time zone other than UTC is a misfeature.

This reimplementation does not do so; the inner value is a `PrimitiveDateTime`, with `.assume_utc()` used to get the operational `OffsetDateTime` value where needed.

> I don't agree that `time` API is simpler. It has no way to encode the timezone in the date-time type (with the exception of the oddly-named `OffsetDateTime`, which hardcodes UTC), so the user has to keep track on their own, and cannot rely on the type system for assurance.

This is clever, but the type parameter of `DateTime` is also one example of added complexity that not every application would necessarily need. `OffsetDateTime` does not hardcode UTC, it supports any possible offset dynamically; the closest equivalent in chrono is `DateTime<FixedOffset>`. `tendermint::Time` takes the job of normalizing to UTC, among other value sanity checks that are necessary for a valid Tendermint timestamp.

An instance of unwanted complexity that bugs me more about chrono is its [handling of leap seconds](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveTime.html#leap-second-handling). It looks like a recipe for a wide variety of bugs, and it's not compatible with Google's definition of a protobuf timestamp, which specifically avoids dealing with leap second times. It's certainly not something I'd feel comfortable basing the protocol library's support for time representation on.

> I also don't agree that there is a real soundness issue here. Yes, the OS can cause a segfault [if you use it wrong](https://twitter.com/richfelker/status/1450300830381445121?s=21). But there are plenty of other ways that I can cause similar "soundness" issues from safe Rust -- for instance, opening up `/proc/mem` and writing into it.

That requires actual malicious intent. By contrast, `std::env::set_var` is a Rust function declared as safe, with only a vague mention of possible induced UB when also using FFI functions, which `chrono` does not give a hint about doing. It's very easy for a newbie to slip in an environment variable modification unaware of the consequences. That, of course, should also be linted against and warrants a security audit notice when found in a published library.

Are there really many other easy ways to violate soundness from safe Rust? Are we living in an illusion of safety?

> > Hiding the inner value of `Time` would help sidestep the debate here; we could feature-gate conversion impls from/to both `chrono` and `time` types so you could use whichever you prefer.

> This option puts extra configuration burden on the library

There is already a number of dependencies managed with feature gates, so this seems like more of the same.

> and forces every user to jump through hoops to get a workable API, which seems like a movement in the opposite direction from the simplification I suggested in #865.

As long as we don't have any other support for:
- the range and validity constraints imposed by RFC 3339 and the specification of protobuf well-known types;
- the specific details of RFC 3339 formatting to conform with the Go implementation;
- implementing the above directly on the DTO types used by prost/tonic;

I think it's necessary to have a newtype with fallible conversions here.

mzabaluev:
> Not sure if `time` fares any better here offhand.

I managed to get the desired behaviour out of it, but unfortunately that currently requires `std`: https://github.com/time-rs/time/issues/400

mzabaluev:
@thanethomson
> I'd recommend targeting the `v0.23.x` branch with this PR for now, since I assume you want this change to make its way into ibc-rs soon?

That's #1036 now.

> I was planning on merging #1022 ASAP, then we'll need to backport these changes to `master`.

Also done in follow-up commits here.

soareschen:
Can we have a parallel branch in `ibc-rs` that fix the breaking changes from this branch?

tendermint/src/time.rs
```diff
@@ -12,80 +11,120 @@ use core::time::Duration;
...
-            .signed_duration_since(other.0)
-            .to_std()
+        let duration = self.0.assume_utc() - other.0.assume_utc();
+        duration
+            .try_into()
             .map_err(|_| Error::duration_out_of_range())
     }
 
     /// Parse [`Time`] from an RFC 3339 date
     pub fn parse_from_rfc3339(s: &str) -> Result<Time, Error> {
```
thanethomson, line 98:
```suggestion
    pub fn parse_from_rfc3339(s: &str) -> Result<Self, Error> {
```

tendermint/src/time.rs
```diff
@@ -12,80 +11,120 @@ use core::time::Duration;
...
-        let date = DateTime::parse_from_rfc3339(s)
-            .map_err(Error::chrono_parse)?
-            .with_timezone(&Utc);
-        Ok(Time(date))
+        let date = OffsetDateTime::parse(s, &Rfc3339)
+            .map_err(Error::time_parse)?
+            .to_offset(offset!(UTC));
+        Time::from_utc(date)
```
thanethomson, line 102:
```suggestion
        Self::from_utc(date)
```

tendermint/src/time.rs
```diff
@@ -12,80 +11,120 @@ use core::time::Duration;
...
     }
 
-    /// Return an RFC 3339 and ISO 8601 date and time string with 6 subseconds digits and Z.
-    pub fn as_rfc3339(&self) -> String {
-        timestamp::as_rfc3339_nanos(&self.0)
+    /// Computes `self + duration`, returning `None` if an overflow occurred.
+    pub fn checked_add(self, duration: Duration) -> Option<Self> {
+        let duration = duration.try_into().ok()?;
+        let t = self.0.checked_add(duration)?;
+        Time::from_utc(t.assume_utc()).ok()
```
thanethomson, line 114:
```suggestion
        Self::from_utc(t.assume_utc()).ok()
```

tendermint/src/time.rs
```diff
@@ -12,80 +11,120 @@ use core::time::Duration;
...
+    }
+
+    /// Computes `self - duration`, returning `None` if an overflow occurred.
+    pub fn checked_sub(self, duration: Duration) -> Option<Self> {
+        let duration = duration.try_into().ok()?;
+        let t = self.0.checked_sub(duration)?;
+        Time::from_utc(t.assume_utc()).ok()
```
thanethomson, line 121:
```suggestion
        Self::from_utc(t.assume_utc()).ok()
```

tendermint/src/time.rs
```diff
@@ -97,47 +136,43 @@ impl FromStr for Time {
...
 }
 
-impl From<DateTime<Utc>> for Time {
-    fn from(t: DateTime<Utc>) -> Time {
-        Time(t)
+impl TryFrom<OffsetDateTime> for Time {
+    type Error = Error;
+
+    fn try_from(t: OffsetDateTime) -> Result<Time, Error> {
+        Time::from_utc(t.to_offset(offset!(UTC)))
```
thanethomson, line 143:
```suggestion
    fn try_from(t: OffsetDateTime) -> Result<Self, Error> {
        Self::from_utc(t.to_offset(offset!(UTC)))
```

tendermint/src/time.rs
```diff
@@ -97,47 +136,43 @@ impl FromStr for Time {
...
-        let res = self
+        let duration = rhs.try_into().map_err(|_| Error::duration_out_of_range())?;
+        let t = self
             .0
-            .checked_add_signed(duration)
+            .checked_add(duration)
             .ok_or_else(Error::duration_out_of_range)?;
-
-        Ok(Time(res))
+        Time::from_utc(t.assume_utc())
```
thanethomson, line 162:
```suggestion
        Self::from_utc(t.assume_utc())
```

tendermint/src/time.rs
```diff
@@ -97,47 +136,43 @@ impl FromStr for Time {
...
-        let res = self
+        let duration = rhs.try_into().map_err(|_| Error::duration_out_of_range())?;
+        let t = self
             .0
-            .checked_sub_signed(duration)
+            .checked_sub(duration)
             .ok_or_else(Error::duration_out_of_range)?;
-
-        Ok(Time(res))
+        Time::from_utc(t.assume_utc())
```
thanethomson, line 175:
```suggestion
        Self::from_utc(t.assume_utc())
```

tendermint/src/time.rs
```diff
@@ -12,80 +11,120 @@ use core::time::Duration;
...
+/// up to 9 digits and no trailing zeros, and the UTC offset denoted by Z.
+/// This reproduces the behavior of Go's `time.RFC3339Nano` format.
+///
+/// [specification]: https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp
+///
 #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
-#[serde(try_from = "Timestamp")]
-pub struct Time(pub DateTime<Utc>);
+#[serde(try_from = "Timestamp", into = "Timestamp")]
+pub struct Time(PrimitiveDateTime);
```
thanethomson, line 36: Given that there's no offset in a `PrimitiveDateTime`, is the assumption here that it'll always be in UTC? If so it may be worthwhile to note this in the struct documentation.

testgen/Cargo.toml
```diff
@@ -23,7 +23,7 @@ ed25519-dalek = { version = "1", default-features = false }
 gumdrop = { version = "0.8.0", default-features = false }
 simple-error = { version = "0.2.1", default-features = false }
 tempfile = { version = "3.1.0", default-features = false }
-chrono = { version = "0.4", default-features = false, features = ["clock"] }
+timelib = { package = "time", version = "0.3", default-features = false, features = ["std"] }
```
thanethomson, line 26: Why the `timelib` alias?

thanethomson (commented):
Just a few minor nits and some questions. Otherwise looks great! :tada:

proto/src/serializers/timestamp.rs
```diff
@@ -30,64 +35,96 @@ where
...
+            minute = t.minute(),
+            second = t.second(),
+        )
+    } else {
+        let mut secfrac = nanos;
+        let mut secfrac_width = 9;
+        while secfrac % 10 == 0 {
+            secfrac /= 10;
+            secfrac_width -= 1;
+        }
```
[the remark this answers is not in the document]
mzabaluev: The existing tests exercise `to_rfc3339` which uses this method, and they caught me out when I did it wrong, so I assume it is covered.

testgen/Cargo.toml
mzabaluev: It was ambiguous with the top-level `time` module, but I've found a better way to disambiguate.

tendermint/src/time.rs
```diff
@@ -12,80 +11,120 @@ use core::time::Duration;
...
+        // Safe to convert to i32 because .nanosecond()
+        // is guaranteed to return a value in 0..1_000_000_000 range.
+        let nanos = t.nanosecond() as i32;
+        Timestamp { seconds, nanos }
     }
 }
 
 impl Time {
+    fn from_utc(t: OffsetDateTime) -> Result<Self, Error> {
+        debug_assert_eq!(t.offset(), offset!(UTC));
```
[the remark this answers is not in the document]
mzabaluev: It's an internal method, all call sites should either produce a value with the UTC offset or convert to it.
The debug check is to catch if a programmer forgot.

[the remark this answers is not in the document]
mzabaluev: I have added a comment on the function.

mzabaluev:
Clippy failures should be fixed by #1041.

thanethomson (approved):
Great, I think we're good to go here. Thanks @mzabaluev!

PR #1035, opened 2021-11-29
ADR: Time API improvements

mzabaluev:
Proposes changes to time API in `tendermint` and some related improvements that would be good to do in a tranche of breaking changes.


thanethomson (approved):
Excellent! Well-reasoned, I'd say. We can revisit the time issue again once the Rust ecosystem's figured out a more coherent strategy for handling time safely/soundly.

Issue #1033, opened 2021-11-26
meta: Backport tracking


thanethomson:
Consider this a tracking issue for all PRs that need to be backported to a particular branch.

### To be backported to `master`:

- [ ] #1030 

### To be backported to `v0.23.x`:

(none right now)

PR #1046, opened 2021-12-08
Use ed25519-consensus instead of ed25519-dalek

### LINKED ISSUE #355: Consider `ed25519-zebra` crate?
Presently the `tendermint` crate uses the [`ed25519-dalek`](https://github.com/dalek-cryptography/ed25519-dalek) vicariously through the [`signatory-dalek`](https://github.com/iqlusioninc/signatory/tree/develop/signatory-dalek) crate.

There's an alternative I think that's worth considering: [`ed25519-zebra`](https://github.com/ZcashFoundation/ed25519-zebra). Both `ed25519-dalek` and `ed25519-zebra` are based on the underlying [`curve25519-dalek`](https://github.com/dalek-cryptography/curve25519-dalek) library. However, `ed25519-zebra` focuses on Ed25519 usage in consensus-critical (i.e. blockchain) contexts, where [discrepancies in Ed25519 validation have been problematic in the past](https://github.com/zcash/zcash/issues/2872). Also, `ed25519-zebra` is written by `curve25519-dalek` author @hdevalence.

There are a few issues at play here: existing Ed25519 libraries don't properly implement [RFC 8032](https://tools.ietf.org/html/rfc8032), and RFC 8032 doesn't go far enough to remove edge cases in verification. Per @hdevalence, he aims for `ed25519-zebra` to be the first Ed25519 library to actually implement RFC 8032 verification logic correctly.

If you do decide to adopt `ed25519-zebra` it seems like you'd want to ensure the Go side has a matching solution, either by implementing the same verification checks that `ed25519-zebra` uses, or by using `ed25519-zebra` directly. I'm guessing the latter might be problematic as it seems unlikely Tendermint would adopt a Rust-based dependency.

More generally, even if you don't wind up adopting `ed25519-zebra`, I think it would make sense to more precisely specify the Ed25519 verification logic Tendermint uses as part of an overall Tendermint specification, as I assume it's presently "whatever Go does".

This seems like the sort of change it might be good to introduce in a chain upgrade (e.g. Stargate). Namely Tendermint only uses Ed25519 signatures for consensus, so previous chains' Ed25519 signatures are irrelevant (only the ECDSA/secp256k1 signed transactions are persisted across chain upgrades). This means introducing stricter Ed25519 verification logic during a chain upgrade carries no risk (to my knowledge) of existing signatures which previously verified but wouldn't verify under the stricter logic, since it's a brand new chain and therefore has no existing Ed25519 signatures to worry about.

/cc @zmanian

hdevalence:
Closes #355 (see that issue for more context; `ed25519-consensus` is a fork of
`ed25519-zebra` that's Zcash-independent).

This change ensures that `tendermint-rs` has the same signature verification as
`tendermint`, which uses the validation criteria provided by the Go
`ed25519consensus` library.

* [x] Wrote tests
* ~~[ ] Added entry in `.changelog/`~~ I did not do this, because I'm not sure what release to target.

### DIFF
--- a/p2p/src/error.rs
+++ b/p2p/src/error.rs
@@ -2,7 +2,6 @@
 
 use flex_error::{define_error, DisplayOnly};
 use prost::DecodeError;
-use signature::Error as SignatureError;
 
 define_error! {
     Error {
@@ -36,7 +35,6 @@ define_error! {
             | _ | { "public key missing" },
 
         Signature
-            [ DisplayOnly<SignatureError> ]
             | _ | { "signature error" },
 
         UnsupportedKey
--- a/p2p/src/secret_connection/protocol.rs
+++ b/p2p/src/secret_connection/protocol.rs
@@ -2,7 +2,6 @@
 
 use std::convert::TryInto;
 
-use ed25519_dalek as ed25519;
 use prost::Message as _;
 
 use x25519_dalek::PublicKey as EphemeralPublic;
@@ -113,8 +112,8 @@ impl Version {
     #[must_use]
     pub fn encode_auth_signature(
         self,
-        pub_key: &ed25519::PublicKey,
-        signature: &ed25519::Signature,
+        pub_key: &ed25519_consensus::VerificationKey,
+        signature: &ed25519_consensus::Signature,
     ) -> Vec<u8> {
         if self.is_protobuf() {
             // Protobuf `AuthSigMessage`
@@ -126,7 +125,7 @@ impl Version {
 
             let msg = proto::p2p::AuthSigMessage {
                 pub_key: Some(pub_key),
-                sig: signature.as_ref().to_vec(),
+                sig: signature.to_bytes().to_vec(),
             };
 
             let mut buf = Vec::new();
@@ -168,8 +167,8 @@ impl Version {
     #[cfg(feature = "amino")]
     fn encode_auth_signature_amino(
         self,
-        pub_key: &ed25519::PublicKey,
-        signature: &ed25519::Signature,
+        pub_key: &ed25519_consensus::VerificationKey,
+        signature: &ed25519_consensus::Signature,
     ) -> Vec<u8> {
         // Legacy Amino encoded `AuthSigMessage`
         let msg = amino_types::AuthSigMessage::new(pub_key, signature);
@@ -184,8 +183,8 @@ impl Version {
     #[cfg(not(feature = "amino"))]
     fn encode_auth_signature_amino(
         self,
-        _: &ed25519::PublicKey,
-        _: &ed25519::Signature,
+        _: &ed25519_consensus::VerificationKey,
+        _: &ed25519_consensus::Signature,
     ) -> Vec<u8> {
         panic!("attempted to encode auth signature using amino, but 'amino' feature is not present")
     }
--- a/tendermint/src/error.rs
+++ b/tendermint/src/error.rs
@@ -208,8 +208,7 @@ define_error! {
             |_| { format_args!("subtle encoding error") },
 
         Signature
-            [ DisplayOnly<signature::Error> ]
-            |_| { format_args!("signature error") },
+            |_| { "signature error" },
 
         TrustThresholdTooLarge
             |_| { "trust threshold is too large (must be <= 1)" },
--- a/tendermint/src/public_key.rs
+++ b/tendermint/src/public_key.rs
@@ -1,6 +1,6 @@
 //! Public keys used in Tendermint networks
 
-pub use ed25519_dalek::PublicKey as Ed25519;
+pub use ed25519_consensus::VerificationKey as Ed25519;
 #[cfg(feature = "secp256k1")]
 pub use k256::ecdsa::VerifyingKey as Secp256k1;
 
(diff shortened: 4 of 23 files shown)


tendermint/src/public_key.rs
```diff
@@ -106,7 +108,9 @@ impl PublicKey {
 
     /// From raw Ed25519 public key bytes
     pub fn from_raw_ed25519(bytes: &[u8]) -> Option<PublicKey> {
-        Ed25519::from_bytes(bytes).map(Into::into).ok()
+        Ed25519::try_from(bytes)
+            .map(|vk| PublicKey::Ed25519(vk))
```
thanethomson, line 112:
```suggestion
            .map(PublicKey::Ed25519)
```

tendermint/src/public_key.rs
```diff
@@ -132,15 +136,14 @@ impl PublicKey {
...
                         Error::signature_invalid(
                             "Ed25519 signature verification failed".to_string(),
                         )
                     }),
-                    Err(e) => Err(Error::signature_invalid(format!(
-                        "invalid Ed25519 signature: {}",
-                        e
+                    Err(_) => Err(Error::signature_invalid(format!(
+                        "Could not parse Ed25519 signature",
                     ))),
```
thanethomson, line 147:
```suggestion
                    Err(_) => Err(Error::signature_invalid(
                        "Could not parse Ed25519 signature".to_string(),
                    )),
```

p2p/src/error.rs
```diff
@@ -36,7 +35,6 @@ define_error! {
             | _ | { "public key missing" },
 
         Signature
-            [ DisplayOnly<SignatureError> ]
             | _ | { "signature error" },
```
thanethomson, line 38: Would it not be worthwhile to bubble up the original error here? Unfortunately it seems like the `ed25519_consensus::Error` type doesn't implement `std::fmt::Display` unless you turn on the `std` feature.

p2p/src/secret_connection/protocol.rs
```diff
@@ -184,8 +183,8 @@ impl Version {
     #[cfg(not(feature = "amino"))]
     fn encode_auth_signature_amino(
```
thanethomson, line 184: I wonder if this method's even necessary. Why would one want to convert a compile-time error to a runtime error like this?

tendermint/src/error.rs
```diff
@@ -208,8 +208,7 @@ define_error! {
             |_| { format_args!("subtle encoding error") },
 
         Signature
-            [ DisplayOnly<signature::Error> ]
-            |_| { format_args!("signature error") },
+            |_| { "signature error" },
```
thanethomson, line 211: Similar to my other comment: should we not find a way here to bubble up the original error?

thanethomson:
> I did not do this, because I'm not sure what release to target.

I'd say target `master` (i.e. v0.24.0 of tendermint-rs). If you'd like this to appear in the v0.23.x series then you'd need to submit a separate PR targeting the `v0.23.x` branch, or make a note of it in #1033 and I'll handle it in the next couple of days.

The changelog entry should probably be listed as a breaking change, right?

p2p/src/error.rs
hdevalence: The motivation for removing the original error was that the code this is replacing used an opaque error type, so there's no additional information conveyed by the original error, and was easier to just remove it rather than patch the `ed25519_consensus::Error` type to have an extra `Display` impl.  But I could add that `Display` impl in another release.

p2p/src/secret_connection/protocol.rs
hdevalence: Not sure, I tried to leave the rest of the code untouched as a drop-in change.

tendermint/src/error.rs
hdevalence: xref https://github.com/informalsystems/tendermint-rs/pull/1046#discussion_r765322068

tendermint/src/public_key.rs
```diff
@@ -106,7 +108,9 @@ impl PublicKey {
 
     /// From raw Ed25519 public key bytes
     pub fn from_raw_ed25519(bytes: &[u8]) -> Option<PublicKey> {
-        Ed25519::from_bytes(bytes).map(Into::into).ok()
+        Ed25519::try_from(bytes)
+            .map(PublicKey::Ed25519)
+            .ok()
```
thanethomson, line 113:
```suggestion
        Ed25519::try_from(bytes).map(PublicKey::Ed25519).ok()
```

cargo fmt complaint

test/src/test/unit/p2p/secret_connection.rs
```diff
@@ -3,7 +3,7 @@ use std::io::Write as _;
 use std::net::{TcpListener, TcpStream};
 use std::thread;
 
-use ed25519_dalek::{self as ed25519};
+use ed25519_consensus;
```
thanethomson, line 6:
```suggestion
```

clippy lint

thanethomson:
Just a changelog entry please and these (hopefully) last few lints, then we can merge.

thanethomson:
Updated and superseded by #1067

Issue #1047, opened 2021-12-08
Dependency on time 0.3.5 conflicts with requirements in ibc-rs dependency tree


mzabaluev:
Version(s) of tendermint-rs: v0.23.x branch pre 0.23.2

### What went wrong?

Changes landed in #1036 break the build for ibc-rs because it requires `rouille`, which pins its dependency on time to `=0.3.2`: https://github.com/tomaka/rouille/issues/246

De-facto `tendermint` now requires 0.3.5 though (which erroneously is not what `Cargo.toml` specifies) due to the API it uses.

### Steps to reproduce

Uncomment the patch section in the ibc-rs project's top-level `Cargo.toml`, change the branch to `v0.23.x` and try to build.

### Definition of "done"

The quickest fix is to remove any use of `time` API post 0.3.2 and make sure nothing in the project requires a higher version.
Luckily, only some non-essential additions currently require that, which can be undone. I'd like to keep the full API In master, though. ibc-rs would have time to address the dependency problems before switching to that.

mzabaluev:
Still no cigar: I left a couple more `checked_add`/`checked_sub` uses in the code. Will fix in another PR.

PR #1048, opened 2021-12-08
Roll back time-rs dependency and remove methods that require version 0.3.5

mzabaluev:
Fixes: #1047

Do not require time 0.3.5, because ibc-rs has another dependency that
pins time at =0.3.2.

This means the recently added methods `Time::checked_add` and
`Time::checked_sub` have to be removed. Fortunately, they haven't made it
into a 0.23.x release yet.

* ~Updated all relevant documentation in docs~
* ~Wrote tests~


mzabaluev:
Weirdly, the two CI failures are due to two different tools failing installation over the network. These do not seem related to the change.

thanethomson (approved):

PR #1061, opened 2021-12-17
rpc: Add temporary fix and tests for `block_results` serialization

### LINKED ISSUE #1021: RPC client fails to decode block_results with non-nil validator_updates
RPC client fails to decode `block_results` response with non empty `validator_updates`.
@mankenavenkatesh has reported this problem, thanks Venkatesh!  It shows on a number of chains when hermes tries to look for begin/end blocker IBC events by calling the `block_results` RPC.
Decoding the response fails when `validator_updates` is not empty. This is reported for some chains at tm-go version `v0.34.14`. I believe the format is different than what we expect, here is one example of `validator_updates` from one chain:

```
    "validator_updates": [
      {
        "pub_key": {
          "Sum": {
            "type": "tendermint.crypto.PublicKey_Ed25519",
            "value": {
              "ed25519": "VqJCr3vjQdffcLIG6RMBl2MgXDFYNY6b3Joaa43gV3o="
            }
          }
        },
        "power": "573929"
      }
    ],
```
I don't think there were recent changes in the crypto PubKey proto definition, maybe this never worked.
Full output of `curl http://18.191.147.51:26657/block_results?height=4555980` can be found [here](https://gist.github.com/ancazamfir/42961787930f310b20731ac83114f419
).

**Steps to reproduce**
Use the above structure for `validator_updates` in the RPC block_results test fixture. The output is:
```
called `Result::unwrap()` on an `Err` value: serde parse error

Caused by:
    missing field `type` at line 103 column 9

```

Be able to decode the `block_results` response with non-nil validator updates.

thanethomson:
Closes #1021 

The important changes are in:

- [`tendermint/src/public_key.rs`](https://github.com/informalsystems/tendermint-rs/pull/1061/files#diff-2674c05b71314dc6aaf0f283b5d45aaad34d44db68c6022a2783d7bb9d715518)
- [`tendermint/src/validator.rs`](https://github.com/informalsystems/tendermint-rs/pull/1061/files#diff-75a0f0afb35852f208256393b2e784c436677dda836424c43c7110406aba856c)

The reason for the massive line count in this PR:

1. I removed the `rpc/tests/support` folder and its fixtures, since those were manually generated, in favour of automatically generated fixtures from the rpc-probe (this PR also introduces a small query plan for interacting with a Gaia node).
2. I generated some fixtures from a local Gaia node. These will be replaced when I can get access to the WebSocket endpoint of a production Gaia node.
3. The output of `curl http://18.191.147.51:26657/block_results?height=4555980` is added in here as a specific test case to ensure it parses.

One simple way of verifying this code is to check out this branch locally and then, from the root of the repository, run:

```bash
# Fetch the block_results response at height 4555980 (previously failed to parse),
# output it to stdout, and pass it through jq for readability.
cargo run -p tendermint-rpc \
    --features cli \
    --bin tendermint-rpc \
    -- \
    -u http://18.191.147.51:26657/ \
    block-results 4555980 | jq
```

* [x] Wrote tests


adizere (approved):
Amazing work Thane!

Confirm this works with the provided tests.

One last thought: Would it be worthwhile to make a PR on Hermes pinned with a dependency on this branch, and then pass it to Injective folks for further end-to-end testing? It seems unlikely we can uncover anything with this additional step, but not sure.

PR #1062, opened 2021-12-20
Bump getrandom to 0.2

tomtau:


rpc/Cargo.toml
```diff
@@ -62,7 +62,7 @@ websocket-client = [
 
 [dependencies]
 bytes = { version = "1.0", default-features = false }
-getrandom = { version = "0.1", default-features = false }
+getrandom = { version = "0.2", default-features = false, features = ["js"] }
```
thanethomson, line 65: From [the docs](https://docs.rs/getrandom/latest/getrandom/#webassembly-support):

> if the "js" Cargo feature is enabled, this crate will assume that you are building for an environment containing JavaScript, and will call the appropriate methods. Both web browser (main window and Web Workers) and Node.js environments are supported, invoking the methods described above using the wasm-bindgen toolchain.
>
> This feature has no effect on targets other than wasm32-unknown-unknown.

Seems reasonable to enable this feature, thanks!

thanethomson (approved):

PR #1071, opened 2022-01-04
Split out verifier parts of `tendermint-light-client` to `tendermint-light-client-verifier`

### LINKED ISSUE #1027: Split out verifier parts of `tendermint-light-client` to a separate no_std compliant crate
The `ibc` crate now uses `tendermint-light-client` to verify IBC light clients. However `tendermint-light-client` also contains many other logic that require use of `std`.

We should move out the verifier part of the code into a crate like `tendermint-light-client-verifier`, so that it is easier to support no_std in that crate and leave `tendermint-light-client` to continue using `std`.

soareschen:
Fixes #1027.

* [ ] Wrote tests


thanethomson (commented):
I've submitted #1074 with some recommendations - what do you think?

If they make sense, I can do the same for your other PR tomorrow to backport those changes to v0.23.x.

soareschen:
> I've submitted #1074 with some recommendations - what do you think?

Looks good. Thanks!

I have also added re-export to the original paths that are exported by the `tendermint-light-client` crate.

light-client/src/lib.rs
```diff
@@ -23,15 +23,18 @@ pub mod errors;
...
 pub mod supervisor;
-pub mod types;
 
 pub(crate) mod utils;
 
+// Re-export `tendermint-light-client-verifier` crate.
+pub use tendermint_light_client_verifier as verifier;
+
+// Re-export for backward compatibility
+pub use verifier::{operations, predicates, types};
```
thanethomson, line 37: I think this is better suited to the v0.23.x series. Since we're dealing with an explicitly breaking change here (tendermint-rs v0.24.0) we may as well abandon the backward compatibility. Especially given that it's basically just ibc-rs that uses this crate at present as far as I can see. See https://crates.io/crates/tendermint-light-client/reverse_dependencies

thanethomson: In other words, I don't think we need https://github.com/informalsystems/tendermint-rs/pull/1071/commits/a4d7dc76831e312d07d42d1d37f2c936e8c8ac2a for `master` at this point. What do you think? Do you think it's necessary for ibc-rs for us to maintain this backward compatibility?

thanethomson: Actually don't worry about it - we can remove these in a subsequent PR. I think we're good to go here.

PR #1078, opened 2022-01-12
Release v0.24.0-pre.1

thanethomson:
:book: [CHANGELOG](https://github.com/informalsystems/tendermint-rs/blob/release/v0.24.0-pre.1/CHANGELOG.md#v0240-pre1)

The changelog includes all changes added from v0.23.0, including "frontports" from v0.23.x. Since there's effectively a lot of repetition in the changelogs between v0.23.x and v0.24.x now, we have a few options when it comes time to retire v0.23.x:

1. Just merge all v0.23.x changelog entries into the changelog on `master`. This will necessarily include a lot of duplication between v0.23.x releases and the v0.24.0-pre.1 release.
2. Drop all changes from v0.23.1 before v0.24.0.
3. Reformat the v0.24.x changelog to deduplicate the entries that already effectively appeared in the v0.23.x series.

I'm currently in favor of option (1) because it's the most accurate one as to what's changed in each release. Are there other options that could be better? Or do others have different preferences?


mzabaluev (approved):
Looks good to me.

PR #1080, opened 2022-01-13
Use Time instead of u64 seconds in tendermint-testgen Header (v0.23)

soareschen:
Fixes https://github.com/informalsystems/ibc-rs/issues/1687

The current tendermint-testgen header records time as `u64` in second precision. This means that any sub-second update will trim to the same second, causing the light client verification to fail if two headers fall in the same second. Using the proper `Time` struct will resolve this issue, allowing the header of a new block to have lower than a second difference than the previous block.

* [ ] Wrote tests

### DIFF
--- a/light-client/src/supervisor.rs
+++ b/light-client/src/supervisor.rs
@@ -585,7 +585,10 @@ mod tests {
 
         let peer_id = provider.unwrap_or("0BEFEEDC0C0ADEADBEBADFADADEFC0FFEEFACADE");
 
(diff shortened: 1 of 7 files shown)


thanethomson (approved):
LGTM! As far as I can see, the testgen data structures are only ever deserialized from JSON fixtures that're already inside the project, and mainly in the `tendermint-light-client` crate, so if the light client integration tests still work it's all good.

PR #1098, opened 2022-03-07
ci: Bump grcov version to 0.8.7

thanethomson:
Minor update to CI.


thanethomson:
Clippy failures will be sorted out in #1097.

PR #1105, opened 2022-03-16
Add support for `tendermint::chain::Id` constants

tony-iqlusion:
In PeggyJv/ocular#18 we are trying to create a registry of "well-known" chain IDs. I proposed using `tendermint::chain::Id` for this, but to do so we'd need to have a const initializer support.

This commit changes the internal representation of `chain::Id` to `Cow<'static, str>`. This permits a `const fn` initializer, but avoids adding a lifetime to `chain::Id` itself.

It also adds `chain::Id::new` which is defined as a `const fn`, which can be used to define chain ID constants.

* [x] Wrote tests

### DIFF
--- a/tendermint/src/chain/id.rs
+++ b/tendermint/src/chain/id.rs
@@ -2,6 +2,7 @@
 
 use crate::error::Error;
 use crate::prelude::*;
+use alloc::borrow::Cow;
 use core::convert::TryFrom;
 use core::{
     cmp::Ordering,
@@ -19,7 +20,7 @@ pub const MAX_LENGTH: usize = 50;
 
 /// Chain identifier (e.g. 'gaia-9000')
 #[derive(Clone)]
-pub struct Id(String);
+pub struct Id(Cow<'static, str>);
 
 impl Protobuf<String> for Id {}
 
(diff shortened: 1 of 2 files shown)


tony-iqlusion:
If you're happy with the general direction I can add some tests for this specific constructor.

thanethomson:
> If you're happy with the general direction I can add some tests for this specific constructor.

Seems good to me!

tony-iqlusion:
Tests and changelog entry added

tendermint/src/chain/id.rs
```diff
@@ -133,38 +147,81 @@ impl<'de> Deserialize<'de> for Id {
...
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::error::ErrorDetail;
 
     const EXAMPLE_CHAIN_ID: &str = "gaia-9000";
+    const MAX_LENGTH_CHAIN_ID: &str = "01234567890123456789012345678901234567890123456789";
+    const OVERLENGTH_CHAIN_ID: &str = "012345678901234567890123456789012345678901234567890";
```
tony-iqlusion, line 173: These need to be constants to test `Id::new`, which accepts a `&'static str` as an argument.

tendermint/src/chain/id.rs
```diff
@@ -133,38 +147,81 @@ impl<'de> Deserialize<'de> for Id {
...
 
     #[test]
-    fn rejects_empty_chain_ids() {
-        match "".parse::<Id>().unwrap_err().detail() {
-            ErrorDetail::Length(_) => {}
-            _ => panic!("expected length error"),
-        }
+    fn from_str_rejects_empty_chain_ids() {
+        assert!(
+            matches!(
```
tony-iqlusion, line 191: Soon: `assert_matches!`

https://github.com/rust-lang/rust/issues/82775

thanethomson (approved):
LGTM!

PR #1110, opened 2022-03-29
Updated spec url

brson:
The tendermint/spec repo is archived and directs people to the URL in this patch.


thanethomson (approved):
Thanks for catching this!

PR #1116, opened 2022-04-08
p2p: Rework transport to be async

xla:
While it was well-intended the reality is that it's going to be
impractical and hinder potential adoption trying to not give in to the
sweet siren sounds of the async ecosystem. The author since has changed
their stance and adopted a conviction that well tuned and guard-railed
async can be workable and lead to maintainable solutions. This
change-set is in preparation of bringing in peer and supervisor
abstractions as described in the ADR.

_Individual commits are significant please review accordingly._

~~* [ ] Wrote tests~~

### DIFF
--- a/docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
+++ b/docs/architecture/adr-009-transport-agnostic-peer-abstraction.md
@@ -2,6 +2,7 @@
 
 ## Changelog
 * 2021-02-05: drafted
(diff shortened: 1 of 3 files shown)


xla:
Fails because of the unmaintained serde_cbor advisory, temporarily addressed in #1119

thanethomson (approved):
Oh those sweet siren sounds :mermaid:

PR #995, opened 2021-09-29
light-client: Replace `Io` with async variant

xla:
Individual commits are significant please review accordingly.

Closes #953 
Supersedes #965

* [x] Referenced an issue explaining the need for the change
* [x] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [ ] ~~Wrote tests~~
* [x] Added entry in `.changelog/`

PR #1121, opened 2022-04-10
ci: Update CI coverage check to use llvm-cov

thanethomson:
Follows from https://github.com/informalsystems/tendermint-rs/pull/995#issuecomment-1094168166

This updates our coverage check in CI to use https://github.com/taiki-e/cargo-llvm-cov and Rust stable.


xla (approved):
(*´∀`)♪

PR #1124, opened 2022-05-04
Release v0.24.0 pre.2

thanethomson:
:book: [CHANGELOG rendered](https://github.com/informalsystems/tendermint-rs/blob/release/v0.24.0-pre.2/CHANGELOG.md#v0240-pre2)


hdevalence:
The breaking change to `prost` is unfortunate, since it's no longer possible to use `0.24.x` (the only published versions with the ABCI domain types) with any of the other ecosystem crates (e.g., `ibc-rs`), or with our crates, which use a different version of `prost`.  (It also broke our build, since we were depending on `0.24.0-pre.1`, which is the only published version that includes ABCI support).

thanethomson:
@hdevalence I'm not sure I understand. Is there something that you need from [v0.24.0-pre.2](https://github.com/informalsystems/tendermint-rs/blob/master/CHANGELOG.md#v0240-pre2) that doesn't involve the prost upgrade?

Also, the current released version(s) of ibc-rs [depend on the v0.23.x series of tendermint-rs](https://github.com/informalsystems/ibc-rs/blob/181b14a05721673055d8564608d6fac520d5c86c/relayer/Cargo.toml#L76).

ibc-rs currently targets Tendermint Core v0.34.x, and the tendermint-rs v0.24.x series [targets Tendermint Core v0.35.x](https://github.com/informalsystems/tendermint-rs#versioning).

Are you patching ibc-rs while using this pre-release version of tendermint-rs v0.24.x?

thanethomson:
I've logged #1127 to clarify and track what's happening with the v0.24.x series of tendermint-rs at present.

PR #1126, opened 2022-05-06
chore: Bump protos to those provided by Tendermint v0.35.4

thanethomson:
Doesn't seem like anything substantial changed - prost's code generation is just nondeterministic unfortunately.


xla (approved):
^_^

PR #1130, opened 2022-05-09
Bump `sha2` to v0.10; replace `ripemd160` with `ripemd` v0.1

tony-iqlusion:
These crates are both based on `digest` v0.10, which is also the same version used by `k256` v0.11 (#1129).

Note that the `ripemd160` crate is now deprecated:

https://docs.rs/ripemd160/latest/ripemd160/

The `ripemd` crate is the replacement. It has the same maintainers (@RustCrypto), but now combines the former `ripemd160` and `ripemd320` crates into a single crate.


xla (approved):
✍(◔◡◔)

thanethomson (approved):

PR #1131, opened 2022-05-24
Fix JSON deserialization of abci::ResponseInfo

mzabaluev:
Fixes: #1132

The fields of the tendermint-go struct equivalent to `abci::ResponseInfo` are annotated with `json:"omitempty"`, so give them default values for `Deserialize`. The `/abci_info` endpoint is known to omit the `app_version` field in some recent revisions of Gaia.

* [ ] Wrote tests

### DIFF
--- a/tools/proto-compiler/src/constants.rs
+++ b/tools/proto-compiler/src/constants.rs
@@ -98,6 +98,7 @@ pub static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
         ".tendermint.abci.ResponseInfo.last_block_height",
         QUOTED_WITH_DEFAULT,
     ),
+    (".tendermint.abci.ResponseInfo.last_block_app_hash", DEFAULT),
     (
         ".tendermint.abci.ResponseInfo.last_block_app_hash",
         BYTES_SKIP_IF_EMPTY,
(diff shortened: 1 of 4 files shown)


tools/proto-compiler/src/constants.rs
```diff
@@ -85,14 +86,22 @@ pub static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
...
+    (".tendermint.abci.ResponseInfo.data", DEFAULT),
+    (".tendermint.abci.ResponseInfo.version", DEFAULT),
+    (
+        ".tendermint.abci.ResponseInfo.app_version",
+        QUOTED_WITH_DEFAULT,
+    ),
+    (
+        ".tendermint.abci.ResponseInfo.last_block_height",
+        QUOTED_WITH_DEFAULT,
+    ),
```
mzabaluev, line 100: Not sure if it's best to fix it here with compiler tweaks, or maybe move the serde annotations to the `AbciInfo` struct, where the serde traits are really used, while currently being delegated to `tendermint_proto::abci::ResponseInfo`.
We have to customize serde for every field of the involved struct in either case, so doing it directly in source would be cleaner.

greg-szabo (approved):
Interesting. You'd think "default" de/serializations have been around long enough that this is already listed in the `consts` list, but alas...

Anyway, if this works for now, cool.

thanethomson (approved):

PR #1143, opened 2022-06-10
Tendermint RPC 0.34 compatibility API

mzabaluev:
Add a feature-gated `v0_34` module in the tendermint-rpc crate, providing compatibility with Tendermint nodes using the RPC protocol as implemented in Tendermint Go version 0.34.

The `v0_34` module provides APIs that exhibit the protocol differences, mainly due to the changes in the subscription event data structure:

- `Event`
- `Subscription`
- `SubscriptionClient`
- `WebSocketClient`

* [x] Wrote tests


rpc/src/v0_34/client/subscription.rs
```diff
@@ -0,0 +1,92 @@
...
+    Stream,
+};
+use pin_project::pin_project;
+
+use crate::{client::subscription::Inner, prelude::*, query::Query, v0_34::event::Event, Error};
+
+/// A client that exclusively provides [`Event`] subscription capabilities,
+/// without any other RPC method support.
+#[async_trait]
+pub trait SubscriptionClient {
```
thanethomson, line 17: One of the benefits of this approach here is that the API can evolve across different versions of Tendermint, but one of the drawbacks is that callers need to refer to Tendermint version-specific client traits.

An alternative approach here would be to have one `SubscriptionClient` trait for all supported versions of Tendermint, and have it be generic over the `Event` type produced by the subscription client. Then the caller could also be generic over the event type. I'd be interested to know whether you considered and ruled out such an approach? (I haven't actually done the hard work here, so I'm not aware yet of the ripple effects throughout the codebase of such an approach.)

thanethomson (commented):
Wow, this seems like a substantial amount of work @mzabaluev, thanks so much! I didn't realize we'd need to make the client machinery generic over the data structures that differ between Tendermint versions, but it totally makes sense.

Two questions, and it'd be great to also have a changelog entry for this, but otherwise LGTM :+1:

rpc/src/v0_34/client/subscription.rs
mzabaluev: I think it's possible to introduce `Event` as an associated type for the `SubscriptionClient` trait, but I was not sure if the more substantial API change is desirable. Currently, `v0_34` can be treated like a bolt-on with no complications to the definitions elsewhere. In the relayer, we'll have to explicitly specialize for both event types anyway, because parsing for old subscription events is (unnecessarily) different.

thanethomson: Makes sense :+1: I'd say we should keep your changes as-is then, and we can always look at merging the traits at some point in a future version of tendermint-rs if necessary.

mzabaluev:
In a meeting with folks from the Tendermint team at Interchain, it has been suggested that we don't need to expose the pre-0.35 event data structure and its separate subscription stream API, what we can instead do is to transform the "flattened" key-value map into a list of ABCI events using the knowledge of how Tendermint 0.34 nodes encode event types and tags into the flattened keys.

With this, there are fewer reasons to provide a statically different `v0_34` API. The remaining ones that come to my mind are:

- Not hiding dynamically switched behaviour that would be needed to support 0.35+ and 0.34 dialects through the single API. The performance impact of that should not be something to worry about, though.
- Statically expose the lack of `/block_by_hash` endpoint in 0.34, by having a `v0_34::Client` trait that lacks the `block_by_hash` method. Not sure how useful that would be in practice.
- As long as 0.34 support is feature-gated, altering behaviour via compile-time features is [a bit of an anti-pattern](https://users.rust-lang.org/t/behavior-altering-features-considered-harmful/68587).

mzabaluev:
I'm closing this for now due to lack of urgent need for this change (given non-adoption of Tendermint 0.35 in the upcoming release of Cosmos SDK) and the alternative idea of transforming the pre-0.35 RPC subscription event data into the uniform current API.

PR #1145, opened 2022-06-17
Remove direct ed25519 crate dependency

ValarDragon:
Mild dependency cleanup that I noticed in a crate that imported tendermintrs

The ed25519 crate dependency is unused as far as I can tell, and this is just used within ed25519_consensus. The point is a bit moot, since ed25519_consensus imports ed25519 as well, but one less version bump that needs to be maintained later.

* [ ] Wrote tests


xla:
@ValarDragon Looks like this change, broke a bunch of imports. Can you review again if it's truly not used?

ValarDragon:
Ah sorry! I forgot to check CI after trying this, my bad.

Issue #1134, opened 2022-05-26
generate privval gRPC server and provide a high-level pluggable default implementation


tomtau:
Version(s) of tendermint-rs: v0.24.0-pre.2
### Description

"privval" is an interface used by a Tendermint process (on validator nodes) to communicate with a signing backend (e.g. YubiHSM).
Previously, "privval" was over a custom socket protocol (implemented e.g. in [tmkms](https://github.com/iqlusioninc/tmkms)) that leveraged Unix domain sockets or Tendermint P2P over TCP where Tendermint acted as a server, and a signing backend connected to it as a client.
In Tendermint 0.35, a new method for "privval" was introduced: [the signer is a standard gRPC server/service](https://github.com/tendermint/tendermint/blob/master/proto/tendermint/privval/service.proto), and Tendermint connects to it as a client.
In Tendermint 0.36, we expect the old "privval" custom socket protocol will be removed and only the gRPC-based "privval" interface will be provided.

Given these circumstances, I think it makes sense for tendermint-rs to support this interface via the following features:

1. provide a "raw" service API definition by enabling `build_server(true)` in tonic_build, probably via a feature-flag (as not every usage of tendermint-proto needs it)
2. provide a high-level wrapper over the raw service (likely in a new "tendermint-validator" crate) which should do the following: 

- provide a basic gRPC server construction options for [tonic's Server](https://docs.rs/tonic/0.7.2/tonic/transport/struct.Server.html) (e.g. TLS)
- do basic validation of requests:
  - use the existing domain types
  - check chain ids against the configured ones
  - check the provided signer's state against the state in a request (see `update_consensus_state ` in tmkms https://github.com/iqlusioninc/tmkms/blob/main/src/chain/state.rs#L66 ) 
- return a cached public key or request a signature from a signing backend, and return corresponding responses
- provide a simple interface that can be implemented by different signer backends

Here's a sketch of the potential interface:

```rust
#[async_trait]
pub trait SignerProvider {
    async fn sign(&self, signable_bytes: &[u8]) -> Result<Signature, Error>;
    async fn load_pubkey(&self) -> Result<PublicKey, Error>;
    async fn load_state(&self) -> Result<consensus::State, Error>;
    async fn persist_state(&mut self, new_state: &consensus::State) -> Result<(), Error>;
}
```

[...]

PR #1147, opened 2022-06-27
tendermint: Added conversions from signing requests to consensus states

tomtau:
This is a sub-PR for [ADR-011](https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-011-validator-crate.md).
(ref: [#1134](https://github.com/informalsystems/tendermint-rs/issues/1134))

* [x] Wrote tests


tomtau:
@thanethomson I noticed the private key definition in the Tendermint-rs crate (tendermint/src/private_key.rs) didn't a `sign` function which would be useful, so I added it, but not sure whether there was any specific reason not to include it before.

tomtau:
@thanethomson I noticed there were these two deprecated methods for converting votes and proposals to consensus states:
https://github.com/informalsystems/tendermint-rs/blob/b23d626c9ca45dec8f5ca3e31bb493d810bafc5d/tendermint/src/proposal.rs#L101
https://github.com/informalsystems/tendermint-rs/blob/b23d626c9ca45dec8f5ca3e31bb493d810bafc5d/tendermint/src/vote.rs#L140
which appear to have a similar aim -- should I "undeprecate"  them and fix them?

tony-iqlusion (approved):
This seems useful

romac (changes requested):
This PR was still targeting the old `master` branch. I switched the base branch to `main` but now it needs a rebase.

PR #1152, opened 2022-06-30
Add a `tendermint-validator` crate

### LINKED ISSUE #1134: generate privval gRPC server and provide a high-level pluggable default implementation

tomtau:
fixes #1134 
* [x] Wrote tests

### DIFF
--- a/privval/src/server.rs
+++ b/privval/src/server.rs
@@ -64,26 +64,23 @@ fn check_state(
 
 impl<S: SignerProvider, VS: ValidatorStateProvider> PrivvalService<S, VS> {
     /// Creates a new server instance by loading all providers.
-    pub async fn new(
-        providers: BTreeMap<chain::Id, (S, VS)>,
-        config: BasicServerConfig,
-    ) -> Result<Self, async_signature::Error> {
+    pub async fn new(providers: BTreeMap<chain::Id, (S, VS)>, config: BasicServerConfig) -> Self {
         info!("creating a new KMS server");
         let mut pubkeys = BTreeMap::new();
         for (chain_id, (signer, _)) in providers.iter() {
-            let pubkey = signer.load_pubkey().await?;
-            let (address, pubkeyb64) = display_validator_info(&pubkey);
+            let pubkey = signer.verifying_key();
+            let (address, pubkeyb64) = display_validator_info(pubkey);
             info!("[{}] loaded a validator ID: {}", chain_id, address);
             info!("[{}] public key: {}", chain_id, pubkeyb64);
-            pubkeys.insert(chain_id.clone(), pubkey);
+            pubkeys.insert(chain_id.clone(), *pubkey);
         }
 
-        Ok(Self {
+        Self {
             providers: Mutex::new(providers),
             pubkeys,
             max_heights: BTreeMap::new(),
             config,
-        })
+        }
     }
 
     fn check_max_height(&self, chain_id: &chain::Id, new_height: block::Height) -> Result<(), ()> {
@@ -381,7 +378,7 @@ mod tests {
             (signer, state_provider),
         );
         let config = BasicServerConfig::new(None, GrpcSocket::Unix("/tmp/test.socket".into()));
-        PrivvalService::new(providers, config).await.unwrap()
+        PrivvalService::new(providers, config).await
     }
 
     #[tokio::test]
--- a/privval/src/signer.rs
+++ b/privval/src/signer.rs
@@ -1,27 +1,13 @@
 //! Signing-related interface and the sample software-only implementation.
 
-use async_signature::AsyncSigner;
+use async_signature::{AsyncKeypair, AsyncSigner};
 use ed25519_consensus::SigningKey;
 use rand_core::{CryptoRng, RngCore};
 use tendermint::{account, PrivateKey, PublicKey, Signature};
 
 /// The trait for different signing backends
 /// (HSMs, TEEs, software-only, remote services etc.)
-#[tonic::async_trait]
-pub trait SignerProvider: AsyncSigner<Signature> {
-    async fn load_pubkey(&self) -> Result<PublicKey, async_signature::Error>;
-}
-
-#[tonic::async_trait]
-impl<T> SignerProvider for T
-where
-    T: AsyncSigner<Signature>,
-    PublicKey: for<'a> From<&'a T>,
-{
-    async fn load_pubkey(&self) -> Result<PublicKey, async_signature::Error> {
-        Ok(self.into())
-    }
-}
+pub trait SignerProvider: AsyncKeypair<Signature, VerifyingKey = PublicKey> {}
 
 /// A helper function that will convert the validator public key
 /// into textual forms that are needed in different operational contexts
@@ -40,21 +26,42 @@ pub fn display_validator_info(pubkey: &PublicKey) -> (String, String) {
 /// The default software-only implementation of [`SignerProvider`].
 /// (Not recommended for production use, but it is useful for testing
 /// or in combination with additional isolation from the host system, e.g. TEE.)
-pub type SoftwareSigner = PrivateKey;
+pub struct SoftwareSigner(PrivateKey, PublicKey);
+
+impl AsRef<PublicKey> for SoftwareSigner {
+    fn as_ref(&self) -> &PublicKey {
+        &self.1
+    }
+}
+
+#[tonic::async_trait]
+impl AsyncSigner<Signature> for SoftwareSigner {
+    async fn sign_async(&self, msg: &[u8]) -> Result<Signature, async_signature::Error> {
+        self.0.sign_async(msg).await
+    }
+}
+
+impl AsyncKeypair<Signature> for SoftwareSigner {
+    type VerifyingKey = PublicKey;
+}
+
+impl SignerProvider for SoftwareSigner {}
 
 /// A helper to generate a new random private key.
 pub fn generate_ed25519<R: RngCore + CryptoRng>(rng: R) -> SoftwareSigner {
     let key = SigningKey::new(rng);
-    PrivateKey::Ed25519(key)
+    let secret = PrivateKey::Ed25519(key);
+    let public = secret.public_key();
+    SoftwareSigner(secret, public)
 }
 
 #[cfg(test)]
 mod test {
-    use async_signature::AsyncSigner;
+    use async_signature::{AsyncKeypair, AsyncSigner};
     use ed25519_consensus::SigningKey;
     use tendermint::PublicKey;
 
-    use crate::{generate_ed25519, SignerProvider};
+    use crate::generate_ed25519;
 
     #[test]
     pub fn test_display_validator_info() {
@@ -72,7 +79,7 @@ mod test {
         let signer = generate_ed25519(rng);
         let signable_bytes = b"test message";
         let signature = signer.sign_async(signable_bytes).await.expect("sign");
-        let pubkey = signer.load_pubkey().await.expect("pubkey");
+        let pubkey = signer.verifying_key();
         assert!(pubkey.verify(signable_bytes, &signature).is_ok());
     }
 }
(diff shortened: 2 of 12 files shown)


tomtau:
this PR may be a bit big, so will wait for https://github.com/informalsystems/tendermint-rs/pull/1147 to be merged + will split this one up (perhaps traits and server mods to be in two separate PRs)

tony-iqlusion:
A note on the `tendermint-validator` name: to me that name implies this code implements a validating full node, so it feels a little misleading.

Perhaps `tendermint-privval` would be a better name to indicate this code implements the server-side of a "privval" connection coming *from* a validator?

tendermint/src/private_key.rs
```diff
@@ -38,6 +38,13 @@ impl PrivateKey {
             PrivateKey::Ed25519(signing_key) => Some(signing_key),
         }
     }
+
+    /// Sign a message with this private key
+    pub fn sign(&self, signable_msg_bytes: &[u8]) -> Signature {
+        match self {
+            PrivateKey::Ed25519(signing_key) => signing_key.sign(signable_msg_bytes).into(),
+        }
+    }
```
tony-iqlusion, line 47: This might be a good application for the [`signature::Signer` trait](https://docs.rs/signature/latest/signature/trait.Signer.html), which would allow `tendermint::PrivateKey` to be used in places which require a generic parameter `S: Signer<Signature>`.

validator/src/server.rs
```diff
@@ -0,0 +1,463 @@
...
+
+use crate::{
+    config::{BasicServerConfig, GrpcSocket},
+    signer::{display_validator_info, SignerProvider},
+    state::ValidatorStateProvider,
+};
+
+/// Validator gRPC high-level server implementation
+#[derive(Debug)]
+pub struct KMSServer<S: SignerProvider, VS: ValidatorStateProvider> {
```
tony-iqlusion, line 32: Per the [Rust API naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html):

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn

I would suggest the following capitalization instead:

```suggestion
pub struct KmsServer<S: SignerProvider, VS: ValidatorStateProvider> {
```

validator/src/server.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// Signer and state providers for individual chains.
+    /// Given the traits require mutable access to the state provider
+    /// across awaits, we use tokio's mutex here.
+    providers: Mutex<HashMap<chain::Id, (S, VS)>>,
+    /// A cache of public keys loaded during the initialization.
+    pubkeys: HashMap<chain::Id, PublicKey>,
+    /// Optional setting for the maximum height,
+    /// after which the server will stop signing for a particular network.
+    max_heights: HashMap<chain::Id, block::Height>,
```
tony-iqlusion, line 41: These seem like cases that don't benefit from an unordered map.

Perhaps use a `BTreeMap` instead, which provides deterministic ordering?

validator/src/signer.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::error::Error;
+
+/// The trait for different signing backends
+/// (HSMs, TEEs, software-only, remote services etc.)
+#[tonic::async_trait]
+pub trait SignerProvider {
+    type E: std::error::Error;
+    async fn sign(&self, signable_bytes: &[u8]) -> Result<Signature, Self::E>;
+    async fn load_pubkey(&self) -> Result<PublicKey, Self::E>;
+}
```
tony-iqlusion, line 16: FWIW this is close to the trait offered by [`async_signature::Signer`](https://docs.rs/async-signature/latest/async_signature/trait.AsyncSigner.html).

validator/src/server.rs
tony-iqlusion: Also if the crate is renamed to `tendermint-privval`, this should probably be called `PrivvalService`

tomtau:
Thanks @tony-iqlusion ! Yes, `tendermint-privval` may be a better name.

As for the `SignerProvider` trait, I guess there are a few options:

1. use [`async_signature::Signer`](https://docs.rs/async-signature/latest/async_signature/trait.AsyncSigner.html) directly instead and require public keys as a separate explicit input argument to the privval service (this may be a bit more error-prone and one will find out any signer-related issues only after the first signing request arrives)
2. a new version of `async_signature` that will add public key retrieval functions and use that directly instead
3. change `SignerProvider` to take [`async_signature::Signer`](https://docs.rs/async-signature/latest/async_signature/trait.AsyncSigner.html) as a type parameter 

What do you think?

tony-iqlusion:
@tomtau for "public key retrieval", in `cosmrs` we use a helper trait with a blanket impl:

https://docs.rs/cosmrs/latest/cosmrs/crypto/secp256k1/trait.EcdsaSigner.html

Upstreaming something like this into `signature`/`async-signature` is something I've considered.

tomtau:
@tony-iqlusion that's the third option and it seems good: if it's upstreamed into `signature`/`async-signature` in the future, it shouldn't be difficult to change. So, I switched `SignerProvider` to be a helper trait.

@thanethomson this PR should be good to go (sans some questions on the other PR: https://github.com/informalsystems/tendermint-rs/pull/1147#issuecomment-1170799635 ) -- I could possibly add those `signature` trait implementations in #1147 if it's preferable to split it that way.

privval/src/server.rs
```diff
@@ -0,0 +1,467 @@
...
+        let req = SignVoteRequest::try_from(request.into_inner())
+            .map_err(|e| Status::invalid_argument(e.to_string()))?;
+        debug!(
+            "[{}] received a vote signing request: {:?}",
+            req.chain_id, req
+        );
+        self.check_max_height(&req.chain_id, req.vote.height)
+            .map_err(|_| Status::failed_precondition("max height exceeded"))?;
+        let mut providers = self.providers.lock().await;
+        let resp = if let Some((signer, state_provider)) = providers.get_mut(&req.chain_id) {
```
tony-iqlusion, line 240: Seems like this would be a bit more readable if you hoisted the error handling which is currently on [L269](https://github.com/informalsystems/tendermint-rs/pull/1152/files#diff-beb011520d9feeec8b734a23ccbae4e32ba3c5e635a23970bb620fbe20a3cd68R269) up here.

privval/src/server.rs
```diff
@@ -0,0 +1,467 @@
...
+        let req = SignProposalRequest::try_from(request.into_inner())
+            .map_err(|e| Status::invalid_argument(e.to_string()))?;
+        debug!(
+            "[{}] received a proposal signing request: {:?}",
+            req.chain_id, req
+        );
+        self.check_max_height(&req.chain_id, req.proposal.height)
+            .map_err(|_| Status::failed_precondition("max height exceeded"))?;
+        let mut providers = self.providers.lock().await;
+        let resp = if let Some((signer, state_provider)) = providers.get_mut(&req.chain_id) {
```
tony-iqlusion, line 290: Similar comment to [this one](https://github.com/informalsystems/tendermint-rs/pull/1152/files#r927788402) here as well

privval/src/signer.rs
```diff
@@ -0,0 +1,74 @@
...
+use ed25519_consensus::SigningKey;
+use rand_core::{CryptoRng, RngCore};
+use tendermint::{account, PrivateKey, PublicKey, Signature};
+
+/// The trait for different signing backends
+/// (HSMs, TEEs, software-only, remote services etc.)
+#[tonic::async_trait]
+pub trait SignerProvider: AsyncSigner<Signature> {
+    async fn load_pubkey(&self) -> Result<PublicKey, async_signature::Error>;
+}
```
tony-iqlusion, line 13: Have you considered a blanket impl similar to this one here?

https://docs.rs/cosmrs/latest/cosmrs/crypto/secp256k1/trait.EcdsaSigner.html#impl-EcdsaSigner

The nice thing about that is it would allow this trait to "just work" with types that impl the `Signer` trait, leveraging a blanket impl of `AsyncSigner` for `Signer`:

https://docs.rs/async-signature/latest/async_signature/trait.AsyncSigner.html#implementors

tony-iqlusion:
Did a quick pass and this is looking good now!

I'll try to more thoroughly review this next week.

tony-iqlusion:
@tomtau here's a PR to add a `Keypair` trait to the `signature` crate:

https://github.com/RustCrypto/traits/pull/1080

It could have a corresponding `AsyncKeypair` trait in the `async-signature` crate.

tony-iqlusion:
Added `AsyncKeypair`: https://docs.rs/async-signature/0.2.0/async_signature/trait.AsyncKeypair.html

tony-iqlusion (approved):

tony-iqlusion:
It seems this is blocked on #1225

romac (changes requested):
I switched the base branch from `master` to `main` so now this PR needs a rebase.

tomtau:
I'm not sure when I'll get to this PR. Are you looking into merging it soon or is anyone interested in taking over it?

romac:
No rush from our side, so feel free to come back to it whenever it suits you if you are still interested in getting this merged. And sorry for not being very reactive with this PR, we've had and still have limited capacity to work on tendermint-rs this past quarter.

PR #1157, opened 2022-07-21
clippy: Fix lints for Rust 1.62 (v0.23.x)

thanethomson:
Fixes some minor clippy lints.


tarcieri:
you might want to pin the rustc version CI uses for clippy to prevent unexpected breakages

thanethomson:
> you might want to pin the rustc version CI uses for clippy to prevent unexpected breakages

I may get lazy if I do that :slightly_smiling_face: I'll consider it though!

Issue #1169, opened 2022-07-28
Support for HTTP Basic authentication in HTTP and WebSocket RPC clients


romac:
Version(s) of tendermint-rs: 

### Summary

Add support for [HTTP Basic authentication](https://datatracker.ietf.org/doc/html/rfc7617) in HTTP and WebSocket RPC clients.

### Problem definition

On the [Hermes](https://github.com/informalsystems/ibc-rs), we have heard a request from some operators to allow them to connect to a full node protected by HTTP Basic authentication.

[We initially thought](https://github.com/informalsystems/ibc-rs/issues/2459) that simply embedding the username and password in the RPC/WebSocket URL (eg. `http://username:password@example.com:1234`) would work [but it does not](https://github.com/informalsystems/ibc-rs/issues/2459#issuecomment-1197823813).

It seems like [`hyper`](https://lib.rs/hyper) does not automatically parse and use the username/password embedded in the URL. Instead, we have to manually add an `Authorization` header to the request.

Same thing with ([`async-`](https://lib.rs/async-tungstenite))[`tungstenite`](https://lib.rs/tungstenite), we have to manually set the `Authorization` header when building the initial HTTP request that upgrades the connection to WebSocket.

### Proposal

Implement support for HTTP Basic authentication for the HTTP and WebSocket clients in `tendermint-rpc`.

### Acceptance criteria

1. Both the HTTP and WebSocket clients can connect to a node protected by HTTP Basic authentication.
2. Hermes can connect to a node protected by HTTP Basic authentication. See [how to reproduce this](https://github.com/informalsystems/ibc-rs/issues/2459#issuecomment-1197823813) on the linked issue.

thanethomson:
Closed by #1170

PR #1170, opened 2022-07-28
v0.23.x: Add support for HTTP Basic authentication to HTTP and WebSocket RPC clients

romac:
Closes: #1169

---

* [x] Wrote tests

### DIFF
--- a/rpc/src/rpc_url.rs
+++ b/rpc/src/rpc_url.rs
@@ -51,8 +51,6 @@ impl FromStr for Scheme {
 pub struct Url {
     inner: url::Url,
     scheme: Scheme,
-    username: Option<String>,
-    password: Option<String>,
     host: String,
     port: u16,
 }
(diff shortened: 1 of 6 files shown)


rpc/src/rpc_url.rs
```diff
@@ -98,15 +99,22 @@ impl Url {
...
     /// Get the password associated with this URL, if any.
     pub fn password(&self) -> Option<&str> {
         self.inner.password()
     }
 
+    /// Get the authority associated with this URL, if any.
+    /// The authority is the username and password separated by a colon.
+    pub fn authority(&self) -> Option<String> {
+        self.username()
+            .map(|user| format!("{user}:{}", self.password().unwrap_or_default()))
```
thanethomson, line 115: Seems like Rust nightly doesn't like this?

rpc/src/rpc_url.rs
```diff
@@ -179,7 +187,7 @@ mod test {
                     host: "127.0.0.1".to_string(),
```
thanethomson, line 187: Could you please add a couple more test cases to `SUPPORTED_URLS` here? e.g. one with just a username, one with username and password.

thanethomson (commented):
On first pass this mostly looks good to me! Just a couple of minor things.

rpc/src/rpc_url.rs
romac: Fixed

romac:
Should be good to go :)

thanethomson (approved):
Excellent, thanks @romac!

PR #1193, opened 2022-09-20
Side-by-side support for Tendermint 0.34 and 0.37

mzabaluev:
ADR: https://github.com/informalsystems/tendermint-rs/pull/1274

A cross-chain bridge agent, such as a relayer, may need to communicate with nodes using different Tendermint versions. To enable this use with a single linked instance of `tendermint-rpc` ~/tendermint-abci~ crate, and also to disentangle tendermint-rs versioning from the versions of the Tendermint Core protocol, change `tendermint-proto` to have generated protobuf structs for all supported tendermint versions in side-by-side modules, and adapt the higher level client APIs to be able to select the version expected from the node.

* [x] Wrote tests


mzabaluev:
in my understanding, an MVP of multi-version tendermint-rs may need the following changes:

- Adapt the `tendermint-abci` client to be able to talk both versions.
  - It is not used by Hermes, are there other customers for this?
  - The `Application`/ `Server` part can just serve the latest supported version; I see no use case for multi-version here.
- Add support for the RPC endpoints new in 0.37 to tendermint-rpc and conditionally disable them for the 0.34 mode, or just rely on the node to respond with an error.

hdevalence:
I don't think it's currently possible to implement support for both 0.34 and 0.37 on `main`, because `main` does not even have support for 0.34 -- ABCI support only exists for 0.35. Before trying to add support for 0.37, the `main` branch should have support for 0.34.  Merging something like this now will generate even more massive conflicts.

mzabaluev:
I think it should be possible to have limited legacy support for 0.34 in tendermint-rs, by generating the `v0_34` structs in `tendermint-proto`, implementing domain type conversions for these, and adapting `tendermint-rpc` to be able to interop with 0.34 nodes (which should be pretty much there due to backward compatibility). As we discussed with @thanethomson, we are not interested in multi-version adaptations for `tendermint-abci` at the moment.

xla:
@mzabaluev What's left to be done to get this changeset completed?

mzabaluev:
I have updated the PR.

The last significant change I wanted to try out was version-specific `Client` traits in `tendermint-rpc`, and it seems to have landed well. I think this is good to go now. Please review the changes as I'm trying to summarize them (briefly) in a changelog entry and describe them in more detail in an upcoming ADR.

aakoshh:
Hi! I want to use the new ABCI++ features in `v0.37.0-rc2`; this PR is just perfectly timed. 

I tried to run the [abci example](https://github.com/informalsystems/tendermint-rs/tree/mikhail/multi-tc-version-support/abci) with a Tendermint I built from the source but ran into some errors immediately after starting `tendermint`.

This is the error from the ABCI process:

```console
2023-01-11T10:30:13.777266Z ERROR tendermint_abci::server: Failed to read incoming request from client 127.0.0.1:35778: error encoding protocol buffer

Caused by:
    failed to decode Protobuf message: Request.value: buffer underflow
```

Maybe it's has nothing to do with this PR, I haven't tried with the stable versions yet.

aakoshh:
The error [above](https://github.com/informalsystems/tendermint-rs/pull/1193#issuecomment-1378558778) comes from [decode_varint](https://github.com/informalsystems/tendermint-rs/blob/mikhail/multi-tc-version-support/abci/src/codec.rs#L168-L175). According to the comments it's probably because https://github.com/tendermint/tendermint/issues/5783 has landed. 

I see that you striked out updating the `tendermint_abci` "client" to support both versions. I'm not sure if it also means not updating the server to be talked to by both versions.

abci/src/application.rs
```diff
@@ -98,6 +98,14 @@ pub trait Application: Send + Clone + 'static {
...
         Default::default()
     }
+
+    fn prepare_proposal(&self, _request: RequestPrepareProposal) -> ResponsePrepareProposal {
+        Default::default()
+    }
+
+    fn process_proposal(&self, _request: RequestProcessProposal) -> ResponseProcessProposal {
+        Default::default()
+    }
```
aakoshh, line 108: Just to note that the defaults will probably not work (they will propose empty blocks and think [reject](https://github.com/informalsystems/tendermint-rs/pull/1193/files#diff-f98b59e5344ef9d1c053ec7d1dd516b0627d1e390a9ba8813ab004c70b340c9fR15) all proposals). The spec has [instructions](https://github.com/tendermint/tendermint/blob/v0.37.0-rc2/spec/abci/abci%2B%2B_tmint_expected_behavior.md#adapting-existing-applications-that-use-abci) for adaptation.

tendermint/src/v0_37/abci/request.rs
```diff
@@ -0,0 +1,223 @@
...
+impl TryFrom<Request> for ConsensusRequest {
+    type Error = Error;
+    fn try_from(req: Request) -> Result<Self, Self::Error> {
+        match req {
+            Request::InitChain(x) => Ok(Self::InitChain(x)),
+            Request::BeginBlock(x) => Ok(Self::BeginBlock(x)),
+            Request::DeliverTx(x) => Ok(Self::DeliverTx(x)),
+            Request::EndBlock(x) => Ok(Self::EndBlock(x)),
+            Request::Commit => Ok(Self::Commit),
+            _ => Err(Error::invalid_abci_request_type()),
```
aakoshh, line 97: The new `PrepareProposal` and `ProcessProposal` variants are missing from `ConsensusRequest`.

abci/src/application.rs
mzabaluev: Right; sorry about this, `tendermint-abci` has not been my top priority for updating to 0.37. Will fix.

tendermint/src/v0_37/abci/request.rs
aakoshh: Here are the ones I found: https://github.com/informalsystems/tendermint-rs/compare/mikhail/multi-tc-version-support...aakoshh:tendermint-rs:mikhail/multi-tc-version-support

mzabaluev: Thank you!

abci/src/application.rs
mzabaluev: As I understand it, this whole default implementation only provides the most trivial default responses.
It's up to the application to fill the methods in where necessary.

mzabaluev:
> The error [above](https://github.com/informalsystems/tendermint-rs/pull/1193#issuecomment-1378558778) comes from [decode_varint](https://github.com/informalsystems/tendermint-rs/blob/mikhail/multi-tc-version-support/abci/src/codec.rs#L168-L175). According to the comments it's possibly because [tendermint/tendermint#5783](https://github.com/tendermint/tendermint/issues/5783) has landed. I see the the new variant was already part of the codebase [once before](https://github.com/informalsystems/tendermint-rs/pull/1022/files#diff-06e2ec8e521bf5e8f9dba0ee4544a2bad5dab15d3915f10ee21fa7ef83b3f13c).
> 
> I see that you striked out updating the `tendermint_abci` "client" to support both versions. I'm not sure if it also means not updating the server to be talked to by both versions.

Would it be acceptable to only provide the latest version implementation on the server side, but optionally support the legacy wire protocol on the client?

abci/src/application.rs
aakoshh: Yes, I agree; what I'm trying to say is that by leaving it on defaults, filling in these methods will always be necessary. 

I can't vouch for all the other cases, but I would expect for those methods the default values represent an acceptable behaviour, like `check_tx` not finding any error, or `end_block` not changing the power table. 

But with these two new methods, the defaults are not accepted by Tendermint:
* If we return default from `prepare_proposal`, that effectively will remove all transactions Tendermint wanted to propose, resulting in an empty block instead. 
* If then we return `ProcessProposal::Unknown` (which is the default, although it could be `Accept`) instead of `ProcessProposal::Accept`, we get the following error: `CONSENSUS FAILURE!!!                         module=consensus err="ProcessProposal responded with status UNKNOWN"`. That is because Tendermint now considers us insane: we just asked it to propose a block we ourselves do not accept.

So, if we want to have a transition to v0.37 where users of the `Application` trait don't have to override unless they have something to change from the default behaviour, and they expect Tendermint to just work, then the default method implementations have a bit more work to do than usual.

(NB the specs explicitly say that TC _may_ pass more transactions in `PrepareProposal` than what fits into the maximum size; otherwise the implementation would be a trivial copy).

aakoshh:
> Would it be acceptable to only provide the latest version implementation on the server side, but optionally support the legacy wire protocol on the client?

This must be a question to your fellow maintainers, but here's my 2 cents:

As I understand the client is mostly for testing, because it directly talks to the ABCI interface, while normally only TC does that. So I would assume it only has to support what the server does, there's no need for optional support for alternative formats.

If the server only supports the new format, then developers who don't want to or can't yet move to TC 0.34 _and_ they use the ABCI server have to refrain from updating to the `tendermint-rs`, or they would not be able to run their blockchain any more. 

From what I gathered, the multi-version support is primarily meant for relayers who need to talk to different versions of TC at the same time, but they would call the RPC interface rather than the ABCI directly, so for them it doesn't matter.

I don't know what's the most tasteful solution (a feature flag?), but a similar question will arise in `tower-abci` as well.

mzabaluev:
I have changed the encoding to unsigned varint in https://github.com/informalsystems/tendermint-rs/pull/1193/commits/620aa99114d2495d827b292cb43ba69c07613857
This corresponds to the earlier update to the 0.37 protobuf, without alternatives, in `tendermint-abci`.
Please comment if alternative approaches are needed; I will also add this to the upcoming ADR describing these changes.

romac:
It would be great if we could add some integration tests against 0.34 and 0.37, for the `tendermint-rpc` crate on top of the existing fixture-based tests we have already in `tendermint-rpc/tests`.

What do you think @mzabaluev @thanethomson?

mzabaluev:
We should, at minimum, have fixtures for the new requests in 0.37, and test that `v0_37::Client` interoperates with these.
Old fixtures can be tested against both `Client` trait implementations, but the extra code that will exercise is pretty trivial.

mzabaluev:
@thanethomson What would it take to re-acquire fixtures from Gaia or simapp based on tendermint 0.37 with rpc-probe, as was done in e562d4d7e74710f9bd6f80330a5f1ce9bb9d2ed6?

thanethomson:
> What would it take to re-acquire fixtures from Gaia or simapp based on tendermint 0.37 with rpc-probe

You'd need a version of Gaia integrated with the latest v0.47.x release of the Cosmos SDK (https://github.com/cosmos/cosmos-sdk/releases/tag/v0.47.0-rc1). I'd recommend speaking with @mpoke about this.

ljoss17:
Blocking https://github.com/informalsystems/hermes/issues/3002

rpc/src/endpoint/tx_search.rs
```diff
@@ -36,21 +36,23 @@ impl Request {
...
+impl<S: Dialect> crate::Request<S> for Request {
+    type Response = Response<S::Event>;
+}
+
+impl<S: Dialect> crate::SimpleRequest<S> for Request {}
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct Response {
-    pub txs: Vec<tx::Response>,
+pub struct Response<Ev> {
```
mzabaluev, line 52: This is annoying to adapt to in client code, while the data type the client has to deal with does not really have to be different: the only problem we're solving with the generics is differences in serialization. I'm inclined to get this back to `Response` using domain types, with `DialectResponse<Ev>` alongside for JSON deserialization. This, however, will require changes to the `Request` trait.

romac:
Left some initial comments after a quick skim for us to discuss during the sync review. Will do another, deeper pass over the PR before the sync review.

abci/src/codec.rs
```diff
@@ -164,14 +164,3 @@ where
         Ok(Some(res))
     }
 }
-
-// encode_varint and decode_varint will be removed once
-// https://github.com/tendermint/tendermint/issues/5783 lands in Tendermint.
```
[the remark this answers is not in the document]
hdevalence: No, it's a breaking change from 0.34. It was originally supposed to be included in 0.35. I don't know if it's slated for inclusion in 0.37.

proto/src/serializers.rs
```diff
@@ -52,8 +52,10 @@
 
 // Todo: remove dead_code allowance as soon as more types are implemented
 #![allow(dead_code)]
+
+pub mod allow_null;
 pub mod bytes;
-pub mod evidence;
+mod evidence;
```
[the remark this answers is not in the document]
mzabaluev: No detailed memory of this now, but these serializers should only used internally by the serde implementations derived in the crate. The contents of the module have changed in breaking ways, too, so I think it's better to hide it.

abci/src/codec.rs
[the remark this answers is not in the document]
mzabaluev: The encoding has changed in 0.37, and in this PR, `tendermint-abci` is converted completely to 0.37 with no compatibility support. We can revert all of this and declare `tendermint-abci` at end of life if that's better.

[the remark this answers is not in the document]
mzabaluev: We could also choose to commit to multi-version support here as well, but this will require more work.

abci/src/application.rs
romac: @aakoshh Would that work for you or would you rather leave out the default impls?

proto/src/serializers.rs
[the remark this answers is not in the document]
romac: For the sake of consistency, let's perhaps make them all public here but make the `serializers` module private.

abci/src/application.rs
aakoshh: Thanks for asking @romac, my vote goes for having a default implementation that copies transactions and accepts proposals, unless overridden, like @mzabaluev said. This is so people don't have to come up with boilerplate to do this every time they implement the trait, similar to how they don't have to override `check_tx` unless they want to actually perform some logic. 

For example I forked `tower-abci` to point it at this branch of `tendermint-rs` and had to implement this logic [for the kvstore example](https://github.com/consensus-shipyard/tower-abci/pull/1/files#diff-3064912e722a6c3a27e57368a923034ba4dff6408ddfc4923ef4e00ed68b38bc), where it seems a bit accidental among all the other `Default::default()`s. After that I took inspiration from the `Application` trait in `tendermint-rs` and provided an [async version](https://github.com/consensus-shipyard/fendermint/blob/master/fendermint/abci/src/application.rs) which takes care of these defaults, so I think it's the sensible thing to do.

mzabaluev: Thank you @aakoshh for your feedback.
The more elaborate (but still very naive) default behaviours have now been implemented in https://github.com/informalsystems/tendermint-rs/pull/1193/commits/7e79039558b50fcdd3ca47512b4e58ed6624132f

rpc/src/client/transport/http.rs
```diff
@@ -39,6 +48,14 @@ use crate::{client::Client, prelude::*, Error, Scheme, SimpleRequest, Url};
...
     inner: sealed::HttpClient,
+    compat: CompatMode,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct HttpConfig {
+    /// Compatibility mode for Tendermint RPC protocol.
+    pub compat: CompatMode,
+    pub proxy_url: Option<HttpClientUrl>,
 }
```
mzabaluev, line 59: This is not the best design for maintainability, and perhaps a better solution would be a builder pattern for HttpClient.

rpc/src/client/transport/websocket.rs
```diff
@@ -52,8 +56,13 @@ const RECV_TIMEOUT: Duration = Duration::from_secs(RECV_TIMEOUT_SECONDS);
...
 
-/// Low-level WebSocket configuration
-pub use async_tungstenite::tungstenite::protocol::WebSocketConfig;
+#[derive(Default, Debug)]
+/// WebSocket client configuration
+pub struct WebSocketConfig {
+    pub compat: CompatMode,
+    /// Low-level protocol configuration
+    pub transport: Option<async_tungstenite::tungstenite::protocol::WebSocketConfig>,
+}
```
mzabaluev, line 65: Also a builder would be better here, instead of introducing another struct with public members.

tools/proto-compiler/src/constants.rs
```diff
@@ -2,11 +2,30 @@
...
+
+/// All Tendermint versions to generate code for
+pub const TENDERMINT_VERSIONS: &[TendermintVersion] = &[
+    TendermintVersion {
+        ident: "v0_34",
+        commitish: "v0.34.24",
+    },
+    TendermintVersion {
+        ident: "v0_37",
+        commitish: "v0.37.0-alpha.1",
```
mzabaluev, line 26: This is the last tag in Tendermint Core where the proto files had bundled Google well known types, before switching to buf to refer to dependencies. Adapting to buf will require more changes to the proto-compiler tool, and I have already done this already once in an older branch.

Here I'm leaving the migration to buf, and switching to CometBFT, as follow-on tasks.
There is no difference in protobuf message formats with the latest state in CometBFT branch `v0.37.x` (as per commit d26bdfe1d424ff24c9676fee5bef9a180870315e).

rpc/src/client/compat.rs
```diff
@@ -0,0 +1,81 @@
...
+use tendermint::Version;
+
+use crate::prelude::*;
+use crate::Error;
+
+/// Protocol compatibility mode for a Tendermint RPC client.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum CompatMode {
+    /// Use the latest version of the protocol (v0.37 in this release).
+    Latest,
```
romac, line 12: What's the reasoning behind using `Latest` here instead of `V0_37`? 

Having the datatype variant change meaning over releases sounds brittle to me, and it would be more explicit to have it denote the actual release that is supported.

If there are no good reasons against it, how about we rename this to `V0_37` and add a `CompatMode::latest()` method that returns the latest compat mode. Let's also change the `Default` instance to return `CompatMode::latest()`.

mzabaluev: Good idea. There may be a case for the `Latest` variant alongside both explicitly selective versions, but perhaps this will produce complications on matching. So let's do "latest" as an associated const or fn.

mzabaluev: Redone in https://github.com/informalsystems/tendermint-rs/pull/1193/commits/f3e161e02b4bc42a9c60a369a4b0b7ecd779da99

rpc/src/client/transport/http.rs
mzabaluev: Redone in https://github.com/informalsystems/tendermint-rs/pull/1193/commits/84d9cad8c8e41e9e0343633a06ae8288101b0ad3

rpc/src/client/transport/websocket.rs
mzabaluev: Redone in https://github.com/informalsystems/tendermint-rs/pull/1193/commits/84d9cad8c8e41e9e0343633a06ae8288101b0ad3

rpc/src/client/bin/main.rs
```diff
@@ -488,6 +492,7 @@ async fn recv_events(mut subs: Subscription, max_events: Option<u32>) -> Result<
     let mut event_count = 0u64;
     while let Some(result) = subs.next().await {
         let event = result?;
+        let event: DialectEvent<<LatestDialect as Dialect>::Event> = event.into();
```
romac, line 495: Shall we discover the compat and then use the appropriate dialect here?
I don't who is or has ever used this rpc binary, but if someone is they may be surprised if they try running it against a 0.34 node. Or maybe it'll just work since one can decode a 0.37 event using the 0.34 schema (but not the other way around).

rpc/src/client/transport/http.rs
```diff
@@ -39,6 +48,63 @@ use crate::{client::Client, prelude::*, Error, Scheme, SimpleRequest, Url};
 #[derive(Debug, Clone)]
 pub struct HttpClient {
     inner: sealed::HttpClient,
+    compat: CompatMode,
+}
+
+/// The builder pattern constructor for [`HttpClient`].
```
romac, line 54: Very nice!

romac (approved):
Amazing work there @mzabaluev! I think we can finally get this in :)

I have one question left, but I don't believe it should block this PR from being merged.

rpc/src/client/transport/http.rs
```diff
@@ -73,33 +73,30 @@ impl Builder {
...
-        match &self.proxy_url {
+    pub fn build(self) -> Result<HttpClient, Error> {
+        match self.proxy_url {
             None => Ok(HttpClient {
                 inner: if self.url.0.is_secure() {
-                    sealed::HttpClient::new_https(self.url.clone().try_into()?)
+                    sealed::HttpClient::new_https(self.url.try_into()?)
                 } else {
-                    sealed::HttpClient::new_http(self.url.clone().try_into()?)
+                    sealed::HttpClient::new_http(self.url.try_into()?)
```
mzabaluev, line 88: Hm, OK. I did not think it worthwhile to change to the less usable self-consuming builder, just to save a few clone operations on small data structures.
Or is there a bigger reason?

rpc/src/client/bin/main.rs
mzabaluev: This is only used to display the events by converting them back into JSON. I think it's actually better to show events in plain text even if the original JSON is from 0.34.

That said, this binary does not do version discovery at all now, so the clients would always decode as 0.37. This we shall change.

mzabaluev: Added version discovery for the CLI in https://github.com/informalsystems/tendermint-rs/pull/1193/commits/2d4a3336458f3d0cea6ae90928ab3f2aabdfb4ef

rpc/src/client/transport/http.rs
romac: The `&mut pattern` does work very well for our use in the event monitor, and since that's the main consumer of the library I figured we could switch to a self-consuming one. It's slightly less flexible in the sense that one may have to re-assign it to variable when setting things conditionally but otherwise it makes for a more fluent API in our current use case. The lack of cloning is just a side benefit from consuming the builder at the end, as most builders do.

romac (approved):

PR #1203, opened 2022-09-23
Rebase ABCI domain types onto main

mzabaluev:
Builds on #1185 to revert some data type changes that targeted Tendermint RPC 0.35.


.changelog/unreleased/breaking-changes/1030-remove-chrono.md
```diff
@@ -0,0 +1,11 @@
...
+  ([#1030](https://github.com/informalsystems/tendermint-rs/issues/1030)):
+  * The struct content is made private.
+  * The range of acceptable values is restricted to years 1-9999
+    (as reckoned in UTC).
+  * Removed conversions from/to `chrono::DateTime<chrono::Utc>`.
+  * Changes in error variants: removed `TimestampOverflow`, replaced with
+    `TimestampNanosOutOfRange`; removed `ChronoParse`, replaced with `TimeParse`.
+- `[tendermint-rpc]` Use `OffsetDateTime` and `Date` types provided by the `time` crate
+  in query operands instead of their `chrono` counterparts.
+  ([#1030](https://github.com/informalsystems/tendermint-rs/issues/1030))
```
thanethomson, line 11: This file should probably be deleted, since it was merged in v0.23.2.

.changelog/unreleased/improvements/1030-new-time-api.md
```diff
@@ -0,0 +1,11 @@
...
+  ([#1030](https://github.com/informalsystems/tendermint-rs/issues/1030))
+- `[tendermint]` Improve `tendermint::Time`
+  ([#1030](https://github.com/informalsystems/tendermint-rs/issues/1030)):
+  * Restrict the validity range of `Time` to dates with years in the range
+    1-9999, to match the specification of protobuf message `Timestamp`.
+    Add an `ErrorDetail` variant `DateOutOfRange` to report when this
+    restriction is not met.
+  * Added a conversion to, and a fallible conversion from,
+    `OffsetDateTime` of the `time` crate.
+  * Added `Time` methods `checked_add` and `checked_sub`.
```
thanethomson, line 11: This file should also be deleted.

thanethomson (commented):
@mzabaluev did you perhaps rerun the proto compiler tool to generate these protos?

mzabaluev:
> @mzabaluev did you perhaps rerun the proto compiler tool to generate these protos?

I tried that, the differences were superficial.

.changelog/unreleased/breaking-changes/1030-remove-chrono.md
mzabaluev: Done

thanethomson:
Since these are going to be breaking changes, I assume we'll be targeting a v0.26.0 release with them, right? I'd recommend capturing some more detailed changelog entries for where there are breakages, e.g. the transition from `Vec<u8>` to `Bytes` in the ABCI protos.

Also don't worry about the nightly coverage breakage - it seems codecov's having an outage of some kind.

mzabaluev:
> the transition from `Vec<u8>` to `Bytes` in the ABCI protos.

I'd actually remove this change altogether and maybe put it up for an ADR.
However, @hdevalence said in https://github.com/informalsystems/tendermint-rs/pull/1185#discussion_r976645069 that existing users of ABCI already use it. I presume it was only true for the users of 0.24?

.changelog/unreleased/breaking-changes/1203-bytes-for-abci-proto.md
```diff
@@ -0,0 +1,2 @@
+- `[tendermint-proto]` Use `Bytes` for byte array fields of ABCI protobuf types.
+  ([#1203](https://github.com/informalsystems/tendermint-rs/pull/1203))
```
mzabaluev, line 2: @thanethomson I've added a notice about this change, but as stated already, I'm not sure we need to go ahead with it for 0.26. This is certainly unrelated to domain types, applied piecemeal to a single proto module not as a consistent change throughout tendermint-proto, and has the overall smell of a premature optimization.

mzabaluev:
> 'd recommend capturing some more detailed changelog entries for where there are breakages, e.g. the transition from Vec<u8> to Bytes in the ABCI protos.

Besides this (subject to discussion above) and the ABCI domain types rework that is the subject of this PR, I have removed any other breaking changes.

.changelog/unreleased/breaking-changes/1203-bytes-for-abci-proto.md
hdevalence: This is related to the domain types, because the ABCI domain types use `Bytes` internally, so unwinding this part of the ABCI domain types will break existing users of the ABCI domain types. This is not meaningfully a breaking change for anyone else, because there weren't users of the ABCI proto types other than the stub ABCI demo application.  It's also not a breaking change relative to `master`, because this code has been merged on that branch (albeit unreleased) for about a year.

If this change is unwound out, you'll have to edit all of the ABCI domain types, and then all the users will have to edit all their uses of those domain types. I don't think it would be good to put this on the critical path for doing a semver release of the `tendermint` crates that contains ABCI domain types.

If the decision, later, is that the domain types should **not** use the `Bytes` type, it would be much better to do that as a later change that can be modeled with semver.  (For the record, I don't think this is a good decision, but in any case I think it should be separate from getting this code backported as-is).

mzabaluev: Thank you @hdevalence for the detailed exposition.

As a late-to-the-scene comment, if I were into micro-optimizing domain type conversions to protobuf, I'd start with the `Protobuf` trait forcing a clone of the entire structure, because the domain-to-raw conversion is implemented via `From<Self>`, but the `encode*` methods take `&self`.

abci/src/codec.rs
```diff
@@ -146,6 +146,7 @@ where
         Ok(len) => len,
         // We've potentially only received a partial length delimiter
         Err(_) if src_len <= MAX_VARINT_LENGTH => return Ok(None),
+        // Err(e) => return Err(Error::decode(e)),
```
thanethomson, line 149: For interest's sake, what's this line for?

thanethomson (approved):

abci/src/client.rs
```diff
@@ -113,11 +113,6 @@ impl Client {
         perform!(self, Commit, RequestCommit {})
     }
 
-    /// Request that the application set an option to a particular value.
-    pub fn set_option(&mut self, req: RequestSetOption) -> Result<ResponseSetOption, Error> {
-        perform!(self, SetOption, req)
-    }
-
```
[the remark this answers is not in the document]
hdevalence: I think this slipped through backporting. We could add it back in, or leave it out.  There's no documentation on what it's actually supposed to do, so I'm not sure how anyone would use it.

abci/src/codec.rs
hdevalence: Don't know, it should probably be removed!

abci/src/client.rs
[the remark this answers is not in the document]
mzabaluev: I have restored the method, but added a `[deprecated]` annotation to it.

abci/src/client.rs
```diff
@@ -114,6 +114,10 @@ impl Client {
     }
 
     /// Request that the application set an option to a particular value.
+    ///
+    /// This request lacks specification and should not be used.
+    /// It will be removed in ABCI 0.37.
+    #[deprecated]
```
thanethomson, line 120:
```suggestion
    /// It will be removed in Tendermint Core v0.37.
    #[deprecated(note = "The set_option ABCI method will be removed in Tendermint Core v0.37")]
```

thanethomson (approved):

Issue #1090, opened 2022-02-11
rpc: Remove ABCI-related types


thanethomson:
Version(s) of tendermint-rs: `master`

### Description

Follows from #1022, which introduced domain types into the `tendermint` crate for ABCI. The RPC crate used to have ABCI-specific data structures to use in its responses, but these have been superseded by the ABCI domain types in the `tendermint` crate.

[These old ABCI RPC data structures](https://github.com/informalsystems/tendermint-rs/blob/904862c7c2db6a2b7a926af2ac8e3d7428f8779f/rpc/src/abci.rs) must be removed.

### Definition of "done"

When all of the [old ABCI RPC data structures](https://github.com/informalsystems/tendermint-rs/blob/904862c7c2db6a2b7a926af2ac8e3d7428f8779f/rpc/src/abci.rs) have been removed.

mzabaluev:
Some questions while working on #1204:

* In the [response struct](https://github.com/informalsystems/tendermint-rs/pull/1204/files#diff-8b372f30a4017d2a935c3db771835904c928aa48fced8fd30e55c8890581c8d2R43) for `/broadcast_tx_commit`, do we care to change the `check_tx` and `deliver_tx` method to the corresponding domain types from the `tendermint` crate? Serde for `TxResult` reflects some quirks that may be only relevant for this endpoint, and also features the `mempool_error` field which is consciously omitted from the `abci::response::CheckTx` comain type.

* The `Code` enum is more useful in idiomatic Rust

thanethomson:
> In the [response struct](https://github.com/informalsystems/tendermint-rs/pull/1204/files#diff-8b372f30a4017d2a935c3db771835904c928aa48fced8fd30e55c8890581c8d2R43) for `/broadcast_tx_commit`, do we care to change the `check_tx` and `deliver_tx` fields to the corresponding domain types from the `tendermint` crate? Serde for `TxResult` reflects some quirks that may be only relevant for this endpoint, and also features the `mempool_error` field which is consciously omitted from the `abci::response::CheckTx` domain type.

The `mempool_error` field actually should be in `abci::response::CheckTx`. It's relevant to Tendermint v0.34.21+.

I'd say we should use the domain types as far as practically possible.

> The `Code` enum is more useful in idiomatic Rust than an integer. It explains itself in match contexts, it enforces the convention, and even has the same size as an `u32`. Would it make sense to promote it into the idiomatic types in `tendermint`?

[...]

PR #1092, opened 2022-02-11
rpc: Replace RPC ABCI types with ABCI domain types 

thanethomson:
Closes #1090 and #947

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Added entry in `.changelog/`

Issue #947, opened 2021-08-09
Consolidate `hash::Hash` and `abci::transaction::Hash` in the `tendermint` crate?


thanethomson:
At present, it seems to me like we could consolidate the [`hash::Hash`](https://github.com/informalsystems/tendermint-rs/blob/877e408a5692dbb13b08da6d0c79193a87181db5/tendermint/src/hash.rs#L26) and [`abci::transaction::Hash`](https://github.com/informalsystems/tendermint-rs/blob/877e408a5692dbb13b08da6d0c79193a87181db5/tendermint/src/abci/transaction/hash.rs#L17) structs. The latter also seems quite specific to SHA256.

We could easily:

1. Create a new `Sha256` struct in the `hash` module to represent a SHA256 hash.
2. Implement all necessary functionality that was on `abci::transaction::Hash` on the `Sha256` struct.
3. Replace the [variant tuple type](https://github.com/informalsystems/tendermint-rs/blob/877e408a5692dbb13b08da6d0c79193a87181db5/tendermint/src/hash.rs#L28) with `Sha256`.
4. Use either the `Hash` struct throughout the codebase, or the `Sha256` struct in places where we know for sure that we only support/expect a SHA256 hash.

**What's the definition of "done" for this issue?**

Once we've:

- [ ] Decided on whether consolidation makes sense.
- [ ] Implemented 1 through 4 above.

PR #1204, opened 2022-09-29
Replace RPC ABCI types with ABCI domain types

mzabaluev:
Closes https://github.com/informalsystems/tendermint-rs/issues/1090 and https://github.com/informalsystems/tendermint-rs/issues/947

Picked up from #1092 by rebasing the branch onto `main`.

* [ ] Wrote tests


tendermint/src/validator.rs
```diff
@@ -318,7 +319,11 @@ impl ProposerPriority {
     }
 }
 
-/// Updates to the validator set
+/// A change to the validator set.
+///
+/// Used to inform Tendermint of changes to the validator set.
+///
+/// [ABCI documentation](https://docs.tendermint.com/master/spec/abci/abci.html#validatorupdate)
```
thanethomson, line 326:
```suggestion
/// [ABCI documentation](https://github.com/tendermint/tendermint/blob/v0.34.x/spec/abci/abci.md#validatorupdate)
```

We currently do not make the spec available via the docs site, and the one from `master` is no longer relevant. We also keep the specification relevant to the specific Tendermint release series on each release branch, so it's easier to see diffs in the specs between major releases of Tendermint.

thanethomson: I think we can probably keep this as-is for this PR, but in a follow-up PR we should update all the spec references to rather point to the `v0.34.x` branch of the Tendermint repo. I've created #1224 to track this.

thanethomson (approved):
Thanks for disentangling all this @mzabaluev!

PR #1205, opened 2022-09-30
support parsing `unix abstract socket`

ktmlm:
tendermint core can accept `unix abstract socket`, but this crate can NOT parse it.

this pr fix this issue.

ref: https://golang-examples.tumblr.com/post/92025745979/pitfall-of-abstract-unix-domain-socket-address-in-go

### DIFF
--- a/config/src/net.rs
+++ b/config/src/net.rs
@@ -87,6 +87,13 @@ impl FromStr for Address {
     type Err = Error;
 
     fn from_str(addr: &str) -> Result<Self, Error> {
+        // unix abstract socket address, `man 7 unix` for system descriptions
+        if addr.starts_with("unix://@") {
+            return Ok(Self::Unix {
+                path: addr.trim_start_matches("unix://").to_owned(),
+            });
+        }
+
         let prefixed_addr = if addr.contains("://") {
             addr.to_owned()
         } else {


config/src/net.rs
```diff
@@ -87,6 +87,13 @@ impl FromStr for Address {
     type Err = Error;
 
     fn from_str(addr: &str) -> Result<Self, Error> {
+        // unix abstract socket address, `man 7 unix` for system descriptions
+        if addr.starts_with("unix://@") {
+            return Ok(Self::Unix {
+                path: addr.trim_start_matches("unix://").to_owned(),
+            });
+        }
```
tony-iqlusion, line 95: It might make sense to file this logic below with the rest of the code that handles the `unix://` prefix.

```suggestion
        if let Some(path) = addr.strip_prefix("unix://@") {
            return Ok(Self::Unix {
                path: path.to_owned(),
            });
        }
```

ktmlm: yes, the '@' should be kept, your suggestions can not run.

__'moving this code block to the rest of the code that handles the `unix://` prefix'__, this is a good idea, but a `unix://@...` is very likely to be denied by the following `Url::parse`

ktmlm: the change I commit are the only one case that get tested in my practice

thanethomson (approved):
Seems reasonable to me. I'd be very interested to know who actually uses abstract Unix domain sockets with Tendermint.

PR #1089, opened 2022-02-10
Add block_by_hash RPC endpoint and tests

hansl:
Links #832.

Implements the block_by_hash RPC which returns an optional block.

* [x] Referenced an issue explaining the need for the change
* [x] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [x] Wrote tests
* [x] Added entry in `.changelog/`

Issue #942, opened 2021-08-06
`tx` method on the rpc client fails with InvalidParams error


jstuczyn:
Hi! I've been trying to use the rpc client to search for transaction details by its hash (using the `tx` method), unfortunately it's been failing for me with the following error:

```
'called `Result::unwrap()` on an `Err` value: Error { code: InvalidParams, message: "Invalid params", data: Some("error converting json params to arguments: illegal base64 data at input byte 64") }'
```

In my tests I have used `tendermint-rpc 0.21.0` run against `wasmd 0.17` (https://github.com/CosmWasm/wasmd) based validators which under the hood use `cosmos-sdk 0.42.5`.

**Steps to reproduce**

1. Start up a local (or could be remote for all it matters, I've checked it against both live and local instances) validator.

2. Run the following piece of code:
```
    let validator = "http://127.0.0.1:26657";
    let client = HttpClient::new(validator).unwrap();
    let tx_hash = "D716F496764E6D9D24DDCB55841D6E5226261106BFAA4EE31DAC5258DE60B62A"
        .parse()
        .unwrap();
    let res = client.tx(tx_hash, false).await.unwrap();
```

2.1. If you prefer to use the same environment in which it failed for me, the following live validator can also be used:
```
    let validator = "https://testnet-milhon-validator1.nymtech.net";
    let client = HttpClient::new(validator).unwrap();
    let tx_hash = "84A0AD8A48D5CBCC0E551EF29FF96FF9D164699F6C5A53E3B7584A0FC6402E6D"
        .parse()
        .unwrap();
    let res = client.tx(tx_hash, false).await.unwrap();
```

3. It's going to produce the error shown at the beginning of the issue description.

Note: I'm certain the block with the used hash exists as `curl http://127.0.0.1:26657/tx\?hash\=0xD716F496764E6D9D24DDCB55841D6E5226261106BFAA4EE31DAC5258DE60B62A` returns the relevant data. I'm including the response so that perhaps you might be able to spot invalid characters there or something:
```
{
  "jsonrpc": "2.0",
  "id": -1,
  "result": {
    "hash": "D716F496764E6D9D24DDCB55841D6E5226261106BFAA4EE31DAC5258DE60B62A",
    "height": "10802",
    "index": 0,
    "tx_result": {
      "code": 0,
      "data": "CgkKB2V4ZWN1dGU=",
[...]

Issue #832, opened 2021-03-22
Add support for remaining Tendermint RPC endpoints


thanethomson:
We need to implement client support for the full set of [Tendermint RPC endpoints](https://docs.tendermint.com/master/rpc/). (Is it necessary to implement the [unsafe methods](https://docs.tendermint.com/master/rpc/#/Unsafe) too?)

**What's the definition of "done" for this issue?**

When we have RPC client support for the following additional endpoints:

- [ ] [`/block_by_hash`](https://docs.tendermint.com/master/rpc/#/Info/block_by_hash)
- [ ] [`/dump_consensus_state`](https://docs.tendermint.com/master/rpc/#/Info/dump_consensus_state)
- [x] [`/consensus_params`](https://docs.tendermint.com/master/rpc/#/Info/consensus_params) (#972)
- [ ] [`/unconfirmed_txs`](https://docs.tendermint.com/master/rpc/#/Info/unconfirmed_txs)
- [ ] [`/num_unconfirmed_txs`](https://docs.tendermint.com/master/rpc/#/Info/num_unconfirmed_txs)
- [ ] [`/block_search`](https://docs.tendermint.com/master/rpc/#/Info/block_search) (#991)
- [x] [`/tx`](https://docs.tendermint.com/master/rpc/#/Info/tx) (#863)
- [ ] [`/check_tx`](https://docs.tendermint.com/master/rpc/#/Tx/check_tx)

thanethomson:
Not sure why GitHub closed this issue.

xla:
> Partially resolves #832.

The little robots look for certain keywords, `resovles` is one of them.

thanethomson:
> The little robots look for certain keywords

The little robots are pretty dumb, because they missed the most important word in that sentence :smile:

hansl:
@thanethomson I made a PR that addresses the `block_by_hash` RPC endpoint: #1089. How can I get started on a review?

hansl:
Is there any update on this? After the yanking of TM 0.35 a lot of these endpoints were removed from the latest TM-rs 0.25 release.

hansl:
`/block_by_hash` can be re-added with https://github.com/informalsystems/tendermint-rs/pull/1206 but it needs a reviewer.

romac:
@mzabaluev Can you please take a look at this and update the issue accordingly to the changes introduced in #1193?

mzabaluev:
[...]

PR #1206, opened 2022-10-04
Re-add `block_by_hash` Tendermint endpoint

fmorency:
Links https://github.com/informalsystems/tendermint-rs/issues/832.

Re-implements the `block_by_hash` RPC, which returns an optional block.

I applied the patch from https://github.com/informalsystems/tendermint-rs/pull/1089 and fixed the `tendermint::Hash` encoding/decoding to accommodate discrepancies in how the Tendermint RPC encodes this field for different RPC interfaces.

Relates to #942 

CC @hansl

* [x] Wrote tests


thanethomson (approved):
Thanks for this @fmorency!

PR #1212, opened 2022-10-13
docs: Fix unclosed HTML tags errors

### LINKED ISSUE #1211: docs: Build fails
For the current docs on `main`, the documentation build fails, as per https://github.com/informalsystems/tendermint-rs/actions/runs/3236423129/jobs/5302211119 (ref: #1209).

Seems to be due to new checks in nightly.

thanethomson:
Closes #1211 

* [ ] Wrote tests

### DIFF
--- a/proto/src/serializers/bytes.rs
+++ b/proto/src/serializers/bytes.rs
@@ -1,4 +1,4 @@
-//! Serialize/deserialize bytes (Vec<u8>) type
+//! Serialize/deserialize bytes (`Vec<u8>`) type
 
 /// Serialize into hexstring, deserialize from hexstring
 pub mod hexstring {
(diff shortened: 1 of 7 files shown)


proto/src/serializers/bytes.rs
```diff
@@ -18,7 +18,7 @@ pub mod hexstring {
             .map_err(serde::de::Error::custom)
     }
 
-    /// Serialize from T into hexstring
+    /// Serialize from `T` into hexstring
```
mzabaluev, line 21: I'd just say "a byte slice", it's generic.

proto/src/serializers/bytes.rs
```diff
@@ -68,14 +68,14 @@ pub mod base64string {
     }
 }
 
-/// Serialize into Vec<base64string>, deserialize from Vec<base64string>
+/// Serialize into `Vec<base64string>`, deserialize from `Vec<base64string>`
```
mzabaluev, line 71: To avoid pseudo type references and generalize over serde formats:

```suggestion
/// Serialize into and deserialize from a sequence of _base64string_.
```

mzabaluev (approved):
Good fix, I only have a few stylistic suggestions to improve these docs while you are at it.

Issue #1213, opened 2022-10-17
Trait-based cryptographic routines API to delegate implementation to downstream crates


blasrodri:
### Description

The goal is to be able to introduce an abstraction that encapsulates a set of everyday cryptographic operations needed
which do not necessarily share the exact underlying implementation details. This is precisely the case when we want to extend the functionality of the `Light Client Verifier` to support different light clients. Ideally, this could be used within any chain to verify the state of any other chain.

Based on [this pr/discussions](https://github.com/informalsystems/tendermint-rs/pull/1138) we've identified a few elements that we want to resolve:

- an agreed-upon interface (see below for a proposed one by @tony-iqlusion)
- reduce or eliminate the number of dependencies that are introduced
- be generic enough on the definition of methods/types so that we're not constraint to a specific implementation (see the [discussion](https://github.com/informalsystems/tendermint-rs/pull/1138#discussion_r901787307) about returning a boolean in the signature verification)

Proposed interface;

```rs
// https://github.com/informalsystems/tendermint-rs/pull/1138#issuecomment-1161835971

use digest::FixedOutput;
use digest::{consts::U32, Digest};
use tendermint::signature::Signer;
use tendermint::signature::Verifier;

pub trait CryptoProvider {
    type Sha256: Digest + FixedOutput<OutputSize = U32>;

    type EcdsaSecp256k1Signer: Signer<k256::ecdsa::Signature>;
    type EcdsaSecp256k1Verifier: Verifier<k256::ecdsa::Signature>;

    type Ed25519Signer: Signer<ed25519::Signature>;
    type Ed25519Verifier: Verifier<ed25519::Signature>;
}

```

### Definition of "done"

The trait is merged into the main branch, all tests are passed, and there is an external implementation of the trait 
for another light client that is successfully implemented using this interface.

adizere:
Copying @thanethomson and @mzabaluev as potentially having feedback on this API. Thanks Blas for shepherding this!

blasrodri:
I [drafted an implementation](https://github.com/informalsystems/tendermint-rs/pull/1214) of the trait for something similar to what Seun initially proposed. At least this proves that the design of the `CryptoProvider` is generic enough for us.

mzabaluev:
The design looks good to me. It enshrines `digest` (rust-crypto) as the framework that the host must implement the crypto routines through, but tendermint already depends on `signature` and most implementations in practice are from rust-crypto as well.

[...]

PR #1214, opened 2022-10-19
Add crypto provider

blasrodri:
Draft definition of the `CryptoProvider` trait defined in #1213 + an actual instance of the trait
for a `HostFunction*` implementation

### DIFF
--- a/light-client-verifier/Cargo.toml
+++ b/light-client-verifier/Cargo.toml
@@ -33,5 +33,9 @@ serde = { version = "1.0.106", default-features = false }
 time = { version = "0.3", default-features = false }
 flex-error = { version = "0.4.4", default-features = false }
 
+k256 = { version = "0.11", default-features = false, features = ["ecdsa", "sha256"] }
+digest = { version = "0.10", default-features = false }
+ed25519 = { version = "1.3", default-features = false }
+
 [dev-dependencies]
 tendermint-testgen = { path = "../testgen", default-features = false }
--- a/light-client-verifier/Cargo.toml
+++ b/light-client-verifier/Cargo.toml
@@ -36,6 +36,7 @@ flex-error = { version = "0.4.4", default-features = false }
 k256 = { version = "0.11", default-features = false, features = ["ecdsa", "sha256"] }
 digest = { version = "0.10", default-features = false }
 ed25519 = { version = "1.3", default-features = false }
+signature = { version = "1", default-features = false }
 
 [dev-dependencies]
 tendermint-testgen = { path = "../testgen", default-features = false }
--- a/light-client-verifier/Cargo.toml
+++ b/light-client-verifier/Cargo.toml
@@ -39,3 +39,5 @@ ed25519 = { version = "1.3", default-features = false }
 
 [dev-dependencies]
 tendermint-testgen = { path = "../testgen", default-features = false }
+sha2 = { version = "0.9", default-features = false }
+
--- a/light-client-verifier/src/host_functions.rs
+++ b/light-client-verifier/src/host_functions.rs
@@ -42,6 +42,16 @@ mod tests {
 
     impl<D: Digest, S: signature::Signature> DigestVerifier<D, S> for SubstrateSignatureVerifier<D> {
         fn verify_digest(&self, digest: D, signature: &S) -> Result<(), ed25519::Error> {
+            // TODO; having issues here
+            /*
+                        error[E0277]: the trait bound `VerifyingKey: DigestVerifier<D, _>` is not satisfied
+              --> light-client-verifier/src/host_functions.rs:46:38
+               |
+            46 |             self.inner.verify_digest(digest, signature)
+               |                        ------------- ^^^^^^ the trait `DigestVerifier<D, _>` is not implemented for `VerifyingKey`
+               |                        |
+               |                        required by a bound introduced by this call
+                         */
             self.inner.verify_digest(digest, signature)
         }
     }
@@ -102,7 +112,6 @@ mod tests {
             // Self::secp256k1_verify(sig, message, public)
             let verifier =
                 <<Self as CryptoProvider>::EcdsaSecp256k1Verifier>::from_bytes(public).unwrap();
-            // TODO: probably should name the verifier properly - not sure how to do it better
             let signature = k256::ecdsa::Signature::from_der(sig).unwrap();
             Ok(verifier.verify(message, &signature).unwrap())
         }
--- a/light-client-verifier/src/host_functions.rs
+++ b/light-client-verifier/src/host_functions.rs
@@ -7,7 +7,7 @@ pub trait CryptoProvider {
     type Sha256: Digest + FixedOutput<OutputSize = U32>;
 
     // type EcdsaSecp256k1Signer: Signer<k256::ecdsa::Signature>;
-    // type EcdsaSecp256k1Verifier: Verifier<k256::ecdsa::Signature>;
+    type EcdsaSecp256k1Verifier: Verifier<k256::ecdsa::Signature>;
 
     // type Ed25519Signer: Signer<ed25519::Signature>;
     // type Ed25519Verifier: Verifier<ed25519::Signature>;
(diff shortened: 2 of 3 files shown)


tony-iqlusion:
It might make more sense to put this in the `tendermint` crate itself, since it already contains high-level functions for working with private and public keys as well as signing and verifying messages.

Then downstream crates like `light-client-verifier` can use those high-level APIs.

This keeps the core cryptographic functionality in one crate, instead of spread out.

tendermint/Cargo.toml
```diff
@@ -51,14 +52,14 @@ tendermint-proto = { version = "0.25.0", default-features = false, path = "../pr
 time = { version = "0.3", default-features = false, features = ["macros", "parsing"] }
 zeroize = { version = "1.1", default-features = false, features = ["zeroize_derive", "alloc"] }
 flex-error = { version = "0.4.4", default-features = false }
-k256 = { version = "0.11", optional = true, default-features = false, features = ["ecdsa", "sha256"] }
+k256 = { version = "0.11", default-features = false, features = ["ecdsa", "sha256"] }
```
blasrodri, line 55: Having this one optional depends on whether `CryptoProvider` will be under a feature flag.

blasrodri:
> It might make more sense to put this in the `tendermint` crate itself, since it already contains high-level functions for working with private and public keys as well as signing and verifying messages.
> 
> Then downstream crates like `light-client-verifier` can use those high-level APIs.
> 
> This keeps the core cryptographic functionality in one crate, instead of spread out.

Moved it on [100fc53](https://github.com/informalsystems/tendermint-rs/pull/1214/commits/100fc53bb64fa706afde750029412e681dbffaa5)

mzabaluev:
The new API looks good in general, will review once more when undrafted.
Would there be an example of how this can be used in a light client verifier?

blasrodri:
> The new API looks good in general, will review once more when undrafted.
> Would there be an example of how this can be used in a light client verifier?

I'll add this, and un-draft it :)

tendermint/src/host_functions.rs
```diff
@@ -0,0 +1,134 @@
+use crate::signature::Verifier;
```
tony-iqlusion, line 1: As a general thought, "host functions" that contains only a `CryptoProvider` trait still feels like YAGNI to me.

I would personally suggest putting it under a `crypto` module (which should probably have been created a long time ago, with all of the existing cryptography code in the `tendermint` crate factored into it).

tony-iqlusion:
The trait itself looks OK, aside from the commented out part.

However, it isn't terribly useful unless the existing cryptography in the crate(s) is changed to use it.

And that's where this sort of change might get a bit onerous: you either need to add a generic type parameter with a default, or a generic type parameter with feature-gated type aliases. I'd probably suggest the latter.

blasrodri:
> The new API looks good in general, will review once more when undrafted. Would there be an example of how this can be used in a light client verifier?

Please check it out here: https://github.com/ComposableFi/near-rs/pull/16/

mzabaluev:
> However, it isn't terribly useful unless the existing cryptography in the crate(s) is changed to use it.

Is this PR meant to start a piecemeal introduction of the larger set of changes first submitted in #1138?

> add a generic type parameter with a default, or a generic type parameter with feature-gated type aliases.

I think a ready-to-use implementation of the crypto functions, using widely adopted pure Rust dependencies, should be provided in the tendermint crate for those users who don't need to bother with the alternatives. In the interests of dependency management, it needs to be feature-gated.

blasrodri:
Can you please be more specific on how to make if feature-gated friendly?
Happy to implement it

mzabaluev:
> Can you please be more specific on how to make if feature-gated friendly? Happy to implement it

Cribbing from #1138 again, you could have a generic `CommitValidator` parameterized over the `CryptoProvider` trait, and a feature-gated production instance:

```rust
#[cfg(feature = "rust-crypto")]
pub type ProdCommitValidator = CommitValidator<RustCryptoProvider>;
```

But perhaps it's work for a follow-up PR. I will look into https://github.com/ComposableFi/near-rs/pull/16 and the fork of tendermint-rs used in Hyperspace.

blasrodri:
@mzabaluev gave my first shot on feature gating this. Would def appreciate some guidance :)

light-client-verifier/src/operations/commit_validator.rs
```diff
@@ -11,6 +14,9 @@ use crate::{
 /// Validates the commit associated with a header against a validator set
 pub trait CommitValidator: Send + Sync {
     /// Perform basic validation
+    #[cfg(feature = "rust-crypto")]
+    type RustCryptoProvider: CryptoProvider;
+
```
mzabaluev, line 19: Feature-gating trait members is not a good idea, I think.

Perhaps I did not explain the idea more carefully. It's OK to work the abstract `CryptoProvider` generics wherever needed without feature-gating. So you could have (with tentative naming):

```rust
// Kill the CommitValidator trait, the methods become inherent members of the struct:

pub struct CommitValidator<C> {
    // ...
}

impl<C: CryptoProvider> CommitValidator<C> {
    // The API that belonged to CommitValidator trait goes here
}

/// The batteries-included validator, for when you don't mind the dependencies on
/// the full rust-crypto stack.
#[cfg(feature = "rust-crypto")]
pub type ProdCommitValidator = CommitValidator<RustCryptoProvider>;
```

@thanethomson I think all those traits having a single empty `ProdFoo` implementation don't carry much weight in the current design, since all the logic is provided in the trait methods. So it's fair to replace them with generic types parameterized around the crypto provider, like in the example above.

tendermint/src/crypto.rs
```diff
@@ -0,0 +1,127 @@
...
+        Digest::update(&mut hasher, msg);
+        self.inner.try_sign_digest(hasher)
+    }
+}
+
+trait DefaultHostFunctions: CryptoProvider {
+    fn sha2_256(preimage: &[u8]) -> [u8; 32];
+    fn ed25519_verify(sig: &[u8], msg: &[u8], pub_key: &[u8]) -> Result<(), ()>;
+    fn secp256k1_verify(sig: &[u8], message: &[u8], public: &[u8]) -> Result<(), ()>;
+}
```
mzabaluev, line 101: The purpose of this trait is not clear. Is it meant to be a "god trait" for all future functions that could be host-provided? Now, it only adds the crypto methods that I think could well be defined in `CryptoProvider` itself, because the implementation is tightly coupled with the associated type definitions there.

I agree with @tony-iqlusion that this umbrella trait is not necessarily needed until we realise a clear purpose for it.

light-client-verifier/Cargo.toml
```diff
@@ -35,3 +35,4 @@ flex-error = { version = "0.4.4", default-features = false }
 
 [dev-dependencies]
 tendermint-testgen = { path = "../testgen", default-features = false }
+sha2 = { version = "0.9", default-features = false }
```
mzabaluev, line 38: Is this needed by default, or only when the "rust-crypto" feature is on? Or perhaps no longer needed as a dependency here at all, since the crypto functionality in light-client-verifier should be completely abstracted via `CryptoProvider`.

tendermint/src/crypto.rs
```diff
@@ -0,0 +1,127 @@
...
+use signature::{DigestSigner, DigestVerifier, Signer};
+
+pub trait CryptoProvider {
+    type Sha256: Digest + FixedOutput<OutputSize = U32>;
+
+    type EcdsaSecp256k1Signer: Signer<k256::ecdsa::Signature>;
+    type EcdsaSecp256k1Verifier: Verifier<k256::ecdsa::Signature>;
+
+    // type Ed25519Signer: Signer<ed25519::Signature>;
+    // type Ed25519Verifier: Verifier<ed25519::Signature>;
```
mzabaluev, line 13: This should be removed, to be added later when/if needed.

tendermint/Cargo.toml
mzabaluev: As I understand it now, we lock the signature type in `CryptoProvider` to the one defined by `k256`. If we cannot abstract this associated type to less specific rust-crypto bounds ("must be fixed array of 256 bits"?) and still keep it usable in generic code, then `k256` should become a hard dependency. But I thought dependency management was the whole purpose of this rework.

mzabaluev (changes requested):

light-client-verifier/Cargo.toml
mzabaluev: Sorry, it's dev-dependencies, ignore this.

mzabaluev:
I think the functionality abstracted by `CryptoProvider` should cut deeper, if the goals of this change are to be realized. One instance where `Sha256` is still used unconditionally is the hasher, which simply calls into intrinsic `hash` methods of tendermint domain types. An alternative hasher cannot replicate that without duplicating the implementation of those methods. I'm doing some experiments to see if we can come up with a more comprehensive redesign and maybe simplify the generics in light-client-verifier at the same time.

mzabaluev:
Here's my crack at it: #1238

romac:
Superseded by #1238

Issue #1222, opened 2022-11-04
Request for tendermint go like APIs (`VerifyCommitLight()` and `VerifyCommitLightTrusting()`)


hu55a1n1:
### Description
Would be nice to have rust equivalents for Tendermint go's [VerifyCommitLight()](https://github.com/tendermint/tendermint/blob/a6dd0d270abc3c01f223eedee44d8b285ae273f6/types/validator_set.go#L722) and [VerifyCommitLightTrusting()](https://github.com/tendermint/tendermint/blob/a6dd0d270abc3c01f223eedee44d8b285ae273f6/types/validator_set.go#L775).

### Definition of "done"
Publicly accessible verifier methods similar to the go impl -> 
```rust
verifier.verify_commit_light();
verifier.verify_commit_light_trusting();
```

PR #1226, opened 2022-11-10
Implement verify_light() and verify_light_trusting()

hu55a1n1:
Closes: #1222

* [ ] Wrote tests

### DIFF
--- a/light-client-verifier/src/operations/voting_power.rs
+++ b/light-client-verifier/src/operations/voting_power.rs
@@ -69,12 +69,12 @@ pub trait VotingPowerCalculator: Send + Sync {
 
     /// Check against the given threshold that there is enough signers
     /// overlap between an untrusted header and untrusted validator set
-    fn check_signers_overlap(
+    fn check_enough_signers_overlap(
         &self,
         untrusted_header: &SignedHeader,
         untrusted_validators: &ValidatorSet,
+        trust_threshold: TrustThreshold,
     ) -> Result<(), VerificationError> {
-        let trust_threshold = TrustThreshold::TWO_THIRDS;
         let voting_power =
             self.voting_power_in(untrusted_header, untrusted_validators, trust_threshold)?;
 
@@ -85,6 +85,19 @@ pub trait VotingPowerCalculator: Send + Sync {
                 voting_power,
             ))
         }
+    }
+
+    /// Check if there is 2/3rd overlap between an untrusted header and untrusted validator set
+    fn check_signers_overlap(
+        &self,
+        untrusted_header: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
+    ) -> Result<(), VerificationError> {
+        self.check_signers_overlap_level(
+            untrusted_header,
+            untrusted_validators,
+            TrustThreshold::TWO_THIRDS,
+        )
     }
 
     /// Compute the voting power in a header and its commit against a validator set.
--- a/light-client-verifier/src/operations/voting_power.rs
+++ b/light-client-verifier/src/operations/voting_power.rs
@@ -93,7 +93,7 @@ pub trait VotingPowerCalculator: Send + Sync {
         untrusted_header: &SignedHeader,
         untrusted_validators: &ValidatorSet,
     ) -> Result<(), VerificationError> {
-        self.check_signers_overlap_level(
+        self.check_enough_signers_overlap(
             untrusted_header,
             untrusted_validators,
             TrustThreshold::TWO_THIRDS,
--- a/light-client-verifier/src/operations/voting_power.rs
+++ b/light-client-verifier/src/operations/voting_power.rs
@@ -67,14 +67,13 @@ pub trait VotingPowerCalculator: Send + Sync {
         }
     }
 
-    /// Check against the given threshold that there is enough signers
-    /// overlap between an untrusted header and untrusted validator set
-    fn check_enough_signers_overlap(
+    /// Check if there is 2/3rd overlap between an untrusted header and untrusted validator set
+    fn check_signers_overlap(
         &self,
         untrusted_header: &SignedHeader,
         untrusted_validators: &ValidatorSet,
-        trust_threshold: TrustThreshold,
     ) -> Result<(), VerificationError> {
+        let trust_threshold = TrustThreshold::TWO_THIRDS;
         let voting_power =
             self.voting_power_in(untrusted_header, untrusted_validators, trust_threshold)?;
 
@@ -85,19 +84,6 @@ pub trait VotingPowerCalculator: Send + Sync {
                 voting_power,
             ))
         }
-    }
-
-    /// Check if there is 2/3rd overlap between an untrusted header and untrusted validator set
-    fn check_signers_overlap(
-        &self,
-        untrusted_header: &SignedHeader,
-        untrusted_validators: &ValidatorSet,
-    ) -> Result<(), VerificationError> {
-        self.check_enough_signers_overlap(
-            untrusted_header,
-            untrusted_validators,
-            TrustThreshold::TWO_THIRDS,
-        )
     }
 
     /// Compute the voting power in a header and its commit against a validator set.
--- a/light-client-verifier/src/predicates.rs
+++ b/light-client-verifier/src/predicates.rs
@@ -194,7 +194,7 @@ pub trait VerificationPredicates: Send + Sync {
         calculator: &dyn VotingPowerCalculator,
         trust_threshold: TrustThreshold,
     ) -> Result<(), VerificationError> {
-        calculator.check_signers_overlap_level(
+        calculator.check_enough_signers_overlap(
             untrusted_sh,
             untrusted_validators,
             trust_threshold,
--- a/light-client-verifier/src/predicates.rs
+++ b/light-client-verifier/src/predicates.rs
@@ -185,23 +185,6 @@ pub trait VerificationPredicates: Send + Sync {
         Ok(())
     }
 
-    /// Check that there is enough signers overlap between the given, untrusted validator set
-    /// and the untrusted signed header based on the specified trust threshold.
-    fn has_specified_signers_overlap(
-        &self,
-        untrusted_sh: &SignedHeader,
-        untrusted_validators: &ValidatorSet,
-        calculator: &dyn VotingPowerCalculator,
-        trust_threshold: TrustThreshold,
-    ) -> Result<(), VerificationError> {
-        calculator.check_enough_signers_overlap(
-            untrusted_sh,
-            untrusted_validators,
-            trust_threshold,
-        )?;
-        Ok(())
-    }
-
     /// Check that the hash of the next validator set in the trusted block matches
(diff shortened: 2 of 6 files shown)


light-client-verifier/src/operations/voting_power.rs
```diff
@@ -87,6 +87,19 @@ pub trait VotingPowerCalculator: Send + Sync {
         }
     }
 
+    /// Check if there is 2/3rd overlap between an untrusted header and untrusted validator set
+    fn check_signers_overlap(
+        &self,
+        untrusted_header: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
+    ) -> Result<(), VerificationError> {
+        self.check_signers_overlap_level(
```
mzabaluev, line 96: I guess this should be a call to the `check_enough_signers_overlap` method above...

light-client-verifier/src/predicates.rs
```diff
@@ -185,6 +185,23 @@ pub trait VerificationPredicates: Send + Sync {
...
+    /// Check that there is enough signers overlap between the given, untrusted validator set
+    /// and the untrusted signed header based on the specified trust threshold.
+    fn has_specified_signers_overlap(
+        &self,
+        untrusted_sh: &SignedHeader,
+        untrusted_validators: &ValidatorSet,
+        calculator: &dyn VotingPowerCalculator,
+        trust_threshold: TrustThreshold,
+    ) -> Result<(), VerificationError> {
+        calculator.check_signers_overlap_level(
```
mzabaluev, line 197: And likewise here...

mzabaluev (commented):
The changes look good so far. It would be nice if the names matched so that the code compiles.
We may also need some more tests for the newly added functionality.

light-client-verifier/src/operations/voting_power.rs
hu55a1n1: 01c5112f3625905bb00db13c030254532410adf4

light-client-verifier/src/predicates.rs
hu55a1n1: 01c5112f3625905bb00db13c030254532410adf4

thanethomson (commented):
This seems good to me, thanks @hu55a1n1! Could you please add a changelog entry for this?

light-client/src/tests.rs
```diff
@@ -152,25 +152,42 @@ pub fn verify_single(
...
+    }
+
+    impl<T> IntoResult<T, Verdict> for Verdict {
+        fn into_result(self, ok: T) -> Result<T, Verdict> {
+            match self {
+                Verdict::Success => Ok(ok),
+                error => Err(error),
+            }
+        }
+    }
```
mzabaluev, line 166: Nit: the purpose of this could look more obvious if this were just a helper function. Then again, method chaining is neat.

mzabaluev (commented):
Looks good!

light-client/src/tests.rs
mzabaluev: As an aside, not related to this PR: why did there have to be a non-idiomatic `Verdict`, rather than a `Result<(), VerificationError>` where `VerificationError` would be an enum or otherwise easily matchable to that variant?
Basically, what the existing `impl From<Result<(), VerificationError>> for Verdict` does could be just how the `Result` returned value is supposed to be used if you care about the `NotEnoughTrust` case. Cc @romac

mzabaluev: > Agreed, let's do that.

I didn't mean this to be done in scope for this PR, which so far has not introduced breaking changes, but if you think it would be good and easy thing to do in the next release, I'm all for it.

romac: Indeed, let's target this for 0.27.

hu55a1n1:
Thanks, @mzabaluev, @thanethomson and @romac for reviewing this. @ancazamfir pointed out that the misbehaviour handling in IBC used the same logic as the update handling (i.e. `ProdVerifier::verify`), but it was done in a way that required us to separate the expensive checks from the inexpensive ones and call them alternatively for each header in the misbehaviour evidence/msg. I ended up extracting 4 functions from the `verify()` method, making them publicly visible and implementing `verify()` itself using these component methods.

light-client-verifier/src/verifier.rs
```diff
@@ -111,119 +111,163 @@ where
...
+        self.predicates.has_sufficient_signers_overlap(
+            untrusted.signed_header,
+            untrusted.validators,
+            &self.voting_power_calculator,
+        )?;
+
+        Ok(())
+    }
+
+    /// Validate a `UntrustedBlockState`, based on the given `TrustedBlockState`, `Options` and
```
ancazamfir, line 168:
```suggestion
    /// Validate an `UntrustedBlockState`, based on the given `TrustedBlockState`, `Options` and
```

light-client-verifier/src/verifier.rs
```diff
@@ -111,119 +111,163 @@ where
...
+    /// current time.
+    pub fn validate_trusting(
```
ancazamfir, line 170:
```suggestion
    pub fn validate_unntrusted_against_trusted(
```

light-client-verifier/src/verifier.rs
```diff
@@ -111,119 +111,163 @@ where
...
+    /// - Ensure the header next validator hashes match the given next validators
+    /// - Additional implementation specific validation via `commit_validator`
+    /// - Check that the untrusted block is more recent than the trusted state
+    /// - If the untrusted block is the very next block after the trusted block, check that their
+    ///   (next) validator sets hashes match.
+    /// - Otherwise, ensure that the untrusted block has a greater height than the trusted block.
+    ///
+    /// **NOTE**: If the untrusted state's `next_validators` field is `None`,
+    /// this will not (and will not be able to) check whether the untrusted
+    /// state's `next_validators_hash` field is valid.
```
ancazamfir, line 259: These comments need updates to reflect the new code.

light-client-verifier/src/verifier.rs
```diff
@@ -111,119 +111,163 @@ where
...
+        &self,
+        untrusted: UntrustedBlockState<'_>,
+        trusted: TrustedBlockState<'_>,
+        options: &Options,
+        now: Time,
+    ) -> Verdict {
+        verdict!(self.validate_untrusted(&untrusted));
+        verdict!(self.validate_trusting(&untrusted, &trusted, options, now));
+        verdict!(self.verify_commit(&untrusted));
+        verdict!(self.verify_commit_trusting(&untrusted, &trusted, options));
```
ancazamfir, line 270: To keep the original order that we had in `verify()`:
```suggestion
        verdict!(self.verify_commit_trusting(&untrusted, &trusted, options));
        verdict!(self.verify_commit(&untrusted));
```

ancazamfir, line 270: It will be good to have the companion IBC PR (https://github.com/cosmos/ibc-rs/pull/215) updated to understand the API usage. It seems that the one used by the update client handler (`verify()` that returns `Verdict`) and these new APIs that will be used by misbehavior (return `Result<(), VerificationError>`)  have inconsistent signatures.

ancazamfir (changes requested):
Thanks @hu55a1n1! Left a few comments but in general it looks good. Could we see https://github.com/cosmos/ibc-rs/pull/215 updates before we merge this?

light-client-verifier/src/verifier.rs
hu55a1n1: I renamed `validate_trusting` to `validate_against_trusted` and `verify_commit_trusting` to `verify_commit_against_trusted` since I thought the `untrusted` part in the name was implicit, but please let me know if you prefer to keep it. 
https://github.com/informalsystems/tendermint-rs/pull/1226/commits/4555afee69b1dd4290fa644fd5ce747bd10f5452

light-client-verifier/src/verifier.rs
hu55a1n1: Sorry for the delay here. https://github.com/cosmos/ibc-rs/pull/215

light-client-verifier/src/verifier.rs
```diff
@@ -111,119 +111,160 @@ where
...
+        self.predicates.validator_sets_match(
             untrusted.validators,
             untrusted.signed_header.header.validators_hash,
             &self.hasher,
-        ));
+        )?;
 
         // TODO(thane): Is this check necessary for IBC?
         if let Some(untrusted_next_validators) = untrusted.next_validators {
             // Ensure the header next validator hashes match the given next validators
```
ancazamfir, line 126: Yes, the check is necessary so we can remove the comment
```suggestion
        // Ensure the header next validator hashes match the given next validators
        if let Some(untrusted_next_validators) = untrusted.next_validators {
```

light-client-verifier/src/verifier.rs
ancazamfir: Thanks!
It is not clear why `verify()` returns a `Verdict` while the new APIs return `VerificationError`. We should use one or the other if there is no good explanation.

hu55a1n1: https://github.com/informalsystems/tendermint-rs/pull/1226/commits/c1b5c341f6df258b84d33aa79ade5e4dc554a3a8
I decided to return `Verdict` from all verification methods but this required me to add a convenience macro. Related - https://github.com/informalsystems/tendermint-rs/issues/1236

light-client-verifier/src/verifier.rs
```diff
@@ -217,13 +236,50 @@ where
...
+impl<P, C, V, H> Verifier for PredicateVerifier<P, C, V, H>
+where
+    P: VerificationPredicates,
+    C: VotingPowerCalculator,
+    V: CommitValidator,
+    H: Hasher,
+{
+    /// Validate the given light block state by performing the following checks ->
+    ///
+    /// - Validate latest untrusted header
```
ancazamfir, line 252:
```suggestion
    /// - Validate untrusted header
```

light-client-verifier/src/verifier.rs
```diff
@@ -217,13 +236,50 @@ where
...
+    ///     - Ensure the header validator hashes match the given validators
+    ///     - Ensure the header next validator hashes match the given next validators
+    ///     - Ensure the header matches the commit
+    ///     - Ensure commit is valid
+    /// - Validate latest untrusted header
```
ancazamfir, line 257:
```suggestion
    /// - Validate the untrusted header against the trusted header
```

ancazamfir (approved):
I will approve as it looks great, thanks @hu55a1n1 
If you can make the API name [in this discussion](https://github.com/cosmos/ibc-rs/pull/215#discussion_r1031494609) it's great. I also added two very minor comments.

hu55a1n1:
> If you can make the API name https://github.com/cosmos/ibc-rs/pull/215#discussion_r1031494609 it's great.
Done -> https://github.com/informalsystems/tendermint-rs/pull/1226/commits/c0f2469caf9d1f38b1c6af2f4232d22867165fec

thanethomson (approved):

Issue #1225, opened 2022-11-10
tendermint-proto: `grpc` feature lost in Tendermint v0.35 shuffle


tony-iqlusion:
@tomtau added a `grpc` feature to the `tendermint-proto` crate in #1137

However it seem it was lost in the shuffle around Tendermint v0.35.

It's needed to implement gRPC "privval" support, as in #1152.

tony-iqlusion:
I opened #1127 which cherry-picks the changes from #1137 onto `main` with some slight modifications

tony-iqlusion:
Note: the `privval` side of things is blocked on https://github.com/tendermint/tendermint/issues/9256

erwanor:
@tony-iqlusion I opened #1338 which includes the server definitions for the rpc and abci service, but not for `privval` because it's blocked on #9256. I just realized that I used `tonic-build@0.9` instead of `tonic-build@0.8` in that PR, I can change that. 

My motivation is that we want to add gRPC in `tower-abci` and unfortunately the `privval` service is blocked by cometbft upstream.

PR #1137, opened 2022-05-30
tendermint-proto: Add a feature flag and generate gRPC server definitions

tomtau:
- `proto-compiler` uses `tonic-build`
- tendermint-proto can be compiled with std for the `grpc` feature
(as required by the generated `tonic` code)
- the primary motivation is the new gRPC-based PrivVal interface
in Tendermint 0.35
(see https://github.com/informalsystems/tendermint-rs/issues/1134)

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [x] Added entry in `.changelog/`

PR #1227, opened 2022-11-10
tendermint-proto: (re-)add `grpc` feature and server definitions

tony-iqlusion:
See #1225.

Adds a "grpc" feature to the crate along with gRPC server definitions for the "privval" interface used by e.g. TMKMS to support remote signing.

These were originally added in #1137 by @tomtau but lost in the shuffle of Tendermint v0.35-related changes.

They have been updated to use `tonic` v0.8, which the `tendermint-proto` crate updated to after the original #1137 PR landed.

* [ ] Wrote tests

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -1,6 +1,6 @@
 //! tendermint-proto library gives the developer access to the Tendermint proto-defined structs.
 
-#![no_std]
+#![cfg_attr(not(feature = "grpc"), no_std)]
 #![deny(warnings, trivial_casts, trivial_numeric_casts, unused_import_braces)]
 #![allow(clippy::large_enum_variant)]
 #![forbid(unsafe_code)]
--- a/tools/proto-compiler/src/main.rs
+++ b/tools/proto-compiler/src/main.rs
@@ -80,7 +80,15 @@ fn main() {
         "super::super::google::protobuf::Timestamp",
     );
     println!("[info] => Creating structs.");
-    pb.compile_protos(&protos, &proto_includes_paths).unwrap();
+    tonic_build::configure()
+        .out_dir(&out_dir)
+        .build_client(false)
+        .build_server(true)
+        .server_mod_attribute("tendermint.abci", "#[cfg(feature = \"grpc\")]")
+        .server_mod_attribute("tendermint.privval", "#[cfg(feature = \"grpc\")]")
+        .server_mod_attribute("tendermint.rpc.grpc", "#[cfg(feature = \"grpc\")]")
+        .compile_with_config(pb, &protos, &proto_includes_paths)
+        .unwrap();
 
     println!("[info] => Removing old structs and copying new structs.");
     copy_files(&out_dir, &target_dir); // This panics if it fails.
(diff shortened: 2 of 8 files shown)


tony-iqlusion:
Note: the gRPC service definitions are otherwise verbatim from #1137 which generated them using `tonic` v0.7. I had to tweak `accept_compression_encodings` and `send_compression_encodings` to get the code to compile with `tonic` v0.8.

They should probably be completely regenerated with `tonic` v0.8, although I'm not sure offhand the best way to do that with `tendermint-proto`, especially with the `grpc` feature gating.

FWIW the process for this, including adding the `#[cfg(feature = "grpc"))]` annotations to the relevant code, is completely automated in `cosmos-sdk-proto`'s `proto-build`: https://github.com/cosmos/cosmos-rust/blob/bee6a6b/proto-build/src/main.rs#L398-L418

tomtau (commented):
The current main seems to be based on v0.34.x: https://github.com/informalsystems/tendermint-rs/blob/main/proto/src/tendermint.rs#L65 which doesn't have the gRPC privval support (those service definitions don't exist here: https://github.com/tendermint/tendermint/blob/v0.34.23/proto/tendermint/privval/types.proto).

I'm not sure whether there's a plan to backport it to 0.34 and have a release with it.

tony-iqlusion:
Oh wow, so it is. That's unfortunate

tony-iqlusion:
@thanethomson any idea if/when the gRPC privval interface will get re-added upstream in Tendermint?

thanethomson:
> any idea if/when the gRPC privval interface will get re-added upstream in Tendermint?

It's in our [prioritized backlog](https://github.com/orgs/tendermint/projects/15/views/8). We have a bunch of burning, high-priority issues to deal with this quarter (see [this board](https://github.com/orgs/tendermint/projects/15/views/5)), so those things take precedence. As we have capacity we pick things off the prioritized backlog to work on.

tony-iqlusion:
Aha cool, here's the relevant issue: https://github.com/tendermint/tendermint/issues/9256

proto/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! tendermint-proto library gives the developer access to the Tendermint proto-defined structs.
 
-#![no_std]
+#![cfg_attr(not(feature = "grpc"), no_std)]
```
mzabaluev, line 3: This effect could be made explicitly seen also in `Cargo.toml` by making this conditional on the "std" feature, which "grpc" would depend on. In practice though, "grpc" is the only feature that requires std for now.

tools/proto-compiler/src/main.rs
```diff
@@ -80,7 +80,15 @@ fn main() {
         "super::super::google::protobuf::Timestamp",
     );
     println!("[info] => Creating structs.");
-    pb.compile_protos(&protos, &proto_includes_paths).unwrap();
+    tonic_build::configure()
+        .out_dir(&out_dir)
+        .build_client(false)
+        .build_server(true)
```
mzabaluev, line 86: What if we want to enable client-side gRPC bindings later, and feature-gate client and server sides independently? I think it's possible to introduce "grpc-server" and "grpc-client" features later, with "grpc" dependent on both, so this would be backward compatible.

mzabaluev (approved):
Looks good to me.

mzabaluev:
If this is still needed, can you rebase it and fix the conflicts?

tony-iqlusion:
@mzabaluev it's blocked on https://github.com/tendermint/tendermint/issues/9256 (which I now see was closed as stale).

I would definitely still love to see all of this happen, if possible though.

mzabaluev:
Should this be closed in favor of #1338?

tony-iqlusion:
Sure

PR #1229, opened 2022-11-15
tendermint-rpc: Improve `Query` API

romac:
Closes: #1230

* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/improvements/1230-query-api.md
+++ b/.changelog/unreleased/improvements/1230-query-api.md
@@ -1,0 +1,3 @@
+- [`tendermint-rpc`] Make `tendermint_rpc::Query`'s fields
+  public and add a `Condition::key(&self) -> &str` method
+  ([#1230](https://github.com/informalsystems/tendermint-rs/issues/1230))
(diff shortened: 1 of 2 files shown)


.changelog/unreleased/improvements/1230-query-api.md
```diff
@@ -0,0 +1,3 @@
+- [`tendermint-rpc`] Make `tendermint_rpc::Query`'s fields
```
thanethomson, line 1:
```suggestion
- `[tendermint-rpc]` Make `tendermint_rpc::Query`'s fields
```

With the other format, it's interpreted as a Markdown link.

rpc/src/query.rs
```diff
@@ -442,16 +442,31 @@ pub enum Condition {
     Exists(String),
 }
 
+impl Condition {
+    /// Return the key that this condition applies to
+    pub fn key(&self) -> &str {
```
thanethomson, line 447: Given that a key is required for every condition, would it make sense to perhaps change `Condition` to be a struct that contains a key and a separate enum for the condition variants?

romac: As long as we don't foresee a new type of condition which does not apply to a key, then yeah I guess it would make sense. Happy to take care of that.

romac: Here's a PR for this, targeting this branch: https://github.com/informalsystems/tendermint-rs/pull/1231

mzabaluev (approved):

thanethomson (approved):

Issue #1095, opened 2022-02-16
abci: Change hashes' type from `bytes::Bytes` to `tendermint::Hash`


thanethomson:
Version(s) of tendermint-rs: `master`

### Description

Related to #1090.

At the moment, several data structures in the `tendermint` crate's `abci` module have `hash` members of type `bytes::Bytes` (e.g. [`BeginBlock`](https://github.com/informalsystems/tendermint-rs/blob/33a48411a080692b3b962330d9f4492829a25c60/tendermint/src/abci/request/begin_block.rs#L20)). I'd like to rather make these of type [`tendermint::Hash`](https://github.com/informalsystems/tendermint-rs/blob/33a48411a080692b3b962330d9f4492829a25c60/tendermint/src/hash.rs#L27).

We can easily provide a conversion for non-empty hashes to `bytes::Bytes`.

cc @hdevalence 

### Definition of "done"

When all `hash` fields in structs in the `tendermint::abci` module are of type `tendermint::Hash`.

mzabaluev:
Questions:

- The response structures `Info` and `InitChain` implement `Default` (`Info` also defaults the missing field values to the ones provided by the `Default` impl). If we change the app hash fields to `AppHash`, is it OK to have the default value of an empty byte array, like it already is de-facto?
- The `hash` field of `Snapshot` is documented as an arbitrary value compared by equality, so the semantics for neither `tendermint::Hash` nor `tendermint::AppHash` obviously apply. I think it's ok to leave it as `Bytes`.

PR #1232, opened 2022-11-16
abci: Change hashes' type from `Bytes` to `tendermint::Hash` or `tendermint::AppHash`

mzabaluev:
Fixes: #1095

Both questions in https://github.com/informalsystems/tendermint-rs/issues/1095#issuecomment-1279223745 are resolved with "Yes".

* [x] Wrote tests

### DIFF
--- a/tendermint/src/abci/response/info.rs
+++ b/tendermint/src/abci/response/info.rs
@@ -1,9 +1,8 @@
-use crate::{block, prelude::*, Error};
+use crate::{block, prelude::*, AppHash, Error};
 use core::convert::TryFrom;
 use tendermint_proto::abci as pb;
 use tendermint_proto::Protobuf;
 
-use bytes::Bytes;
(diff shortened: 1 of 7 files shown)


tendermint/src/abci/response/info.rs
```diff
@@ -19,8 +18,7 @@ pub struct Info {
     /// The latest block for which the app has called [`Commit`](super::super::Request::Commit).
     pub last_block_height: block::Height,
     /// The latest result of [`Commit`](super::super::Request::Commit).
-    // XXX(hdevalence): fix this, should be apphash?
-    pub last_block_app_hash: Bytes,
+    pub last_block_app_hash: AppHash,
```
mzabaluev, line 21: Mysteriously, this is serialized as an array of bytes rather than a hex string.
This PR does not change it.

Issue #1233, opened 2022-11-16
Decouple tendermint-rpc from tendermint-proto


mzabaluev:
### Description

After #1203 and #1090, a few direct uses of tendermint-proto remain in tendermint-rpc, so I see an opportunity to remove the direct API link in preparation to #1193. The only significant change is to replace direct use of a generated `TxProto` struct with a domain type. That type will also use `Hash` in all proper places.

### Definition of "done"

tendermint-rpc no longer lists tendermint-proto in its dependencies, its public API only exposes `tendermint` types and functions (some of which may be re-imports of tendermint-proto).

PR #1234, opened 2022-11-16
Decouple tendermint-rpc from tendermint-proto

mzabaluev:
Fixes: #1233

Replace remaining use of `prost-build`-generated structs in `tendermint-rpc` with domain types newly defined in `tendermint`.

In `tendermint`, `merkle::proof::Proof` was named inconsistently with regard to protobuf, which has it as `crypto.ProofOps`, yet also the `crypto.Proof` message is defined, which we make use of now. Renamed that domain type to `ProofOps` to make way for the newly added `Proof`. 

* [ ] Wrote tests


thanethomson (approved):
This is great, thanks @mzabaluev!

PR #1242, opened 2022-12-05
light-client-js: remove the syn version pin

mzabaluev:
The bug this used to work around is closed and the problem
no longer seems to persist.

### DIFF
--- a/light-client-js/src/lib.rs
+++ b/light-client-js/src/lib.rs
@@ -24,22 +24,18 @@ use wasm_bindgen::{prelude::*, JsValue};
 
 /// Check whether a given untrusted block can be trusted.
 #[wasm_bindgen]
-pub fn verify(untrusted: &JsValue, trusted: &JsValue, options: &JsValue, now: &JsValue) -> JsValue {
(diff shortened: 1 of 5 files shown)


light-client-js/src/lib.rs
```diff
@@ -25,46 +25,50 @@ use wasm_bindgen::{prelude::*, JsValue};
...
-            )
-        },
-    );
-    JsValue::from_serde(&result).unwrap()
+    let result = deserialize_params(
+        untrusted.clone(),
+        trusted.clone(),
+        options.clone(),
+        now.clone(),
+    )
```
mzabaluev, line 33: I wonder if the signature of `verify` can be changed to take the arguments by value.

thanethomson (approved):
LGTM!

mzabaluev:
> Would this be considered a breaking change, given the dependency update?

I don't think so. A wasm bundle has all its Rust dependencies baked in, and the JavaScript signature of the entry point did not change either.

That said, I need to give it a suitable changelog entry.

PR #1245, opened 2022-12-06
deps:  Use ed25519-consensus instead of ed25519-dalek (#1067)

### LINKED ISSUE #355: Consider `ed25519-zebra` crate?

hdevalence:
Backports this change to the new `main` (was #1067).

* Use ed25519-consensus instead of ed25519-dalek

Closes #355 (see that issue for more context; `ed25519-consensus` is a fork of `ed25519-zebra` that's Zcash-independent).

This change ensures that `tendermint-rs` has the same signature verification as `tendermint`, which uses the validation criteria provided by the Go `ed25519consensus` library.

* clippy fixes

Co-authored-by: Thane Thomson <connect@thanethomson.com>

* Remove redundant dependency

Signed-off-by: Thane Thomson <connect@thanethomson.com>

* cargo fmt

Signed-off-by: Thane Thomson <connect@thanethomson.com>

* Add changelog entries

Signed-off-by: Thane Thomson <connect@thanethomson.com>

Co-authored-by: Henry de Valence <hdevalence@penumbra.zone>

### DIFF
--- a/p2p/src/error.rs
+++ b/p2p/src/error.rs
@@ -6,7 +6,6 @@
 
 use flex_error::{define_error, DisplayOnly};
 use prost::DecodeError;
-use signature::Error as SignatureError;
 
 define_error! {
     Error {
@@ -40,7 +39,6 @@ define_error! {
             | _ | { "public key missing" },
 
         Signature
-            [ DisplayOnly<SignatureError> ]
             | _ | { "signature error" },
 
         UnsupportedKey
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -183,6 +183,7 @@ impl Handshake<AwaitingAuthSig> {
             .and_then(|key| key.sum)
             .ok_or_else(Error::missing_key)?;
 
+        #[allow(clippy::match_wildcard_for_single_variants)]
         let remote_pubkey = match pk_sum {
             proto::crypto::public_key::Sum::Ed25519(ref bytes) => {
                 ed25519_consensus::VerificationKey::try_from(&bytes[..])
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -183,13 +183,12 @@ impl Handshake<AwaitingAuthSig> {
             .and_then(|key| key.sum)
             .ok_or_else(Error::missing_key)?;
 
-        #[allow(clippy::match_wildcard_for_single_variants)]
         let remote_pubkey = match pk_sum {
             proto::crypto::public_key::Sum::Ed25519(ref bytes) => {
                 ed25519_consensus::VerificationKey::try_from(&bytes[..])
                     .map_err(|_| Error::signature())
             },
-            _ => Err(Error::unsupported_key()),
+            proto::crypto::public_key::Sum::Secp256k1(_) => Err(Error::unsupported_key()),
         }?;
 
         let remote_sig = ed25519_consensus::Signature::try_from(auth_sig_msg.sig.as_slice())
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -16,7 +16,6 @@ use chacha20poly1305::{
     aead::{generic_array::GenericArray, AeadInPlace, NewAead},
     ChaCha20Poly1305,
 };
-use ed25519_dalek::{self as ed25519, Signer, Verifier};
 use merlin::Transcript;
 use rand_core::OsRng;
 use subtle::ConstantTimeEq;
@@ -61,7 +60,7 @@ pub struct Handshake<S> {
 
 /// `AwaitingEphKey` means we're waiting for the remote ephemeral pubkey.
 pub struct AwaitingEphKey {
-    local_privkey: ed25519::Keypair,
+    local_privkey: ed25519_consensus::SigningKey,
     local_eph_privkey: Option<EphemeralSecret>,
 }
 
@@ -71,7 +70,7 @@ pub struct AwaitingAuthSig {
     kdf: Kdf,
     recv_cipher: ChaCha20Poly1305,
     send_cipher: ChaCha20Poly1305,
-    local_signature: ed25519::Signature,
+    local_signature: ed25519_consensus::Signature,
 }
 
 #[allow(clippy::use_self)]
@@ -79,7 +78,7 @@ impl Handshake<AwaitingEphKey> {
     /// Initiate a handshake.
     #[must_use]
     pub fn new(
-        local_privkey: ed25519::Keypair,
+        local_privkey: ed25519_consensus::SigningKey,
         protocol_version: Version,
     ) -> (Self, EphemeralPublic) {
         // Generate an ephemeral key for perfect forward secrecy.
@@ -151,9 +150,9 @@ impl Handshake<AwaitingEphKey> {
 
         // Sign the challenge bytes for authentication.
         let local_signature = if self.protocol_version.has_transcript() {
-            sign_challenge(&sc_mac, &self.state.local_privkey)?
+            self.state.local_privkey.sign(&sc_mac)
         } else {
-            sign_challenge(&kdf.challenge, &self.state.local_privkey)?
+            self.state.local_privkey.sign(&kdf.challenge)
         };
 
         Ok(Handshake {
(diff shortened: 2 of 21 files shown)


p2p/src/error.rs
```diff
@@ -40,7 +39,6 @@ define_error! {
             | _ | { "public key missing" },
 
         Signature
-            [ DisplayOnly<SignatureError> ]
```
mzabaluev, line 43: This error is deliberately opaque, but we can't make the `Display` impl transparent in flex-error, so we have our own message. OK.

p2p/src/secret_connection.rs
```diff
@@ -184,24 +183,26 @@ impl Handshake<AwaitingAuthSig> {
...
             .ok_or_else(Error::missing_key)?;
 
+        #[allow(clippy::match_wildcard_for_single_variants)]
         let remote_pubkey = match pk_sum {
             proto::crypto::public_key::Sum::Ed25519(ref bytes) => {
-                ed25519::PublicKey::from_bytes(bytes).map_err(Error::signature)
+                ed25519_consensus::VerificationKey::try_from(&bytes[..])
+                    .map_err(|_| Error::signature())
             },
-            proto::crypto::public_key::Sum::Secp256k1(_) => Err(Error::unsupported_key()),
```
mzabaluev, line 191: I would rather keep the exhaustive match here, to force changes when new key types get added into protobuf.

Also, we're still working directly with proto, not through domain types. Not to be fixed in this PR, though.

tendermint/src/private_key.rs
```diff
@@ -63,5 +71,17 @@ where
...
+    // cached copy of the public key as the second half.  This is somewhat
+    // dangerous, since there's no validation that the two parts are consistent
+    // with each other, so we ignore the second half and just check consistency
+    // with the re-derived data.
+    let signing_key = Ed25519::try_from(&keypair_bytes[0..32])
+        .map_err(|_| D::Error::custom("invalid signing key"))?;
+    let verification_key = VerificationKey::from(&signing_key);
+    if &keypair_bytes[32..64] != verification_key.as_bytes() {
+        return Err(D::Error::custom("keypair mismatch"));
+    }
```
mzabaluev, line 84: Great!

tendermint/src/public_key.rs
```diff
@@ -21,6 +22,9 @@ use tendermint_proto::{
 
 use crate::{error::Error, prelude::*, signature::Signature};
 
+#[cfg(feature = "secp256k1")]
+use signature::Verifier as _;
```
mzabaluev, line 26: So with the "secp256k1" feature on, the trait's `verify` method is brought in scope, but our `VerificationKey` has the inherent `verify` method and it wins without ambiguity. Dodgy, but it works.

test/Cargo.toml
```diff
@@ -14,10 +14,10 @@ authors     = ["Alexander Simmerl <a.simmerl@gmail.com>"]
 test = true
 
 [dev-dependencies]
-ed25519-dalek = { version = "1", default-features = false, features = ["rand"] }
+ed25519-consensus = { version = "2", default-features = false }
 flex-error = { version = "0.4.4", default-features = false }
 flume = { version = "0.10", default-features = false }
-rand_core = { version = "0.5", default-features = false, features = ["std"] }
+rand_core = { version = "0.6", default-features = false, features = ["std"] }
```
mzabaluev, line 20: Huh, isis still has not updated their rand version? Another reason to switch.

mzabaluev (changes requested):
Generally looks good.

One thing which I find potentially problematic (though not a change I'm requesting now) is that ed25519-consensus does not integrate into the `signature` framework. Why is that?

hdevalence:
> Generally looks good.
> 
> One thing which I find potentially problematic (though not a change I'm requesting now) is that ed25519-consensus does not integrate into the `signature` framework. Why is that?

There's no particular reason; there wasn't a need to do it when the crate was written, and it wasn't possible to implement for trivial reasons -- AFAIR, it was because the `Signature` trait requires `AsRef<[u8]>`, but internally the `ed25519_consensus::Signature` type has two `[u8; 32]` arrays because that was slightly more convenient as an internal representation.

mzabaluev:
@hdevalence I think it's OK to not conform to the `signature` traits, as we're not going to make the Ed25519 implementation replaceable in the near term as per the previous discussion. The API is also close enough to not require significant rework if we decide we need it.

So, I think this just needs a change to do the exhaustive match for proto's key type enum, and we can merge it.

tony-iqlusion:
> AFAIR, it was because the Signature trait requires AsRef<[u8]>, but internally the ed25519_consensus::Signature type has two [u8; 32] arrays because that was slightly more convenient as an internal representation.

@hdevalence that bound is removed in a forthcoming `signature` v2.0 release, FWIW:

https://docs.rs/signature/2.0.0-pre.2/signature/index.html

hdevalence:
> @hdevalence I think it's OK to not conform to the `signature` traits, as we're not going to make the Ed25519 implementation replaceable in the near term as per discussion on Slack (recap: it's not supported by Web Crypto, corner cases addressed differently by other implementations so we need the precise behavior). The API is also close enough to not require significant rework if we decide we need it.
> 
> So, I think this just needs a change to do the exhaustive match for proto's key type enum, and we can merge it.

Cool, I updated the exhaustive match. I'd be happy to try to support the `signature` traits in a later release of the crate if it becomes relevant.

mzabaluev (approved):
Needs merging from main, but is otherwise good to go.

hdevalence:
Rebased.

thanethomson:
Was the failing check here due to clippy errors from Rust v1.66? If so, #1253's been merged and that should fix things.

hdevalence:
Rebased and did another round of clippy fixes.

xla (approved):
(͠≖ ͜ʖ͠≖)👌

mzabaluev:
Thank you @hdevalence and @xla!

hdevalence:
Thanks for merging this! Are there other changes in flight at the moment, or would it be possible to get it into a release soon?

PR #1247, opened 2022-12-06
light-client: Fix verification of blocks between two trusted heights

### LINKED ISSUE #1246: light-client: Verification fails for target height between two trusted heights
Imagine the light store contains two trusted light blocks at heights H1 and H2 (where H2 > H1). Attempting to verify a light block at height V that falls between H1 and H2 (e.g., H1 < V < H2) will incorrectly fall back to backwards verification and in case the `unstable` feature is not enabled, it will incorrectly fail with the `target_lower_than_trusted_state` error.

Given that both H1 and H2 are trusted, there is actually no reason to need backwards verification in this case as it should be enough to start at H1 and perform the usual bisection verification algorithm. It seems [this is also what the Go implementation does](https://github.com/tendermint/tendermint/blob/49502dae92047cd1fa8cf3b6b2ba348fb6de8da3/light/client.go#L594-L602).

The Rust implementation does not seem to have a nicely-contained implementation of bisection, but it instead relies on the light block store for tracking the "highest trusted light block" and all the various places are querying it.

My proposal for implementing this with minimum disruption is to add two methods to the `LightStore` trait, namely:
```rust
    /// Get the light block of greatest hight before the given height with the given status.
    fn highest_before(&self, height: Height, status: Status) -> Option<LightBlock>;

    /// Get the first light block before the given height with the trusted or verified status.
    fn highest_trusted_or_verified_before(&self, height: Height) -> Option<LightBlock>;
```
Where the second method is provided as a default implementation on the `LightStore` trait (similar to the other ones).

Then only slight modifications are needed to the stores, light client and scheduler to make this work. I have a draft implementation of this in a local branch but need to figure out how to add some longer verification tests with the current framework.

kostko:
Fixes #1246 

* [x] Wrote tests

### DIFF
--- a/.changelog/unreleased/bug-fixes/1246-verify-between-trusted.md
+++ b/.changelog/unreleased/bug-fixes/1246-verify-between-trusted.md
@@ -1,0 +1,3 @@
+- `[tendermint-light-client]` Fix verification of blocks between two trusted
+  heights
+  ([#1246](https://github.com/informalsystems/tendermint-rs/issues/1246))
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,7 +81,14 @@ jobs:
         with:
           toolchain: stable
           override: true
+      # NOTE: We test with default features to make sure things work without "unstable".
       - uses: actions-rs/cargo@v1
+        name: Test with default features
+        with:
+          command: test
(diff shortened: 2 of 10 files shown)


romac (approved):
Amazing work, thanks!

PR #1238, opened 2022-11-21
Make crypto implementations replaceable

mzabaluev:
Resolves: #1213

I took some of the changes from #1214 and made more extensive use of them in tendermint hashing.

* [x] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [x] Wrote tests
* [x] Added entry in `.changelog/`

PR #1252, opened 2022-12-16
Investigate some design tweaks over Crypto Implementation

Farhad-Shabani:
As a reference for suggestion made in #1238

* [ ] Wrote tests


Wizdave97:
This is a much better implementation for substrate chains

tendermint/src/merkle.rs
```diff
@@ -61,88 +61,31 @@ pub trait MerkleHash {
...
-        Digest::update(self, right);
-
-        // Finalize the digest, reset the hasher state
-        let digest = self.finalize_reset();
-
-        copy_to_hash(digest)
+    /// Compute a simple Merkle root from vectors of arbitrary byte vectors.
+    /// The leaves of the tree are the bytes of the given byte vectors in
+    /// the given order.
+    fn simple_hash_from_byte_vectors(&mut self, byte_vecs: &[Vec<u8>]) -> Hash {
```
mzabaluev, line 68: Why does this need to be converted to a method?
With the freestanding function, we can't misuse the API by e.g. calling it on a digest object with non-reset digest state.
`simple_hash_from_byte_vectors` is a one-stop-shop function that creates the digest state internally, hashes the input and returns the result.

tendermint/src/merkle.rs
```diff
@@ -61,88 +61,31 @@ pub trait MerkleHash {
...
+        self.hash_byte_vectors(byte_vecs)
     }
 }
 
 /// A wrapper for platform-provided host functions which can't do incremental
 /// hashing. One unfortunate example of such platform is Polkadot.
-pub struct NonIncremental<H>(PhantomData<H>);
-
-impl<H> Default for NonIncremental<H> {
+pub struct DefaultMerkleHash;
```
mzabaluev, line 75: I think this is wrong, we specifically don't want a stateless function to be the basis for default hashing implementation, when in most programming environments we can use the `sha2` crate.

tendermint/src/merkle.rs
```diff
@@ -1,49 +1,49 @@
...
-pub trait MerkleHash {
+pub trait MerkleHash: DigestSha256 {
     // tmhash({})
     // Pre and post-conditions: the hasher is in the reset state
     // before and after calling this function.
-    fn empty_hash(&mut self) -> Hash;
+    fn empty_hash(&mut self) -> Hash {
+        // Get the output of an empty digest state.
+        Self::digest_sha256(&Vec::with_capacity(0))
+    }
```
mzabaluev, line 24: Why bake the implementation into the trait?
This pattern is used elsewhere, for example, in `VerificationPredicates`, but I don't think it's needed here.

tendermint/src/merkle.rs
```diff
@@ -154,7 +97,7 @@ mod tests {
         let empty_tree_root = &hex::decode(empty_tree_root_hex).unwrap();
         let empty_tree: Vec<Vec<u8>> = vec![vec![]; 0];
 
-        let root = simple_hash_from_byte_vectors::<Sha256>(&empty_tree);
+        let root = DefaultMerkleHash.simple_hash_from_byte_vectors(&empty_tree);
```
mzabaluev, line 100: This looks non-idiomatic, the object instance has no purpose here, as explained above.

tendermint/src/merkle.rs
```diff
@@ -61,88 +61,31 @@ pub trait MerkleHash {
...
-    fn empty_hash(&mut self) -> Hash {
-        // Get the output of an empty digest state.
-        let digest = self.finalize_reset();
-        copy_to_hash(digest)
-    }
-
-    fn leaf_hash(&mut self, bytes: &[u8]) -> Hash {
-        // Feed the data to the hasher, prepended with 0x00.
-        Digest::update(self, [0x00]);
-        Digest::update(self, bytes);
```
mzabaluev, line 86: Note how this does not allocate any memory temporarily or copy data around. This is what we want most of our users to actually use, while the few platforms where the `NonIncremental` wrapper is needed could be kept as integration freak cases.

mzabaluev (commented):
Some initial comments, mostly flak. I think we'll need to come up with a minimal set of changes that make sense and reduce coupling between implementations and specific hash functions.

tendermint/Cargo.toml
```diff
@@ -50,16 +50,15 @@ tendermint-proto = { version = "0.28.0", default-features = false, path = "../pr
 time = { version = "0.3", default-features = false, features = ["macros", "parsing"] }
 zeroize = { version = "1.1", default-features = false, features = ["zeroize_derive", "alloc"] }
 flex-error = { version = "0.4.4", default-features = false }
-sha2 = { version = "0.10", optional = true, default-features = false }
+sha2 = { version = "0.10", default-features = false }
```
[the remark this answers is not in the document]
Farhad-Shabani: Yeah, comparatively, it doesn't buy much. I looked it as a side objective

tendermint/src/merkle.rs
```diff
@@ -1,49 +1,49 @@
...
-where
-    H: MerkleHash + Default,
-{
-    let mut hasher = H::default();
-    hasher.hash_byte_vectors(byte_vecs)
-}
-
 /// Implementation of Merkle tree hashing for Tendermint.
-pub trait MerkleHash {
+pub trait MerkleHash: DigestSha256 {
```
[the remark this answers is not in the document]
Farhad-Shabani: It already do so. With the `DigestSha256` trait (included types from higher-level `HashType` trait), basically you can introduce any type of hashing functions. but I caught the issue. It should be renamed to make more sense (e.g.as DigestHash)

tendermint/src/merkle.rs
```diff
@@ -1,49 +1,49 @@
...
+    }
 
     // tmhash(0x00 || leaf)
     // Pre and post-conditions: the hasher is in the reset state
     // before and after calling this function.
-    fn leaf_hash(&mut self, bytes: &[u8]) -> Hash;
+    fn leaf_hash(&mut self, bytes: &[u8]) -> Hash {
+        let mut buf = Vec::with_capacity(1 + bytes.len());
+        buf.push(0x00);
+        buf.extend_from_slice(bytes);
```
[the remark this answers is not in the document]
Farhad-Shabani: Did so, looking for a minimal set of changes compared to the orig files and hoping this be covered in a separate PR for this purpose. but Right, if there is a demand and a performance matter, would be better to get it back and also have it here.

tendermint/src/merkle.rs
Farhad-Shabani: Likewise, I had looked at it from the opposite angle, but with the same concern. Having a bare function, so Merkle hasher could be triggered by a fake similar one. I encapsulated it to operate only on that data. btw, could be investigate more or even leave it untouched for now. alright

tendermint/src/merkle.rs
Farhad-Shabani: I don’t know if I catch you correctly, but as it stands here, it is bounded by `sha2::Sha256`

Farhad-Shabani:
In light of the need Mikhail aptly raised for incremental hashing and the comments mostly arose because of losing it, I've reverted that part. 
Here, just remains the suggestions you may want to consider, including the refactored parts for verifying signatures verification through an interface.

tony-iqlusion:
@Farhad-Shabani I don't understand what advantage that signature verification interface is supposed to provide over [`signature::Verifier`](https://docs.rs/signature/latest/signature/trait.Verifier.html) and similar other signature verification traits in the `signature` crate.

It seems needlessly convoluted, and also has the disadvantage that it isn't natively supported by commonly used Rust ecosystem cryptography crates.

Farhad-Shabani:
@tony-iqlusion agree with you on this. Actually, I've tried it first, but having `PublicKey` as enum, supporting different key/signature variants made it tricky to refactor and provide the needed interface for `validator/Info`. 
But, thanks bringing this up, made me rethought and came up with a new idea. Check that please. Now the `validator/Info`  can provide the generic interface through `verify_with` method. if it works overall, it can be fastened

PR #1257, opened 2023-01-04
tendermint: fix minimal-versions correctness for `bytes`

tony-iqlusion:
The `tendermint` crate uses features of `bytes` introduced in version 1.2, namely a `From<Vec<u8>>` impl for `Bytes`.

See https://github.com/cosmos/cosmos-rust/pull/314 for additional context


xla (approved):
(͠≖ ͜ʖ͠≖)👌

Issue #1219, opened 2022-10-27
Support `LightClientAttackEvidence`


mzabaluev:
### Description

Properly encode `LightClientAttackEvidence` variant of `Evidence` messages with data provided by the client.
Currently the `Evidence` domain type lacks required data in its `LightClientAttackEvidence` variant.

Required by https://github.com/informalsystems/hermes/issues/2772

### Definition of "done"

- A `LightClientAttackEvidence` struct and a `Evidence::LightClientAttack` variant carrying it is provided in `tendermint`, with encoding support from/to the protobuf `LightClientAttackEvidence` message and serde implementations for RPC.
- The supervisor code in `tendermint-light-client` is changed to report `LightClientAttackEvidence` data upon detecting a fork.

PR #1259, opened 2023-01-10
Fill in the `LightClientAttackEvidence` struct  and conversions to/from Protobuf

romac:
Closes: #1219

* [ ] Wrote tests

### DIFF
--- a/tendermint/src/evidence.rs
+++ b/tendermint/src/evidence.rs
@@ -9,39 +9,34 @@ use serde::{Deserialize, Serialize};
 use tendermint_proto::{
     google::protobuf::Duration as RawDuration,
     types::{
-        evidence::{Sum as RawSum, Sum},
-        DuplicateVoteEvidence as RawDuplicateVoteEvidence, Evidence as RawEvidence,
-        EvidenceList as RawEvidenceList, EvidenceParams as RawEvidenceParams,
+        evidence::Sum as RawSum, DuplicateVoteEvidence as RawDuplicateVoteEvidence,
+        Evidence as RawEvidence, EvidenceList as RawEvidenceList,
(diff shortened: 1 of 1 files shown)


ancazamfir:
Besides these changes I think we also need to put an API in place to be used by hermes. imho `tendermint-rs` should be filling in `Evidence::LightClientAttackEvidence` (e.g. the `byzantine_validators`) based on the two conflicting and maybe a common light block.
Also not clear to me if hermes will call the `broadcast_evidence()` RPC or we do this via the new API (call to `report_evidence()`?) and in the latter case if there are more changes required.

So maybe we need to work in parallel with the hermes PR here, you probably have better idea.

romac:
> Besides these changes I think we also need to put an API in place to be used by hermes. imho tendermint-rs should be filling in Evidence::LightClientAttackEvidence (e.g. the byzantine_validators) based on the two conflicting and maybe a common light block.

Yes, I am working on that right now :)

romac:
> Also not clear to me if hermes will call the broadcast_evidence() RPC or we do this via the new API (call to report_evidence()?) and in the latter case if there are more changes required.

What's this new API? Do you have a link to it?

Also should we keep submitting the `MisbehaviorEvidence` message to the chain like we do currently?

ancazamfir:
> > Also not clear to me if hermes will call the broadcast_evidence() RPC or we do this via the new API (call to report_evidence()?) and in the latter case if there are more changes required.
> 
> What's this new API? Do you have a link to it?

the new API to be used by hermes to submit evidence to the node. 
> 
> Also should we keep submitting the `MisbehaviorEvidence` message to the chain like we do currently?
yes, that should stay.

romac:
Superseded by #1260

PR #1264, opened 2023-02-03
Derive PartialEq, Eq on ProdCommitValidator

mzabaluev:
Hermes needs `PartialEq` and `Eq` implemented on `ProdVerifier`. No idea why, but it's simple to add the derives
rather than fighting this downstream.

Also add test code that ensures `ProdVerifier` supports the common derived traits.

* [x] Wrote tests


ancazamfir (approved):
lgtm! thanks @mzabaluev!

xla (approved):

PR #1272, opened 2023-02-24
debug: improve Debug impl for Ed25519 VerificationKeys

hdevalence:
Since the `ed25519_consensus::VerificationKey` is no longer used, the `VerificationKey`s go back to having raw debug output that's hard to read when it appears in logs.  This commit restores the hex-formatting previously inherited from `ed25519_consensus`, allowing nice output like (in use in `pd` downstream):

```
2023-02-24T03:34:57.002766Z DEBUG abci:EndBlock{height=187}:end_block:staking:build_tendermint_validator_updates: updates=[Update { pub_key: Ed25519(7423aa71eb844cdde4dc0b8c72ef1225931dc54c21a102cbe2fded2645d4e814), power: Power(50000000000) }]
```

In the future, another option could be to have the key be base64-encoded, to match Tendermint's data structures, but this is a less obtrusive change (not requiring a base64 encoder) to restore functionality first.


xla (approved):
🦾

PR #1273, opened 2023-02-24
Prepare `0.29.1` release of `tendermint`

hdevalence:
This commit only bumps the patch version of the `tendermint` crate; since it's compatible and will be auto-selected, it didn't seem important to bump all of the versions of all of the other crates, or bump their dependency specifications.


thanethomson:
The trouble with only bumping the one crate's patch version is, say you need to make a patch update to a different crate at a later stage (e.g. the tendermint-abci crate) after having pushed the `v0.29.1` tag to cut the release - what tag do you push to the Git repo to trigger the next release? Then you'll already have used `v0.29.1`.

This is partly why we version all crates in lockstep.

The only other alternative is to delete the `v0.29.1` tag and push it again at a later stage if you bump a different crate's patch version to `v0.29.1` and want to trigger its release. We haven't tested this approach, however, with our current release script.

hdevalence:
Sounds reasonable, I replaced all of the `0.29.0` versions by `0.29.1` instead.

xla (approved):
🦾

thanethomson (approved):
LGTM! Feel free to merge and then tag whenever you're ready.

PR #1285, opened 2023-03-10
Bump `ed25519` => v2; `k256` => v0.13; `signature` => v2

tony-iqlusion:
Closes: #1290

Bumps all of the above dependencies to their latest versions, which use the new `signature` v2 release:

https://github.com/RustCrypto/traits/pull/1211

* [ ] Wrote tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,3 @@ exclude = [
 [profile.release.package.tendermint-light-client-js]
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
-
-# WORKAROUND for fizyk20/generic-array#133
-# Docs build is broken without this until above PR is merged/released
-[patch.crates-io.generic-array]
-git = "https://github.com/fjarri/generic-array"
-branch = "zeroize-docs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,9 @@ exclude = [
 [profile.release.package.tendermint-light-client-js]
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
+
+# WORKAROUND for fizyk20/generic-array#133
+# Docs build is broken without this until above PR is merged/released
+[patch.crates-io.generic-array]
+git = "https://github.com/fjarri/generic-array"
+branch = "zeroize-docs"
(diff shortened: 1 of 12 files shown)


tony-iqlusion:
Note: there's a PR to use the `signature` v2 crate's traits with `ed25519-consensus` here:

https://github.com/penumbra-zone/ed25519-consensus/pull/6

tony-iqlusion:
Hmmm, the docs failure seems to be coming from `generic-array`:

https://github.com/informalsystems/tendermint-rs/actions/runs/4386941224/jobs/7681624555#step:4:290

tony-iqlusion:
Appears to be a bug in the `generic-array` crate, unfortunately.

I can see if I can configure the docs job to skip that crate.

tony-iqlusion:
D'oh, `cargo doc --exclude` only works for crates in the same workspace.

Not sure how to fix the doc build without an upstream fix in `generic-array`. The best I can do is point it at that PR using `patch.crates-io`

Cargo.toml
```diff
@@ -24,3 +24,9 @@ exclude = [
 [profile.release.package.tendermint-light-client-js]
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
+
+# WORKAROUND for fizyk20/generic-array#133
+# Docs build is broken without this until above PR is merged/released
+[patch.crates-io.generic-array]
+git = "https://github.com/fjarri/generic-array"
+branch = "zeroize-docs"
```
tony-iqlusion, line 32: I added this as a workaround which fixes the docs build in CI, but it's not a great fix

romac: The fix was [just released](https://github.com/fizyk20/generic-array/pull/133#issuecomment-1486018064) in `generic-array` 0.14.7, so I took the liberty of removing the patched crate.

tony-iqlusion:
The build failure is a semi-recent nightly regression. 

The docs build on `nightly-2023-03-01`, for example. (I'm not sure why it's using nightly in the first place?)

tony-iqlusion:
Pinned the nightly used to build the docs and it passed, but that seems like an unrelated build failure which would impact any branch

romac:
> I'm not sure why it's using nightly in the first place?

I believe we are building the docs on nightly to match [what docs.rs is doing](https://docs.rs/about/builds).

romac (approved):

PR #1293, opened 2023-04-05
Parse and fetch proto dependencies from `buf.lock` in the repository source

### LINKED ISSUE #1289: Can not generate tendermint-proto from source with buf dependencies
### Description

Originally reported as https://github.com/cometbft/cometbft/issues/565

tendermint-proto-compiler needs to be able to process buf dependencies in the `buf.lock` file in order to fetch third-party
protobuf modules used by CometBFT proto module source in releases starting from 0.37.0.
For #1193 I have scrounged by generating off the `v0.37.0-alpha.1` release and checking that the proto source has not changed between then and the eventual 0.37.x releases, but to adapt to any future changes we need to adapt to the buf way of managing protobuf dependencies.

### Definition of "done"

The tool that's used to generate code for tendermint-proto (either t-p-compiler or, if you feel ambitious, switch completely to buf.build) can deal with the buf dependencies in CometBFT releases since `v0.37.0`.

mzabaluev:
Resolves #1289

* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/improvements/1293-buf-proto-deps.md
+++ b/.changelog/unreleased/improvements/1293-buf-proto-deps.md
@@ -1,0 +1,3 @@
+- [`tools/proto-compiler`] Parse and fetch proto dependencies as listed in the
+  `buf.lock` file
+  ([\#1293](https://github.com/informalsystems/tendermint-rs/pull/1293)).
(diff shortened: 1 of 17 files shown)


romac (approved):
Great stuff!

PR #1297, opened 2023-04-16
Prepare release `v0.32.0`

erwanor:
Follow-up to #1296, trivial API change.

* [ ] Wrote tests


hdevalence (changes requested):
Shouldn't this be 0.31.1 ? There's no breaking changes.

erwanor:
Ah yes, let's do that.

Issue #1294, opened 2023-04-13
light-client: new lib API for misbehaviour header verification


ancazamfir:
### Description
Currently ibc-rs uses the `verify()` function to verify headers received in `MsgUpdateClient`
https://github.com/informalsystems/tendermint-rs/blob/db9cb212974937a38fb64af4f1170c3070b10dc6/light-client-verifier/src/verifier.rs#L260-L273
There is a need to verify headers received in `MsgSubmitMisbehaviour`. The verification for these headers is a bit more relaxed in order to catch FLA attacks. In particular the "header in the future" check for the header should be skipped from `validate_against_trusted()`.
https://github.com/informalsystems/tendermint-rs/blob/db9cb212974937a38fb64af4f1170c3070b10dc6/light-client-verifier/src/verifier.rs#L165-L170

Currently the ibc-rs makes explicitly these calls from `verify()`:
```
        ...verify_validator_sets(&untrusted);
        ...verify_commit_against_trusted(&untrusted, &trusted, options));
        ...verify_commit(&untrusted));
```
and also does all the checks from `validate_against_trusted()` except the header in the future one. Code is hard to follow as the checks are spread across multiple functions and also one needs to dive into the library to understand what it is being checked and where. 

This [PR](https://github.com/informalsystems/tendermint-rs/pull/1292) reorganizes the code around client update/ misbehaviour. New code will temporarily call the same `verify()` API for both update and misbehaviour headers but we need to eventually provide the new API. 
Intuitively this should also be needed for the light client detector in `tendermint-rs` for header verification from witnesses.
 

### Definition of "done"
- [ ] new `verify_misbehaviour_header` API implementation and documentation available

PR #1300, opened 2023-04-17
Add `Verifier::verify_misbehaviour_header` for verifying headers coming from a misbehaviour evidence

romac:
Closes: #1294

* [ ] Wrote tests

### DIFF
--- a/light-client-verifier/src/verifier.rs
+++ b/light-client-verifier/src/verifier.rs
@@ -52,7 +52,7 @@ impl From<Result<(), VerificationError>> for Verdict {
 /// - [TMBC-VAL-COMMIT.1]
 pub trait Verifier: Send + Sync {
     /// Verify a header received in a `MsgUpdateClient`.
-    fn verify(
+    fn verify_update_header(
         &self,
         untrusted: UntrustedBlockState<'_>,
         trusted: TrustedBlockState<'_>,
@@ -278,7 +278,7 @@ where
     /// `trusted.next_validators.hash() == trusted.next_validators_hash`,
     /// as typically the `trusted.next_validators` validator set comes from the relayer,
     /// and `trusted.next_validators_hash` is the hash stored on chain.
-    fn verify(
+    fn verify_update_header(
         &self,
         untrusted: UntrustedBlockState<'_>,
         trusted: TrustedBlockState<'_>,
@@ -369,7 +369,7 @@ mod tests {
             clock_drift: Default::default(),
         };
 
-        let verdict = vp.verify(
+        let verdict = vp.verify_update_header(
             light_block_2.as_untrusted_state(),
             light_block_1.as_trusted_state(),
             &opt,
--- a/light-client-verifier/src/verifier.rs
+++ b/light-client-verifier/src/verifier.rs
@@ -220,18 +220,6 @@ where
         Verdict::Success
     }
 
-    /// Validate an `UntrustedBlockState` coming from a misbehaviour evidence,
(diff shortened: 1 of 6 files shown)


light-client-verifier/src/verifier.rs
```diff
@@ -146,29 +158,51 @@ where
...
-            options.trusting_period,
-            now,
-        ));
-
         // Ensure the header isn't from a future time
         verdict!(self.predicates.is_header_from_past(
             untrusted.signed_header.header.time,
             options.clock_drift,
             now,
         ));
```
ancazamfir, line 176: I was wondering why this check is in a function called "validate_against_trusted" as it doesn't need the trusted state.

romac: Maybe that's cleaner indeed!

romac: 9e0a4c851cbd9894f12b5806a57e0eb8ee757ef9

light-client-verifier/src/verifier.rs
```diff
@@ -266,6 +299,28 @@ where
...
+    fn verify_misbehaviour_header(
+        &self,
+        untrusted: UntrustedBlockState<'_>,
+        trusted: TrustedBlockState<'_>,
+        options: &Options,
+        now: Time,
+    ) -> Verdict {
+        ensure_verdict_success!(self.verify_validator_sets(&untrusted));
+        ensure_verdict_success!(
+            self.validate_misbehaviour_against_trusted(&untrusted, &trusted, options, now)
```
plafer, line 322: I think it would be clearer to use `self.validate_against_trusted()` directly to see the correspondence with the `verify()` implementation

light-client-verifier/src/verifier.rs
```diff
@@ -266,6 +299,28 @@ where
     ) -> Verdict {
         ensure_verdict_success!(self.verify_validator_sets(&untrusted));
         ensure_verdict_success!(self.validate_against_trusted(&untrusted, &trusted, options, now));
+        ensure_verdict_success!(self.check_header_is_from_past(&untrusted, options, now));
+        ensure_verdict_success!(self.verify_commit_against_trusted(&untrusted, &trusted, options));
+        ensure_verdict_success!(self.verify_commit(&untrusted));
```
plafer, line 304: Since this implementation is a subset of `verify_misbehaviour_header()`, maybe we could call `verify_misbehaviour_header()`, and then `check_header_is_from_past()`? This would make it clear that there's only one additional check between an update and a misbehaviour header.

Although then it might be confusing to call a "misbehaviour" in an "update" verification? Or maybe we do some renaming? Not sure...

light-client-verifier/src/verifier.rs
romac: f8324587

light-client-verifier/src/verifier.rs
```diff
@@ -51,32 +51,44 @@ impl From<Result<(), VerificationError>> for Verdict {
 /// - [TMBC-VAL-CONTAINS-CORR.1]
 /// - [TMBC-VAL-COMMIT.1]
 pub trait Verifier: Send + Sync {
-    /// Perform the verification.
+    /// Verify a header received in a `MsgUpdateClient`.
     fn verify(
```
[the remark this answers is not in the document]
romac: 1c1bf009

light-client-verifier/src/verifier.rs
romac: I would rather keep it as is, since it's only a few lines of code and the one check that is omitted is clearly documented in the doc comment.

PR #1303, opened 2023-04-26
Remove nightly-only options in rustfmt config

romac:
These options cannot be used on stable Rust, which we use for developing the library and on CI for checking the formatting. As such they should be removed from the `.rustfmt.toml` config, as they can otherwise be picked up by some IDEs which will automatically use the nightly `rustfmt` to honor the unstable options.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,19 +1,6 @@
-unstable_features = true
-
-# comments
-comment_width = 110
-normalize_comments = true
-wrap_comments = true
-
 # imports
-group_imports = "StdExternalCrate"
-imports_granularity = "Crate"
 reorder_imports = true
 
 # strings
-format_strings = false
 max_width = 100
 match_block_trailing_comma = true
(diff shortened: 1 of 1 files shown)


thanethomson (approved):
Any idea when these formatting options will make their way into stable? This file's been configured like this for ages now, and these formatting conventions have served us pretty well.

But I'm also all for not depending on nightly :slightly_smiling_face:

romac:
We're actually formatting and checking the code with the stable options only so those had no effect up to now afaict. At least it on my machine, running `cargo fmt` or formatting with rust-analyzer on stable does not pick them up.

I only noticed this because my new Neovim distrib (AstroNvim) seems to use nightly rustfmt by default, hence this PR to ensure it does not reformat all the code I touch using the unstable settings.

romac:
No idea as to when these options will become stable but I wouldn't hold my breath given that they've unstable for years now.

PR #1318, opened 2023-05-22
fix(rpc): `validator_updates` in block results might be explicit null

xoac:
* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/bug-fixes/1318-deserialize_block_results.md
+++ b/.changelog/unreleased/bug-fixes/1318-deserialize_block_results.md
@@ -1,0 +1,1 @@
+- `[tendermint-rpc]` - null is deserialized as empty `value` for `EventAttribute` (abci v0_37)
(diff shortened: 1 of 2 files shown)


mzabaluev (changes requested):
It seems that the change does not match the title of the pull request. `validator_updates` is a field in `ResponseEndBlock`, isn't it?

Can you file an issue detailing the problem?

Also, it would be good to provide a test fixture, hopefully one obtained from a real-world node setup.

xoac:
@mzabaluev replaced this one with #1318 as the issue  was that HttpClient ignore compact node when query v0.34 last block results. 

I think this PR is also valid as `value` in `EventAttribute` can be explicitly null.

mzabaluev:
> I think this PR is also valid as value in EventAttribute can be explicitly null.

I have not observed that with 0.37 yet, and I believe the serialization code in CometBFT has been changed to not produce null values. The 0.34 dialect deserialization does allow null, though. Can you provide an example where a null value is produced?

xoac:
Probably my wrong assumption. For now I query only version 0.34

PR #1321, opened 2023-06-14
RPC: Provide earliest block information from /status

mdyring:
This is a trivial addition of missing /status RPC fields, specifically the "earliest block" information.

This information is useful and tells us the range of blocks that can be queried from a node. To be used for a Cosmos block crawler/indexer project.

I deemed the change so minor as to not warrant a changelog entry.

* [x] Wrote tests

### DIFF
--- a/rpc/src/endpoint/status.rs
+++ b/rpc/src/endpoint/status.rs
@@ -41,6 +41,20 @@ impl crate::Response for Response {}
 /// Sync information
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct SyncInfo {
+    /// Earliest block hash
+    #[serde(with = "tendermint::serializers::hash")]
+    pub earliest_block_hash: Hash,
+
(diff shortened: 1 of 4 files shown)


rpc/src/endpoint/status.rs
```diff
@@ -41,6 +41,20 @@ impl crate::Response for Response {}
 /// Sync information
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct SyncInfo {
+    /// Earliest block hash
```
mzabaluev, line 44: Now only the `catching_up` field is not represented in this structure.
Maybe we should add it as well.

mzabaluev (approved):
Thanks for adding this! Looks good.

rpc/src/endpoint/status.rs
mdyring: Catching up is already there (end of SyncInfo struct) 👍

PR #1322, opened 2023-07-04
fix `from_str` serializer

noot:
* [ ] Wrote tests

This fixes the `from_str` serialization. I discovered this while trying to serialize and deserialize the response type for `broadcast_tx_commit` while making a mock tendermint server; the `gas_wanted` field and a few others were serialized as integers, but were decoded as a string, causing an error `invalid type: integer `0`, expected a string at line 1 column 98`. `Display` formats the integer without quotes, but we wanted with quotes.

Before:
```
"{\"jsonrpc\":\"2.0\",\"id\":1,\"result\":{\"check_tx\":{\"code\":0,\"data\":[],\"log\":\"\",\"info\":\"\",\"gas_wanted\":0,\"gas_used\":0,\"events\":[],\"codespace\":\"\",\"sender\":\"\",\"priority\":0,\"mempool_error\":\"\"},\"deliver_tx\":{\"code\":0,\"data\":[],\"log\":\"\",\"info\":\"\",\"gas_wanted\":0,\"gas_used\":0,\"events\":[],\"codespace\":\"\"},\"hash\":\"\",\"height\":\"1\"},\"error\":null}"
```

After:
```
"{\"jsonrpc\":\"2.0\",\"id\":1,\"result\":{\"check_tx\":{\"code\":0,\"data\":null,\"log\":\"\",\"info\":\"\",\"gas_wanted\":\"0\",\"gas_used\":\"0\",\"events\":[],\"codespace\":\"\",\"sender\":\"\",\"priority\":\"0\",\"mempool_error\":\"\"},\"deliver_tx\":{\"code\":0,\"data\":null,\"log\":\"\",\"info\":\"\",\"gas_wanted\":\"0\",\"gas_used\":\"0\",\"events\":[],\"codespace\":\"\"},\"hash\":\"0000000000000000000000000000000000000000000000000000000000000000\",\"height\":\"1\"},\"error\":null}"
```

Also made `rpc::Wrapper::new_with_id` public because I'm using it externally.

### DIFF
--- a/rpc/src/response.rs
+++ b/rpc/src/response.rs
@@ -69,7 +69,6 @@ impl<R> Wrapper<R> {
         }
     }
 
-    #[cfg(test)]
     pub fn new_with_id(id: Id, result: Option<R>, error: Option<ResponseError>) -> Self {
         Self {
             jsonrpc: Version::current(),
--- a/proto/src/serializers/from_str.rs
+++ b/proto/src/serializers/from_str.rs
@@ -23,5 +23,5 @@ where
     S: Serializer,
     T: core::fmt::Display,
 {
-    format!("{value}").serialize(serializer)
+    value.to_string().serialize(serializer)
 }


rpc/src/response.rs
```diff
@@ -69,7 +69,6 @@ impl<R> Wrapper<R> {
         }
     }
 
-    #[cfg(test)]
```
thanethomson, line 72: IIRC this method was only ever supposed to be used in tests. Why the change?

noot: I wanted to use this type externally and there's no other way to construct it afaik. I can revert this and just make my own `Wrapper` type if you'd prefer

thanethomson (approved):

thanethomson:
Thanks @noot!

PR #1340, opened 2023-08-02
Make `total_voting_power` optional

rootulp:
Motivated by https://github.com/celestiaorg/celestia-core/issues/1052

I'm a Rust noob so would appreciate a thorough review and/or feedback on if this is desirable.

* [x] Updated all relevant documentation in docs (I don't see any references to `total_voting_power` in docs)
* [ ] Wrote tests

### DIFF
--- a/tendermint/src/validator.rs
+++ b/tendermint/src/validator.rs
@@ -20,7 +20,7 @@ use crate::{
 pub struct Set {
     validators: Vec<Info>,
     proposer: Option<Info>,
-    total_voting_power: vote::Power,
+    total_voting_power: Option<vote::Power>,
 }
 
 impl Set {
@@ -29,12 +29,14 @@ impl Set {
         Self::sort_validators(&mut validators);
 
         // Compute the total voting power
-        let total_voting_power = validators
-            .iter()
-            .map(|v| v.power.value())
-            .sum::<u64>()
-            .try_into()
-            .unwrap();
+        let total_voting_power = Some(
+            validators
+                .iter()
+                .map(|v| v.power.value())
+                .sum::<u64>()
+                .try_into()
+                .unwrap(),
+        );
 
         Set {
             validators,
@@ -77,7 +79,7 @@ impl Set {
 
     /// Get total voting power
     pub fn total_voting_power(&self) -> vote::Power {
-        self.total_voting_power
+        self.total_voting_power.unwrap()
     }
 
     /// Sort the validators according to the current Tendermint requirements
@@ -277,7 +279,7 @@ tendermint_pb_modules! {
             RawValidatorSet {
                 validators: value.validators.into_iter().map(Into::into).collect(),
                 proposer: value.proposer.map(Into::into),
-                total_voting_power: value.total_voting_power.into(),
+                total_voting_power: value.total_voting_power.unwrap().into(),
             }
         }
     }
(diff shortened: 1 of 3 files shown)


tendermint/src/validator.rs
```diff
@@ -77,7 +79,7 @@ impl Set {
 
     /// Get total voting power
     pub fn total_voting_power(&self) -> vote::Power {
-        self.total_voting_power
+        self.total_voting_power.unwrap()
```
mzabaluev, line 82: Is there any other way to deal with this than turning it into a panic bomb?
Perhaps the getter should return `Option<vote::Power>` if this is really meant to become optional.

tendermint/src/validator.rs
```diff
@@ -277,7 +279,7 @@ tendermint_pb_modules! {
             RawValidatorSet {
                 validators: value.validators.into_iter().map(Into::into).collect(),
                 proposer: value.proposer.map(Into::into),
-                total_voting_power: value.total_voting_power.into(),
+                total_voting_power: value.total_voting_power.unwrap().into(),
```
mzabaluev, line 282: This should perhaps convert a `None` into the default value for the field (0).

tendermint/src/validator.rs
rootulp: Thanks! Makes sense to me so fixed in eab906d1b1287b42d24e552a10876e9357806b36

tendermint/src/validator.rs
```diff
@@ -29,12 +29,14 @@ impl Set {
...
-            .sum::<u64>()
-            .try_into()
-            .unwrap();
+        let total_voting_power = Some(
+            validators
+                .iter()
+                .map(|v| v.power.value())
+                .sum::<u64>()
+                .try_into()
+                .unwrap(),
```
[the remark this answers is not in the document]
rootulp: Done in eab906d1b1287b42d24e552a10876e9357806b36

tendermint/src/validator.rs
rootulp: Done in eab906d1b1287b42d24e552a10876e9357806b36

rootulp:
I can't explicitly request a reviewer but I'd appreciate @thanethomson 's opinion on if it makes sense to do this.

ratankaliani:
Are there any updates on the review states of this PR?

ratankaliani:
Bump to above ^

mzabaluev:
As we are looking closer into this change, it raises some deeper questions about the way this field is serialized.

It's not a good idea to denormalize the total voting power value on the `validator::Set` domain type. Instead, if the field is not present in the deserialized JSON or protobuf, it should be recomputed from the validators' voting power values. There is also an issue of trust with impications on protocol security: if we accept the total voting power value without checking it against the sum of listed validators' powers, this means `validator::Set` cannot be assumed correct by construction.

We had to skip the field to match the JSON schema in at least one place in #1292.
Unfortunately, `validator::Set` has also had `Serialize` and `Deserialize` derived, the `total_voting_power` field included, and this deserialization schema has probably been in use by parties unknown. I think we can relax this into making `total_voting_power` optional on deserialization, but this changes the schema of the values accepted by deserialization. OTOH taking the total voting value without validating that it equals the sum of validator's voting powers implies trust in the sender computing it right anyway, so I doubt we'd be breaking code that was not broken for other reasons.

mzabaluev:
Closing to implement a proper fix for #1348.

PR #1346, opened 2023-09-05
rpc: remove HTTP proxy support

### LINKED ISSUE #1342: RUSTSEC-2023-0052: webpki: CPU denial of service in certificate path building
> webpki: CPU denial of service in certificate path building

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `webpki`                      |
| Version             | `0.22.0`                   |
| Date                | 2023-08-22                         |

When this crate is given a pathological certificate chain to validate, it will
spend CPU time exponential with the number of candidate certificates at each
step of path building.

Both TLS clients and TLS servers that accept client certificate are affected.

This was previously reported in
&lt;https://github.com/briansmith/webpki/issues/69&gt; and re-reported recently
by Luke Malinowski.

`rustls-webpki` is a fork of this crate which contains a fix for this issue
and is actively maintained.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2023-0052.html) for additional details.

mzabaluev:
Closes #1342 by removing all use of hyper-proxy, and the HTTP proxy support in tendermint-rpc client which requires it.

* [ ] Wrote tests

### DIFF
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -40,7 +40,6 @@ http-client = [
   "futures",
   "http",
   "hyper",
-  "hyper-proxy",
   "hyper-rustls",
   "tokio/fs",
   "tokio/macros",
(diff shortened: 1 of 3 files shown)


thanethomson:
@tony-iqlusion, IIRC you needed HTTP proxy support, which drove the work in #820 right?

mzabaluev:
HTTP CONNECT support seems doable in house by following [the doc](https://docs.rs/hyper/latest/hyper/upgrade/index.html) on protocol upgrades.

romac:
How bad is that vulnerability in practice? I would expect that people will use proxies that they know and trust, and worst case that their VM running Hermes coming to a halt in case the proxy server is compromised to not have major consequences but I could be wrong. So perhaps before removing code that might be helpful to some of our users, although I am always partial to deleting  code and tidying up dependencies, we should (a) assess the actual severity of the vulnerability for our use case and (b) see if we can wait it out a bit and push for the community to move to a better maintained alternative. What do you think @mzabaluev @tony-iqlusion  @thanethomson?

mzabaluev:
@romac I have [commented](https://github.com/informalsystems/tendermint-rs/issues/1342#issuecomment-1706516638) on the issue. Indeed the potential for exploiting this is quite limited. I've also started looking into implementing HTTP CONNECT transport on our own, which does not seem like a big thing to do (even though we need to set aside a day or two for this).

mzabaluev:
Let's preserve the functionality and go with #1362 instead.

PR #1350, opened 2023-09-11
Validate construction and deserialization of `validator::Set`

### LINKED ISSUE #1348: Serialization for `validator::Set` and related proto structs is inconsistent, lacks validation
#1340 attempted to fix an issue with deserializing `ValidatorSet` data without the `total_voting_power` field being present in the serialized message.
The field is actually redundant, its value is computable as the sum of voting powers of the validators listed in the message. If the field is not present in the deserialized JSON or protobuf, its value should be recomputed. There is also an issue of trust with impications on protocol security: if we accept the total voting power value without checking it against the sum of listed validators' powers, this means `validator::Set` cannot be assumed correct by construction.

In the serialization bolted onto the tendermint-proto types generated for `ValidatorSet`, we skipped the field to match the serialization schema of misbehavior evidence, where it is not present in at least one place, in #1292.

Unfortunately, `validator::Set` has also had `Serialize` and `Deserialize` derived, the `total_voting_power` field included with no fallback, and this stricter deserialization schema has probably been in use by parties unknown. That serialization is simply derived, with no validations.

### Definition of "done"

The following needs to be implemented and backed by tests:

* In `tendermint`, the `Deserialize` impl for `validator::Set` can tolerate the absence of the `total_voting_power` field by computing it from validator data.
* If the the `total_voting_power` field is present in the deserialized structure, its value is validated against the computed sum of validators' voting powers.
  * A special case can be made for the value of 0, which is easy to fill by default in Go and seems to occur in the wild. This is also the default deserialization for the missing field in protobuf. It should be treated like omission of the field.
* In `tendermint-proto`, the deserialization for the generated `ValidatorSet` structs accepts a present `total_voting_power` field and ~applies the same validation as described above~ represents its value in the struct field. On serialization, the field is omitted to maintain backward compatibility; alternatively (in a technically breaking change), the field is included, but serializations of compound types with validator set fields may bypass this serialization by e.g. using a custom helper.
* The sum computation is protected against integer overflows.
* Validator set serializations where the total voting power exceeds the protocol-defined maximum limit (`u64::MAX / 8`) are rejected as invalid.

_Originally posted by @mzabaluev in https://github.com/informalsystems/tendermint-rs/issues/1340#issuecomment-1708508934_

mzabaluev:
Closes #1348

Use the validating conversion from the protobuf struct `ValidatorSet`, which has the same schema including the now optionally deserialized `total_voting_power` field.
This is to ensure we don't blindly trust `total_voting_power` if it's
present in the serialization, and accept when it's absent.

Also ensure no integer overflows can occur while calculating the total voting power, and check the resulting power against
the protocol-defined sanity limit.

* [x] Wrote tests

### DIFF
--- a/tools/proto-compiler/src/constants.rs
+++ b/tools/proto-compiler/src/constants.rs
@@ -41,6 +41,7 @@ const TYPE_TAG: &str = r#"#[serde(tag = "type", content = "value")]"#;
 /// Predefined custom attributes for field annotations
 const QUOTED: &str = r#"#[serde(with = "crate::serializers::from_str")]"#;
 const QUOTED_WITH_DEFAULT: &str = r#"#[serde(with = "crate::serializers::from_str", default)]"#;
+const QUOTED_ALLOW_NULL: &str = r#"#[serde(with = "crate::serializers::from_str_allow_null")]"#;
 const DEFAULT: &str = r#"#[serde(default)]"#;
 const HEXSTRING: &str = r#"#[serde(with = "crate::serializers::bytes::hexstring")]"#;
 const BASE64STRING: &str = r#"#[serde(with = "crate::serializers::bytes::base64string")]"#;
@@ -197,8 +198,9 @@ pub static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
     ), // https://github.com/tendermint/tendermint/issues/5549
     (
         ".tendermint.types.Validator.proposer_priority",
-        QUOTED_WITH_DEFAULT,
-    ), // Default is for /genesis deserialization
+        QUOTED_ALLOW_NULL,
+    ), // null occurs in some LightBlock data
+    (".tendermint.types.Validator.proposer_priority", DEFAULT), // Default is for /genesis deserialization
     (
         ".tendermint.types.ValidatorSet.total_voting_power",
         QUOTED_WITH_DEFAULT,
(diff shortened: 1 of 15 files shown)


tools/proto-compiler/src/constants.rs
```diff
@@ -98,7 +99,7 @@ pub static CUSTOM_TYPE_ATTRIBUTES: &[(&str, &str)] = &[
     (".tendermint.types.Commit", SERIALIZED),
     (".tendermint.types.CommitSig", SERIALIZED),
     (".tendermint.types.ValidatorSet", SERIALIZED),
-    (".tendermint.crypto.PublicKey", SERIALIZED),
+    (".tendermint.crypto.PublicKey.sum", SERIALIZED),
```
romac, line 102: Is that a typo, or does Prost expect the type name in lowercase? I am asking because that `#[derive]` argument only applies to structs/enums AFAIK. 

```suggestion
    (".tendermint.crypto.PublicKey.Sum", SERIALIZED),
```

romac: In any case, it does not seem to be reflected in the generated protos.

mzabaluev: It actually is reflected and the name must be in lowercase. Without this, the `Sum` type does not get the serde attributes, which it previously "inherited" from the configuration for the containing message. This is how it works with prost-build.

romac: Gotcha!

romac (approved):
Great stuff, thanks!

PR #1351, opened 2023-09-12
Correct and optimize custom serializers for stringized values, hashes, nullable and optional fields

mzabaluev:
The custom field serializers in `tendermint` and `tendermint-proto` have a few problems.
The modules where the deserializers expect the values to be an `Option` (i.e. nullable) do not consistently use `Serializer::serialize_some` when serializing a non-null value. This is not an issue with `serde_json`, but as the serde framework is generic, this implementation will break with other serializers, particularly for non-self-describing formats.
Deserializer functions unconditionally allocate temporary strings where they can potentially be borrowed from the deserializer, achieving zero copies from source data.

These changes correct the serde schema issues and introduce optimization by deserializing strings into `Cow`, allowing use of borrowed string values from deserializers that support it.

* [ ] Wrote tests

### DIFF
--- a/proto/src/serializers/optional.rs
+++ b/proto/src/serializers/optional.rs
@@ -1,6 +1,9 @@
 //! Serialize/deserialize `Option<T>` type where `T` has a serializer/deserializer.
-//! Use `null` if the received value equals the `Default` implementation.
-// Todo: Rename this serializer to something like "default_eq_none" to mirror its purpose better.
+//! Deserialize to `None` if the received value equals the `Default` value.
+//! Serialize `None` as `Some` with the `Default` value for `T`.
+
+// TODO: Rename this serializer to something like "default_eq_none" to mirror its purpose better.
+
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 /// Deserialize `Option<T>`
@@ -19,7 +22,7 @@ where
     T: Default + Serialize,
 {
     match value {
-        Some(v) => v.serialize(serializer),
-        None => T::default().serialize(serializer),
+        Some(v) => serializer.serialize_some(v),
+        None => serializer.serialize_some(&T::default()),
     }
 }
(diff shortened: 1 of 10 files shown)


proto/src/serializers/optional.rs
```diff
@@ -1,6 +1,9 @@
 //! Serialize/deserialize `Option<T>` type where `T` has a serializer/deserializer.
-//! Use `null` if the received value equals the `Default` implementation.
-// Todo: Rename this serializer to something like "default_eq_none" to mirror its purpose better.
+//! Deserialize to `None` if the received value equals the `Default` value.
+//! Serialize `None` as `Some` with the `Default` value for `T`.
```
mzabaluev, line 3: The behaviour of this helper is questionable, before or after this change. If the type has a reasonable default value, why make the domain type field an `Option`? And in the other direction, why uselessly serialize a default value when a `None` would do as well?

Farhad-Shabani (approved):
Just a small nit. Otherwise looks good. Thanks

Farhad-Shabani (approved):
Just a small nit. Otherwise looks good. Thanks

proto/src/serializers/optional.rs
Farhad-Shabani: Yes, It's kind of a head-scratcher!
I see this is derived on fields of certain proto types with `Option<…>` types (e.g. [here](https://github.com/informalsystems/tendermint-rs/blob/bdef9884221a850323f98b3fa2b9b7471f5e8437/proto/src/prost/v0_34/tendermint.types.rs#L200-L201)). Perhaps, those types do not inherently need to be Option. Consequently, they required serde to handle Option this way.

proto/src/serializers/optional_from_str.rs
```diff
@@ -23,11 +24,9 @@ where
     T: FromStr,
     T::Err: Display,
 {
-    let s = match Option::<String>::deserialize(deserializer)? {
+    let s = match Option::<Cow<'_, &str>>::deserialize(deserializer)? {
```
Farhad-Shabani, line 27:
```suggestion
    let s = match Option::<Cow<'_, str>>::deserialize(deserializer)? {
```

mzabaluev: Good catch, thanks!

PR #1354, opened 2023-09-18
Release 0.33.2

mzabaluev:


mzabaluev:
The audit check fails due to #1342, but this can't be addressed without bumping the minor version in 0.x accordingly to semver. The fix is therefore pushed to the next release.

thanethomson (approved):
LGTM!

mzabaluev:
The docs build is fixed in #1355, to be rolled into the release.

PR #1358, opened 2023-09-22
Add support for secp256k1 consensus keys

mkaczanowski:
Closes: #1364

This adds the support for parsing secp256k1 consensus key from `priv_validator_key.json`.

A while ago I've merged support for secp256k1 consensus key type to `tmkms` mainly for band protocol network. It came out that the functionality to import the private key from tendermint format to tmkms doesn't work due to lack of support in `tendermint-rs`.

* https://github.com/iqlusioninc/tmkms/issues/761
* to the best of my knowledge
* [x] Wrote tests


romac (approved):
Thanks!

tendermint/src/private_key.rs
```diff
@@ -30,6 +34,15 @@ pub enum PrivateKey {
...
     Ed25519(Ed25519),
+
+    #[cfg(feature = "secp256k1")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "secp256k1")))]
+    #[serde(
+        rename = "tendermint/PrivKeySecp256k1",
+        serialize_with = "serialize_secp256k1_privkey",
+        deserialize_with = "deserialize_secp256k1_privkey"
+    )]
+    Secp256k1(Secp256k1),
```
mzabaluev, line 45: Feature-guarded enum variants bring fragility. I'd rather make the whole thing non-optional, `secp256k1` support does not add much in terms of dependencies. But the enum is `#[non_exhaustive]` so maybe it's OK for now.

mkaczanowski: since the `secp256k1` is marked almost everywhere I fell with the same approach. But I agree that conditional enums are not as great. Though rust is pretty good with handling conditional compilation via features.

Therefore I hold no strong opinion here

mzabaluev: The problem with conditional enum variants is that they make match expressions fail to compile dependent on features. `#[non_exhaustive]` does a lot to make it less fragile, as you can't have a match that would be exhaustive without certain features and then fail to compile if the features are enabled. But anyone wanting to match on the conditionally-enabled variant would have to enable the corresponding feature. This, perhaps, is no worse than other conditionally defined APIs, so I'm willing to let it go.

tarcieri: FYI, I believe the original motivation for feature-gating was at the time libsecp256k1 was being used and it didn’t work with WASM.

We’ve since moved to the pure Rust `k256` crate so that isn’t an issue.

I think it would probably make more sense to feature gate all the cryptographic functionality under a single feature rather than just secp256k1, that way people who just want e.g. tendermint-rpc for an API client aren’t pulling in a bunch of crypto deps they aren’t using.

PR #1360, opened 2023-09-26
Update `async-tungstenite` to v0.23.0

romac:
This in turns updates `tungstenite` to v0.20.0+, will pulls in v0.20.1, which contains a fix for this security advisory about a DoS attack:

https://github.com/informalsystems/hermes/security/dependabot/33

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 
 members = [
     "abci",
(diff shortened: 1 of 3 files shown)


mzabaluev:
Good; I've done that in #1362 as well to deduplicate dependencies before noticing this PR. I think that PR will be better to go with, as it does not get the dependencies in a state with two copies of tokio-rustls and whatnot.

romac:
Oh great! Let's close this then

Issue #1359, opened 2023-09-25
HTTP client should not try to parse body of response with non-2XX status code


xoac:
### What went wrong?

Concurrent request to `tx_search` fails with `value: serde parse error` 

```
    Finished dev [unoptimized + debuginfo] target(s) in 6.36s
     Running `target/debug/tendermint_tx_search`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: serde parse error

Caused by:
    expected value at line 1 column 1
```

### Steps to reproduce

<details>
<summary>Cargo.toml</summary>

```toml
[package]
name = "tendermint_tx_search"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
# async 
tokio = { version = "1.13.1", features = [
    "rt-multi-thread",
    "time",
    "sync",
    "macros",
] }

# tendermint
tendermint-rpc = { version = "0.33", features = ["http-client"] }
tendermint = { version = "0.33" }
```
</details>

<details>
<summary>main.rs (not working)</summary>

```rs
use tendermint_rpc::{client::Client, query::Query, Order};

#[tokio::main]
async fn main() {
    let mut c = tendermint_rpc::client::HttpClient::new("https://juno-rpc.icycro.org").unwrap();
    c.set_compat_mode(tendermint_rpc::client::CompatMode::V0_37);

    let query = Query::default()
        .and_eq(
            "coin_spent.spender",
            "juno185zc74a7w2pfxkv6d06t3kdja65tngjgg6223x",
        )
        .and_lt("tx.height", 10_187_764);

    let query2 = Query::default()
        .and_eq(
            "coin_received.receiver",
            "juno185zc74a7w2pfxkv6d06t3kdja65tngjgg6223x",
        )
        .and_lt("tx.height", 10_187_764);

    let r_task = c.tx_search(query, false, 1, 12, tendermint_rpc::Order::Descending);
    let r2_taks = c.tx_search(query2, false, 1, 15, Order::Descending);

    let (r, r2) = tokio::try_join!(r_task, r2_taks).unwrap();

    println!("{:?}", r);
}
```

</details>

<details>
<summary>main.rs(working)</summary>

```rs
use tendermint_rpc::{client::Client, query::Query, Order};

#[tokio::main]
async fn main() {
[...]

PR #1361, opened 2023-09-28
Turn non-200 HTTP responses into an error instead of trying to parse the body as a JSON-RPC response

romac:
Closes: #1359

This PR checks for the status code to be 200 OK, if not it returns an error with the actual status code and the body of the response, but does not attempt to parse the response body as a JSON-RPC payload.

CometBFT will [always set the status code](https://github.com/cometbft/cometbft/blob/86c07f20f163c17a0f5b7149cbd441a242b88557/rpc/jsonrpc/server/http_server.go#L151) to 200 OK and the content type to JSON for response which were successfully handled (which does mean that the JSON-RPC response itself is successful).

* [ ] Wrote tests

### DIFF
--- a/rpc/src/client/transport/http.rs
+++ b/rpc/src/client/transport/http.rs
@@ -160,7 +160,7 @@ impl HttpClient {
     {
         let request_body = request.into_json();
 
-        tracing::trace!("outgoing request: {}", request_body);
+        tracing::debug!(url = %self.url, body = %request_body, "outgoing request");
 
         let mut builder = self
             .inner
@@ -182,11 +182,23 @@ impl HttpClient {
     {
         let request = self.build_request(request)?;
         let response = self.inner.execute(request).await.map_err(Error::http)?;
+        let response_status = response.status();
         let response_body = response.bytes().await.map_err(Error::http)?;
-        tracing::trace!(
-            "incoming response: {}",
-            String::from_utf8_lossy(&response_body)
+
+        tracing::debug!(
+            status = %response_status,
+            body = %String::from_utf8_lossy(&response_body),
+            "incoming response"
         );
+
+        // Successful JSON-RPC requests are expected to return a 200 OK HTTP status.
+        // Otherwise, this means that the HTTP request failed as a whole,
+        // as opposed to the JSON-RPC request returning an error,
+        // and we cannot expect the response body to be a valid JSON-RPC response.
+        if response_status != reqwest::StatusCode::OK {
+            return Err(Error::http_request_failed(response_status));
+        }
+
         R::Response::from_string(&response_body).map(Into::into)
     }
 }
--- a/rpc/src/error.rs
+++ b/rpc/src/error.rs
@@ -12,6 +12,12 @@ type ReqwestError = flex_error::DisplayOnly<reqwest::Error>;
 #[cfg(not(feature = "reqwest"))]
 type ReqwestError = flex_error::NoSource;
 
+#[cfg(feature = "reqwest")]
+type HttpStatusCode = reqwest::StatusCode;
+
+#[cfg(not(feature = "reqwest"))]
+type HttpStatusCode = core::num::NonZeroU16;
+
 #[cfg(feature = "tokio")]
 type JoinError = flex_error::DisplayOnly<tokio::task::JoinError>;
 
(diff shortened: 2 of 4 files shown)


rpc/src/client/transport/http.rs
```diff
@@ -367,8 +367,21 @@ mod sealed {
         {
             let request = self.build_request(request)?;
             let response = self.inner.request(request).await.map_err(Error::hyper)?;
+
+            tracing::debug!("Incoming response:");
+            let response_status = response.status();
+            tracing::debug!("  Status: {}", response.status());
             let response_body = response_to_string(response).await?;
-            tracing::debug!("Incoming response: {}", response_body);
+            tracing::debug!("  Body: {}", response_body);
```
mzabaluev, line 375: I'd rather put out a single tracing event, using keywords:

```rust
    tracing::debug!(status = %response_status, body = %response_body, "incoming response");
```

On the other hand, is it too much spam for debug level? Maybe the body dump should be on `trace` level.

rpc/src/client/transport/http.rs
```diff
@@ -367,8 +367,21 @@ mod sealed {
...
+
+            // Successful JSON-RPC requests are expected to return a 200 OK HTTP status.
+            // Otherwise, this means that the HTTP request failed as a whole,
+            // as opposed to the JSON-RPC request returning an error,
+            // and we cannot expect the response body to be a valid JSON-RPC response.
+            if response_status != http::StatusCode::OK {
```
mzabaluev, line 381: There is a method `StatusCode::is_success`, but maybe success statuses that are not exactly 200 should also be seen as non-conformant.

rpc/src/error.rs
```diff
@@ -89,6 +89,15 @@ define_error! {
                 format_args!("method not found: {}", e.method)
             },
 
+        HttpRequestFailed
+            {
+                status: http::StatusCode,
+                body: String
+            }
+            | e | {
+                format_args!("HTTP request failed with non-200 status code: {}, body: {}", e.status, e.body)
```
mzabaluev, line 98: Is there ever a need to dump the whole response body (which could be arbitrarily large) in error handling? Please leave it to the logs.

rpc/src/client/transport/http.rs
romac: Yes it's my understanding from reading both the Comet source and [this spec](https://www.simple-is-better.org/json-rpc/transport_http.html), that only 200 OK responses can be interpreted as a valid JSON-RPC payload. That said, I don't know how official that spec is.

Also it does not deal with redirects, etc. but that was already the case before. Do you know if hyper takes care of those for us?

rpc/src/error.rs
romac: Good point!

romac: [`2dc92cc`](https://github.com/informalsystems/tendermint-rs/pull/1361/commits/2dc92cc35a6beb8afca2c9dd6dd99f8eee178ed0)

rpc/src/client/transport/http.rs
romac: Yeah let's do that. I think it's fine at the debug level, and we're using it as via from Hermes `--debug=rpc` already, so let's leave it like that.

romac: e64df0b3

mzabaluev (approved):

mzabaluev:
Conditional compilation needs to be fixed for wasm, otherwise looks fine.

romac:
I will do this after we merge #1362

romac:
> Conditional compilation needs to be fixed for wasm, otherwise looks fine.

https://github.com/informalsystems/tendermint-rs/pull/1361/commits/0b0f4c482b67b13357b1f0ba40acecb804829f33

romac:
I've tested this from Hermes, and it works as expected.

mzabaluev:
> Conditional compilation needs to be fixed for wasm, otherwise looks fine.

In fact, why do we build RPC stuff for wasm?

romac:
> In fact, why do we build RPC stuff for wasm?

Because one can define a `Client` implementation which uses eg. the browser's `fetch` method and then compile all that to WASM.

PR #1362, opened 2023-09-28
rpc: replace `hyper::Client` with `reqwest::Client`

### LINKED ISSUE #1342: RUSTSEC-2023-0052: webpki: CPU denial of service in certificate path building

mzabaluev:
Closes #1342

In `hyper`, the high-level `Client` implementation is going to be removed in the next major release. The current client lacks built-in support for HTTP proxies, and we want to ditch `hyper-proxy` as it is unmaintained and its `webpki` dependency has a known security issue.

* [ ] Wrote tests


rpc/src/error.rs
```diff
@@ -48,12 +36,12 @@ define_error! {
...
-            [ HttpError ]
+            [ ReqwestError ]
             | _ | { "HTTP error" },
 
-        Hyper
-            [ HyperError ]
-            | _ | { "HTTP error" },
+        InvalidProxy
+            [ ReqwestError ]
+            | _ | { "Invalid proxy configuration" },
```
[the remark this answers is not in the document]
romac: What's the downside of exposing those? Breaking semver compatibility if we update the underlying client crate to another major version or if we change it again?

Since we are already releasing breaking changes very frequently, I don't believe that's a big deal, but I am happy to be convinced otherwise.

On the other hand, the only downside of type-erasing those errors is that whoever is consuming them will have to dynamically cast them if they want to inspect them, which is brittle.

So my vote is for keeping those as-is, but I am open to changing my mind.

rpc/src/client/transport/http.rs
```diff
@@ -111,14 +114,7 @@ impl HttpClient {
...
         let url = url.try_into()?;
-        Ok(Self {
-            inner: if url.0.is_secure() {
-                sealed::HttpClient::new_https(url.try_into()?)
-            } else {
-                sealed::HttpClient::new_http(url.try_into()?)
-            },
-            compat: Default::default(),
-        })
+        Self::builder(url).build()
```
romac, line 117: So much nicer :)

rpc/src/client/transport/http.rs
```diff
@@ -196,7 +222,7 @@ impl Client for HttpClient {
                 // Back-fill with a request to /block endpoint and
                 // taking just the header from the response.
                 let resp = self
-                    .perform_v0_34(endpoint::block::Request::new(height))
+                    .execute::<_, v0_34::Dialect>(endpoint::block::Request::new(height))
```
romac, line 225: I am not a huge fan of functions which won't compile without explicitly passing them type argument, how about passing the dialect as an value-level argument instead?

ie. change the definition of `execute` to:

```rust
    async fn execute<R, S>(&self, request: R, dialect: S) -> Result<R::Output, Error>
    where
        R: SimpleRequest<S>,
        S: Dialect,
    {
```

What do you think?

romac:
Love all that deleted code 🥇

rpc/src/client/transport/http.rs
mzabaluev: Yes, stylistically it's better.

mzabaluev: Done in https://github.com/informalsystems/tendermint-rs/pull/1362/commits/8cc408d5770c41c75bd12da403093ae610ce4340 and https://github.com/informalsystems/tendermint-rs/pull/1362/commits/a41c4abe39cfa9a86d9d7081071a03bf56666894

rpc/src/error.rs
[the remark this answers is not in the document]
romac: Agreed! Let's cross that bridge when we get to it

rpc/src/client/transport/http.rs
```diff
@@ -79,27 +87,23 @@ impl Builder {
...
-                inner: if proxy_url.0.is_secure() {
-                    sealed::HttpClient::new_https_proxy(
-                        self.url.try_into()?,
-                        proxy_url.try_into()?,
-                    )?
+        let builder = reqwest::ClientBuilder::new().user_agent(USER_AGENT);
+        let inner = match self.proxy_url {
+            None => builder.build().map_err(Error::http)?,
+            Some(proxy_url) => {
+                let proxy = if self.url.0.is_secure() {
```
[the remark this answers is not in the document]
mzabaluev: No, the `Proxy` constructors define what kind of protocol you want proxied, not where to connect.
I've made that mistake myself and corrected it in https://github.com/informalsystems/tendermint-rs/pull/1362/commits/5a6398639c00a7e6c355c0ed9f03b3b77d8cbba9

romac (approved):

PR #1363, opened 2023-09-29
Fix cargo warning by explicitly specifying a resolver for the workspace

romac:
Fix the following cargo warning when building the workspace:

```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

* [ ] Wrote tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 
 members = [
     "abci",


mzabaluev (approved):
Oh yeah, I've been constantly annoyed about this myself.

PR #1370, opened 2023-10-11
p2p: replace `x25519-dalek` with `curve25519-dalek-ng`

tony-iqlusion:
The `tendermint-p2p` crate previously pulled in both an obsolete version of `x25519-dalek`/`curve25519-dalek`, but also `curve25519-dalek-ng` by way of `ed25519-consensus`.

This replaces `x25519-dalek` with `curve25519-dalek-ng` which is already a dependency of `ed25519-consensus`. This reduces the overall number of dependencies and ensures `curve25519-dalek-ng` is the main Curve25519 implementation used.

Note: I don't have an open issue for this but since it reduces the overall number of dependencies I thought it made sense.

* [ ] Wrote tests

### DIFF
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -28,16 +28,16 @@ amino = ["prost-derive"]
 
 [dependencies]
 chacha20poly1305 = { version = "0.8", default-features = false, features = ["reduced-round"] }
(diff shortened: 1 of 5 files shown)


romac (approved):
Thanks a lot!

PR #1371, opened 2023-10-13
Avoid enabling flex-error/eyre_tracer feature

mina86:
Don’t enable flex-error/eyre_tracer feature in crates which don’t use
eyre directly.  This allows customers depend on say tendermint crate
without having to pull in eyre crate.  This is useful because eyre
uses mutable static variables (specifically HOOK variable) which is
incompatible with some blockchain runtimes.

* [n/a] Referenced an issue explaining the need for the change
* [n/a] Wrote tests


mina86:
@romac

romac:
Good catch! Thanks

PR #1376, opened 2023-11-09
tendermint: allow null in `EventAttribute` kv JSON

### LINKED ISSUE #1375: The null `value` field of `EventAttribute` should be allowed
An indexer ran into null `value` field of `EventAttribute`, preventing a service from indexing events and resulting in downtime. The exact error was `serde invalid type: null`.

_Originally posted by @daniel-savu in https://github.com/informalsystems/tendermint-rs/issues/1216#issuecomment-1802436901_

mzabaluev:
Fixes #1375

When deserializing `EventAttribute` values, allow null values in the key and value fields. This follows the behaviour in 0.34 RPC (where the key-value data were also base64-encoded), but technically breaks the serde schema of everything that includes ABCI event data.

* [ ] Wrote tests


mzabaluev:
The coverage failure should be fixed by #1377.

romac (approved):

PR #1380, opened 2023-11-14
rpc: Add HTTP timeout, default 30 seconds.

### LINKED ISSUE #1379: rpc: Introduce HTTP timeouts
### Description
Add configurable timeout that is passed to reqwest. The current default is "no timeout", which can lead to requests never finishing.

### Definition of "done"
Add timeout() to request builder and wire it into reqwest.

mdyring:
Simply introduces a HTTP timeout parameter, which is passed to reqwest.

This is highly desirable a the default behaviour is "no timeout", which can result in stalls.

Closes #1379.

* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/improvements/1379-rpc-http-timeout.md
+++ b/.changelog/unreleased/improvements/1379-rpc-http-timeout.md
@@ -1,0 +1,2 @@
+- Add request timeout for the RPC HttpClient
+  ([\#1379](https://github.com/informalsystems/tendermint-rs/issues/1379))
--- a/rpc/src/client/transport/http.rs
+++ b/rpc/src/client/transport/http.rs
@@ -7,6 +7,7 @@ use core::{
 
 use async_trait::async_trait;
 use reqwest::{header, Proxy};
+use std::time::Duration;
 
 use tendermint::{block::Height, evidence::Evidence, Hash};
 use tendermint_config::net;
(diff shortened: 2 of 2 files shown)


mzabaluev (changes requested):
This looks good API-wise.
Can we follow the behavior in `reqwest` and have "no timeout" behavior by default? This will be also the least disruptive to the current users of `HttpClient`.

rllola:
Hey! We are looking forward for this to be merged because we have the issue of our request being stuck.

Anything we can do so it can be merged soon ?

tony-iqlusion:
Note that you can use e.g. `tokio::time::timeout` to enforce timeouts on anything `Future`-based without requiring upstream support.

That said, if you do end up adding this functionality in a first-class manner, I think it would also be good if it were on-by-default.

mdyring:
> This looks good API-wise. Can we follow the behavior in `reqwest` and have "no timeout" behavior by default? This will be also the least disruptive to the current users of `HttpClient`.

I'd argue that "no timeout" behavior is unexpected and thus not a good default (even though that seems to be the default for reqwest).

As @tony-iqlusion mentioned, `tokio::time::timeout` is also an option for library users, so alternatively it could simply be documented that the calls might hang.

rllola:
@tony-iqlusion Thanks for your solution.

I agree with @mdyring, the "no timeout" is unexpected and I am in favor of a default value around 30secs.

tony-iqlusion:
Personally I'm biased towards on-by-default timeouts, but having any sort of built-in timeout would be nice.

romac:
Alright, I'll rest my case then since you guys all agree that having a timeout by default is the preferred behaviour.

romac (approved):

PR #1381, opened 2023-11-19
Add autogenerated protos for `v1.0`

sergio-mena:
This (draft) PR is the first to providing support for CometBFT `v1.0`.

The PR appears to be _massive_ but the reason is the tons of newly auto-generated code it introduces. For this reason, *please don't review it in one shot*; rather use the following description of commits:

* Commit 1: Ignore. It is the result of running the protobuf generation on my machine, which produces a few changes in the auto-generated code.
* Commit 2: Review. First step: add the logic to handle both _Tendermint_ and _CometBFT_ protobufs. We might decide to deprecate the _Tendermint_ part.
* Commit 3: Ignore. The effect in auto-generated code of commit 2.
* Commit 4: Review. Adapted protobuf generation tweaks to CombetBFT protos. Commit 2, just copy-pasted the Tendermint ones. This commit allows to see how I adapted them.
* Commit 5: Ignore. The effect in auto-generated code of commit 4.
* Commit 6: Review. As there are several levels of modules now, the preexisting mechanism to generate the modules was too rigid. This commit introduces and extended version
* Commit 7: Ignore. The effect in auto-generated code of commit 6.
* Commit 8: Review. Fixes needed in the manual code so that the new proto bufs builds correctly
* Commit 9: Review. Changelog.

Note: even if this PR is likely to be ready for review, I decided to leave it a draft PR because we need to discuss whether we should be forking tendermint-rs into cometbft-rs and move this PR there.

* [ ] Wrote tests


tools/proto-compiler/src/constants.rs
```diff
@@ -109,6 +120,40 @@ pub static CUSTOM_TYPE_ATTRIBUTES: &[(&str, &str)] = &[
...
+    (".cometbft.crypto.PublicKey.sum", SERIALIZED),
+    (".cometbft.crypto.PublicKey.sum", TYPE_TAG),
+    (".cometbft.abci.ResponseInfo", SERIALIZED),
+    (".cometbft.types.CanonicalBlockID", SERIALIZED),
+    (".cometbft.types.CanonicalPartSetHeader", SERIALIZED),
+    (".cometbft.types.Validator", SERIALIZED),
+    (".cometbft.types.CanonicalVote", SERIALIZED),
+    (".cometbft.types.BlockMeta", SERIALIZED),
+    (".cometbft.types.TxProof", SERIALIZED),
+    (".cometbft.crypto.Proof", SERIALIZED),
```
sergio-mena: Please ignore these changes: review them in Commit 4

tools/proto-compiler/src/constants.rs
```diff
@@ -231,4 +276,118 @@ pub static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
...
+    (
+        ".cometbft.types.Evidence.sum.light_client_attack_evidence",
+        RENAME_LIGHTCLIENTATTACK,
+    ),
+    (".cometbft.types.TxProof.data", BASE64STRING),
+    (".cometbft.types.TxProof.root_hash", HEXSTRING),
+    (".cometbft.crypto.Proof.index", QUOTED),
+    (".cometbft.crypto.Proof.total", QUOTED),
+    (".cometbft.crypto.Proof.aunts", VEC_BASE64STRING),
+    (".cometbft.crypto.Proof.leaf_hash", BASE64STRING),
```
sergio-mena: Please ignore these changes: review them in Commit 4

tools/proto-compiler/src/functions.rs
```diff
@@ -172,7 +172,7 @@ pub fn copy_files(src_dir: &Path, target_dir: &Path) {
             e.file_type().is_file()
                 && e.file_name()
                     .to_str()
-                    .map(|name| name.starts_with("tendermint."))
+                    .map(|name| name.starts_with(&format!("{project}.")))
```
mzabaluev, line 176: Performance nit: it would be better if the prefix string was formatted once, not with every invocation of the closure.

tools/proto-compiler/src/functions.rs
```diff
@@ -212,12 +212,14 @@ pub fn find_proto_files(proto_path: &Path) -> Vec<PathBuf> {
...
     create_dir_all(target_dir).unwrap();
+
+    let project_dot = format!("{}.", version.project);
     let file_names = WalkDir::new(prost_dir)
         .into_iter()
         .filter_map(|e| e.ok())
         .filter(|e| {
             e.file_type().is_file()
-                && e.file_name().to_str().unwrap().starts_with("tendermint.")
+                && e.file_name().to_str().unwrap().starts_with(&project_dot)
```
mzabaluev, line 278: Right, like here.

mzabaluev (commented):
Good automation for the module tree.

I don't think we need to namespace the cometbft generated files under module `v1_0`.
Remember the idea of ADR-107 is to eventually promote the message definitions to `*.v1` packages at the proto level. So `abci.v1beta4.Request` becomes `abci.v1.Request` and so on. It's better that we generate the root module under `proto/cometbft/lib.rs`.

I don't see a need to hold on to `proto::tendermint`, modules either. We're making a switch for cometbft-rs, the legacy generated files will not be used by the rest of the cometbft-rs crates, so they are pretty much dead code. @thanethomson had the idea that the users who need support for `tendermint.*` protos could still use the tendermint-rs crates as already released, and we can still support those from this current repository in maintenance mode.

sergio-mena:
This is continued in #cometbft/cometbft-rs#2

PR #1389, opened 2024-01-30
proto: Add missing serde derives on Protobuf definitions

romac:
* [ ] Wrote tests


proto/Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_bytes = { version = "0.11", default-features = false, features = ["alloc"] }
 subtle-encoding = { version = "0.5", default-features = false, features = ["hex", "base64", "alloc"] }
 num-traits = { version = "0.2", default-features = false }
-num-derive = { version = "0.3", default-features = false }
+num-derive = { version = "0.4", default-features = false }
```
mzabaluev, line 32: I've been wondering what do we use num-traits for when `prost-build` provides the conversions from `i32` [out of the box](https://docs.rs/prost/latest/prost/#enumerations).

mzabaluev (commented):
Good, we'll need this in cometbft-rs too!
Before we can use the buf plugin to generate the majority of serde impls, that is.

proto/Cargo.toml
romac: https://github.com/informalsystems/tendermint-rs/pull/1389/commits/5d23beadfb2f03f80bd86981f19cffa3c3407090

PR #1391, opened 2024-02-02
rpc: Fix deserialization of `/block_results` response when some tx results are non-ok

romac:
See https://github.com/informalsystems/hermes/issues/3817#issuecomment-1916478908 for background.

---

* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/bug-fixes/1391-block-results-not-ok.md
+++ b/.changelog/unreleased/bug-fixes/1391-block-results-not-ok.md
@@ -1,0 +1,3 @@
+- `[tendermint-rpc]` Fix deserialization of `/block_results` response when some
+  tx results are non-ok ([\#1391](https://github.com/informalsystems/tendermint-
+  rs/pull/1391))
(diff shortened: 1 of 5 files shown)


romac:
@mzabaluev We should probably do something smarter in cometbft-rs, eg. model `ExecTxResult` as an enum rather than have all successful and failing fields in the same struct.

mzabaluev:
@romac
> We should probably do something smarter in cometbft-rs, eg. model `ExecTxResult` as an enum rather than have all successful and failing fields in the same struct.

Good idea. Which fields should be represented in results with `Code != 0`? I see "log" and "codespace" fields in these fixtures. What about "info"?

romac:
Good question, I don't know! The fixture does not feature it in either case. Might have to check out the CometBFT source code to figure when it's included.

mzabaluev:
Note that you can already match on the `abci::Code` domain type. As for whether the other fields may be present in `ExecTxResult` of a failed transaction, I can find no normative information on this. As far as I can see in the CometBFT source, result are only formed by test code, there are no helper methods to guide applications on how to properly report failures.

`codespace` perhaps only makes sense for non-zero codes and could be a member of redesigned `Code::Err`. `log` and `info` could be set with useful information in failure cases as well.
`gas_wanted`, `gas_used`, and `events` only make sense in the success case.

PR #1393, opened 2024-02-16
p2p: fix data corruption on longer packets

### LINKED ISSUE #1392: p2p: failure to read complete message
### What went wrong?

There are ongoing reports of failure to read a complete message in the `tendermint-p2p` crate which seem to be manifesting as an incomplete Protobuf message in TMKMS:

https://github.com/iqlusioninc/tmkms/issues/729

### Steps to reproduce

Unfortunately this seems very chain-specific and I have not been given proper repro steps myself in the aforementioned issue.

### Definition of "done"

It's unclear what actually needs to change here.

A somewhat related issue is the API that `tendermint-p2p` exposes: it's a message-oriented protocol, but the interface uses the `std::io::{Read, Write}` traits, which are stream-oriented. It's possible TMKMS is misusing the API, but it's the wrong abstraction to begin with.

mzabaluev:
The code handling chunking of data frames longer than the configured
maximum was faulty.

May fix #1392 and possibly other occurrences of data corruption.

* [x] Wrote tests

### DIFF
--- a/p2p/src/secret_connection.rs
+++ b/p2p/src/secret_connection.rs
@@ -538,16 +538,7 @@ fn encrypt_and_write<IoHandler: Write>(
     data: &[u8],
 ) -> io::Result<usize> {
     let mut n = 0_usize;
-    let mut remaining = data;
-    while !remaining.is_empty() {
-        let chunk: &[u8];
-        if DATA_MAX_SIZE < remaining.len() {
-            chunk = &remaining[..DATA_MAX_SIZE];
-            remaining = &remaining[DATA_MAX_SIZE..];
-        } else {
-            chunk = remaining;
-            remaining = &[];
-        }
+    for chunk in data.chunks(DATA_MAX_SIZE) {
         let sealed_frame = &mut [0_u8; TAG_SIZE + TOTAL_FRAME_SIZE];
         encrypt(chunk, &send_state.cipher, &send_state.nonce, sealed_frame)
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))?;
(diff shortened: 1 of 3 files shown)


p2p/src/secret_connection.rs
```diff
@@ -538,15 +538,15 @@ fn encrypt_and_write<IoHandler: Write>(
...
-            chunk = &data[..DATA_MAX_SIZE];
-            data_copy = &data_copy[DATA_MAX_SIZE..];
+        if DATA_MAX_SIZE < remaining.len() {
+            chunk = &remaining[..DATA_MAX_SIZE];
+            remaining = &remaining[DATA_MAX_SIZE..];
         } else {
-            chunk = data_copy;
-            data_copy = &[0_u8; 0];
+            chunk = remaining;
+            remaining = &[];
```
rnbguy, line 549: you can use `for chunk in data.chunks(DATA_MAX_SIZE) { ... }` directly !

mzabaluev: https://github.com/informalsystems/tendermint-rs/pull/1393/commits/9767651320ec91aeff6501dacf12691e8e4d79ea

tony-iqlusion (approved):
Worth a shot at least, although I think ideally the buffer management would be changed so this library allocates the buffer and `DATA_MAX_SIZE` is just a sanity limit

mzabaluev:
> ideally the buffer management would be changed so this library allocates the buffer and `DATA_MAX_SIZE` is just a sanity limit

The protocol always sends full AEAD frames, and long messages are split between successive frames. So, regardless of what the doc comment says, `DATA_MAX_SIZE` only defines the size of the encrypted frames. In plaintext, the message is length-prefixed.

PR #1395, opened 2024-02-22
light-client: Optimize voting power calculation by breaking the loop when we have enough voting power

### LINKED ISSUE #235: lite_impl: Commit#voting_power_in func is suboptimal

i-m-aditya:
Fixes #235

* [ ] Wrote tests


romac:
Thanks @i-m-aditya! 💐

romac (approved):

PR #1405, opened 2024-03-29
Add Vec<u8> for EventAttribute for 0.34

penso:
* [ ] Wrote tests


tendermint/src/abci/event.rs
```diff
@@ -111,47 +166,50 @@ where
...
+                (&attr.key, &attr.value).hash(state);
+            },
+        }
     }
 }
 
 impl<K: Into<String>, V: Into<String>> From<(K, V, bool)> for EventAttribute {
     fn from((key, value, index): (K, V, bool)) -> Self {
-        EventAttribute {
+        // TODO: support all versions
```
penso, line 207: @romac How can I get the version currently used here?

romac: We cannot, so I think we can either leave this as-is or remove the instance altogether and leave it up to the call to build the right event attribute version.

romac:
@penso Great work, thanks so much! I did a few follow-up fixes, notably for the `key` field, which I believe is not enforced to be a valid base64-encoded UTF-8 string either as evidence by the Protobuf schema (even if it probably always is in practice).

romac:
Before merging this I would like to get a few more eyes on this, as this is a pretty substantial breaking change and that I am not 100% convinced yet that this works for all uses cases.

@hdevalence @tony-iqlusion @ancazamfir @soareschen If any or all of you could take a look at this PR whenever you have some spare time, it would be much appreciated 💐

rpc/src/dialect/v0_34.rs
```diff
@@ -44,15 +44,17 @@ pub struct EventAttribute {
     /// The event key.
     #[serde(
         serialize_with = "base64string::serialize",
-        deserialize_with = "base64string::deserialize_to_string"
+        deserialize_with = "base64string::deserialize"
     )]
-    pub key: String,
+    pub key: Vec<u8>,
```
romac, line 49: While in practice I suspect all event keys are valid UTF-8 strings encoded as base64, since the Protobuf schema does not enforce this, I believe it warrants changing its type to `Vec<u8>` as well.

romac:
On my side, I will try to integrate this into Hermes before we merge, and see how that goes.

tendermint/src/abci/event.rs
```diff
@@ -21,10 +20,90 @@ pub struct Event {
...
+/// The value of the index flag is non-deterministic and may vary across different nodes in the network.
+///
+/// Before Tendermint v0.37, the key and value can contain arbitrary byte arrays.
+/// Since Tendermint v0.37, the key and value are defined to be valud UTF-8 encoded strings.
+///
+/// IMPORTANT: The order of the two variants below is significant and nmust not be changed.
+/// The `EventAttribute` enum is serialized and deserialized using the `untagged` attribute,
+/// meaning that the first variant is tried first when deserializing, if that fails
+/// then the second variant is tried. This allows us to deserialize v0.37+ events which
+/// are vald UTF-8 strings, and fall back to deserializing v0.34 events which are arbitrary byte arrays.
```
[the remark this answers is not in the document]
penso: Yeah but you're supposed to know which node you're connected to, I'd rather see a attribute `version` added within the serialized EventAttribute than an upper tag named `V034`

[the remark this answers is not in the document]
romac: Unfortunately we cannot change the serialization format of the events, as that would require changes in Tendermint/CometBFT 0.34.x, 0.37.x, 0.38.x and 1.x, which won't happen I think.

penso:
> @penso Great work, thanks so much! I did a few follow-up fixes, notably for the `key` field, which I believe is not enforced to be a valid base64-encoded UTF-8 string either as evidence by the Protobuf schema (even if it probably always is in practice).

Glad it helped, didn't think about the `key` field but agreed if not enforced we should move that as well.

hdevalence:
cc @erwanor

romac:
I wonder if we should get rid of `tendermint::abci::EventAttribute` altogether and define two dialects of `EventAttribute` (and all transitive upstream dependencies): one for v0.34 and one for v0.37+. Then one would have to know statically which version of Tendermint/Comet they are dealing with to manipulate `EventAttributes` or any type which contains one.

erwanor:
@penso Thanks for you work on this! @romac's suggestion to fold the `v034` dialect in its own module, similar to the version based modeling of ABCI messages, makes sense to me.

soareschen:
I took a quick look, and here are some of my thoughts:

- There should be a trait so that one can work with both variants of the `EventAttribute`s in a uniform way. From what I see, bytes are more general than strings, and we can easily convert a string to bytes using `.as_bytes()`. So it should be sufficient to provide trait methods to access the raw bytes of the key/value.
- As @romac suggested, it would be good to provide two separate `EventAttribute` types, which would implement a common trait. That way, user code can still work generically with the two types using the common trait.
- Performance-wise, it may be expensive to perform base64 decoding on all events, regardless of whether the user is interested in the event. Depending on the actual performance, we may want to perform lazy parsing, at least on the value field, so that we can skip parsing events that do not match the key of interest.
  - On the other hand, it may be complicated to implement lazy parsing. So we may want to consider doing this only in future PRs, in case if the performance benefits are worth it.

penso:
@romac @erwanor would love this merged for me to use to fetch blocks from 0.34 chains. As it's currently failing.

romac:
In the interest of getting this merged quickly, let's go ahead with the current solution and see if we can improve on it later on based on @soareschen's suggestions.

romac (approved):

PR #1410, opened 2024-04-12
light-client-verifier: avoid checking signatures multiple times

mina86:
* [ ] Wrote tests


mina86:
@romac, could you give a quick review of this code? We’re continuing to hit performance limits and so need to continue trying to optimise the code. With this change I’m trying to make it so that a) no signature is verified twice and b) signatures are verified from one with most power so hopefully fewer signatures overall will need verification.

One thing to keep in mind is that this is API breaking.

romac:
Thanks for the PR! These are some quite substantial changes to the logic so I unfortunately won't be able to properly review them this week, sorry about that. At first glance this looks alright to me though, I and don't mind the API changes. I'll take a closer look first thing next week.

mina86:
Thanks. One more thing that may be noteworthy is that I’ve filtered out non-commit votes.  They didn’t contribute to voting power so I figured it’s fine to just ignore them.

light-client-verifier/src/operations/voting_power.rs
```diff
@@ -120,55 +164,135 @@ impl<V> Default for ProvidedVotingPowerCalculator<V> {
...
+
+impl NonAbsentCommitVotes {
+    pub fn new(signed_header: &SignedHeader) -> Result<Self, VerificationError> {
+        let mut votes = signed_header
+            .commit
+            .signatures
+            .iter()
+            .enumerate()
+            .flat_map(|(idx, signature)| {
+                let idx = ValidatorIndex::try_from(idx).unwrap();
```
romac, line 239: Maybe add a comment explaining why this is infallible or use an infallible constructor?

light-client-verifier/src/operations/voting_power.rs
```diff
@@ -77,28 +112,37 @@ pub trait VotingPowerCalculator: Send + Sync {
...
         validator_set: &ValidatorSet,
         trust_threshold: TrustThreshold,
     ) -> Result<VotingPowerTally, VerificationError>;
+
+    /// Compute the voting power in a header and its commit against two separate
+    /// validator sets.
+    ///
+    /// This is equivalent to calling [`Self::voting_power_in`] on each set
+    /// separately but may be more optimised.
+    fn voting_power_in_sets(
```
romac, line 140: Do we really need to expose this method in the trait, or can we instead just inline its implementation in `check_enough_trust_and_signers`?

light-client-verifier/src/predicates.rs
```diff
@@ -621,6 +629,8 @@ mod tests {
         }
     }
 
+    // TODO: Fix the test.  DO NOT SUBMIT
```
romac, line 632: Reminder to fix this TODO

romac (changes requested):
Overall looks good to me! Left a few minor questions and notes.

Do you have any numbers as to how much these changes improve gas efficiency compared to your previous PR?

light-client-verifier/src/operations/voting_power.rs
mina86: We don’t have to expose it. I guess it just feels natural to that the two required methods are voting_power_in and voting_power_in_sets.  I guess alternative is to get rid of those two methods and just leave `check_enough_trust_and_signers` and `check_signers_overlap` methods as required.

mina86: Changed to remove vote_power_in and vote_power_in_sets. What do you think?

romac: You are right, after looking at the resulting code, it indeed makes more sense to keep `voting_power_in` and `voting_power_in_sets`. Because the latter is a bit ad-hoc, I was hoping we could do without it in favor of a more generic method like

```rust
trait VotingPowerCalculator {
   // ...

    fn compute_voting_power<V: signature::Verifier>(
        votes: &mut NonAbsentCommitVotes,
        validator_set: &ValidatorSet,
        trust_threshold: TrustThreshold,
        total_voting_power: u64,
    ) -> Result<VotingPowerTally, VerificationError>;
}
```

but that would require us to expose the `NonAbsentCommitVotes` struct publicly and add a `V: signature::Verifier` type parameter either to the `VotingPowerCalculator` trait or to each method, and I'd rather keep that as an implementation detail to allow other means of verifying signatures.

Let's thus revert to the previous commit.

What do you think about adding a default implementation for `voting_power_in_sets` that calls `voting_power_in` twice?

mina86: > What do you think about adding a default implementation for voting_power_in_sets that calls voting_power_in twice?

I wondered about this and I think being a bit user-hostile here is justifiable.  Not having default implementation forces the implementator to think about the optimisation rather than using the old unoptimised code.  To ease the nagging equivalent code could be put in docstring so lazy implementators/people in a hurry could just copy that.

romac: Makes sense!

mina86: I think this is ready for review then.

mina86:
> Do you have any numbers as to how much these changes improve gas efficiency compared to your previous PR?

Sorry, forgot about this question.  At this moment I don’t have concrete numbers.  We were testing various optimisations to get things working on Solana and that was one of them.  At the moment I’ve no numbers for this change in isolation.  I’ll try to get some numbers today but cannot promise.

romac (approved):

romac:
@mina86 Great work, thanks a lot for the care you've put into this PR and for bearing with me, much appreciated!

PR #1413, opened 2024-04-22
light-client-verifier: reuse sign_bytes buffer

mina86:
Rather than allocating a new vector each time canonical vote is encoded, allocate a vector in NonAbsentCommitVotes once and then reuse it each time sign_bytes are needed.  This saves on doing multiple allocations and deallocations.

For overly verbose description of the length calculation:

* CanonicalVote.  Note that length prefix doesn’t count bytes used for encoding the length, e.g. if the CanonicalVote message is 127 bytes than the length prefix is one byte and thus sign_bytes is 128 bytes. The table assumes all fields are set.

      |                 | tag | value |
      |-----------------+-----+-------+
      | (length)        |     |   1–2 | 2 if (length) is > 127
      | type            |   1 |     1 | SignedMsgType is at most 32
      | height          |   1 |     8 | The field is sfixed64
      | round           |   1 |     8 | The field is sfixed64
      | block_id        |   1 | 73–77 | See CanonicalBlockId below
      | timestamp       |   1 |  7—13 | See Timestamp below
      | chain_id        |   1 |   1+N | Assuming N < 128
      |-----------------+-----+-------+
      | (total)         |   105–116+N | If N≤13 than (length) is
      |                 |             | 1 byte and this is 115+N

* CanonicalBlockId.

      |                 | tag | value |
      |-----------------+-----+-------+
      | (length)        |     |     1 |
      | hash            |   1 |    33 | Assuming 32-byte hash.
      | part_set_header |   1 | 37–41 | See CanonicalPartSetHeader
      |-----------------+-----+-------+
      | (total)         |       73–77 |

* CanonicalPartSetHeader.

      |                 | tag | value |
      |-----------------+-----+-------+
      | (length)        |     |     1 |
      | total           |   1 |   1–5 | Varint of u32; at most 5 bytes.
      | hash            |   1 |    33 | Assuming 32-byte hash.
      |-----------------+-----+-------+
      | (total)         |       37–41 |

* Timestamp.  In unlikely event vote was made exactly at a second, nanos is zero and thus the field is not present.  Assuming uniform probability of the nanos, in a bit over 73% of cases the filed will occupy it’s maximum size.

      |                 | tag | value |
      |-----------------+-----+-------+
      | (length)        |     |     1 |
      | seconds         |   1 |     5 | Assuming timestamps around now.
      | nanos           | 0–1 |   0–5 |
      |-----------------+-----+-------+
      | (total)         |     |  7–13 |

* [ ] Wrote tests


light-client-verifier/src/operations/voting_power.rs
```diff
@@ -319,14 +329,18 @@ impl NonAbsentCommitVotes {
         };
 
         if !vote.verified {
-            let sign_bytes = vote.signed_vote.sign_bytes();
+            self.sign_bytes.truncate(0);
+            vote.signed_vote
+                .sign_bytes_into(&mut self.sign_bytes)
+                .unwrap();
```
romac, line 335:
```suggestion
                .expect("the signing buffer should be large enough; perhaps the chain identifier is longer than 44 bytes?");
```

light-client-verifier/src/operations/voting_power.rs
```diff
@@ -297,7 +302,12 @@ impl NonAbsentCommitVotes {
                 pair[0].validator_id(),
             ))
         } else {
-            Ok(Self { votes })
+            // As of protocol 0.38, 160 buffer is enough for chain identifier up
+            // to 44 bytes.
+            Ok(Self {
+                votes,
+                sign_bytes: Vec::with_capacity(160),
```
romac, line 309: Let's move the capacity to a constant, and move the comment there as well?

romac: Also, according to the CometBFT [docs](https://docs.cometbft.com/v0.37/core/using-cometbft#genesis) and [implementation](https://github.com/cometbft/cometbft/blob/v0.37.x/types/genesis.go?plain=1#L20), the ChainID can be of length up to 50 bytes, so let's account for that and increase the capacity to 166?

romac (changes requested):
Thanks a lot for the PR! Looks good, but I left a couple suggestions and believe we should increase the buffer size a little bit to err on the safe side, see my comment above.

light-client-verifier/src/operations/voting_power.rs
mina86: The buffer is resized as needed so this only fails due to memory allocation failures.  See that [remaining_mut](https://docs.rs/bytes/1.5.0/src/bytes/buf/buf_mut.rs.html#1459) for a Vec basically returns isize::MAX and [prost checks that](https://docs.rs/prost/latest/src/prost/message.rs.html#85).

romac (approved):
Thanks again so much for the PR!

romac:
I will release all this in v0.36 later this week.

Issue #1415, opened 2024-04-27
Deserialization of `abci::Event` fails


penso:
### What went wrong?

If you look at https://github.com/cometbft/cometbft/blob/main/api/cometbft/abci/v1/types.pb.go#L3012C82-L3012C91 you'll notice `omitempty` for `Type`. Trying to find previous similar issues it looks like the same as explained by @ebuchman in https://github.com/informalsystems/tendermint-rs/issues/197

Block 12791634 for DYDX has the following block_results:

```json
    "finalize_block_events": [
      {
        "attributes": [
          {
            "key": "",
            "value": "\n0/dydxprotocol.clob.MsgProposedOperationsResponse",
            "index": false
          }
        ]
      }
    ]
```

It therefor doesn't have any `type`, deserialization fails.

hdevalence:
Probably Event's serialize impls should run through the proto type (using Serde's try_from/into attributes) and the proto type should have a ProtoJSON compatible Serde impl via pbjson_build.

hdevalence:
Example of how this works for a random Penumbra data structure: the `CompactBlock` rust type serializes through the proto type:
https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/compact-block/src/compact_block.rs#L20-L21
The corresponding proto type has a generated Serde impl that handles this correctly:
https://github.com/penumbra-zone/penumbra/blob/main/crates/proto/src/gen/penumbra.core.component.compact_block.v1.serde.rs
That code is generated using `pbjson_build`:
https://github.com/penumbra-zone/penumbra/blob/main/tools/proto-compiler/src/main.rs#L135-L147

There's a bit more machinery here but the upshot is that this fixes the problem in every case. (In Penumbra, we never derive Serde formats from the Rust representation, all our data formats are specified only in protos, with no custom "shaping" of the format, to avoid these kinds of inconsistencies).

penso:
The proto file at https://github.com/cometbft/cometbft/blob/main/proto/cometbft/abci/v1/types.proto#L467 shows the `type` field isn't optional, the generated Rust struct from the proto file wouldn't be optional either and same bug would occur. You'd need to add the same fix `serde(default)` for this specific field when generating the Rust struct from the proto file.

The fix would be for the Go implementation to not have `omitempty`.

hdevalence:
[...]

PR #1416, opened 2024-04-27
fix(abci): Add serde `default` for `Event::type`

penso:
Fixes https://github.com/informalsystems/tendermint-rs/issues/1415

* [X] Wrote tests

### DIFF
--- a/tendermint/src/abci/event.rs
+++ b/tendermint/src/abci/event.rs
@@ -18,7 +18,7 @@ pub struct Event {
     ///
     /// Tendermint calls this the `type`, but we use `kind` to avoid confusion
     /// with Rust types and follow Rust conventions.
-    #[serde(rename = "type")]
+    #[serde(rename = "type", default)]
     pub kind: String,
 
     /// A list of [`EventAttribute`]s describing the event.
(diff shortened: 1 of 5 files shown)


tendermint/src/abci/event.rs
```diff
@@ -18,7 +18,7 @@ pub struct Event {
     ///
     /// Tendermint calls this the `type`, but we use `kind` to avoid confusion
     /// with Rust types and follow Rust conventions.
-    #[serde(rename = "type")]
+    #[serde(rename = "type", default)]
```
hdevalence, line 21: I'm not sure this is an optimal solution, because it's recreating the data format by hand rather than deriving it from the source of truth (the proto messages). See https://github.com/informalsystems/tendermint-rs/issues/1415#issuecomment-2081010722

penso: Went for the shortest fix to an immediate issue. But anyone else, please feel free to fix it differently.

romac (approved):

tendermint/src/abci/event.rs
romac: Let's go with this fix for now since we are already re-creating the data format by hand everywhere in tendermint-rs.

Maybe we can address this whole thing at once and move to ProtoJSON in a follow-up PR or in cometbft-rs.

PR #1421, opened 2024-05-18
feat(rpc): expose reqwest client

KaiserKarel:
Previously, the reqwest client for HttpClient would be unconditionally built by the internal builder, meaning that useful middleware such as tower could not be applied.

This commits adds two ways to create the HttpClient with a custom reqwest::Client, either through new_from_parts, or using the builder.

* [ ] Referenced an issue explaining the need for the change -> this is a trivial code change.
* [ ] Wrote tests


romac:
@KaiserKarel Looks great, thanks for the PR. Can you please add a changelog entry for this, mentioning the new `Builder::client` and `HttpClient::new_from_parts` methods?

romac:
Don't worry about the failure on nightly, I'll take care of it.

romac (approved):

PR #1423, opened 2024-05-21
light-client-verifier: 🌱 restore `verify_commit()` interface

cratelyn:
in #1410, alterations were made to the `PredicateVerifier<P, C, V>` to improve performance when verifying commits.

specifically, a call to
`verify_commit_against_trusted(untrusted, trusted, options)` will now make use of new internal interfaces that check validator signatures and signer overlap at the same time.

this is a worthwhile performance improvement, but in the process, it made changes to the public interface of `PredicateVerifier<P, C, V>` that breaks some use cases. as i understand it, there is no strict need to remove the other public interface from the verifier. those that call `verify_commit_against_trusted` can still receive a performance boost, but calling `verify_commit` on just the untrusted state should still work after those changes.

so, this commit restores `verify_commit(untrusted)`, and keeps `verify_commit_against_trusted(untrusted, trusted, options)` under its previous name.

<!--

Thanks for filing a PR!

Before hitting the button, please check the following items.  Please note that
every non-trivial PR must reference an issue that explains the changes in the
PR.

Please also make sure you've targeted the correct branch with your PR. See the
contributing guidelines for details.

### DIFF
--- a/.changelog/unreleased/breaking-changes/1423-verify-commit-restore.md
+++ b/.changelog/unreleased/breaking-changes/1423-verify-commit-restore.md
@@ -1,0 +1,5 @@
+- `[tendermint-light-client-verifier]` Restores the commit verification interfaces of `PredicateVerifier<P, C, V>` from `<= 0.35.0`.
+  * `verify_commit(&self. untrusted: &UntrustedBlockState<'_>)` is restored, as in <= 0.35.0.
+  * `verify_commit(&self, untrusted: &UntrustedBlockState<'_>, trusted: &TrustedBlockState<'_>,)` introduced in 0.36.0 is renamed to `verify_commit_against_trusted`.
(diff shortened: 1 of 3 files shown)


cratelyn:
fixed nightly regression in #1424, i'll rebase when that lands.

romac:
Looks good me, agree we should have been careful not to break the existing API.

@mina86 How does that sound to you?

romac:
@cratelyn Can you please add a ch age log entry while you're at it? 🙏

cratelyn:
> @cratelyn Can you please add a ch age log entry while you're at it? 🙏

@romac sure thing. c0e0f65 :scroll:

romac:
Thanks! Will wait for @mina86 to take a look before merging.

mina86:
LGTM

romac (approved):

PR #1436, opened 2024-06-14
tendermint:  relax `Block` validation rules to avoid assumptions on initial height

### LINKED ISSUE #1435: block: domain type validation is too stringent
### What went wrong?

Comet allows non-zero initial heights, those are useful and should be considered frequent (e.g. Penumbra use them to perform offchain state transitions).

However, the domain type validation logic of `Block` enforces an `initialHeight` of 1 which is too stringent:

https://github.com/informalsystems/tendermint-rs/blame/94a5fc074964fca05df48acde88ce133a58f4ce7/tendermint/src/block.rs#L125

### Definition of "done"

The check is dropped so that it is possible to parse blocks that don't refer to a last commit. Open to better suggestions but that seem straightforward to do.

erwanor:
Fix #1435, by removing extra-validation done on the `Block` domain type.

* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/breaking-changes/1435-block-domain-type.md
+++ b/.changelog/unreleased/breaking-changes/1435-block-domain-type.md
@@ -1,0 +1,2 @@
+- tendermint: relax validation rules on `Block`
+  ([\#1435](https://github.com/informalsystems/tendermint-rs/issues/1435))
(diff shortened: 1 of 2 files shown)


romac (approved):
Thanks! 🙏

PR #1439, opened 2024-06-26
rpc: Add `/genesis_chunked` query

### LINKED ISSUE #1438: Add `/genesis_chunked` RPC endpoint
### Description

Current version of tendermint-rs `v0.37.0` only has the `/genesis` endpoint, https://github.com/informalsystems/tendermint-rs/blob/v0.37.0/rpc/src/method.rs#L58.
If the genesis data is too big this will result in a error, see https://github.com/informalsystems/hermes/issues/3589.

By adding the `/genesis_chunked` endpoint it will be possible to retrieve the data from multiple chunks.

### Definition of "done"

* [ ] `/genesis_chunked` endpoint

ljoss17:
Closes #1438

* [ ] Wrote tests

### DIFF
--- a/.changelog/unreleased/features/1438-genesis-chunked-endpoint.md
+++ b/.changelog/unreleased/features/1438-genesis-chunked-endpoint.md
@@ -1,2 +1,2 @@
-- `[tendermint-rpc]` Add `/genesis_chunked` RPC endpoint
+- `[tendermint-rpc]` Add support for the `/genesis_chunked` RPC endpoint
   ([\#1438](https://github.com/informalsystems/tendermint-rs/issues/1438))
--- a/.changelog/unreleased/features/1438-genesis-chunked-endpoint.md
+++ b/.changelog/unreleased/features/1438-genesis-chunked-endpoint.md
@@ -1,0 +1,2 @@
+- `[tendermint-rpc]` Add `/genesis_chunked` RPC endpoint
+  ([\#1438](https://github.com/informalsystems/tendermint-rs/issues/1438))
--- a/rpc/src/client.rs
+++ b/rpc/src/client.rs
@@ -271,6 +271,14 @@ pub trait Client {
         Ok(self.perform(genesis::Request::default()).await?.genesis)
     }
 
+    /// `/genesis_chunked`: get genesis file in multiple chunks.
+    async fn genesis_chunked(
+        &self,
+        chunk: Option<String>,
+    ) -> Result<genesis_chunked::Response, Error> {
+        self.perform(genesis_chunked::Request::new(chunk)).await
+    }
+
     /// `/net_info`: obtain information about P2P and other network connections.
     async fn net_info(&self) -> Result<net_info::Response, Error> {
         self.perform(net_info::Request).await
--- a/rpc/src/client.rs
+++ b/rpc/src/client.rs
@@ -5,7 +5,6 @@ pub use compat::CompatMode;
 
 #[cfg(any(feature = "http-client", feature = "websocket-client"))]
 mod subscription;
-use futures::Stream;
 #[cfg(any(feature = "http-client", feature = "websocket-client"))]
 pub use subscription::{Subscription, SubscriptionClient};
 
@@ -26,7 +25,6 @@ pub use transport::websocket::{
 pub use transport::mock::{MockClient, MockRequestMatcher, MockRequestMethodMatcher};
 
 use core::fmt;
-use core::pin::Pin;
 
 use async_trait::async_trait;
 use serde::{de::DeserializeOwned, Serialize};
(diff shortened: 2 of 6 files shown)


rpc/src/endpoint/genesis_chunked.rs
```diff
@@ -0,0 +1,52 @@
...
+use alloc::{string::String, vec::Vec};
+use serde::{Deserialize, Serialize};
+use tendermint_proto::serializers;
+
+use crate::{dialect::Dialect, request::RequestMessage};
+
+/// Get the genesis state in multiple chunks for the current chain
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct Request {
+    pub chunk: Option<String>,
```
romac, line 12: In the RPC spec, the chunk parameter is specified as an integer: https://docs.tendermint.com/v0.37/rpc/#/Info/genesis_chunked

rpc/src/endpoint/genesis_chunked.rs
```diff
@@ -0,0 +1,52 @@
...
+    type Output = Response;
+}
+
+/// Block responses
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Response {
+    #[serde(with = "serializers::from_str")]
+    pub chunk: u64,
+    #[serde(with = "serializers::from_str")]
+    pub total: u64,
```
romac, line 47: These two are also supposed to be integers as far as I can tell.

rpc/src/client.rs
```diff
@@ -271,6 +271,14 @@ pub trait Client {
         Ok(self.perform(genesis::Request::default()).await?.genesis)
     }
 
+    /// `/genesis_chunked`: get genesis file in multiple chunks.
+    async fn genesis_chunked(
```
romac, line 275: It would also be nice if we could define a default method to retrieve all parts as once as a [`Stream`](https://docs.rs/futures/latest/futures/stream/trait.Stream.html).

eg. we could add another method with this signature:

```rust
async fn genesis_chunked_stream(&self) -> impl Stream<Item = Result<genesis_chunked::Response, Error>> {
  // ...
}
```

Remains to see exactly how to implement this.

romac (changes requested):

rpc/src/client.rs
romac: Perhaps something like this?

```rust
    async fn genesis_chunked_stream(&self) -> Box<dyn Stream<Item = Result<Vec<u8>, Error>> + '_> {
        Box::new(futures::stream::unfold(Some(0), move |chunk| async move {
            let Some(chunk) = chunk else {
                // An error occurred, abort the stream
                return None;
            };

            match self.genesis_chunked(Some(chunk)).await {
                Ok(response) => {
                    if response.chunk >= response.total {
                        // No more chunks to fetch
                        None
                    } else {
                        // Emit this chunk and fetch the next chunk
                        Some((Ok(response.data), Some(response.chunk + 1)))
                    }
                },
                Err(e) => Some((Err(e), None)), // Abort the stream
            }
        }))
    }
```

rpc/src/endpoint/genesis_chunked.rs
ljoss17: It was also my understanding, but when calling this query with `Option<i32>` I get:

```
outgoing request url=http://localhost:17298/ body={
  "jsonrpc": "2.0",
  "id": "93d95f7c-17ea-478a-96c4-9e11ff198ca0",
  "method": "genesis_chunked",
  "params": {
    "chunk": 0
  }
}
...
incoming response status=200 OK body={"jsonrpc":"2.0","id":"93d95f7c-17ea-478a-96c4-9e11ff198ca0","error":{"code":-32602,"message":"Invalid params","data":"error converting json params to arguments: invalid 64-bit integer encoding \"0\", expected string"}}
```

rpc/src/endpoint/genesis_chunked.rs
ljoss17: I seem to need to convert it or else I'm getting:

```
query to /genesis_chunked failed with error: 
   0: serde parse error
   1: invalid type: string "0", expected i32 at line 1 column 82
```

romac: Alright, my bad, was mislead by the docs.

rpc/src/endpoint/genesis_chunked.rs
romac: I see, so let's take a `usize`, and convert to `Option<String>` internally

romac: ie. 


```rust
async fn genesis_chunked(
    &self,
    chunk: usize,
) -> Result<genesis_chunked::Response, Error> {
    self.perform(genesis_chunked::Request::new(Some(chunk.to_string()))).await
}
```

romac (approved):

.changelog/unreleased/features/1438-genesis-chunked-endpoint.md
```diff
@@ -0,0 +1,2 @@
+- `[tendermint-rpc]` Add `/genesis_chunked` RPC endpoint
```
romac, line 1:
```suggestion
- `[tendermint-rpc]` Add support for the `/genesis_chunked` RPC endpoint
```

Issue #1440, opened 2024-07-11
rpc: Deserializing `/block_results` with empty JSON `consensus_param_updates` fails


ljoss17:
### What went wrong?

If the `/block_results` response has an empty JSON object for the `consensus_param_updates` instead of `null`, the deserialization will fail.

* `{"consensus_param_updates": {}}`: fails
* `{"consensus_param_updates": null}`: succeeds, setting `consensus_param_updates` to `None`

### Definition of "done"

Deserializing an empty JSON object should also result in `consensus_param_updates` being `None`

PR #1441, opened 2024-07-11
rpc: Allow empty JSON object value for `consensus_param_updates` in `/block_results` response

ljoss17:
Closes: #1440 

This PR adds a custom deserializer for the `consensus_param_updates` field in the `/block_results` response to deserialize both `null` and an empty JSON object to `None`.

E.g. both of these will successfully deserialize to `None`:

* `{"consensus_param_updates": {}}`
* `{"consensus_param_updates": null}`

* [ ] Wrote tests


romac (approved):

ljoss17:
> Thanks! Can we also fix this while we are at it?
> 
> https://github.com/informalsystems/tendermint-rs/blob/ed576f2f479e58fd269a2725c5b4730bc2a1776d/tendermint/src/consensus/params.rs#L19-L20

Oh I missed that sorry. I'll fix it

tendermint/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub mod chain;
 pub mod channel;
 pub mod consensus;
 pub mod crypto;
+pub mod deserializers;
```
romac, line 38: Let's add it to the `serializers` instead, to help discovery.

ljoss17: Done in 568f843becc80bce00cf12fa1bf58eebfcba273d

PR #1442, opened 2024-07-13
Release v0.38.0

ljoss17:

PR #1444, opened 2024-07-16
Bump `prost` to v0.13 and `tonic` to v0.12

tony-iqlusion:
Updates these dependencies to their latest versions.

Protos have not been regenerated using the latest `prost-build`/`tonic-build` however seem to be compatible.

See also: #1442


tony-iqlusion:
Well great, I didn't try regenerating the protos as noted in the description/commit message, but it seems if you do regenerate them they no longer compile per the `generated-protos-compile` CI job

tony-iqlusion:
The core problem seems to be that `crate::google::protobuf::Duration` (i.e. `prost_types::Duration`) is no longer a `Copy` type, which means no struct that contains it can be a `Copy` type any more.

I'm not sure why this change was made. It's a simple struct with two public fields which are themselves `Copy` types (`i64` and `i32`)

tony-iqlusion:
Sidebar: in `cosmos-sdk-proto` and `ibc-proto` we've discussed ditching `prost-types` and generating the `google.protobuf` protos directly from the protobuf schema using the same conventions as everything else, which among other things allows integrations with `pbjson` for building JSON serializers /cc @romac 

See discussion here: https://github.com/cosmos/cosmos-rust/pull/459#pullrequestreview-1931444478

romac:
> The core problem seems to be that `crate::google::protobuf::Duration` (i.e. `prost_types::Duration`) is no longer a `Copy` type, which means no struct that contains it can be a `Copy` type any more.
> 
> I'm not sure why this change was made. It's a simple struct with two public fields which are themselves `Copy` types (`i64` and `i32`)

`prost_types::Duration` still implements `Copy` in `prost-types` 0.13.1: https://docs.rs/prost-types/latest/prost_types/struct.Duration.html#impl-Copy-for-Duration

> Sidebar: in `cosmos-sdk-proto` and `ibc-proto` we've discussed ditching `prost-types` and generating the `google.protobuf` protos directly from the protobuf schema using the same conventions as everything else, which among other things allows integrations with `pbjson` for building JSON serializers /cc @romac
> 
> See discussion here: [cosmos/cosmos-rust#459 (review)](https://github.com/cosmos/cosmos-rust/pull/459#pullrequestreview-1931444478)

That's already what we are already doing in tendermint-rs, albeit we do not regenerate them automatically but have instead checked in the generated code: https://github.com/informalsystems/tendermint-rs/blob/main/proto/src/protobuf.rs

These types never implemented `Copy` in the first place, but we can either add it manually or re-generate these specific protos. I can take care of that.

romac:
We should probably generate code for all `google.protobuf` types (ie. including `Any`)  instead of copy-pasting a subset of it, and then use those directly from `ibc-proto-rs` instead of the latter having its own copy of them. 

Otherwise one cannot pass a `tendermint_proto` `Duration` where an `ibc-proto` `Duration` is expected and vice-versa, same for the other *well-known types*.

What do you think?

tony-iqlusion:
@romanc sounds good to me! Seems like the first step towards unifying protos between `tendermint-proto`, `cosmos-sdk-proto`, and `ibc-proto`

I'd just ask to retain `Any::{from_msg, to_msg}`: https://docs.rs/prost-types/latest/prost_types/struct.Any.html#implementations

romanc:
Cc @romac

romac:
> I'd just ask to retain Any::{from_msg, to_msg}: https://docs.rs/prost-types/latest/prost_types/struct.Any.html#implementations

Yes good idea!

Let's do all this in a separate PR.

tony-iqlusion:
@romac I didn't put `tendermint-proto` on the front initially because the changes impact multiple crates

romac (approved):

Issue #1445, opened 2024-07-17
Generate code for all `google.protobuf` types


romac:
@tony-iqlusion:

> Sidebar: in `cosmos-sdk-proto` and `ibc-proto` we've discussed ditching `prost-types` and generating the `google.protobuf` protos directly from the protobuf schema using the same conventions as everything else, which among other things allows integrations with `pbjson` for building JSON serializers /cc @romac 
>
> See discussion here: https://github.com/cosmos/cosmos-rust/pull/459#pullrequestreview-1931444478

@romac:

> We should probably generate code for all `google.protobuf` types (ie. including `Any`)  instead of copy-pasting a subset of it, and then use those directly from `ibc-proto-rs` instead of the latter having its own copy of them. 
>
> Otherwise one cannot pass a `tendermint_proto` `Duration` where an `ibc-proto` `Duration` is expected and vice-versa, same for the other *well-known types*.

_Originally posted by @romac in https://github.com/informalsystems/tendermint-rs/issues/1444#issuecomment-2232582972_

Issue #1053, opened 2021-12-13
Harmonize on proto-types for representing well-known protobuf message types


mzabaluev:
Version(s) of tendermint-rs: v0.23.2, master

### Description

There is code in ibc-rs that [jumps through hoops](https://github.com/informalsystems/ibc-rs/blob/master/modules/src/clients/ics07_tendermint/consensus_state.rs#L73) for the sole reason that the protobuf message structures generated for IBC use proto-types, while tendermint-proto includes a homespun variant of structures representing `google.protobuf`.

This was done in #639 that predates the rearchitecting with domain types and defines serde impls directly on the protobuf message types. This is less preferable than putting the serde concern cleanly onto the domain types, since only the latter can enforce value validity.

### Definition of "done"

`tendermint-proto` uses `proto-types` to represent `Timestamp`, `Duration`, and possibly other well-known message types.
The corresponding domain types in `tendermint` respond to all serde needs.

tony-iqlusion:
We've had some requests downstream in `cosmos-sdk-proto` for better alternatives for representing well-known types:

- https://github.com/cosmos/cosmos-rust/issues/185
- https://github.com/cosmos/cosmos-rust/issues/176

The latter suggests using [`prost-wkt`](https://github.com/fdeantoni/prost-wkt) which is a richer crate for representing well-known types than `prost-types` and it also includes `serde` support.

It also has helper methods for packing and unpacking messages to/from an `Any` type which would be very helpful.

PR #1452, opened 2024-07-30
proto: Add `Any` Protobuf type and improve `Duration` and `Timestamp` Protobuf types

romac:
Closes: https://github.com/informalsystems/tendermint-rs/issues/1445

## Notes

- This PR only includes `Any`, `Duration`, and `Timestamp` for now, but we may add other well-known types if needed.
- All three types come with a `prost::Name` impl, as requested.
- `Any::from_msg` and `Any::to_msg` are provided.
- I chose not to go with `pbjson-build` as it brings a dependency on `chrono` whereas we are using `time` at the moment. Also, there only a few protos and we can just inline the serde implementation generated by `pbjson-build` or write our own manually. I am happy to revisit this if deemed necessary.
- JSON serialization for `Any` is not compliant with the ProtoJSON spec because it is currently impossible to do so without knowledge of all registered types that may be serialized as an `Any`. See [the corresponding `pbjson` issue](https://github.com/influxdata/pbjson/issues/2) for more info. Instead, we currently use the same serialization format as `pbjson`.

## Important

Before we merge this, we need to try out integration in ibc-proto-rs, cosmos-rust, ibc-rs, and Hermes.

---

* [ ] Wrote tests


romac:
@tony-iqlusion Looking forward to hearing what you think :)

tony-iqlusion (approved):
This looks good to me

ash-burnt:
> Unlike [prost_types::Timestamp](https://github.com/tokio-rs/prost/blob/ad5650b0c96b7e5c1f8310462af7f6e35c2a0bfe/prost-types/src/timestamp.rs#L196), our Timestamp struct does not provide a way to construct it from a std::time::SystemTime, is this something we want to add?

@Peartes can you take a pass at this in a follow-up PR?

romac:
> > Unlike [prost_types::Timestamp](https://github.com/tokio-rs/prost/blob/ad5650b0c96b7e5c1f8310462af7f6e35c2a0bfe/prost-types/src/timestamp.rs#L196), our Timestamp struct does not provide a way to construct it from a std::time::SystemTime, is this something we want to add?
> 
> @Peartes can you take a pass at this in a follow-up PR?

I have taken care of it, as well as derivation of impls for `schemars`, `borsh` and `parity-scale-codec`.

romac:
Here is the draft integration PR for ibc-proto: https://github.com/cosmos/ibc-proto-rs/pull/226

Farhad-Shabani (approved):
Thank you @romac. Looks good from ibc-rs end as well. 
Since ibc-proto-rs and tendermint-proto get hormonized, this also resolves [an existing bump](https://github.com/cosmos/ibc-rs/blob/2378cd4ba45094b8ed856ff7dba5f1d0882f59ae/ibc-clients/ics07-tendermint/types/src/consensus_state.rs#L88-L91) in ibc-rs. This PR probably closes #1053 as well?

romac:
Yes once merged it will resolve that as well.

tony-iqlusion:
@romac it looks like both `tendermint` and `tendermint-proto` still include `prost-types` in their Cargo.toml

proto/Cargo.toml
```diff
@@ -15,8 +15,12 @@ description = """
 
 [features]
 default = []
+std = []
```
tony-iqlusion, line 18: I'm getting a bunch of build failures trying to use this feature:

```
$ cargo test --all-features                                                                                       101 ↵
   Compiling tendermint-proto v0.39.0
error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> /Users/tarcieri/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tendermint-proto-0.39.0/src/google/protobuf/timestamp.rs:105:11
```

I think you forgot an `extern crate std` and it seems like CI probably isn't testing it.

tony-iqlusion: Opened an issue to track this: #1454

proto/src/google/protobuf/duration.rs
```diff
@@ -0,0 +1,314 @@
...
+            nanos as i32
+        };
+        let mut duration = Duration { seconds, nanos };
+        duration.normalize();
+        duration
+    }
+}
+
+impl TryFrom<Duration> for core::time::Duration {
+    type Error = core::time::Duration;
```
tony-iqlusion, line 119: Hmm, this previously returned `DurationError`: 

https://docs.rs/prost-types/0.13.1/src/prost_types/duration.rs.html#86-104

```rust
impl TryFrom<Duration> for time::Duration {
    type Error = DurationError;

    /// Converts a `Duration` to a `std::time::Duration`, failing if the duration is negative.
    fn try_from(mut duration: Duration) -> Result<time::Duration, DurationError> {
        duration.normalize();
        if duration.seconds >= 0 && duration.nanos >= 0 {
            Ok(time::Duration::new(
                duration.seconds as u64,
                duration.nanos as u32,
            ))
        } else {
            Err(DurationError::NegativeDuration(time::Duration::new(
                (-duration.seconds) as u64,
                (-duration.nanos) as u32,
            )))
        }
    }
}
```

romac: Good catch, will fix it! As well as replace the `From<core::time::Duration>` instance with a `TryFrom<core::time::Duration>` following `prost-types`'s implementation instead of `pbjson-types` which I lifted those from.

romac: https://github.com/informalsystems/tendermint-rs/pull/1456

PR #1455, opened 2024-08-07
Ensure each and every feature flags can be used independently

### LINKED ISSUE #1454: proto: `std` feature is broken
### What went wrong?

The `std` feature added in #1452 doesn't work. /cc @romac

### Steps to reproduce

```console
$ cd proto
$ cargo test --features std
```

This results in the following compile errors:

```
$ cargo test --features std                                                                                       130 ↵
   Compiling tendermint-proto v0.39.0 (/Users/tarcieri/src/tendermint-rs/proto)
error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> proto/src/google/protobuf/timestamp.rs:105:11
    |
105 | impl From<std::time::SystemTime> for Timestamp {
    |           ^^^ use of undeclared crate or module `std`

error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> proto/src/google/protobuf/timestamp.rs:106:26
    |
106 |     fn from(system_time: std::time::SystemTime) -> Timestamp {
    |                          ^^^ use of undeclared crate or module `std`

error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> proto/src/google/protobuf/timestamp.rs:107:65
    |
107 |         let (seconds, nanos) = match system_time.duration_since(std::time::UNIX_EPOCH) {
    |                                                                 ^^^ use of undeclared crate or module `std`

error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> proto/src/google/protobuf/timestamp.rs:155:29
    |
155 | impl TryFrom<Timestamp> for std::time::SystemTime {
    |                             ^^^ use of undeclared crate or module `std`
```

(truncated)

These errors don't occur when `cargo test --all-features` is used, hence why it wasn't caught by CI, ostensibly because when it's compiled with that some other crate is linking `std`.

It would be good in CI to explicitly test the `std` feature in isolation.

### Definition of "done"

`cargo test --features std` compiles in isolation.

tony-iqlusion:
The `std` feature previously didn't link the `std` crate, and so when used in isolation wouldn't compile.

Fixes #1454

cc @romac


.github/workflows/test.yml
```diff
@@ -48,6 +48,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
+      - run: cargo test -p tendermint-proto
+      - run: cargo test -p tendermint-proto --features std
```
tony-iqlusion, line 52: These caught the failure: https://github.com/informalsystems/tendermint-rs/actions/runs/10290702236/job/28481276076?pr=1455

romac: I've added `cargo hack test --each-feature` to the test suite to ensure we test each feature independently, which caught a few more errors in other crates as well.

tony-iqlusion:
@romac looks good! 👍

romac (approved):

PR #1456, opened 2024-08-08
proto: Align `TryFrom` impls for `Duration` with `prost-types`

romac:


tony-iqlusion (approved):
I think this probably makes more sense than supporting negative durations, at least unless there's an actual use for them in the Cosmos ecosystem somewhere

tony-iqlusion:
@romac with this merged I can confirm I can upgrade `cosmos-sdk-proto` / `cosmrs` 🎉

romac:
Glad to hear! Will do a release tomorrow then

tony-iqlusion:
@romac it'd be good if you can get #1457 in as well

romac:
Will do! Apologies for the oversight and thanks for all the PRs 🙏

rnbguy:
Coming from the above ibc-rs issue. Because this PR, we can't expect `Protobuf<T: From<Self>>`. So the `Protobuf` trait

https://github.com/informalsystems/tendermint-rs/blob/2f94e7f5346a094ec98e9019d0181815ccede7f1/proto/src/lib.rs#L105

should be refactored to something like

```rs
pub trait Protobuf<T: Message + TryFrom<Self> + Default>
where ...
```

Should I open an issue for this? Or I missed anything from the latest releases?

rnbguy:
I just explored the idea of `trait Protobuf<T: Message + TryFrom<Self> + Default>` and realized a lot of refactor needs to be done inside the tendermint domain types. Mainly because the `encode_length` or `encode_vec` -- these two `Protobuf` trait methods become fallible -- which in turn makes methods like `hash`, `sign_bytes`, `into_signable_vec` infallible too.

I am going to open an issue for this and continue the discussion there.

erwanor:
Like Rano said, this broke a general ecosystem pattern of infallible domain type to wire conversion. To route around this, we are newtyping the core duration and making the conversion infallible: https://github.com/penumbra-zone/ibc-types/pull/94. This is kind of ugly but works.

I don't think we should let this change contaminate the rest of the ecosystem types by changing the `Protobuf` marker trait. It's a core ecosystem pattern that is very entrenched, and good. If we change it, we generate work and headwinds for downstreams, with no upside in return.

Without commenting on this specific change, it would be awesome to have more heads up/consensus when we change domain type modeling or APIs. This is because the smallest change has dangerous ripple effects throughout. In most cases, downstream crates can navigate those change nimbly, but they are more perilous for deployed systems that need to plan around consensus/state compatibility. That's the case for Penumbra, Astria, Namada, and I'm sure many others as well.

PR #1457, opened 2024-08-08
proto: remove `prost-types` dependency

tony-iqlusion:
Now that #1452 has landed `prost-types` isn't used.

This removes it as a dependency.


romac (approved):
Thanks! 🙏

PR #1464, opened 2024-08-28
[WIP] p2p: add `AsyncSecretConnection` based on tokio

tony-iqlusion:
Adds an `async` version of `SecretConnection`, currently specialized to TCP-based connections exclusively.

This requires quite a bit of duplication, especially since the existing `SecretConnection` implementation is entirely built around `std::io` traits, which are unusable in an async context.

The implementation returns whole chunks as a `Vec<u8>`, which avoids the need to "double buffer" incoming data as ownership of such data is passed directly to the caller upon completing a read.

* [ ] Wrote tests

### DIFF
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -25,6 +25,7 @@ test = false
 [features]
 default = ["flex-error/std", "flex-error/eyre_tracer"]
 amino = ["prost-derive"]
+async = ["tokio"]
 
 [dependencies]
 chacha20poly1305 = { version = "0.10", default-features = false, features = ["reduced-round"] }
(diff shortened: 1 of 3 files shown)


tony-iqlusion:
Note: I'd like to move TMKMS to an async architecture, which would make it better at concurrency (e.g. supporting multiple YubiHSMs, and pipelining extension signing requests to multiple YubiHSMs) but that's difficult to do when the entire request layer is synchronous.

I thought I'd do this in an additive manner for now, rather than attempting to make any breaking changes to `tendermint-p2p`.

This implementation is currently untested. I'd like to get some tests in to show it can at least interop against the synchronous `SecretConnection` server.

PR #1467, opened 2024-09-17
rpc: Add new dialect for v0.38 in order to fix changes in misbehaviour evidence serialization

romac:
Specifically for these changes ([Issue](https://github.com/cometbft/cometbft/issues/3528), [PR](https://github.com/cometbft/cometbft/pull/3543))

* [x] Wrote tests

### DIFF
--- a/rpc/src/endpoint/block.rs
+++ b/rpc/src/endpoint/block.rs
@@ -102,7 +102,6 @@ pub mod v0_38 {
         pub data: Vec<Vec<u8>>,
 
         /// Evidence of malfeasance
-        // #[serde(with = "self::evidence_serializers")]
         pub evidence: evidence::List,
 
         /// Last commit, should be `None` for the initial block.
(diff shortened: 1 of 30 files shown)


rpc/src/endpoint/block.rs
```diff
@@ -48,3 +61,82 @@ pub struct Response {
...
+    #[serde(try_from = "RawBlock", into = "RawBlock")]
+    pub struct DialectBlock {
+        /// Block header
+        pub header: Header,
+
+        /// Transaction data
+        pub data: Vec<Vec<u8>>,
+
+        /// Evidence of malfeasance
+        // #[serde(with = "self::evidence_serializers")]
```
ljoss17, line 105: Should this be removed?

romac: Yes since it does not have any impact given that we deserialize through RawBlock

ljoss17 (approved):
Looks good to me!

PR #1471, opened 2024-10-31
use Vec::with_capacity avoid realloc mem

nkysg:
use Vec::with_capacity avoid realloc mem
* [ ] Wrote tests


melekes (approved):
👍

PR #1475, opened 2024-11-18
fix(tendermint): Fix deserialization from `serde_json::Value`

### LINKED ISSUE #1474: `tendermint::Hash` can not be used as a return value when used with `jsonrpsee`
We are from [Lumina](https://github.com/eigerco/lumina), a Celestia light node written in Rust. When we started the project we [forked](https://github.com/eigerco/celestia-tendermint-rs) `tendermint-rs` to adjust it to Celestia's modifications. Now we are looking to migrate away form our fork.

### What went wrong?

We are using [jsonrpsee](https://github.com/paritytech/jsonrpsee) for calling into [Celestia's node API](https://node-rpc-docs.celestia.org/). The API is using some types from `tendermint-rs`, such as `tendermint::Hash`.

Any struct that uses `tendermint::Hash` internally (e.g. [`SyncState`](https://github.com/eigerco/lumina/blob/debcadfba38507df7168142302951892d3319d7b/types/src/sync.rs#L6-L26)) fails to deserialize and gives the following error:

```
thread 'sync_state' panicked at rpc/tests/header.rs:109:51:
called `Result::unwrap()` on an `Err` value: ParseError(Error("invalid type: string \"0AAC13B8406BDC81049B4E2F382F75754C5436F7DAC702256E9282EE33A242C4\", expected a borrowed string", line: 0, column: 0))
```

We debugged the issue and this happens because:

1. `jsonrpsee` firsts deserializes to `serde_json::Value` and from there to `SyncState`.
2. `tendermint::Hash` uses `<&str>::deserialize(deserializer)`.
3. `<&str>::deserialize(deserializer)` can not be used on `serde_json::Value` due to borrowing issues.

The pattern that `jsonrpsee` uses (i.e. `&str` -> `serde_json::Value` -> `T`) is valid and it may be used from other crates too.
### Steps to reproduce

```rust
use serde::Deserialize;

#[derive(Debug, Deserialize)]
struct HashTest {
    hash: tendermint::Hash,
}

fn main() {
    let s = r#"{"hash": "9F86D081884C7D659A2FEAA0C55AD015A3BF4F1B2B0B822CD15D6C15B0F00A08"}"#;

    // Works
    serde_json::from_str::<HashTest>(s).unwrap();

    // Fails
    let json_value = serde_json::from_str::<serde_json::Value>(s).unwrap();
    serde_json::from_value::<HashTest>(json_value).unwrap();
}
```

### Definition of "done"

We need to stop using `<&str>::deserialize`. One way is to use `String` but allocates. Another way is to use the [`CowStr`](https://github.com/eigerco/celestia-tendermint-rs/blob/6fc0858e3d291ccd9f3e7d6554ea79b2583e22f5/proto/src/serializers/cow_str.rs#L7) we implemented in our fork, which allocates on demand.

oblique:
Fixes #1474 

* [x] Wrote tests

### DIFF
--- a/.changelog/unreleased/bug-fixes/1475-fix-deserialization-from-serde_json_Value.md
+++ b/.changelog/unreleased/bug-fixes/1475-fix-deserialization-from-serde_json_Value.md
@@ -1,0 +1,4 @@
+- [tendermint-proto] Intoduce `CowStr` deserializer.
+  ([\#1474](https://github.com/informalsystems/tendermint-rs/issues/1474), [\#1475](https://github.com/informalsystems/tendermint-rs/pull/1475))
+- [tendermint] Fix deserialization from `serde_json::Value`.
+  ([\#1474](https://github.com/informalsystems/tendermint-rs/issues/1474), [\#1475](https://github.com/informalsystems/tendermint-rs/pull/1475))
--- a/proto/src/serializers.rs
+++ b/proto/src/serializers.rs
@@ -55,6 +55,7 @@
 
 pub mod allow_null;
 pub mod bytes;
+pub mod cow_str;
 pub mod evidence;
 pub mod from_str;
 pub mod from_str_allow_null;
--- a/proto/src/serializers/bytes.rs
+++ b/proto/src/serializers/bytes.rs
@@ -6,13 +6,14 @@ pub mod hexstring {
     use subtle_encoding::hex;
 
     use crate::prelude::*;
+    use crate::serializers::cow_str::CowStr;
 
     /// Deserialize a hex-encoded string into `Vec<u8>`
     pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
     where
         D: Deserializer<'de>,
     {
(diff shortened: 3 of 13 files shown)


romac:
> Some questions:
> 
> 1. `CowStr` can also be used in any place of `tendermint-rs` that uses `String::deserialize` etc. Should I change all of them?

This can be done in a follow-up PR :)

> 2. Do you like `CowStr` or should I just use `String`?

`CowStr` is fine!

romac:
Can you please add a changelog entry in the appropriate subdirectory in `.changelog/`? Then I'll gladly merge this :)

melekes:
@oblique hey! Sorry for the delay here. Do you have the capacity to finish this?

melekes (approved):
👍

oblique:
@melekes Hey, sorry, I completely forgot about this PR. Thanks for pinging. I added the changelog, I hope it's how you want it.

oblique:
@melekes fixed all the clippy issues of `tendermint-proto` and `tendermint`

PR #1484, opened 2025-01-05
docs: fixed typos

Marcofann:


rpc/src/client/transport/auth.rs
```diff
@@ -1,5 +1,5 @@
 //! This module defines the `Authorization` type for
-//! authorizing a HTTP or WebSocket RPC client using
+//! authorizing an HTTP or WebSocket RPC client using
```
melekes, line 2: I don't think this is correct

Marcofann:
@melekes 
I've reverted the change in `auth.rs` as requested.
Let me know if there's anything else you'd like to adjust.

melekes (approved):
👍

PR #1494, opened 2025-03-28
fix: strip auth info from url to avoid double authorization header

gjermundgaraba:
Closes: #1495

* [x] Wrote tests


romac (approved):
Thank you!

Issue #1500, opened 2025-04-15
Incorrect deserialization of `ResultBroadcastTx`


larry0x:
When broadcasting a transaction using CometBFT RPC, it returns a `ResultBroadcastTx` where the `Data` field is encoded in Hex:

https://github.com/cometbft/cometbft/blob/v0.38.17/rpc/core/types/responses.go#L177

```go
// CheckTx result
type ResultBroadcastTx struct {
	Code      uint32         `json:"code"`
	Data      bytes.HexBytes `json:"data"`  // <------ here!!!
	Log       string         `json:"log"`
	Codespace string         `json:"codespace"`

	Hash bytes.HexBytes `json:"hash"`
}
```

However, tendermint-rs attempts to deserialize this as Base64:

https://github.com/informalsystems/tendermint-rs/blob/main/rpc/src/endpoint/broadcast/tx_sync.rs#L48

```rust
/// Response from either an async or sync transaction broadcast request.
#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct Response {
    /// Code space
    pub codespace: String,

    /// Code
    pub code: Code,

    /// Data
    #[serde(with = "serializers::bytes::base64string")]  // <------ here!!!
    pub data: Bytes,

    /// Log
    pub log: String,

    /// Transaction hash
    pub hash: Hash,
}
```

This causes deserialization to return gibberish data.

larry0x:
Seems like a duplicate of https://github.com/informalsystems/tendermint-rs/issues/1251

larry0x:
Here's a tentative fix that works for us. But we only use `tx_sync`; we're not familiar with `tx_async` and `tx_commit`.

https://github.com/informalsystems/tendermint-rs/compare/v0.40.3...left-curve:tendermint-rs:v0.40.3-grug

PR #1501, opened 2025-04-30
chore: Deserialize data field of broadcast tx response as hex bytes

ancazamfir:
Closes: #1500

* [x] Wrote tests


romac (approved):
Looks good! I double-checked and CometBFT uses hex encoding for those fields since at least v0.34.x, so all good. Thanks for the fix!

PR #1519, opened 2026-05-29
fix(rpc): decode URL userinfo for basic auth

danbryan:
## Summary

- Percent-decode URL username and password fields before constructing Basic Auth credentials
- Preserve existing behavior for username-only and username/password auth
- Add regression coverage for reserved characters in URL userinfo

## Testing

- `cargo test -p tendermint-rpc --features http-client client::transport::auth`
- `cargo fmt --check -p tendermint-rpc`


srdtrk (approved):
lgtm

